PR #22, opened 2012-01-05
xt_recent-echo action

keszybz:
This is incomplete without the stuff on the debian branch, but since each branch must be merged separately, I'll send a second pull request for debian.

### DIFF
--- a/config/action.d/iptables-xt_recent-echo.conf
+++ b/config/action.d/iptables-xt_recent-echo.conf
@@ -1,0 +1,76 @@
+# Fail2Ban configuration file
+#
+# Author: Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>
+#
+# $Revision: 1 $
+#
+
+[Definition]
+
+# Option:  actionstart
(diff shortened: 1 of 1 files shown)


lkraav:
Hey all. Not sure if this is worth opening up a new issue about, so asking here first. If I want the `xt_recent` strategy alone handle when the unban should happen, the way to go is to set `actionunban =` to empty value?

sebres:
yes, but I don't see in this PR xt_recent knows the bantime, so the default timeout (if have it) will be used there. 
Additionally, we'll get a ban time increment (prolonged ban for bad IPS) in newer version of f2b, where bantime is variable, in this case how xt_recent should know the unban time?

lkraav:
> Additionally, we'll get a ban time increment (prolonged ban for bad IPS) in newer version of f2b, where bantime is variable, in this case how xt_recent should know the unban time?

I think the variable bantime at f2b level helps. I'm thinking that a prolonged tarpit strategy where the ban time increases with each violation is helpful.

PR #38, opened 2012-03-02
"reban" action + dshield-matches implementation

stefdude:
After banning, I still like to log attempts (via a iptables/shorewall droplog target).
I want to do something more with IPs that insist in connecting (DROP on all ports, make the ban last longer, ...)

The first patch adds "actionreban": an optional action to perform when an IP triggers the criteria while already being banned.
Secondly, it updates "matches" and the "timestamp" of the ban. This avoids a BAN-UNBAN-BAN-UNBAN cycles if the bantime is shorter than the "attack time".

Secondly, as I defined a (custom) filter to read the iptables-logging in the kernel logs, and the DShield service supports this native iptables-format, I use the 'matches' to report to DShield.

### DIFF
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -1,0 +1,39 @@
+# Fail2Ban configuration file
+#
+# Author: Xavier Devlamynck
+#
+# $Revision$
+#
+
+
+[INCLUDES]
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition]
+
+# Option:  failregex
+# Notes.:  regex to match the password failures messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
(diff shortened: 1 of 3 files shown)


leeclemens:
It seems the ultimate intent has been met with the recidive jail? (dshield can still be reported to, as well.) I'd recommend this be closed withing a short time.

sebres:
A `reban` functionality gets implemented within #2588, just I guess it could serve a bit different purposes, but at the same time it has more to offer here: for example auto-reban in failed case, improved invariant-check (involving reban), etc.
No idea at the moment whether dshield-action from this PR (or rather its `actionreban`) is then "compatible" to new version (or how we can port it to newest version). It looks to me more that `updateBanTicket` and `actionreban` from this PR are intended to something like do some prolongation (so must reban in case of longer bantime)? If so then probably parameter `actionprolong` may be better suitable for this purposes (see #1460).

PR #40, opened 2012-03-22
Added option to choose between network or local socket

ledgr:
You can now choose if you want to run fail2ban on local unix socket or network socket. Using of network socket makes it possible to easily pass commands from central syslog server (where client shoud be) to firewall or other machine (where server is), making it possible to stop attackers from scanning your whole server farm.

P.S.
Some minor changes also made, like client receive buffer size which was 6 bits? Maybe a typing error? There should be *at least* 64 bits, otherwise every single command from client to server is split to multiple parts.

TODO:
- make it possible to *properly* send commands from one client to multiple servers
- add some kind of firewall (by IP) at server side (almost done, but I need to rewrite client/server configuration part)
- rewrite client/server configuration part. Client should have separate serverList option. Server should have separate allowedClients option, and be able to configure itself without a client.
~ encrypt traffic?
~ use password to authenticate against server?

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -17,7 +17,7 @@ client/filterreader.py
 client/actionreader.py
 client/__init__.py
 client/configurator.py
-client/csocket.py
+client/clientcommunicator.py
 server/asyncserver.py
 server/filter.py
 server/filterpyinotify.py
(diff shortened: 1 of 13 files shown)


leeclemens:
Related code has been changed, some even recently..closing as superseded

szepeviktor:
@yarikoptic Do you anyone to contact for developing a - locally queuing - zeromq notification action together?

yarikoptic:
Did you see https://github.com/buanzo/fail2ban-zmq-tools ?

yarikoptic:
Did you see https://github.com/buanzo/fail2ban-zmq-tools ?

szepeviktor:
No.
Thank you.

PR #88, opened 2012-11-08
Ipv6 support.

Th4nat0s:
Seems really to work :)


sebres:
Will close this, because we've introduced another IPv6 solution in 0.10 branch (see #1410)

PR #351, opened 2013-09-10
Renamed test files

craneworks:
All testfiles have now the prefix test_.
So they can be find straight by py.test.


yarikoptic:
I will close this one for now -- we indeed might want to rename test files but this PR is way too stale now

PR #399, opened 2013-10-15
Socket create directory if absent + error reporting improvements

grooverdan:
create socket directory and enhance error message. Base very loosely on #168. I'm thinking its quite lasy to depend on initscripts or the user manually to do things that the program can doo.

BF/ENH: remove race conditions by trying to remove force remove socket. Report errors if cannot remove 
socket

Tests:

Creates directory and socket and communicates

<pre>
$ rm -rf /tmp/nonexist && ./fail2ban-client -s /tmp/nonexist/s.sock  -c config/ start
2013-10-15 10:53:17,453 fail2ban.server : INFO   Starting Fail2ban v0.8.10.dev
2013-10-15 10:53:17,453 fail2ban.server : INFO   Starting in daemon mode
$ ./fail2ban-client -s /tmp/nonexist/s.sock stop && rm -rf /tmp/nonexist
Shutdown successful
</pre>

Reports error if can't create a directory

<pre>
$ ./fail2ban-client -s /etc/cant_write/s.sock  -c config/ start
ERROR  While attempting to create directory /etc/cant_write: [Errno 13] Permission denied: '/etc/cant_write'
$ ./fail2ban-client -s /tmp/s.sock  -c config/ start
2013-10-15 10:54:14,101 fail2ban.server : INFO   Starting Fail2ban v0.8.10.dev
2013-10-15 10:54:14,102 fail2ban.server : INFO   Starting in daemon mode
</pre>

Starting a second instance fails when it could ping the remote

<pre>
$ ./fail2ban-client -s /tmp/s.sock  -c config/ start
ERROR  Server already running
</pre>

The force option had no effect above because the server responded to a ping. Correct behaviour?

<pre>
$ ./fail2ban-client -s /tmp/s.sock -x   -c config/ start
ERROR  Server already running
</pre>

Starting second instance when fist instance killed off still fails

<pre>
$ ./fail2ban-client -s /tmp/s.sock   -c config/ start
2013-10-15 11:57:28,998 fail2ban.server : INFO   Starting Fail2ban v0.8.10.dev
2013-10-15 11:57:28,998 fail2ban.server : INFO   Starting in daemon mode
$ killall -9  fail2ban-server
$ ./fail2ban-client -s /tmp/s.sock   -c config/ start
2013-10-15 11:57:43,699 fail2ban.server : INFO   Starting Fail2ban v0.8.10.dev
2013-10-15 11:57:43,700 fail2ban.server : INFO   Starting in daemon mode
[...]

PR #403, opened 2013-10-22
BF: remove race condition on socket removal with -x

grooverdan:
The non-messy parts of #399


leeclemens:
I'm not convinced this hasn't been superseded and should simply be closed, thoughts?

yarikoptic:
nah -- we still have ```if``` check so would well be worth finishing it up this one -- would only be of a benefit.

sebres:
IMHO, this race condition is not longer an issue. 
Could you possibly explain which steps exactly needed to reproduce?

PR #444, opened 2013-11-19
added support for matching IPv6 addresses in loglines

skoef:
- adjusted HOST regexp to match IPv6 addresses
- prevented fail2ban to try to resolve IPv6 addresses instead of trreat them as plain addresses

### DIFF
--- a/server/failregex.py
+++ b/server/failregex.py
@@ -41,7 +41,7 @@ class Regex:
 		self._matchCache = None
 		# Perform shortcuts expansions.
 		# Replace "<HOST>" with default regular expression for host.
-		regex = regex.replace("<HOST>", "(?:::f{4,6}:)?(?P<host>[\w\-.^_]*\w)")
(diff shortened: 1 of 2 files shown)


Namsep:
Would like to see IPv6 support aswell, having systems that are IPv6 only it's kinda needed. Same applies to OSSEC, for which is also an IPv6 pull I'm currently testing. 

On another note, Cento 7 Firewalld is not very usefull on servers as it gets way to complex. I would suggest to stick to IPtables and use IPset. As there are some hacks to have persistent blocks for OSSEC and I would love to do persistent blocks, there is also an IPset-service -which saves the IPset list to disk- now on Fedora 20. So hopefully soon on Rhel and Centos. If that's the case it would appear to me that it is way easy to do permanent blocks and keeping IP's blocked during server / firewall restarts.

https://bugzilla.redhat.com/show_bug.cgi?id=1136257

sebres:
0.10

Issue #30, opened 2012-01-13
upon restart look only into the tail of log file up to 'bantime' back


yarikoptic:
Source distributions from 0.8.4 until 0.8.6 (exclusive) looked straight into the last line in the logfile, thus ignoring all possible errors back (VCS still had "look at the whole file" upon restart).  in 0.8.6 behavior changed to look through whole file banning while encountering matching entries.  That is an expensive operation if log files are not rotated frequently.  Ideally upon restart f2b should look only in the tail of the file as far back in time as as 'bantime'. 

Probably it is ok to just read them in the reversed order.

The easiest and least inefficient implementation could use 'reversed()' while reading whole file into RAM, thus might be prohibitively inefficient.   For smarter implementations see

http://stackoverflow.com/questions/2301789/read-a-file-in-reverse-order-using-python/2301867#2301867
http://stackoverflow.com/questions/260273/most-efficient-way-to-search-the-last-x-lines-of-a-file-in-python

sebres:
I believe this was implemented in 0.10 (known as seekToTime)...
I will close it now, just reopen if I'm wrong

PR #480, opened 2013-12-07
Add fail2ban persistent data storage

kwirk:
In relation to #2 and #30. Also could be used to support recidive and xarf/blocklist.de, as database stores all matched lines.

Needs some polish and some documenting, but thought I'd raise pull request to gain some comments and feedback.

### DIFF
--- a/fail2ban/server/database.py
+++ b/fail2ban/server/database.py
@@ -227,23 +227,41 @@ class Fail2BanDb(object):
 					"failures": ticket.getAttempt()}))
 
 	@commitandrollback()
-	def getBans(self, cur, jail=None, bantime=None):
-		query = "SELECT ip, timeofban, data FROM bans"
+	def _getBans(self, cur, jail=None, bantime=None, ip=None):
+		query = "SELECT ip, timeofban, data FROM bans WHERE 1"
 		queryArgs = []
 
 		if jail is not None:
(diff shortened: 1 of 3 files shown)


toonmaster:
Am I misunderstanding the fix? I am using Fail2ban 0.9.7 in CentOS 7. Whenever the fail2ban-client is reloaded or restarted, the IP it finds in the log to ban is banned beginning at bantime. In other words, if the IP has already been banned for two hours, the bantime isn't bantime-2hrs, it is reset to bantime. So, an IP that has been banned for two hours would be reset to four hours, having banned it for six hours, not four, as defined by bantime. The 'time served' should be saved in the database so that after a service restart/reload the IP continues to serve the remaining time rather than being reset.

sebres:
This merge is done in 2013 and is outdated a bit, current realization considering the end of ban time, which is `start_of_ban + bantime`, so it works properly. This as regards the expiration date of the ticket self (due to restore from DB process). In case of restart/reload the time of unban is earlier calculated.

The interpolation of bantime self in 0.9 is fixed config-variable, so value substituted from `<bantime>` in action does not see this at all (does not consider it in 0.9, so some action if used it may get longer bantime in this case).
But this was changed too in newest versions (let alone bantime is variable there, due to ban-time-incr feature).

Issue #544, opened 2013-12-30
ENH: trapping disobeying bots by apache-badbots filter


truxoft:
On my server I created a honeypot-trap for bots - it is a secret folder, nowhere linked to, and nowhere published, except of the ban in the robots.txt files of all domains on my server. I added the following lines to the robots.txt which instructs them not to enter and index the given directory:

<pre>User-agent: *
Disallow: /forbidden-666-do-not-enter</pre>

I usually place a simple index file into the directory, where I also hide a honeypot-trap email address for identifying spammers who pick it, and start sending spam to it. 

With the robots.txt trap ready, I can then start banning all bots and script kiddies who attempt to access any URL including the forbidden folder name. I do it by adding the following failregex to the apache-badbots filter:

<pre>^&lt;HOST&gt; -.*"(GET|POST).*/forbidden-666-do-not-enter.*$</pre>

Of, course the name of the forbidden directory (or a file) should not be exactly the same as above on all servers using the filter, otherwise bots could learn avoiding them. The best would be if the name could be passed from the jail.conf/jail.local as a variable. In such case though, the variable would have to have a default value, because failure to set it would lead to fail2ban matching all lines.

Ivo

PR #557, opened 2014-01-03
Apache botsearch

grooverdan:
for #544 and http://www.fail2ban.org/wiki/index.php/Fail2ban:Community_Portal#apache_webmail_phishing_jail_-_regex_and_filter_2 and http://www.fail2ban.org/wiki/index.php/Fail2ban:Community_Portal#Bad_robots_.2F_malicious_users_trolling_robots.txt

### DIFF
--- a/config/filter.d/apache-botsearch.conf
+++ b/config/filter.d/apache-botsearch.conf
@@ -22,6 +22,7 @@ before = apache-common.conf
 [Definition]
 
 failregex = ^%(_apache_error_client)s ((AH001(28|30): )?File does not exist|(AH01264: )?script not found or unable to stat): <webroot><block>(, referer: \S+)?\s*$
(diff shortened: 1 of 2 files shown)


config/filter.d/apache-botsearch.conf
```diff
@@ -0,0 +1,48 @@
...
+
+ignoreregex = 
+
+
+[Init]
+
+# Webroot represents the webroot on which all other files are based
+webroot = /var/www/
+# Block is the actual non-found directories to block
+block = (<webmail>|<phpmyadmin>|<wordpress>)[^,]*
```
yarikoptic: @grooverdan  would you remember why you have decided to plug all those into explicit variables here instead of just defining them for string interpolations, how we did in common.conf for other generic common patterns?

PR #661, opened 2014-03-21
ENH: log also threadName -- helps a lot while debugging multiple jails

yarikoptic:
but may be it should be in effect only in DEBUG mode?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,12 @@ ver. 0.9.1 (2014/xx/xx) - better, faster, stronger
    * fail2ban-regex failed to work in python3 due to space/tab mix
    * journalmatch for recidive incorrect PRIORITY
 
+- New features:
+
+
+- Enhancements
+   * Fail2ban-regex - add print-all-matched option. Closes gh-652
+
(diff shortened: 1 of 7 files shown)


leeclemens:
Sorry if I'm just missing something, but why is "debug mode which is pretty much unsupported for recidive" not supported by a specific jail? I admit I haven't checked this out in the source...but it just seems too strange to be true.

yarikoptic:
On Tue, 07 Jul 2015, Lee Clemens wrote:

>    Sorry if I'm just missing something, but why is "debug mode which is
>    pretty much unsupported for recidive" not supported by a specific jail? I
>    admit I haven't checked this out in the source...but it just seems too
>    strange to be true.

recidive is based on parsing the fail2ban.log file... if we increase
loglevel to debug -- it would cause "snake eating its own tail"
situation while recidive jail reading new lines from the log file and
reporting back on that into the same file...

-- 
Yaroslav O. Halchenko, Ph.D.

leeclemens:
fundamentally flawed, as I've already mentioned.

yarikoptic:
yeah -- let's dump for now ;) Cheers

PR #662, opened 2014-03-23
Removed challenge from regex

echabot:
A challenge is just a request being made not if it was successful or not. SIP auth challenges should only be checked for in a DOS filter.


sebres:
Seems to be fixed in e995d5a0b6ae416b1d13091361dde7ce33668026 (filter implementing modes now). Thus closed.

PR #677, opened 2014-04-02
Create named-ddos.conf

eschmidbauer:
This stuff probably needs to be cleaned up.


leeclemens:
I suggest this be closed, as it seems generally the wrong approach to a perceived problem - at best. Worse case, it needs to be fixed up quite a bit and hasn't been in almost a year.

yarikoptic:
thus closing for now.  I am also adding a new tag "closed-for-no-activity"

PR #716, opened 2014-05-05
Ban time incr

sebres:
New feature "automatically exponential increasing ban time".
WARNING: by first start the server upgrades sqlite database (table "bans" will recreated with another schema);
Option "bantime.increment" in jail.conf allows to use database for searching of previously banned ip's to increase a default ban time using special formula, by default each next ban increase banTime * 1, 2, 4, 8, 16, 32...
see "jail.conf" for some other options of "bantime.___";
additional we can configure a little randomization of ban time, to prevent "clever" botnets calculate exact time IP can be unbanned.


sebres:
will pro-rata fix a FR from #1020

yarikoptic:
Still interested to finalize it ? (look through my unaddressed comments)

sebres:
Finalize? How can I find all of these "unaddressed" comments at once?
I'm already pretty far away with my private branches, based on it. And this one is outdated for me, but I don't want to enlarge this request, that already large enough.
I can try to rebase my better branch relative current master, but it will be again a "bigish PR"...

yarikoptic:
all those comments are on this page! just look for my comment which doesn't have your comment after it

Is your better branch solely about this feature or a collection of various changes?  if this feature -- new PR would be ok, if collection -- would be impossible to provide proper review etc.  If that better one is based on this one, after we finalize this one your other changes would become leaner so possible to review

leeclemens:
It seems like there are some potentially independent improvements made there (log file search algo, ban time ranking, etc). It may be better if they were split up (assuming not dependent).

yarikoptic:
On Fri, 03 Jul 2015, Lee Clemens wrote:

>    It seems like there are some potentially independent improvements made
>    there (log file search algo, ban time ranking, etc). It may be better if
>    they were split up (assuming not dependent).

splitting would be too much work to achieve... if this one could be
reviewed/finalized in the current set of introduced features, imho would
be ok, if there is no objection against any particular "feature"

leeclemens:
@yarikoptic I think `git cherry-pick` was meant for that :)  First example that comes to mind, that could benefit recent issues without pulling in tons of other changes, is the fix for seeking (#1016) which references sebres@53a30a2 and sebres@c1637e9. I may be wrong (and will happily admit such), but using a different algorithm to read a file initially shouldn't be dependent upon a bunch of changes to increment the ban time (as #716 suggests).

leeclemens:
@yarikoptic or see #1048 and the subsequent reference to this PR to solve the issue...which specific change is that comment referring too? Showing  37 changed files  with 1,984 additions and 145 deletions. Or is a config argument not recognized without changing 37 files? (bad ... very bad.)

Please reference any software development or version control best practices documentation for more clarification.

leeclemens:
My frustration is because I love it, but I also hate it :)   So many good changes, lumped together as one massive change.

sebres:
<blockquote>if fail2ban actually required 37 files to be changed in order to recognize a single configuration element's value being changed.</blockquote>
@leeclemens Well,  the problem is not a `dbpurge` self, but that I've said already several times, purge will at all never called (only in test case):
https://github.com/fail2ban/fail2ban/search?utf8=%E2%9C%93&q=%22purge%28%29%22&type=Code

leeclemens:
@sebres thanks for bearing with me...I see a lot of value in aspects of this PR and want to do some actual testing (have just glossed over commit diffs so far).

yarikoptic:
could you please push @sebres new version with current master merged for yet another attempt to review etc ;-)

sebres:
master branch merged to ban-time-incr (in the state where it was, without performance mod, etc).
soon I will try to review the code, for possible bug fixing and better config handling in my future branch.
To safe it state, I will push it in separate poll request.

sebres:
Because it's very outdated I will close this PR (as it is here I cannot support it more). The new one coming soon on sebres-fork.

szepeviktor:
@yarikoptic @leeclemens Do you have anything against this PR (or its new version)?

leeclemens:
@szepeviktor This PR was closed on Jul 24, almost 3 months ago. My previous comments still exist if unaddressed.

yarikoptic:
The problem was indeed that pr included multiple new features (incremental ban, string times, use of observer purpose of which seems wasn't cleared out, and I think there was something else cool), so adequate review was to difficult and some comments were lost/unanswered. Wanna help @szepeviktor? Split out eg first string time commits/changes into separate PR... Then may be we could proceed to the next feature, and so on

Issue #720, opened 2014-05-14
strptime bug fix: if gmtoff is None we have 1 hour increment of time (europe/berlin zone)


sebres:
related to pull request #710.
By tests I see all the time wrong lines will be found wit pooling backend.

Example I have in auth.log folowing line:

<pre>
May 14 10:56:52 ... test:auth[22715]: pam_unix(test:auth): authentication failure; logname= uid=0 euid=0 tty=test ruser= rhost=1.2.3.5
</pre>

So with findtime=600 it should be found not later <b>11:06:52</b>, but this line will be found still <b>11:47</b>, defying database will be used also, here is a log-extract:

<pre>
2014-05-14 11:47:29,630 fail2ban.server.datedetector[23174]: DEBUG   Got time 1400061412.000000 for "u'May 14 10:56:52'" using template (?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?
2014-05-14 11:47:29,630 fail2ban.server.filter[23174]: DEBUG   Processing line with time:1400061412.0 and ip:1.2.3.5
2014-05-14 11:47:29,630 fail2ban.server.filter[23174]: INFO    [pam-generic] Found 1.2.3.5
</pre>

If I adds a date output to found line as in branch "strptime-bug", I see (+1 hour):

<pre>
2014-05-14 11:47:29,630 fail2ban.server.filter[23174]: INFO    [pam-generic] Found 1.2.3.5 - 2014-05-14 11:56:52,
</pre>

If I use "utcfromtimestamp" as suggested from @kwirk (in #710) the result is even worse (-1 hours):

<pre>
2014-05-14 11:47:29,630 fail2ban.server.filter[23174]: INFO    [pam-generic] Found 1.2.3.5 - 2014-05-14 09:56:52,
</pre>

Within my branch "strptime-bug" (#710) it seems to be correct.
The pull request was closed from @kwirk, I don't know how I can reopen it again.

PR #721, opened 2014-05-14
strptime bug

kwirk:
Thanks to @sebres with #720 for highlighting. Most changes are testcase reverts from #349 which was merged in #368. I believe this is now correct, but please give this thorough review :smile:


sebres:
`cherrypick-to-0.8` was a nice label (at its time) :smile:

PR #741, opened 2014-06-18
ASSP Filter Update

Silvenga:
Added/Updated detection of:

* RelayAttempt (Scanners looking for open relays, drop to reduce attack surface)
* Auth error (Brute Force attempts, drop to reduce attack surface)
* Too frequent connections (Seen when attempting to deliver SPAM or Brute Force, drop to reduce server load)

Regex lines 12-14 are untested with newest release of ASSP. I personally have not seen anything in my logs that is caught by the Regex (For ASSP1?). 

ASSP Homepage has been down for a while, falling back the Sourceforge project page.

Fixed spelling

I recommend setting the ban time for ASSP to 2+ hours.

### DIFF
--- a/config/filter.d/assp.conf
+++ b/config/filter.d/assp.conf
@@ -1,50 +1,34 @@
 # Fail2Ban filter for Anti-Spam SMTP Proxy Server also known as ASSP
 # 
-#    Honmepage:   http://sourceforge.net/projects/assp/
+#    Homepage:   http://sourceforge.net/projects/assp/
 #    ProjectSite: http://sourceforge.net/projects/assp/?source=directory
 #
 #
 
 [Definition] 
 
-__assp_actions = (?:dropping|refusing)
-
-failregex = ^(:? \[SSL-out\])? <HOST> max sender authentication errors \(\d{,3}\) exceeded -- %(__assp_actions)s connection - after reply: \d{3} \d{1}\.\d{1}.\d{1} Error: authentication failed: \w+;$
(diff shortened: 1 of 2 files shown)


leeclemens:
Agree that any failregex should be backwards compatible, and no sample log lines should necessarily ever be removed. Closing due to that, and this is almost a year old.

rhardy613:
ASSP V1 development stopped at the end of 2014 and it is now deprecated. All users were urged to upgrade to ASSP V2 which is still actively developed. For some reason fail2ban 0.9.5 (and trunk) still have code which only understands ASSP V1 logs. This patch seems to fix this and this ticket was closed and not commited. Please consider revisiting this security issue. I shouldn't be able to have ASSP protection enabled in fail2ban and have it completely miss brute force attacks. Which happens now with a current unpatched 0.9.5 fail2ban system.

rhardy613:
Correction this patch did not work during my tests. I've attached a patch against fail2ban 0.9.5 which makes fail2ban properly detect brute force with ASSP on my Ubuntu 16.04 systems:
[fix-asspV2-brute-force-741.diff.gz](https://github.com/fail2ban/fail2ban/files/392121/fix-asspV2-brute-force-741.diff.gz)

yarikoptic:
Would you mind sending it as a PR to ease review etc

rhardy613:
OK I will work on that. I don't use git for PRs very frequently so please be patient.
The patch has changed since that one anyways as there were more patterns to match.

PR #746, opened 2014-06-27
Please add one more date template for opentftpd-sp server logs

drom007:
Please add one more logs format for opentftpd-sp server logs. They looks like this: 
[25-Jun-14 18:09:41] Client xxxxxxxxx:pppp, Access violation


sebres:
Since we've an opportunity to provide `datepattern` in filter/jail, I see no benefits to extend default patterns with more formats (that are not used everywhere).
So just extend the filter with this parameter.

PR #751, opened 2014-07-03
Update postfix-sasl.conf: Different log output on centos.

Pixinger:
In our logfile in some cases the message is "...SASL login..." instead of "...SASL LOGIN...".
Maybe it makes sense to change the whole "SASL (?:LOGIN|login|PLAIN|(?:CRAM|DIGEST)-MD5)" block to "SASL.*", because when browsing through the internet I have seen other possibilities then "CRAM,DIGEST,..." but I'm not sure how strict the regex should be handled.
Running Centos 6 - happens only sometimes and only if the domainname to the ip is unknown for postfix.

### DIFF
--- a/config/filter.d/postfix-sasl.conf
+++ b/config/filter.d/postfix-sasl.conf
@@ -1,5 +1,8 @@
 # Fail2Ban filter for postfix authentication failures
 #
+# 04.07.2014: Pixinger77 modified regex because of undetected log message, where "LOGIN" was written in lower case as "login" - CentOS6.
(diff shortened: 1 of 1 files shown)


leeclemens:
The regex seems to match both LOGIN and login, so I am unsure what your specific issue is.

Before expanding the regex to match \S+, I'd be interested in seeing what log lines you have that are not being matched. Can you provide any? I'm guessing there is some performance impact by broadening the regex?

yarikoptic:
@Pixinger your input is awaited for to respond to https://github.com/fail2ban/fail2ban/pull/751#issuecomment-70437889

yarikoptic:
issue was addressed actually by bcfcefa203ee381446f3db143f63b2a7570c1deb already, so closing

PR #794, opened 2014-08-22
Removed explicit bantime specification

mrc0mmand:
Explicit bantime spec. is unnecessary since 3d7504c19eb04d4756808805ad34998ed4dbe8e6


sebres:
Since 8403eab2c4d71c8662a58ae5fda642aca473ba36 (0.10-incr-dynamic-bantime) merged into #1460 this is not relevant anymore.
I'll merge #1460 in 0.10 soon, because in-between it is stable enough.

PR #831, opened 2014-10-24
add zpanel-auth filter and jail

alexcreek:
I've added a filter for the zpanel web hosting panel login page.  When a login request fails, 'GET /?invalidlogin' is logged to the access log.  

In zpanel's default configuration, it's apache vhost doesn't specify logging so the requests are sent to /var/log/apache2/other_vhosts_access.log as a default action.  In this configuration, each log entry is prefixed with the respective vhost and listening port.  The failregex checks for the vhost:port prefix as well as the traditional apache combined format.

### DIFF
--- a/config/filter.d/zpanel-auth.conf
+++ b/config/filter.d/zpanel-auth.conf
@@ -1,0 +1,33 @@
+# Fail2Ban configuration file
+#
+# Author: Alex Creek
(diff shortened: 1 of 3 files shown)


yarikoptic:
ok -closing as wontfix then for now

PR #887, opened 2014-12-20
Update postfix-sasl.conf

syntacs:
the actual version doesnt work with postfix 2.11.0-1 (part of fully patched Ubuntu 14.04.1 LTS)
auth faiilures look like that:
Dec  7 11:34:38 deep postfix/smtpd[5551]: warning: unknown[75.127.12.202]: SASL LOGIN authentication failed: authentication failure

didn't do that an my own, credits to
http://www.howtoforge.com/forums/member.php?u=81691


syntacs:
I did some checks to be sure about minimal change - and found it working also with my log lines
-> ignore the pull request

PR #905, opened 2015-01-01
Firewallcmd-multiport.conf and Firewallcmd-allports.conf 

TorontoMedia:
cleaner pull requests

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -317,7 +317,7 @@ some obscure corner of the Internet.
 Many thanks to our contributors for this release Daniel Black, Yaroslav
 Halchenko, Steven Hiscocks, Mark McKinstry, Andy Fragen, Orion Poplawski,
 Alexander Dietrich, JP Espinosa, Jamyn Shanley, Beau Raines, François
-Boulogne and others who have helped on IRC and mailing list, logged issues
+Boulogne, Donald Yandt and others who have helped on IRC and mailing list, logged issues
 and bug requests.
 
 - IMPORTANT incompatible changes:
@@ -422,7 +422,8 @@ and bug requests.
    * filter.d/uwimap-auth - filter for uwimap-auth IMAP/POP server
   Steven Hiscocks and Daniel Black
    * filter.d/selinux-{common,ssh} -- add SELinux date and ssh filter
-
+  Donald Yandt
+   * action.d/firewallcmd-{multiport,allports}.conf --added two new actions
 - Enhancements:
   François Boulogne and Frédéric
    * filter.d/lighttpd - auth regexs for lighttpd-1.4.31
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,9 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      (opposite to simply set failregex/ignoreregex that overwrites it),
      see gh-867.
    - Monit config for fail2ban in /files/monit
+   - New actions:
+     - action.d/firewallcmd-multiport Thanks Donald Yandt
+     - action.d/firewallcmd-multiport Thanks Donald Yandt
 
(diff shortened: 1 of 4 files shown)


ChangeLog
```diff
@@ -422,7 +422,8 @@ and bug requests.
    * filter.d/uwimap-auth - filter for uwimap-auth IMAP/POP server
   Steven Hiscocks and Daniel Black
    * filter.d/selinux-{common,ssh} -- add SELinux date and ssh filter
-
+  Donald Yandt
+   * action.d/firewallcmd-{multiport,allports}.conf --added two new actions
```
yarikoptic: I am puzzled -- why are you changing changelog for the very old release? ;)  modify the one on top of the file for 0.9.2, not in the middle ;-)

config/action.d/firewallcmd-allports.conf
```diff
@@ -0,0 +1,52 @@
...
+
+actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -j f2b-<name>
+             firewall-cmd --direct --remove-rules ipv4 filter f2b-<name>
+             firewall-cmd --direct --remove-chain ipv4 filter f2b-<name>
+
+
+# Note:  uses regular expression word boundaries '\b' 
+# Example actioncheck: firewall-cmd --direct --get-chains ipv4 filter | grep -q '\bf2b-recidive\b'
+
+actioncheck = firewall-cmd --direct --get-chains ipv4 filter | grep -q '\bf2b-<name>\b'
```
yarikoptic: I guess ```\b``` is not quite proper here:
```
> /bin/ls | grep '\bfail2ban-regex\b' 
fail2ban-regex-0.9
fail2ban-regex-oldmaster
```
if it is at the end of the line in the output of firewall-cmd, previously used '$' was a better choice.  So why not just to make it ```\sf2b-<name>$```?

yarikoptic: btw -- just a reminder -- no need for a new PR. just add commit with fixes and push to the same branch.  Cheers!

ChangeLog
TorontoMedia: Sorry, I have a serious flu right now so I'm not thinking straight. I'll try and get everything straightened out.

config/action.d/firewallcmd-allports.conf
TorontoMedia: Okay I wasn't sure exactly when actioncheck was called but if it's only called after adding a chain then '\s' & '$' would be better suited. I was using firewallcmd and grep -c to come up with a regexp. 

Also the reason I tried to clean up a bit was because I'm not using Git and I ended up with 4 patches since I'm not use to the web api...so please bear with me lol.

ChangeLog
```diff
@@ -30,6 +30,9 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      (opposite to simply set failregex/ignoreregex that overwrites it),
      see gh-867.
    - Monit config for fail2ban in /files/monit
+   - New actions:
+     - action.d/firewallcmd-multiport Thanks Donald Yandt
```
yarikoptic: I guess one of them must be allports...  could be listed in a single entry,  with a single thanks  :-)

TorontoMedia: Sorry, I shouldn't be on a computer right now lol.

TorontoMedia:
Is there something holding my PR back that I need to fix?

yarikoptic:
nothing holding,  just fell off the  radar,  thanks for the buzz

TorontoMedia:
No problem :)

TorontoMedia:
Also thank you for merging my PR.

yarikoptic:
Thank you for contributing!

Issue #755, opened 2014-07-04
Error in FilterPyinotify


ghost:
Hello!
Iam using here Ubuntu 14.04 (x64) and Fail2ban 0.8.13-1. Since a while I get the error: 

fail2ban.filter [22302]: ERROR   Error in FilterPyinotify callback: _strptime_time

I really dont have a idea what to do.
Thanks.

leeclemens:
Pull request: https://github.com/fail2ban/fail2ban/pull/906

playeren:
I just installed fail2ban on a Debian 8.6 with apt-get. I got fail2ban v0.8.13-1 with python-pyinotify v0.9.4-1, and got the `Error in FilterPyinotify callback: 'module' object has no attribute '_strptime_time'` in the fail2ban.log.

Did I get an older version without the fix, or am I doing something wrong?

I can make the error go away by setting backend=polling in jail.{conf|local}

szepeviktor:
BTW you may download Fail2ban from sid as it is an `-all` package.
https://packages.debian.org/sid/fail2ban

playeren:
Thanks @szepeviktor but I'm a bit worried about the hassle of keeping everything updated when mixing stable packages with SID and backports. I'm a wuss :) I'll stick to 'polling' for now, and wait until the fix hits me from a normal upgrade.

chdh:
I also got `Error in FilterPyinotify callback: 'module' object has no attribute '_strptime_time'` on Debian 8.7 + Jessie Backports (fail2ban 0.8.13-1, python-pyinotify 0.9.5-1) and had to set `backend = polling`.

NET-057:
Did all these changes [bug fixed](https://github.com/fail2ban/fail2ban/pull/906/files/d0bcacd97be8d2eac24c048d0f58b5b93ee90340#r22415265) . now that error is fix but not blocking any ip.

szepeviktor:
@NET-057 Are you sure it is not blocking in the last 4 years?

NET-057:
@szepeviktor 3 days ago. i installed fail2ban (v0.8.14) from [Link](https://www.fail2ban.org/wiki/index.php/Downloads). it works fine for CentOS. but when i tried it with ubuntu 14.04 it shows `ERROR Error in FilterPyinotify callback: _strptime_time`. during log parse  

i added 
`from time import strptime
strptime("2012", "%Y")` 

[...]

PR #906, opened 2015-01-02
Fix strptime thread safety issue

leeclemens:
strptime thread safety hack-around - http://bugs.python.org/issue7980

### DIFF
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -430,6 +430,9 @@ class ServerExecutionException(Exception):
 	pass
 
 if __name__ == "__main__": # pragma: no cover - can't test main
+	from time import strptime
+	# strptime thread safety hack-around - http://bugs.python.org/issue7980
+	strptime("2012", "%Y")
 	client = Fail2banClient()
 	# Exit with correct return value
 	if client.start(sys.argv):
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -430,9 +430,6 @@ class ServerExecutionException(Exception):
 	pass
 
 if __name__ == "__main__": # pragma: no cover - can't test main
-	from time import strptime
-	# strptime thread safety hack-around - http://bugs.python.org/issue7980
-	strptime("2012", "%Y")
 	client = Fail2banClient()
 	# Exit with correct return value
 	if client.start(sys.argv):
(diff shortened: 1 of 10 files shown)


leeclemens:
Closes Issue https://github.com/fail2ban/fail2ban/issues/755

bin/fail2ban-client
```diff
@@ -430,6 +430,9 @@ class ServerExecutionException(Exception):
 	pass
 
 if __name__ == "__main__": # pragma: no cover - can't test main
+	from time import strptime
+	# strptime thread safety hack-around - http://bugs.python.org/issue7980
+	strptime("2012", "%Y")
```
yarikoptic: ha -- I think I had considered a similar hack at some point but don't remember why I haven't PRed it

shouldn't it be sufficient just to do once in ```fail2ban/__init__.py```.  that one I believe is imported by all the tools and we don't thread in the tools explicitely

leeclemens:
My understanding is that it is best to call it as soon as possible - always before any other thread calls (possibly imports?) strptime - so I felt three separate calls in each of the __main__ sections made sense, otherwise I would agree a centralized common point of execution would be better.

yarikoptic:
the irony is that they first import fail2ban, so ```__init__``` would take effect before running ```__main__``` ;-) so ```__init__.py``` will result in the eariiest point ;)

leeclemens:
I misread what you said as putting it into `def __init__` for each, rather than `__init__.py`. I haven't experimented with that approach. Given the bug is fairly difficult to reproduce (race condition) and I have experienced this method consistently working, I chose to use it. If you think putting it into `__init__.py` would be better, I will happily move it.

yarikoptic:
I do think that 1 place is better than 3 copies.
But I can't bloody recall why I haven't fixed it myself, although I remember looking into it... May be I just could not reproduce it and postponed

leeclemens:
I think that should do it

yarikoptic:
good.  changelog?  :-)

leeclemens:
Kind of messy, want me to delete branch and re-submit as one commit? (Not sure how to merge after I pushed to my remote/origin)

yarikoptic:
yeah -- that was a long way -- but it is the end what matters most ;)  cheers -- merging!

PR #915, opened 2015-01-19
port[s] typo fixed in jail.conf/nginx-http-auth, issue gh-913

### LINKED ISSUE #913: Config error in nginx-http-auth jail
The definition of the `nginx-http-auth` jail in `jail.conf` is slightly wrong. This is due to a wrong port definition.
The keyword has to be `port` instead of `ports`, which leads to blocking of all access to the system instaed of just `http,https`

sebres:
closes #913

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
 -----------
 
 - Fixes:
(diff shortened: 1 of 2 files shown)


yarikoptic:
if you have time it would be nice to create a unittest  for checking that no ports  defined in any jail....  Sure they could theoretically be there since it is just a  variable used  in action  but  at current moment  we don't have them...  merging this one now

PR #917, opened 2015-01-20
infinite busy loop on _escapedTags match in substituteRecursiveTags gh-907

### LINKED ISSUE #907: Infinite loop on _escapedTags match in substituteRecursiveTags
Line 399 of server/action.py continues if an escaped tag matches; but this just repeats the surrounding 'while m:' loop without updating m. It therefore loops forever. This prevents using the blocklist_de action on Fedora 21, as it includes <matches> in the actionban string.

This patch seems to work for me:
```
------------------------------------------------------
commit c86c6b16e548911339ba3333d5242979cdb2f8a5
Author: Tony Thompson <athompson@samarc.org.uk>
Date:   Sun Jan 4 16:28:14 2015 +0000

    Step to next on _escapedTag match
    
    Previous code just repeated the while() loop with no change to the
    condition, and so looped forever. Step to the next match instead.

diff --git a/fail2ban/server/action.py b/fail2ban/server/action.py
index da7517f..9d8e9c8 100644
--- a/fail2ban/server/action.py
+++ b/fail2ban/server/action.py
@@ -395,8 +395,8 @@ class CommandAction(ActionBase):
                                        #logSys.log(5, 'recursion fail tag: %s value: %s' % (tag, value) )
                                        return False
                                elif found_tag in cls._escapedTags:
-                                       # Escaped so won't match
-                                       continue
+                                       # Escaped so won't match; move to next match
+                                       m = t.search(value, m.start() + 1)
                                else:
                                        if tags.has_key(found_tag):
                                                value = value.replace('<%s>' % found_tag , tags[found_tag])
------------------------------------------------------
```
Patch is available in branch master of https://github.com/TonyThompson/fail2ban-patch.git, if this looks like an appropriate fix.

sebres:
closes #907

### DIFF
--- a/fail2ban/server/action.py
+++ b/fail2ban/server/action.py
@@ -394,20 +394,16 @@ class CommandAction(ActionBase):
 					# recursive definitions are bad
 					#logSys.log(5, 'recursion fail tag: %s value: %s' % (tag, value) )
 					return False
-				elif found_tag in cls._escapedTags:
-					# Escaped so won't match
+				if found_tag in cls._escapedTags or not tags.has_key(found_tag):
+					# Escaped or missing tags - just continue on searching after end of match
+					# Missing tags are ok - cInfo can contain aInfo elements like <HOST> and valid shell
+					# constructs like <STDIN>.
+					m = t.search(value, m.end())
 					continue
-				else:
-					if tags.has_key(found_tag):
-						value = value.replace('<%s>' % found_tag , tags[found_tag])
-						#logSys.log(5, 'value now: %s' % value)
-						done.append(found_tag)
-						m = t.search(value, m.start())
-					else:
-						# Missing tags are ok so we just continue on searching.
-						# cInfo can contain aInfo elements like <HOST> and valid shell
-						# constructs like <STDIN>.
-						m = t.search(value, m.start() + 1)
+				value = value.replace('<%s>' % found_tag , tags[found_tag])
+				#logSys.log(5, 'value now: %s' % value)
+				done.append(found_tag)
+				m = t.search(value, m.start())
 			#logSys.log(5, 'TAG: %s, newvalue: %s' % (tag, value))
 			tags[tag] = value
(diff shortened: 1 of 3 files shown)


fail2ban/server/action.py
```diff
@@ -394,20 +394,16 @@ def substituteRecursiveTags(cls, tags):
 					# recursive definitions are bad
 					#logSys.log(5, 'recursion fail tag: %s value: %s' % (tag, value) )
 					return False
-				elif found_tag in cls._escapedTags:
-					# Escaped so won't match
+				if found_tag in cls._escapedTags or not tags.has_key(found_tag):
+					# Escaped or missing tags - just continue on searching after end of match
+					# Missing tags are ok - cInfo can contain aInfo elements like <HOST> and valid shell
+					# constructs like <STDIN>.
+					m = t.search(value, m.end())
```
yarikoptic: why ```m.end``` and not ```m.start()+1```  - couldn't it be some obscure "embbeded" tag e.g. "<<MATCH_ME>OST>" where we would define ```MATCH_ME``` to be either H or IPV6H and assume that it would work? ;-)  (just hypothesising I guess since we seems to have such unittests)

sebres: such doubtful configurations should be avoided, imho.

sebres: and it's faster - start/reload of my f2b (with ca. 20 jails) improved up to 0.5 second faster.

yarikoptic: >    and it's faster - start/reload of my f2b (with ca. 20 jails) improved up
>    to 0.5 second faster.

may be just by missing something it should've not missed? ;)  I think
such a speedup should still be in place even if you revert only this one
to the start+1

sebres: whatever, your example is anyway not correct, because the regular expression `r'<([^ >]+)>'` will prevent using "embedded" tags, like you specified...

PR #918, opened 2015-01-20
Issue/161

### LINKED ISSUE #161: asyncore: (9, 'Bad file descriptor')
```
Testing using /usr/bin/python2.6
Fail2ban 0.8.8 test suite. Python 2.6.8 (unknown, Jan 26 2013, 14:35:25) [GCC 4.7.2]. Please wait...
.........................................Exception in thread Thread-61:
  File "/usr/lib/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.6/threading.py", line 484, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/yoh/deb/gits/fail2ban/server/asyncserver.py", line 144, in start
    asyncore.loop(use_poll = False) # fixes the "Unexpected communication problem" issue on Python 2.6 and 3.0
  File "/usr/lib/python2.6/asyncore.py", line 210, in loop
    poll_fun(timeout, map)
  File "/usr/lib/python2.6/asyncore.py", line 140, in poll
    r, w, e = select.select(r, w, e, timeout)
error: (9, 'Bad file descriptor')

.....................................F....
```

Never ran into this one before (or at least not for a while ;) )

sebres:
closes #161
workaround for the "Bad file descriptor" issue on Python 2.7, gh-161 : asyncore.loop() using poll
by the way, prevents to write "'build/bdist.linux-x86_64' does not exist -- can't clean it" into stderr;

### DIFF
--- a/fail2ban/server/asyncserver.py
+++ b/fail2ban/server/asyncserver.py
@@ -149,8 +149,11 @@ class AsyncServer(asyncore.dispatcher):
 		self.__init = True
 		# TODO Add try..catch
 		# There's a bug report for Python 2.6/3.0 that use_poll=True yields some 2.5 incompatibilities:
-		logSys.debug("Detected Python 2.6 or greater. asyncore.loop() not using poll")
-		asyncore.loop(use_poll=False) # fixes the "Unexpected communication problem" issue on Python 2.6 and 3.0
+		if sys.version_info >= (2, 7) and sys.version_info < (2, 8): # if python 2.7 ...
+			logSys.debug("Detected Python 2.7. asyncore.loop() using poll")
+			asyncore.loop(use_poll=True) # workaround for the "Bad file descriptor" issue on Python 2.7, gh-161
+		else:
+			asyncore.loop(use_poll=False) # fixes the "Unexpected communication problem" issue on Python 2.6 and 3.0
 	
 	##
 	# Stops the communication server.
--- a/fail2ban/tests/misctestcase.py
+++ b/fail2ban/tests/misctestcase.py
@@ -113,7 +113,7 @@ class SetupTest(unittest.TestCase):
 			# clean up
 			shutil.rmtree(tmp)
 			# remove build directory
-			os.system("%s %s clean --all >/dev/null"
+			os.system("%s %s clean --all >/dev/null 2>&1"
 					  % (sys.executable, self.setup))
 
 class TestsUtilsTest(unittest.TestCase):


sebres:
sorry, PR #917 has been also in my master (checked in together).

sebres:
sys.version_info is a tuple in 2.6, also recommited.

fail2ban/server/asyncserver.py
```diff
@@ -149,8 +149,11 @@ def start(self, sock, force):
 		self.__init = True
 		# TODO Add try..catch
 		# There's a bug report for Python 2.6/3.0 that use_poll=True yields some 2.5 incompatibilities:
-		logSys.debug("Detected Python 2.6 or greater. asyncore.loop() not using poll")
-		asyncore.loop(use_poll=False) # fixes the "Unexpected communication problem" issue on Python 2.6 and 3.0
+		if sys.version_info >= (2, 7) and sys.version_info < (2, 8): # if python 2.7 ...
+			logSys.debug("Detected Python 2.7. asyncore.loop() using poll")
+			asyncore.loop(use_poll=True) # workaround for the "Bad file descriptor" issue on Python 2.7, gh-161
```
yarikoptic: could you check the history/log -- I remember we explicitly switched for ```use_poll=False``` to overcome some other problem.  so we better verify that we are not raising an old devil here

yarikoptic: I have reread #161 ... may be indeed it would be feasible to workaround by simply making it 2.7-specific BUT I am not sure if we wouldn't hit the same issue with e.g. popping out "fail2ban.server : ERROR Unexpected communication error".  I guess we would just need to test-drive it with recent Python's  @fail2ban/developers -- additional opinion?

sebres: I understand that "unexpected communication problem" previously occurred only in 2.6.
Possibly it was already fixed in other versions?
Other workarround may be a try/catch around, because according to <a href="http://bugs.python.org/issue10878">issue10878</a> it occurred in `asyncore` by `close`. But imho it is not so pretty...

yarikoptic: >    I understand that "unexpected communication problem" previously occurred
>    only in 2.6.

that is where it was detected... I don't remember if anyone was already
running 2.7 at that point.

>    Possibly it was already fixed in other versions?

who knows? ;) could be

>    Other workarround may be a try/catch around, because according to
>    issue10878 it occurred in asyncore by close. But imho it is not so
>    pretty...

I think I have tried something like that at some point but didn't
succeed to full solution

yarikoptic:
Ok -- I will swallow the bullet and will merge it as soon as ChangeLog entry describing this appears... If anything -- change is minimal so we could revert it if side-effects appear

yarikoptic:
Ok -- I will merge it and describe in ChangeLog myself ;)

PR #919, opened 2015-01-20
Issue/907 + embedded

sebres:


fail2ban/server/action.py
```diff
@@ -378,38 +378,43 @@ def substituteRecursiveTags(cls, tags):
...
+					value = value.replace('<%s>' % found_tag , tags[found_tag])
+					#logSys.log(5, 'value now: %s' % value)
+					done.append(found_tag)
+					m = t.search(value, m.start())
+				#logSys.log(5, 'TAG: %s, newvalue: %s' % (tag, value))
+				# was substituted?
+				if tags[tag] != value:
+					# check again later if embedded-recursive substitution:
+					if value.startswith('<') and value.endswith('>'):
+						repFlag = True
```
yarikoptic: I guess I haven't yet grasped the full flow so decided to test more
https://github.com/sebres/fail2ban/pull/2
and it passed... hm... ;-)

yarikoptic:
if no objections or further recommendations from other @fail2ban/contributors -- I will merge this one in later on

fail2ban/server/action.py
```diff
@@ -378,38 +378,43 @@ def substituteRecursiveTags(cls, tags):
...
-				found_tag = m.group(1)
-				#logSys.log(5, 'found: %s' % found_tag)
-				if found_tag == tag or found_tag in done:
-					# recursive definitions are bad
-					#logSys.log(5, 'recursion fail tag: %s value: %s' % (tag, value) )
-					return False
-				elif found_tag in cls._escapedTags:
+		t = re.compile(r'<([^ <>]+)>')
+		while True:
+			repFlag = False
```
yarikoptic: btw -- should it may be be within 'for tag' loop?  could you provide brief docstring outlining its meaning/purpose?

leeclemens:
Was master/HEAD ever tested in either related Issue (#907 and #885)? It is unclear to me (after admittedly basic review) how the `while m:` loop was identified to be the culprit (strace?). This behavior seems localized to a specific version given the downgrade resolving the issue in Issue #885, but I haven't seen any notes of commits where the issue was introduced (between those versions). Just my two cents.

sebres:
<blockquote>how the `while m`: loop was identified to be the culprit</blockquote>@leeclemens By the issue #907 was the `continute` without moving `m` further (if ignore tag found): so endless loop.

fail2ban/server/action.py
sebres: Nop, it should be within `while`, because that is a flag to repeat substitution again - if anything was replaced and we have still tags in value.

yarikoptic:
ok.  no major objections, tests still pass and were extended -- sounds like a win/win -- merging! Thanks @sebres !

sebres:
Thanks @yarikoptic, especially for his great test cases >)

yarikoptic:
:clap: :clap: :clap: :bouquet:

sebres:
@yarikoptic хотя думается мне, в свое время мы за это (embedded after embedded subst) еще огребем по полной программе.

PR #924, opened 2015-01-23
Add extended info to status output using Cymru

leeclemens:
Add new `fail2ban-client status <JAIL>` optional argument, [FLAVOR].
'basic' performs same behavior.
Implemented 'cymru' flavor to query Cymru's DNS API to gather ASN, Country and RIR data for each banned IP using dnspython.


fail2ban/server/banmanager.py
```diff
@@ -26,6 +26,9 @@
 
 from threading import Lock
 
+import dns.exception
```
yarikoptic: what about making this all optional, e.g. delay imports till the call, and then if not available -- spitting out informative message and then report old style limited output?

fail2ban/tests/banmanagertestcase.py
```diff
@@ -39,19 +38,58 @@ def setUp(self):
...
+		self.__ban_ip = "93.184.216.34"
+		self.__asn = "15133"
+		self.__country = "EU"
+		self.__rir = "ripencc"
+		self.__ticket = BanTicket(self.__ban_ip, 1167605999.0)
+		self.__banManager = BanManager()
+		self.assertTrue(self.__banManager.addBanTicket(self.__ticket))
+
+	def tearDown(self):
+		"""Call after every test case."""
```
yarikoptic:
```Python
pass
```
?

fail2ban/tests/utils.py
```diff
@@ -107,6 +107,7 @@ def addTest(self, suite):
 	tests.addTest(unittest.makeSuite(failmanagertestcase.AddFailure))
 	# BanManager
 	tests.addTest(unittest.makeSuite(banmanagertestcase.AddFailure))
+	tests.addTest(unittest.makeSuite(banmanagertestcase.StatusExtendedCymruInfo))
```
yarikoptic: Let's condition on having that external dependency present?

ChangeLog
```diff
@@ -41,7 +41,11 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    - Monit config for fail2ban in /files/monit
    - New actions:
      - action.d/firewallcmd-multiport and action.d/firewallcmd-allports Thanks Donald Yandt
-     
+   - New status commands:
+     - fail2ban-client status <jail> extended
+       - prints Cymru data (ASN, Country RIR) per banned IP
+       - Requires dnspython or dnspython3
```
yarikoptic: README.md should get that optional (see below) dependency listed

fail2ban/server/banmanager.py
leeclemens: Informative message in the output, like "Banned ASN list: Failed to import dnspython" ? or just in the log and have it print out an empty list?

fail2ban/tests/banmanagertestcase.py
leeclemens: copied from AddFailure :)

fail2ban/server/banmanager.py
yarikoptic: >  > +import dns.exception
>    Informative message in the output, like "Banned ASN list: Failed to import
>    dnspython" ? 

something like

Banned ASN list:  [please install dnspython for this functionality]

fail2ban/tests/banmanagertestcase.py
yarikoptic: >  > +     def tearDown(self):
>  > +             """Call after every test case."""

>    copied from AddFailure :)

surprised that it works... may be  it was a while back when it wouldn't
work -- I am that old now ;)

fail2ban/server/banmanager.py
leeclemens: The only issue with that approach is that with a munin-plugin (or other external program) parsing that data would be misleading. It is human useful, but blending the error message into the output isn't computer readable friendly imho

fail2ban/tests/banmanagertestcase.py
leeclemens: I like having the explicit `pass` better, added to both instances

fail2ban/tests/utils.py
leeclemens: Agreed. It is installed for Travis tests, but added import test around it. a80ff26e151c295bb5cc1661fa5402c0115a8e97

ChangeLog
leeclemens: Agreed, I thought there was a better place to add dependency info :)  8fd9d6a770bdfc90d59da6c1493d82e0b1a2752e

fail2ban/server/banmanager.py
leeclemens: What if a single "error" was returned for each? That could be more evident (munin graph with only "error" being printed), then check logs and see the module/import error

README.md
```diff
@@ -33,6 +33,7 @@ Optional:
   - Linux >= 2.6.13
 - [gamin >= 0.0.21](http://www.gnome.org/~veillard/gamin)
 - [systemd >= 204](http://www.freedesktop.org/wiki/Software/systemd)
+- [dnspython or dnspython3](http://www.dnspython.org/)
```
yarikoptic: I guess no need to specify version specific one since we don't do for the other ones (e.g. pyinotify)

fail2ban/server/banmanager.py
yarikoptic: well, still would be lovely to spit out somehow (stderr) and informative message to the user, and then yes embed error into "Banned ASN list"

fail2ban/protocol.py
```diff
@@ -55,6 +55,7 @@
 ["start <JAIL>", "starts the jail <JAIL>"], 
 ["stop <JAIL>", "stops the jail <JAIL>. The jail is removed"], 
 ["status <JAIL>", "gets the current status of <JAIL>"],
+["status <JAIL> extended", "gets the current status of <JAIL> with extended info"],
```
yarikoptic: this extended status, is not just 'extended' it is specific to cymru, right?

I just wonder if it needs to be spelled out as such (e.g. s/extended/cymru/ throughout) since someone might come up with another "extended" status output based on another service

leeclemens: That is fair, I was thinking extended would be any additional info, and someone could add additional output to the same 'extended' status from any source (all calls to Cymru are made once, then calls to parse the data). Therefore someone could add GeoIP sourced data with a method `geBanListExtendedCountryGeoIP` and add that to the existing `statusExtended` method's return.

README.md
leeclemens: I wanted to, but I'm thinking the ability to query a text record has always been present, so I didn't want to specify a more recent version without verifying.

leeclemens:
I believe all comments have been addressed or responded to:

Only outstanding may be related to having fail2ban-client status <JAIL> extended <EXT_SOURCE> to return Cymru data and potentially other source data in the future. 

My reply: `statusExtended` can be updated in the future to include other data, if more sources become useful, this can be changed - otherwise they can be called from `statusExtended` together.

README.md
yarikoptic: I just meant that it is sufficient to list dnspython without dnspython3

fail2ban/protocol.py
yarikoptic: I am still toying with this idea... so theoretically original status is just a 'basic' status and 'extended' one should output 'basic' + 'everything else we can tell' (which now would be cymru).  With current implementation it would be only cymru.  What if instead of adding a new public method per each such additional status info it would be just a kwarg "flavor" which could be "basic" (default), "extended", "cymru" and then implemented within ```status```?

btw -- could you provide an example of current output?

leeclemens: That was what I was originally thinking, but I like the idea of generalizing/abstracting it more.

Here is a sample output with some randomly generated IPs. RFC 1918 and multicast, etc come back with NXDOMAIN.

https://gist.github.com/anonymous/e09a9328564b30b555a6

README.md
leeclemens: Ahh, understood

fail2ban/protocol.py
leeclemens: Also to keep in mind, my ultimate goal is to have fail2ban-client status <JAIL> ??? mapped to munin graphs, which need to know how to call fail2ban-client as well as which line to parse (ASN, Country, etc). One use-case may be with plugin names: fail2ban_basic (current status, counts per jail), fail2ban_cyrmu_asn, fail2ban_cyrmu_rir, fail2ban_XXX_YYY - where munin calls fail2ban-client status <JAIL> XXX and parses lines dealing with YYY (code determines)

yarikoptic: I left a comment in that gist but I guess it doesn't announce on those:

    should unique names be reported instead? May be with counts

    side note: we should really come up with Json output some time

yarikoptic:
@leeclemens thanks again and I don't mind merging it in this state.  But still wonder, because of it being used by external tools, if we should converge on optimal interface (extendable for future additions) and format (unique or not, json?) first

fail2ban/protocol.py
leeclemens: I was thinking having the values all line up with each other, so someone could tie them together if they wanted to (match IP to RIR from parsing the output).

I think an overhaul/expansion of `fail2ban-client status` would be a good thing, but I'd rather not expand this enhancement to include it (which seems okay with you from your latest comment).

leeclemens:
I was thinking of the kwarg approach you mentioned, defaulting to 'basic'. Actually, the munin plugin I have working relies on that (just hardcoded at this point to use 'extended') - not that I couldn't change it.

And as I mentioned in the line comment a few minutes ago, later I can work on a more robust status interface overall. (Any thoughts on timeline of next release?)

yarikoptic:
next release -- https://github.com/fail2ban/fail2ban/milestones/0.9.2 looks searingly full.  So I guess we should just move of those not to be closed issues toward 0.9.3 and have 0.9.2 cut now with already enough number of fixes and enhancements

leeclemens:
Cool, I have the flavor stuff working, just working through some merge/commit/push issues

leeclemens:
okay, I think that should do it?

yarikoptic:
looks lovely,  Travis is happy so what to wait for?  :-)

leeclemens:
Beautiful! I'm already starting to think of ways to fixup the entire 'status' stack :)

sebres:
wtf going on?!

yarikoptic:
Do you mean a coverage report on a merged PR? Don't know... Just ignore I guess

yarikoptic:
I would be OK to disable coveralls and leave just codecov

sebres:
Sometimes it is very helpfull, and both play up from time to time, so let him still be...

PR #925, opened 2015-01-26
add nsupdate action

rumple010:
Adds a new action file that uses nsupdate to dynamically update a BIND
zone file with a TXT resource record representing a banned IP address.
Resource record is deleted from the zone when the ban expires.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -41,6 +41,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    - Monit config for fail2ban in /files/monit
    - New actions:
      - action.d/firewallcmd-multiport and action.d/firewallcmd-allports Thanks Donald Yandt
+     - action.d/nsupdate Thanks Andrew St. Jean
      
 - Enhancements:
    * Enable multiport for firewallcmd-new action.  Closes gh-834
--- a/config/action.d/nsupdate.conf
+++ b/config/action.d/nsupdate.conf
@@ -20,6 +20,9 @@
 # keyfile	Full path to TSIG key file used for authentication between
 #		nsupdate and BIND.
 #
+# Create an nsupdate.local to set at least the <domain> and <keyfile>
+# options as they don't have default values.
+#
 # The ban and unban commands assume nsupdate will authenticate to the BIND
 # server using a TSIG key. The full path to the key file must be specified
 # in the <keyfile> parameter. Use this command to generate your TSIG key.
--- a/config/action.d/nsupdate.conf
+++ b/config/action.d/nsupdate.conf
@@ -1,0 +1,110 @@
+# Fail2Ban configuration file
+#
+# Author: Andrew St. Jean
+#
+# Use nsupdate to perform dynamic DNS updates on a BIND zone file.
+# One may want to do this to update a local RBL with banned IP addresses.
+#
+# Options
+#
+# domain	DNS domain that will appear in nsupdate add and delete
+#		commands.
+#
+# ttl		The time to live (TTL) in seconds of the TXT resource
+#		record.
+#
+# rdata		Data portion of the TXT resource record.
+#
+# nsupdatecmd	Full path to the nsupdate command.
+#
+# keyfile	Full path to TSIG key file used for authentication between
+#		nsupdate and BIND.
+#
+# The ban and unban commands assume nsupdate will authenticate to the BIND
+# server using a TSIG key. The full path to the key file must be specified
+# in the <keyfile> parameter. Use this command to generate your TSIG key.
+#
+# dnssec-keygen -a HMAC-MD5 -b 256 -n HOST <key_name>
+#
+# Replace <key_name> with some meaningful name.
+#
+# This command will generate two files. Specify the .private file in the
+# <keyfile> option. Note that the .key file must also be present in the same
+# directory for nsupdate to use the key.
+#
+# Don't forget to add the key and appropriate allow-update or update-policy
+# option to your named.conf file.
+#
+
+[Definition]
+
+# Option:  actionstart
+# Notes.:  command executed once at the start of Fail2Ban.
+# Values:  CMD
+#
+actionstart =
+
+
+# Option:  actionstop
+# Notes.:  command executed once at the end of Fail2Ban
+# Values:  CMD
+#
+actionstop =
+
+
+# Option:  actioncheck
+# Notes.:  command executed once before each actionban command
+# Values:  CMD
+#
+actioncheck =
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionban = echo <ip> | awk -F. '{print "prereq nxrrset "$4"."$3"."$2"."$1".<domain> TXT"; print "update add "$4"."$3"."$2"."$1".<domain> <ttl> IN TXT \"<rdata>\""; print "send"}' | <nsupdatecmd> -k <keyfile>
+
+# Option:  actionunban
+# Notes.:  command executed when unbanning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionunban = echo <ip> | awk -F. '{print "update delete "$4"."$3"."$2"."$1".<domain>"; print "send"}' | <nsupdatecmd> -k <keyfile>
+
+[Init]
+
+# Option:  domain
+# Notes.:  DNS domain that nsupdate will update.
+# Values:  STRING
+#
+domain = 
+
+# Option:  ttl
+# Notes.:  time to live (TTL) of TXT resource record added by nsupdate.
+# Values:  NUM
+#
+ttl = 600
+
+# Option:  rdata
+# Notes.:  data portion of the TXT resource record added by nsupdate.
+# Values:  STRING
+#
+rdata = Your IP has been banned
+
+# Option:  nsupdatecmd
+# Notes.:  specifies the full path to the nsupdate program that dynamically
+#          updates BIND zone files.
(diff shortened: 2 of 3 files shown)


yarikoptic:
Looks neat... I have no way to test it since do not host bind DNS servers locally. Anyone of @fail2ban/contributors have had experience?
meanwhile -- could you add a ChangeLog entry and yourself to THANKS?

buanzo:
I will test it.

rumple010:
I'm using it on a FreeBSD 9.2 box so testing on another platform will be the most helpful.

leeclemens:
It may be worth noting in the comments that a .local file will be required to set required with no default options (at least): domain and keyfile. For package management issues, you won't want to update the .conf file directly.

Where as:
/etc/fail2ban/action.d/nsupdate.local

[Init]
domain = example.com
keyfile = /root/nsupdate.key

leeclemens:
I think the default TTL should be very low, to avoid caching past bantime (think querying a second before bantime expires, and then having that query cached for an additional 'bantime'). Not sure how to dynamically set the TTL, since it is relative to the external query, not it being set.

szepeviktor:
Could you explain how to use the updated BIND zone in another fail2ban installation?

rumple010:
Lee,

A TTL of 10 minutes doesn't seem excessively long to me. Worst case scenario is that an IP is banned for a maximum of 20 minutes. 
The flip side is that I don't want the TTL to be too low. If it is very low and a malicious IP keeps showing up in the log fail2ban is monitoring, it could lead to excessive DNS updates to BIND.
Having said that, perhaps I should just remove the default value for TTL altogether and let users decide what is an appropriate value.

rumple010:
szepeviktor,

I'm not sure what you mean by 'use the updated BIND zone in another fail2ban installation'.

I'm using nsupdate to dynamically update a local RBL DNS domain. I run a mail server that uses SMTP Auth to allow users to send email to the world. Every so often, some IP out there runs a dictionary attack against my mail server trying to guess a correct username and password so they can send spam through my server. These authentication failures show up in a log file. Fail2ban monitors this log file. When too many auth failures show up from a single IP, the nsupdate action fires and creates a DNS TXT record in my RBL. My mail server checks the RBL whenever an IP connects to it over SMTP. If the source IP appears in the RBL, the connection is dropped. Once the bantime as defined in my jail.local file expires, the TXT record is removed from the RBL and the IP can connect to my mail server once more. Until it again causes too many auth failures, that is.

szepeviktor:
Thank you! This is nice solution.
I think the [postfix filter](https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/postfix.conf) with an iptables action serves the same purpose providing you have only one server.

I am coming from the *fail2ban network* issue and looking for a way to distribute (any kind of) bans across multiple servers.

buanzo:
Indeed, we could start a collaborative DNSBL with this.

szepeviktor:
How would ban IP-s on other servers by a DNSBL?
Do you propose to do the banning (make DNSBL queries) by different dameons e.g. ssh, postfix, apache?
I'd prefer to communicate from fail2ban to fail2ban.

buanzo:
You can already do that using custom actions, a jail for fail2ban.log
itself and a secure means to transport ban details such as authenticated
https, a messaging system for clusters, etc. the recwiving system qould
have a jail for a log file creates by the messaging system.

Feeding a dnsbl using fail2ban so lots of services could benefit drom it is
scalable.

leeclemens:
@rumple010 The default bantime is 10 minutes, so banning an IP for an additional 10 minutes is 100% longer than configured/expected (relative duration being my concern). Also, since it is a ban action, it would only be updating DNS when it is banned - not for subsequent failed attempts, so the TTL would not be related to any 'excessive DNS updates'.

rumple010:
I didn't realize that the default ban time was 10 minutes. What would you suggest is a reasonable default TTL?

leeclemens:
It was a loaded comment, and is a loaded question...here's a reference:

http://freedns.afraid.org/faq/

They set the TTL to 60 seconds for dynamic DNS. Arguments regarding ease of cache poisoning suggest longer, I've seen 1 minute and 30 seconds recommended when using RR (zytrax). It's a debated topic.

RFC 2181 clarifies that the TTL value may have "a minimum value of 0, and a maximum value of 2147483647".

I think a better understanding of what may be querying the DNS records would dictate this. This update is merely fail2ban adding Dynamic DNS records - assuming something will consume those records later. My gut says 60 seconds makes sense; while lower is interesting, I wouldn't want to go too much higher personally.

buanzo:
IMHPO, I would think the default value should equal the default ban time,
as simplification. DNSBL operators should publish their own rules/values
for their target community.

leeclemens:
@buanzo the problem is the TTL isn't the bantime, it's the bantime + the TTL - so if it's the bantime, the bantime becomes 2*bantime.

That gets into what I had asked before, which was about what would be consuming these records. If it's another jail, it will have its own bantime and the TTL should be as low as possible to avoid artificially increasing it (see previous statement).

buanzo:
fail2ban should not be using a dnsbl it helps generate i think.

leeclemens:
That doesn't answer my question, but it's a start (something would not consume it) :)

I think the key difference between bantime and TTL is that bantime starts expiring once the host bans it, while the TTL starts expiring when any one of N servers query its record. Therefore, setting the TTL to bantime gives no indication of how long the host should be banned for, so it is misleading at best.

buanzo:
until the host is no longer in the dnsbl?

the ttl value will just suggest how much time a caching dns server should
cache that response.

buanzo:
i would mimic what other dnsbl consummers do, too.

buanzo:
http://www.engardelinux.org/modules/index/list_archives.cgi?list=postfix-users&page=0568.html&month=2012-04

FTL: "But consider this: the TTL of a DNSBL listing is a feature. Sometimes
legitimate sites will be listed, for example, in the CBL. Once they
clean up the problem, do you still want to block them?"

leeclemens:
Again, I'm asking what the intent of this is - i.e. the consumers. The OP is to update dynamic dns records using nsupdate. Very simple, basic action which can be used in a variety of ways. It makes no specific mention to DNSBL/CBL/etc.

@buanzo In reference to your post: Do you still want to block them?

leeclemens:
I ask, because you are suggesting banning them for up to 2 times the bantime, rather than as close to the bantime as possible. I am curious what your thinking is for that, and I'm afraid I haven't seen an explanation to that specific question.

buanzo:
I do NOT believe TTL to mean any thing other than "time this record expires
from a caching dns server's cache".
bantime is in no way related to TTL, and I don't think it should.

buanzo:
Hi Lee,

If you use fail2ban to populate a DNSBL (which is I believe the purpose of
this nsupdate feature), then your consumers will use whatever policy is
[pre]configured, and query and cache DNSBL records per their own desires.

When DNSBL-populating-fail2ban-instance unbans the host, for whatever
reasons, being bantime reached, or any other kind of policy put in place by
the sysadmin, then consumers, when they finally decide to query the DNSBL
for that particular HOST again, will receive a 'this host is not listed in
this dnsbl' response.

With that context in place, and the previous responses in this thread, I
think the relationship between bantime and TTL is clearly exposed.

If this does not answer your question, I suggest you rephrase it a bit if
you don't mind? Maybe I am confused or missing something. Thank you and
sorry!

leeclemens:
In your opinion, should the host be 'unbanned' from DNS as quickly as possible (i.e. TTL expired), such that the shortest period of time possible exists between the banning server unbanning the host and a consumer querying the DNS record?

I'm starting to suspect we are seeking the same end goal, but I'm not sure why you want a longer TTL to accomplish it.

buanzo:
Ah. I see your point.

I do not want a long TTL. I want an appropiate one. Not very short. Not
very long. Sysadmins or policy makers should setup their own values.

OTOH, consumers are free to query the dnsbl as often as they want anyway
(as long as they do not DoS the service, heh).

We are not really in control of that, aren't we?

Hence my opinion.

I would review dnsbl ttl values and come up with a list, for the sake of
homogeneity.

leeclemens:
My opinion is let the admins or the applications interpret...let DNS records be as timely / accurate as possible for this particular use-case. This is not strictly about a DNSBL, but about updating dynamic dns records (hence my reference to freedns.afraid.org).

What happens after DDNS is updated is up to whomever decides to query it. This isn't meant to be an API to fail2ban's ban list for other applications to query hosts and ban times, but to simply update a DDNS record when a given host is banned.

It should reflect who is banned now, as closely as possible (hence lower TTL), while utilizing the performance of DNS queries/caching.

How many queries per second do your DNS servers service where a 60 second TTL would DoS it?.

ChangeLog
```diff
@@ -41,6 +41,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    - Monit config for fail2ban in /files/monit
    - New actions:
      - action.d/firewallcmd-multiport and action.d/firewallcmd-allports Thanks Donald Yandt
+     - action.d/nsupdate Thanks Andrew St. Jean
```
yarikoptic: please describe here very briefly what is this for since "ns" is a loaded acronym ;-)

config/action.d/nsupdate.conf
```diff
@@ -0,0 +1,114 @@
...
+#
+# Options
+#
+# domain	DNS domain that will appear in nsupdate add and delete
+#		commands.
+#
+# ttl		The time to live (TTL) in seconds of the TXT resource
+#		record.
+#
+# rdata		Data portion of the TXT resource record.
```
yarikoptic: I guess some tabs crawled in... just indent using spaces here

ChangeLog
yarikoptic: ok - will do myself
cheers

rumple010:
@yarikoptic 

Sorry for the delayed response. I would have gotten to your suggestions but you beat me to it.

PR #926, opened 2015-01-26
ENH: use non-UTC date invocation (without -u) and report offset for local zone

### LINKED ISSUE #911: Emails from fail2ban 0.9.1 are timestamped with UTC time instead of local time
I first reported this here:
https://bugzilla.redhat.com/show_bug.cgi?id=1180850
but was redirected here.  Unsure if this is considered a bug, but this behavior surprised me.

Description of problem:
Whenever fail2ban sends an email it has a UTC time stamp instead of a local time zone time stamp (like most other local logs use).

Version-Release number of selected component (if applicable):
I'm using the fail2ban-0.9.1-2.el7 RPM on CentOS 7

How reproducible:
Every time

Steps to Reproduce:
1. Enable a jail (like sshd)
2. Use one of the email actions (like "%(action_mwl)s")
3. Create a ban that generates an email (fail2ban-client set sshd banip 1.2.3.4)

Actual results:
Get an email that has a UTC time stamp

Expected results:
Get an email that has a local time stamp

Additional info:
Since I"m only using the "%(action_mwl)s" action I only need to edit one file to change fail2ban to use local time.
vi /etc/fail2ban/action.d/sendmail-whois-lines.conf
-Prepend a # to the start of the actionban date line.
-For example find this section:
<pre>
actionban = printf %%b "Subject: [Fail2Ban] <name>: banned <ip> from `uname -n`
            Date: `LC_TIME=C date -u +"%%a, %%d %%h %%Y %%T +0000"`
</pre>

-And insert a # at the start of the second line so it looks like this:
<pre>
actionban = printf %%b "Subject: [Fail2Ban] <name>: banned <ip> from `uname -n`
#            Date: `LC_TIME=C date -u +"%%a, %%d %%h %%Y %%T +0000"``
</pre>
-This simple fix seems to work well for me, but since I don't understand what they're trying to do with "LC_TIME=C date" here YMMV.

I'm not sure if this is a bug or a feature.  I found a similar Debian bug report from 10 years ago, so at least I'm not the only one who doesn't like UTC dated emails.  See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=336449 for the other bug report.

yarikoptic:
Closes #911

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -48,9 +48,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
 - Enhancements:
    * Enable multiport for firewallcmd-new action.  Closes gh-834
    * files/debian-initd migrated from the debian branch and should be
-     suitable for manual installations now (thanks Juan Karlo de
-	 Guzman)
-   * action.d/{sendmail-*,xarg-login-attack}.conf - report local
+     suitable for manual installations now (thanks Juan Karlo de Guzman)
+   * action.d/{sendmail-*,xarf-login-attack}.conf - report local
      timezone not UTC time/zone. Closes gh-911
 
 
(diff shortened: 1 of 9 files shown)


yarikoptic:
If no objections, I would appreciate if someone else from @fail2ban/developers would merge the PR

szepeviktor:
Note the xarg typo in Chagelog.

Issue #928, opened 2015-01-27
Trouble with using pam_unix for auth failures


opoplawski:
When using sssd (or ldap, or ..) for authentication you will see messages like:
````
Jan 27 08:53:23 mail auth: pam_unix(dovecot:auth): authentication failure; logname= uid=0 euid=0 tty=dovecot ruser=USER rhost=111.111.111.111  user=USER
Jan 27 08:53:23 mail auth: pam_sss(dovecot:auth): authentication success; logname= uid=0 euid=0 tty=dovecot ruser=USER rhost=111.111.111.111 user=USER
````
So instead of pam_unix in regexs, we should have something like %(pam_auth)s so you can set the pam authentication mechanism.  Perhaps fail2ban could parse /etc/nsswitch.conf to automatically determine what auth mechanism was in use:
````passwd:     files sss````

Current usage:
````
/etc/fail2ban/filter.d/dovecot.conf:failregex = ^%(__prefix_line)s(pam_unix(\(dovecot:auth\))?:)?\s+authentication failure; logname=\S* uid=\S* euid=\S* tty=dovecot ruser=\S* rhost=<HOST>(\s+user=\S*)?\s*$
/etc/fail2ban/filter.d/vsftpd.conf:__pam_re=\(?pam_unix(?:\(\S+\))?\)?:?
/etc/fail2ban/filter.d/pam-generic.conf:__pam_re=\(?pam_unix(?:\(\S+\))?\)?:?
/etc/fail2ban/filter.d/pam-generic.conf:# _daemon = \S*\(?pam_unix\)?
/etc/fail2ban/filter.d/wuftpd.conf:__pam_re=\(?pam_unix(?:\(wu-ftpd:auth\))?\)?:?
````
At least the vsftpd and wuftpd filters have a __pam_re that can be overridden.

sebres:
agree

opoplawski:
This sets up the variable.  No idea how to make this auto-generated.

sebres:
I think it should not necessarily be "auto-generated" - just set default to `pam_unix`, and if someone wants anything else, it can be set simple to other value directly in jail.local.

opoplawski:
There are other issues with pam_sss unfortunately - see https://www.centos.org/forums/viewtopic.php?t=1718 for an example.  The upshot is that there is no pam_sss authentication failed messages for unknown users.

opoplawski:
[...]

PR #929, opened 2015-01-27
Add filter variable __pam_auth

opoplawski:
Add filter variable __pam_auth to allow easier changing of pam auth backend.

This is an initial step for fixing issue #928 - would still be nice to auto set this.

### DIFF
--- a/config/filter.d/common.conf
+++ b/config/filter.d/common.conf
@@ -53,4 +53,8 @@ __bsd_syslog_verbose = (<[^.]+\.[^.]+>)
 # This can be optional (for instance if we match named native log files)
 __prefix_line = \s*%(__bsd_syslog_verbose)s?\s*(?:%(__hostname)s )?(?:%(__kernel_prefix)s )?(?:@vserver_\S+ )?%(__daemon_combs_re)s?\s%(__daemon_extra_re)s?\s*
 
+# PAM authentication mechanism check for failures, e.g.: pam_unix, pam_sss,
+# pam_ldap
+__pam_auth = pam_unix
+
 # Author: Yaroslav Halchenko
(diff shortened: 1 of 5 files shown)


yarikoptic:
shouldn't we just list them all (```pam_(unix|sss)```) or just allow ```pam_\S+```?

sebres:
Yep, `pam_\S+` would be indeed good also.
But why not both: 
- we can use everywhere in filter a `%(__pam_auth)s`;
- and in `config/filter.d/common.conf` default value can be `__pam_auth  = pam_\S+`

So if needed, everyone is capable to specify jail own pam-suffix just with overwriting of `%(__pam_auth)s`;

opoplawski:
Please read issue #928 - the problem is that for network users pam_unix will always return an auth failure, which would then trigger a ban.

yarikoptic:
@opoplawski ah -- right.  ok then.. just add a ChangeLog for that .

Not sure how 'autoset'ing could work here though, so why not just to consider #928 closed by this one?

opoplawski:
Here's my idea for auto generating (which could be a separate issue if we want to pursue it):

You don't want to key off of errors in any earlier entries in /etc/nssswitch.conf as they may be superseded by later modules.  You probably just want the last one.  I suspect 99% of entries are like:
````passwd: files sss````
or
````passwd: files ldap```` where you only want to key off of the last entry.  If __pam_auth were set to pam_sss or pam_ldap by default in these cases, I think you would cover most cases automatically.  "files" would be special case and would result in "pam_unix".

yarikoptic:
@opoplawski   sorry -- I am still a bit missing how we would automate that at fail2ban level

PR #930, opened 2015-01-28
Conditionally log Ignore IP, and pass in reason (Closes #916)

### LINKED ISSUE #916: Log cause of Ignore
INFO logs occur if IP is in ignoreip or ignorecommand returns false (and presumably ignoreregex matches), but the log only says "Ignore <ip>". I think the cause of the IP being ignored is also valuable.

leeclemens:


fail2ban/server/filter.py
```diff
@@ -369,16 +373,20 @@ def inIgnoreIPList(self, ip):
...
 			if a == b:
+				self.logIgnoreIp(ip, log_ignore, ignore_source="ignoreip by addr")
 				return True
 
 		if self.__ignoreCommand:
 			command = CommandAction.replaceTag(self.__ignoreCommand, { 'ip': ip } )
 			logSys.debug('ignore command: ' + command)
-			return CommandAction.executeCmd(command)
+			ret_ignore = CommandAction.executeCmd(command)
+			self.logIgnoreIp(ip, log_ignore and ret_ignore, ignore_source="ignorecommand")
```
yarikoptic: why not to make source values "dns", "ip", "command"? and adjust log msg accordingly to include "by ", thus overall unifying the msg?

yarikoptic:
and could you add a rudimentary check if logged whatever is meant to be

fail2ban/server/filter.py
leeclemens: I was thinking there were two reasons, one of which had two ways to determine (dns/ip)...but I like the simpler approach you suggested - https://github.com/leeclemens/fail2ban/commit/f4341999cd5e5695e600e051ac7fdf0edffbb93c

leeclemens:
Seems unittest.TestCase.assertLogs wasn't added until Python 3.4. I will add test cases to check DNS, IP and Command - I believe only IPs are tested currently. Thoughts on testing the error messages line up?

leeclemens:
Nevermind, ignorecommand is also tested in the unittests.

leeclemens:
I did my own testing before committing, but I'm not sure how to have the unittests check the log of why something was ignored.

yarikoptic:
just test calls to logIgnoreIp  function itself -- does it log only when it should, either logging works correctly?  just a really minor unittest

leeclemens:
I found self._is_logged, which was the missing piece I needed. Should be good now.

yarikoptic:
oh -- merged but then recalled that there is no ChangeLog entry -- please add!

leeclemens:
Thanks! Added https://github.com/fail2ban/fail2ban/pull/945

PR #932, opened 2015-01-29
Dovecot - dovecot auth failure from EL7

opoplawski:
Updated the dovecot regex to match:

Jan 29 05:32:50 mail dovecot: auth-worker(304): pam(username,1.2.3.4): pam_authenticate() failed: Authentication failure (password mismatch?)

which I'm seeing on EL7 with dovecot 2.2.10

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -83,6 +83,7 @@ Michael Hanselmann
 Mika (mkl)
 Nick Munger
 onorua
+Orion Poplawski
 Paul Marrapese
 Paul Traina
 Noel Butler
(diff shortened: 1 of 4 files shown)


THANKS
```diff
@@ -83,6 +83,7 @@ Michael Hanselmann
 Mika (mkl)
 Nick Munger
 onorua
+Orion Poplawski
```
yarikoptic: oh man -- you weren't there yet? shame! ;-)

PR #933, opened 2015-01-29
Add jail nginx-botsearch

mrc0mmand:
Jail blocks requests for predefined non-existent folders. Based on
apache-botsearch jail.

### DIFF
--- a/config/filter.d/nginx-botsearch.conf
+++ b/config/filter.d/nginx-botsearch.conf
@@ -4,7 +4,7 @@
 [Definition]
 
 failregex = ^<HOST> \- \S+ \[\] \"(GET|POST) \/<block> \S+\" 404 .+$
-			^ \[error\] \d+#\d+: \*\d+ (\S+ )?\"\S+\" (failed|is not found) \(2\: No such file or directory\), client\: <HOST>\, server\: \S*\, request: \"(GET|POST) \/<block> \S+\"\, .*?$
+            ^ \[error\] \d+#\d+: \*\d+ (\S+ )?\"\S+\" (failed|is not found) \(2\: No such file or directory\), client\: <HOST>\, server\: \S*\, request: \"(GET|POST) \/<block> \S+\"\, .*?$
 
 ignoreregex = 
 
--- a/config/filter.d/nginx-botsearch.conf
+++ b/config/filter.d/nginx-botsearch.conf
@@ -1,0 +1,29 @@
+# Fail2Ban filter to match web requests for selected URLs that don't exist
+#
+
+[Definition]
+
+failregex = ^<HOST> \- \S+ \[\] \"(GET|POST) \/<block> \S+\" 404 .+$
+			^ \[error\] \d+#\d+: \*\d+ (\S+ )?\"\S+\" (failed|is not found) \(2\: No such file or directory\), client\: <HOST>\, server\: \S*\, request: \"(GET|POST) \/<block> \S+\"\, .*?$
+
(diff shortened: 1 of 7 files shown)


szepeviktor:
If you have a real WordPress site you could set up a robot trap.
https://github.com/szepeviktor/wordpress-plugin-construction/blob/master/mu-nofollow-robot-trap/nofollow-robot-trap.php

config/filter.d/nginx-botsearch.conf
```diff
@@ -0,0 +1,29 @@
+# Fail2Ban filter to match web requests for selected URLs that don't exist
+#
+
+[Definition]
+
+failregex = ^<HOST> \- \S+ \[\] \"(GET|POST) \/<block> \S+\" 404 .+$
+			^ \[error\] \d+#\d+: \*\d+ (\S+ )?\"\S+\" (failed|is not found) \(2\: No such file or directory\), client\: <HOST>\, server\: \S*\, request: \"(GET|POST) \/<block> \S+\"\, .*?$
```
yarikoptic: better to indent with spaces here to avoid such formatting jumps

mrc0mmand: Yeah, you're right, I just forgot to switch my editor back to spaces. Will do now.

config/filter.d/nginx-botsearch.conf
```diff
@@ -0,0 +1,29 @@
...
+
+ignoreregex = 
+
+
+[Init]
+
+# Block is the actual non-found directories to block
+block = \/?(<webmail>|<phpmyadmin>|<wordpress>|cgi-bin|mysqladmin)[^,]*
```
yarikoptic: I am asking in https://github.com/fail2ban/fail2ban/pull/557/files#r23978582  on why we have decided to go the way of explicit fail2ban variables here instead of just relying on string interpolations... anyways -- since all those match with apache-botsearch (or am I missing difference) we should have now botsearch-common.conf which would define them all and then only get included into apache- and ngnix- to be used in failregex's

mrc0mmand: Yes, all these vars are same as in apache-botsearch, so botsearch-common sound like a good idea.

yarikoptic: great -- so move then this [Init] portion into common, which you include on top in INCLUDES section.

mrc0mmand: Should I edit the apache-botsearch filter as well?

yarikoptic: >    Should I edit the apache-botsearch filter as well?

SURE!

config/jail.conf
```diff
@@ -294,6 +294,12 @@ maxretry = 1
 port    = http,https
 logpath = %(nginx_error_log)s
 
+[nginx-botsearch]
+
+port     = http,https
+logpath  = %(nginx_error_log)s
+maxretry = 10
+findtime = 300
```
yarikoptic: remove custom findtime, set maxretry to low number like already in apache-botsearch

mrc0mmand: Okay, done.

yarikoptic:
and damn me -- always forget it -- ChangeLog entry please ;-)

mrc0mmand:
I added the ChangeLog entry and I definitely should've used `rebase` instead of `merge`...

yarikoptic:
Thanks!  Merge is perfectly fine with me in this case!

PR #940, opened 2015-02-02
New jail: apache-fakegooglebot

leeclemens:
Detects fake googlebot user agents in apache access log


ChangeLog
```diff
@@ -49,6 +50,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
        - empty or "basic" works as-is
        - "cymru" additionally prints (ASN, Country RIR) per banned IP
        - Requires dnspython or dnspython3
+   - New ignorecommand:
+     - apache-fakegooglebot.sh  Thanks Lee Clemens
```
yarikoptic: I don't think it is beneficial to list it here as well.  IMHO New filter entry is sufficient (And you could extend with information on this ignorecommand or just omit it)

config/filter.d/ignorecommands/apache-fakegooglebot.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# Based on: https://isc.sans.edu/forums/diary/When+Google+isnt+Google/15968/
```
yarikoptic: well -- if based, then you would need first to clarify the copyright/license terms with original author.  Otherwise -- recode

config/jail.conf
```diff
@@ -277,6 +277,14 @@ logpath  = %(apache_error_log)s
 maxretry = 2
 
 
+[apache-fakegooglebot]
+
+port     = http,https
+logpath  = %(apache_access_log)s
+maxretry = 1
+ignorecommand = %(ignorecommands_dir)s/apache-fakegooglebot.sh <ip>
```
yarikoptic: as I have mentioned in another issue, better to not ship implementation specific suffixes here (i.e.  remove ```.sh``` from the name of that command i.e. rename the file as well)

config/paths-common.conf
```diff
@@ -61,3 +61,6 @@ dovecot_log = %(syslog_mail_warn)s
 solidpop3d_log = %(syslog_local0)s
 
 mysql_log = %(syslog_daemon)s
+
+# Directory with ignorecommand scripts
+ignorecommands_dir = /etc/fail2ban/filter.d/ignorecommands
```
yarikoptic: what for if it is hardcoded in setup.py anyways... and I guess jail.conf wouldn't be appropriate level anyways (fail2ban.conf would be I Guess... but let's just hardcode for now?)

setup.py
```diff
@@ -96,7 +96,7 @@
 		'bin/fail2ban-client',
 		'bin/fail2ban-server',
 		'bin/fail2ban-regex',
-		'bin/fail2ban-testcases',
+		'bin/fail2ban-testcases'
```
yarikoptic: why?  it was perfectly ok and I usually prefer trailing comma to minimize next diff happen we need to add a new entry

in general -- better to retain only changes related to introduced new functionality in such PRs

leeclemens: Missed this one, it may have been from a refactor

config/filter.d/ignorecommands/apache-fakegooglebot.sh
leeclemens: It is a forum post with no copyright/license terms I could see. I removed the parsing of the log file parts since fail2ban is providing the IP as an argument. What is left is basically just comparing results of `host` and parsing the DNS answer.

ChangeLog
leeclemens: removed

config/jail.conf
leeclemens: Done

config/paths-common.conf
leeclemens: My thought of having it here was so it could be referenced in the jail and changed in one place (not counting setup.py). If three more jails were added that use it, it would reduce the instances to change if need be.

config/filter.d/ignorecommands/apache-fakegooglebot.sh
yarikoptic: ok -- I am rewriting it... ;-)

yarikoptic: https://github.com/leeclemens/fail2ban/pull/1/

yarikoptic: thinking about it, we should have made ignorecommand more generic, just requiring specification as "command <IP>" and any other options. then we could just create this generic helper "ignoreifnamematching" which would get a hostname pattern as the first argument and IP as the 2nd, i.e. ```ignoreifnamematching 'crawl-.*\.googlebot\.com' <IP>``` and be done but also scale easily for other similar cases... what do you think?

yarikoptic: NB that long ugly cmd name in above is just the first thing which came into my mind ;)

config/paths-common.conf
yarikoptic: ah -- gotcha, thanks for explanation, makes sense indeed !

config/filter.d/ignorecommands/apache-fakegooglebot.sh
leeclemens: It's interesting to make it more generic, I'm trying to think of a use-case where similar matching makes sense (smtp comes to mind, but that's built in and configurable mostly within the MTA).

I'm inclined to suggest, if someone uses the same logic of rDNS name matching in the future, we should abstract them to reduce duplicate code in ignorecommands scripts. I'm afraid creating a new argument name (and implementing its appropriate getters/setters/testcases) with only one implementation creates more sprawl than it hopes to prevent. Just my two cents thinking about it initially.

leeclemens:
coverage was 91.86% and now says it DECREASED to 91.89% ?

setup.py
```diff
@@ -124,6 +124,9 @@
 		('/etc/fail2ban/filter.d',
 			glob("config/filter.d/*.conf")
 		),
+		('/etc/fail2ban/filter.d/ignorecommands',
+			glob("config/filter.d/ignorecommands/*")
+		),
```
yarikoptic: and may be we should adjust MANIFEST*'s?

leeclemens: Added to both, broader question - should MANIFEST be in version control at all?

yarikoptic: let it be there for now ;)

yarikoptic:
I am ok to merge this one, unless any other @fail2ban/developers has any reservation/further comments.

setup.py
leeclemens: fair enough :)

config/filter.d/ignorecommands/apache-fakegooglebot.sh
leeclemens: Worth noting I like the idea of doing it all in python, the thought of an external command kept me from seeing that as a better solution.

yarikoptic:
ok -- no concerns, so let's go!

PR #941, opened 2015-02-02
Clarify filter.DNSUtils functions' terminoligy and add unittests

leeclemens:


fail2ban/server/filter.py
```diff
@@ -898,22 +898,22 @@ def textToIp(text, useDns):
...
 		"""
 		# 32-bit IPv4 address mask
 		MASK = 0xFFFFFFFFL
-		return ~(MASK >> n) & MASK & DNSUtils.addr2bin(i)
+		return ~(MASK >> cidr) & MASK & DNSUtils.addr2dec(ipstring)
 
 	@staticmethod
-	def addr2bin(string):
-		""" Convert a string IPv4 address into an unsigned integer.
+	def addr2dec(ipstring):
```
yarikoptic: who says it is a decimal? ;) it is just a binary of type int as far as I see it

fail2ban/server/filter.py
```diff
@@ -898,22 +898,22 @@ def textToIp(text, useDns):
 		return ipList
 
 	@staticmethod
-	def cidr(i, n):
-		""" Convert an IP address string with a CIDR mask into a 32-bit
-			integer.
+	def netAddr2Dec(ipstring, cidr):
```
yarikoptic: if anything, we just might want to merge it into add2bin/dec, which would apply cidr if not None.  otherwise -- all other methods are  lower cases...

if you like to improve this -- just RF all the DNSUtils into a separate submodule  (e.g. dnsutils) and remove this unnecessary @static madness... I even have done it at some point -- must be in one of the "lost branches" ;)

fail2ban/tests/filtertestcase.py
```diff
@@ -1011,6 +1011,23 @@ def testTextToIp(self):
 			else:
 				self.assertEqual(res, [])
 
+	def testNetAddr2Dec(self):
```
yarikoptic: I love tests too -- we should hangout together!

fail2ban/server/filter.py
leeclemens: Okay, this comment is in binary too :)

It returns a long, whereas typical conversion to binary would return a string of 0's and 1's - as they would otherwise be represented as a decimal (or long).

My thinking was, do you get back a binary string or an integer? Can that more accurately be described as "bin" or "dec"?

Binary value of IP Address "10.0.0.0" is "00001010 00000000 00000000 00000000".  Hex: 0A:00:00:00 Decimal: 167772160.

fail2ban/server/filter.py
leeclemens: Combined the two functions, with kwarg cidr=None

fail2ban/server/filter.py
yarikoptic: >    My thinking was, do you get back a binary string or an integer? Can that
>    more accurately be described as "bin" or "dec"?

>    Binary value of IP Address "10.0.0.0" is "00001010 00000000 00000000
>    00000000". Hex: 0A:00:00:00 

both above are specific "views" of such a binary number

> Decimal: 167772160.

and this one a typical decimal printout of an integer as well.  If it
was a string with a "view" in integer form -- it would have been 'dec',
but now it is just an integer, which is just stored as a number and we
use digital computers which internally use binary arithmetics and
storage...  that is what was meant behind "bin" here -- just pure binary
representation, not really a "view/printout" which as if it was a
string -- it can be hex or any other view if someone "prints it out".
So I would just vote to leave it alone as is -- 'dec' would be imho more
confusing that helpful.

fail2ban/server/filter.py
```diff
@@ -360,11 +360,11 @@ def inIgnoreIPList(self, ip):
...
 			elif "." in s[1]: # 255.255.255.0 style mask
 				s[1] = len(re.search(
-					"(?<=b)1+", bin(DNSUtils.addr2bin(s[1]))).group())
+					"(?<=b)1+", bin(DNSUtils.addr2dec(s[1]))).group())
 			s[1] = long(s[1])
 			try:
-				a = DNSUtils.cidr(s[0], s[1])
-				b = DNSUtils.cidr(ip, s[1])
+				a = DNSUtils.addr2dec(s[0], cidr=s[1])
+				b = DNSUtils.addr2dec(ip, cidr=s[1])
```
yarikoptic: please keep it 'addr2bin' for now -- I have replied to outdated now diff - so not sure if reached

fail2ban/server/filter.py
leeclemens: As the author of IPCalcEx (Android App), I may be biased (my app uses binary string and decimal as I portrayed them here). With regards to fail2ban, that argument isn't as relevant regarding a function's name - so I will cordially disagree, since it is trivial in comparison to the argument it would require to even potentially persuade you :) I see your point, and I see mine :)

For the record, I still think decimal is a more accurate description to the value returned :)

fail2ban/server/filter.py
leeclemens: Updated, was replying as you replied here :)

fail2ban/server/filter.py
```diff
@@ -898,22 +898,18 @@ def textToIp(text, useDns):
...
-		MASK = 0xFFFFFFFFL
-		return ~(MASK >> n) & MASK & DNSUtils.addr2bin(i)
-
-	@staticmethod
-	def addr2bin(string):
-		""" Convert a string IPv4 address into an unsigned integer.
-		"""
-		return struct.unpack("!L", socket.inet_aton(string))[0]
+		if cidr is None:
+			return struct.unpack("!L", socket.inet_aton(ipstring))[0]
```
yarikoptic: see -- even python's standard lib  agrees with me

```
Convert an IP address in string format (123.45.67.89) to the 32-bit packed
binary format used in low-level network functions.
```
for inet_aton function ;)    if we were to find a middle ground -- I might agree on "num" (for numeric) but may be let's not do it ;)

leeclemens: You had to go there ;) I thought I just agreed with you :)

yarikoptic: here is another prove that it has nothing to do with 'dec':

```
*In [10]: '%x' % struct.unpack('!L', socket.inet_aton('10.1.2.3'))[0]
Out[10]: 'a010203'
```

as you can see -- it just packs those 4 numbers into 4 bytes.  I really see no relation to 'dec' here.

leeclemens: By "I just agreed with you" I actually mean "I just agreed with you"

yarikoptic: Please the reached agreement by a change  few lines down below (I thought I have left a comment there but it is gone -- I need to get away from keyboard) ;-)

leeclemens: In the sense that the value is "packed", I believe means it returns a single decimal value? At least that's what I get back from executing it. Having the word "binary" included in the description isn't as descriptive as you may think.

yarikoptic: 'decimal' is "A number expressed in the scale of tens".   From executing it you get 4 bytes representing an integer (internally represented in scale of twos, thus binary :-P) -- it is just printed on the screen usually  by default in a decimal format and that doesn't make it a decimal value. So let's just go to your comment 2 comments back ;)

leeclemens: Understood. I think I should step back as well.

As a note for later:

Is the issue the comment? "Convert a string IPv4 address into decimal form" should be "Convert a string IPv4 address into binary form" ?

yarikoptic: BTW -- are you receiving emails to droid@lc-cs.com ?

leeclemens: Yes, I receive email there. And thank you for the inclusion in @fail2ban/developers

leeclemens:
@yarikoptic While I do receive email there, it isn't my main address for github/fail2ban- I'm not sure where you got that from...while I do use it in some places. java@leeclemens.net is the address I use for github and fail2ban-users lists

fail2ban/server/filter.py
yarikoptic: >    Understood. I think I should step back as well.

>    As a note for later:

>    Is the issue the comment? "Convert a string IPv4 address into decimal
>    form" should be "Convert a string IPv4 address into binary form" ?

yeah.  I still think that may be 'numeric' is the most appropriate here,
but heck -- binary it is ;)

leeclemens:
I think my update crossed paths with your comment. I'd agree numeric is the best compromise - if you still agree I can update it.

fail2ban/server/filter.py
```diff
@@ -898,22 +898,18 @@ def textToIp(text, useDns):
...
+		else:
+			MASK = 0xFFFFFFFFL
+			return ~(MASK >> cidr) & MASK & DNSUtils.addr2bin(ipstring)
 
 	@staticmethod
-	def bin2addr(addr):
-		""" Convert a numeric IPv4 address into string n.n.n.n form.
+	def dec2addr(ipdec):
```
yarikoptic: uff -- keep it bin2addr.... number of commits reaching critical mass for a given number of changes -- could you filter or squash them?

leeclemens:
Take #2 https://github.com/fail2ban/fail2ban/pull/946

PR #949, opened 2015-02-06
Configure Syslog Socket Path (closes #814)

### LINKED ISSUE #814: Hard coded path to syslog socket in server/server.py
server/server.py uses the hard coded path "/dev/log" as the path to syslog's unix socket. But that path is OS dependent, for FreeBSD it is /var/run/log and for Max OS X it is /var/run/syslog.

I have no idea how to code that OS independent. For FreeBSD I patch the path during installation.

leeclemens:
fail2ban.conf now includes syslogsocket value, default = /dev/log
Can be overridden in fail2ban.local

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -69,6 +69,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      timezone not UTC time/zone. Closes gh-911
    * Conditionally log Ignore IP with reason (dns, ip, command). Closes gh-916
    * Absorbed DNSUtils.cidr into addr2bin in filter.py, added unittests
+   * Added syslogsocket configuration to fail2ban.conf
 
 
 ver. 0.9.1 (2014/10/29) - better, faster, stronger
--- a/config/fail2ban.conf
+++ b/config/fail2ban.conf
@@ -36,8 +36,9 @@ logtarget = /var/log/fail2ban.log
 
 # Option: syslogsocket
 # Notes: Set the syslog socket file. Only used when logtarget is SYSLOG
-# Values: [ FILE ]  Default: /dev/log
-syslogsocket = /dev/log
+#        auto uses platform.system() to determine predefined paths
+# Values: [ auto | FILE ]  Default: auto
+syslogsocket = auto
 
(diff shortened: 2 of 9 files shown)


yarikoptic:
Looks good to me, although setting in the config file explicitly precludes "automatic detection" based on OS... may be worth adding "auto" (Default) which if matched would result in an attempt to detect?

leeclemens:
The default is set to /dev/log (which is currently hardcoded). I could see an "auto" config option with a few known paths being a good special-case default - would be a little overhead at startup for distros that occurred later in the "auto" test list - so they may wish to config their path anyway

leeclemens:
btw, that was my first glance assessment..."auto" would help anyone without /dev/log not need to reconfigure during packaging (assuming their path is in fail2ban's hardcoded list of paths...)

yarikoptic:
no -- if a custom path explicitly provided -- just choose it (so no matching to hardcoded list).  If not, and only if SYSLOG is selected as the log target, then deduce based on platform which file should be used.  If deduced path doesn't exist -- puke and exit directing user to specify path explicitly

leeclemens:
How would the path be deduced by platform? From what I could tell, the platform paths aren't available inside fail2ban config's reader; that is why I floated the idea of iterating a list of known potential paths.

yarikoptic:
```
python -c "import platform; print {'Darwin':  '/var/run/syslog', 'Linux': '/var/log'}[platform.system()]"
```
we just need to discover what is output on windows and freebsd ;)  for windows then provide "None" so it would go over socket

mrc0mmand:
This is what I get on Windows 7 (with Python3):
```
>python -c "import platform; print(platform.system())
Windows
```

leeclemens:
Thanks @mrc0mmand for the info!

I'll go with this hardcoded dict() or "auto" as default, should be able to update soon.

leeclemens:
Okay, using "auto" as the default, the path will be determined (and set) by the dict(). I excluded Windows, since it would be None anyway - and any absent value from the infinite possible value from platform.system() results in the same error to set syslock socket.

Issue #893, opened 2014-12-22
FreeBSD: fail2ban should not stick forever on ipfw entries deleted by hand


ache:
Currently if some entry is deleted by ipfw manually, it stick forever in sqlite database and fail2ban attempts to delete it over and over. Fail2ban simple should understand this ipfw error message (ESRCH - "No such process") as _success_ to prevent it:
014-12-22 20:22:46,060 fail2ban.action         [57792]: ERROR   ipfw table 2 delete 117.21.225.176 -- stderr: 'ipfw: setsockopt(IP_FW_TABLE_XDEL): No such process\n'

ache:
For address adding you also need to handle this error as success:
ERROR   ipfw table 2 add 62.210.178.242 -- stderr: 'ipfw: setsockopt(IP_FW_TABLE_XADD): File exists\n'

leeclemens:
What if you use fail2ban-client to unban it, rather than modifying ipfw directly (which fail2ban won't know about)?

ache:
This is not limited to direct modification. Another example is restart after machine crash (ipfw tables are empty, but fail2ban will try forever from this moment to unban already unbanned). Another example is fail2ban crash (I see it few times in python sqlite3 module). Etc. Moreover, with each such case the number of unneeded attempts and junk log entries increases over time (since fail2ban don't know a way to decrease it).

The general rule for all complex cases is treating attempt to delete of already deleted as OK and attempt to add already present as OK too.

leeclemens:
My response was based upon your previous comments. Does fail2ban-client unban resolve this particular issue? (I realize you may see other issues, in addition to what you originally raised here.)

ache:
No, both fail2ban client and daemon fails with errors I already quoted exactly in the start of thread.
First error is delete already deleted. Second error is add already added.
This two should be no errors at all!

Nothing from fail2ban side can drop this stale entries afterwards (excepting external removing perfectly valid fail2ban database and rebooting the machine) and its number increased with each incident.

ache:
To fix that you need to check not only ipfw exit code but the error string
'ipfw: setsockopt(IP_FW_TABLE_XDEL): No such process\n'
on deleting and the error string
'ipfw: setsockopt(IP_FW_TABLE_XADD): File exists\n'
on adding.

leeclemens:
Care to submit a PR?

ache:
Does fail2ban have PRs system? I don't now how and where.

[...]

PR #952, opened 2015-02-08
Update bsd-ipfw.conf

ache:
Deleting not existent is not error.
Adding already present is not error.
Otherwise all those entries becomes stale forever, not removed and its number increases over time.

### DIFF
--- a/config/action.d/bsd-ipfw.conf
+++ b/config/action.d/bsd-ipfw.conf
@@ -38,7 +38,7 @@ actioncheck =
 # Values:  CMD
 #
 # requires an ipfw rule like "deny ip from table(1) to me"
-actionban = ipfw table <table> add <ip>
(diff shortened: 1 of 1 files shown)


ache:
Close #893

yarikoptic:
Thanks @ache -- looks sane to me but I wonder if robust, e.g. wouldn't message depend on locale?
alternatively:  we don't have 'actioncheck' so not sure how, but -- is there a way to list corresponding table and then simply do not run the command if corresponding ban is already in place (or already removed)?  that imho could be cleaner

ache:
1) Currently ipfw don't call setlocale() so always operates with "C" locale.
2) It is possible to list: 
ipfw table <num> list
but this list can be really big, causing unnecessary I/O, better relay on one line error diagnostic.

yarikoptic:
@fail2ban/contributors anyone on FreeBSD could check this one out?

PR #955, opened 2015-02-13
BF: fail2ban-regex does not read '.local' file of given filter (gh-954)

### LINKED ISSUE #954: fail2ban-regex support for .local files?
For real filter .conf files, a .local file can override specific settings, etc, which is great.

But, fail2ban-regex does not appear to look for a .local version of the regex filename, or have any way to explicitly provide one.  This diminishes fail2ban-regex's ability to be used to test or predict the behavior of fail2ban with a set of filter.d/foo{.conf,.local} files.

sebres:
closes #954

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      issue (gh-161)
    * filter.d/postfix-sasl.conf - tweak failregex and add ignoreregex to ignore
      system authentication issues
+   * fail2ban-regex reads filter file(s) completely, incl. '.local' file etc. (gh-954)
 
 - New Features:
    - New filter:
--- a/bin/fail2ban-regex
+++ b/bin/fail2ban-regex
@@ -220,6 +220,7 @@ class Fail2banRegex(object):
 		self._datepattern_set = False
 		self._journalmatch = None
 
+		self.share_config=dict()
 		self._filter = Filter(None)
 		self._ignoreregex = list()
 		self._failregex = list()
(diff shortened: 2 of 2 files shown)


sebres:
@fail2ban/contributors what's about `KeyError: 'categories'` in `BadIPsActionTest`?
sporadical? can be fixed?

yarikoptic:
Thanks @sebres.  (Eventually we should move all functionality of cmdline tools under fail2ban/cmdline and just invoke from the scripts -- that would allow us to unittest it)
FWIW -- it is better to make commit messages imperative, i.e. to describe what was done (and then it fixed a known problem), not just a statement of a problem itself.  And apparently travis fails are legit but not related to this one

yarikoptic:
submitted https://github.com/fail2ban/fail2ban/pull/956  to at least troubleshoot that issue with the response we receive... nothing else imho for us to do

ChangeLog
```diff
@@ -36,6 +36,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      issue (gh-161)
    * filter.d/postfix-sasl.conf - tweak failregex and add ignoreregex to ignore
      system authentication issues
+   * fail2ban-regex does not read '.local' file of given filter (gh-954)
```
yarikoptic: here as well -- better to make msg descriptive of what has happened (changed, or fixed), not just state the problem in present tense.

sebres:
@yarikoptic amend commit with log entry changed.

bin/fail2ban-regex
```diff
@@ -260,38 +261,47 @@ class Fail2banRegex(object):
...
-						journalmatch = command[3]
-						self.setJournalMatch(shlex.split(journalmatch))
-					elif command[2] == 'datepattern':
-						datepattern = command[3]
-						self.setDatePattern(datepattern)
+		if os.path.isfile(value) or os.path.isfile(value + '.conf'):
+			if os.path.basename(os.path.dirname(value)) == 'filter.d':
+				## within filter.d folder - use standard loading algorithm to load filter completely (with .local etc.):
+				basedir = os.path.dirname(os.path.dirname(value))
+				value = os.path.splitext(os.path.basename(value))[0]
```
yarikoptic: why the same logic can't apply if reading not from filter.d/?

sebres: Because this logic is implemented in JailReader/FilterReader/DefinitionInitConfigReader...
With another words: If we use `read` (not `readexplicit`) you should creates an instance of FilterReader with name of filter (not a filename):
```
FilterReader('sshd', 'fail2ban-regex-jail', {}, share_config=self.share_config, basedir='/etc/f2b/config')
```
And `filter.d` will be added indirect inside of FilterReader or DefinitionInitConfigReader ... 

BTW, I've not implemented it so :)

sebres: But I stumbled every time by implementing of sharing/caching of config files... :)

yarikoptic:
eh, ok -- let it be! ;)

PR #956, opened 2015-02-13
ENH: check badips.com response on presence of "categories" in it

yarikoptic:
As https://travis-ci.org/fail2ban/fail2ban/jobs/50609529 shows
query might fail in that response would not contain "categories".
With this change we will handle it explicitly and will spit out ValueError,
providing information about the response so it could be
troubleshooted


yarikoptic:
travis failure is a fluke, no objection from others -- merging! ;)

PR #958, opened 2015-02-13
Allow use of Python virtual environment in Redhat sysv init script

leeclemens:
Use /etc/default/fail2ban to determine Python virtual environment and activate it (also dynamically determine path to `fail2ban-client` using `which`.

If /etc/default/fail2ban does not exist, or defined virtual environment does not exist - falls back to existing behavior (but uses `which` to find `fail2ban-client`).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      issue (gh-161)
    * filter.d/postfix-sasl.conf - tweak failregex and add ignoreregex to ignore
      system authentication issues
+   * files/redhat-initrd - activate python virtual environment
+     if available in /etc/default/fail2ban
 
 - New Features:
    - New filter:
--- a/files/redhat-initd
+++ b/files/redhat-initd
@@ -21,10 +21,13 @@
 # Source function library.
 . /etc/rc.d/init.d/functions
 
+[ -f /etc/default/fail2ban ] && source /etc/default/fail2ban
+[ "$PYTHON_VENV" ] && [ -f ${PYTHON_VENV}/bin/activate ] && source ${PYTHON_VENV}/bin/activate
+
 # Check that the config file exists
 [ -f /etc/fail2ban/fail2ban.conf ] || exit 0
 
-FAIL2BAN="/usr/bin/fail2ban-client"
+FAIL2BAN=$(which fail2ban-client)
 prog=fail2ban-server
 lockfile=${LOCKFILE-/var/lock/subsys/fail2ban}
 socket=${SOCKET-/var/run/fail2ban/fail2ban.sock}


yarikoptic:
Although a sound idea (thanks), not sure if syadmins/package managers would appreciate ambigously specified paths in init. @opoplawski -- what do you think?

leeclemens:
I think /etc/default and /etc/sysconfig are both fairly standard places, perhaps /etc/sysconfig more-so with RedHat - /etc/sysconfig/httpd would be one example of the latter.

But I'm open to suggestions on how to improve it.

yarikoptic:
my concern is not about location of where to introduce configuration, but rather about the fact of using 'which'.  Happen I user/admin have augmented PATH in my environment pointing to e.g. development tree of fail2ban for some reason (e.g. was testing smth) then restarting service with such init.d would invoke fail2ban from that PATH instead of system-wide one, as those init files are assumed to do

opoplawski:
/etc/default is a debian thing.  /etc/sysconfig was a redhat thing, but it's nearly gone now.  Just put stuff in /etc/fail2ban would be my suggestion.

sysv init is also nearly gone replaced by systemd.  systemd unit files require hard coded paths to executables.

 What's the point of all this?  Is this something that fail2ban really needs to support?

leeclemens:
I was actually led to using /etc/default instead of /etc/sysconfig from some systemd documentation.

I'd question the statement that "sysv init is also nearly gone replaced by systemd"...How many RHEL 5/6, CentOS 5/6 boxes are still out there compared to 7? (I don't have the stats, just asking if you do.)

The point would be to allow fail2ban to run in a Python virtual environment, to facilitate using other versions of Python. If I'm not mistaken, RHEL 6.6 support ends Oct 31, 2016.

 It has no impact on any stock installation (using system's Python).

Also, if /etc/service is used, the environment should be cleaned anyway, so the concern of using `which` shouldn't play a role.

If you guys don't think it's useful, that's fine...it's in my repo and I doubt many other updates will be made to the sysv script moving forward. I just thought it might be useful to someone out there, so I wanted to contribute it back.


Unrelated, are you saying systemd will not support running fail2ban using a Python virtual environment?

opoplawski:
I'm just saying that in the systemd service file you would need to somehow construct the full path to the python interpreter, you wouldn't be able to use "which".  I'm not familiar with python virtual envs so not sure if this is possible or not.  And whatever method is decided on for configuring it, I think it should be the same for sysv init and systemd.

Why is it useful to run fail2ban with other versions of python? (Genuinely curious)

leeclemens:
It would be nice if it were the same for sysv and systemd. I'm not too familiar with the inner-workings of systemd to know how to configure it to source a file or use a variable in the config like is pretty standard for sysv/python virtual environments.

I use virtual environments for a number of reasons: consistency across platforms (versions, locations, modules installed), to keep fail2ban out of other Python application's environments (minimal modules installed, less overlap in case of reinstall/upgrade/downgrade) and basically any other benefit to running something in a segregated environment.

leeclemens:
I will gladly change the config to `/etc/sysconfig/fail2ban` as that may be more relevant for RedHat/CentOS as @opoplawski pointed out.

Can we agree on the following? If not, please let me know.

Using `which` and replacing all `FAIL2BAN` variables with `fail2ban-client` are effectively the same (I will gladly make that change if deemed useful).

Executing `/sbin/service` is recommended, as opposed to executing the init script directly, because it cleans environment variables that may lead to unanticipated executables of `fail2ban-client` (e.g. `PATH` as @yarikoptic mentioned as a potential issue) (I have not tested this in depth, so if you see different results than I suspect please provide details and we can work to a better solution.).
From `man service`:
> service  runs  a System V init script in as predictable environment as possible, removing most environment variables and with current working directory set to /.

We should not withhold an improvement to SysV's init script simply because there is no clear way to implement the same improvement to systemd's init configuration.

Using a config file as done here is extensible and should facilitate any future implementation in systemd's init configuration based on the UNIX Philosophy's Rule of Extensibility.

opoplawski:
On 02/18/2015 08:36 PM, Lee Clemens wrote:
> I will gladly change the config to |/etc/sysconfig/fail2ban| as that may be
> more relevant for RedHat/CentOS as @opoplawski <https://github.com/opoplawski>
> pointed out.

Actually, I believe I suggested something in /etc/fail2ban as being the most
distro neutral.

> 
> Can we agree on the following? If not, please let me know.
> 
> Using |which| and replacing all |FAIL2BAN| variables with |fail2ban-client|
> are effectively the same (I will gladly make that change if deemed useful).

No.  Using which makes it conditional on whatever PATH happens to be, which
strikes be as terribly insecure, buggy, unstable, etc.  You presumably know
which python you want to use, so use it explicitly.

> Executing |/sbin/service| is recommended, as opposed to executing the init
> script directly, because it cleans environment variables that may lead to
> unanticipated executables of |fail2ban-client| (e.g. |PATH| as @yarikoptic
> <https://github.com/yarikoptic> mentioned as a potential issue) (I have not
> tested this in depth, so if you see different results than I suspect please
> provide details and we can work to a better solution.).
> From |man service|:
> 
>     service runs a System V init script in as predictable environment as
>     possible, removing most environment variables and with current working
>     directory set to /.

Yes.

> We should not withhold an improvement to SysV's init script simply because
> there is no clear way to implement the same improvement to systemd's init
> configuration.

I'd say we just haven't looked at it closely enough, but I would be hesitant
to add a feature that will just go away as people move to modern distros.

-- 
Orion Poplawski
Technical Manager                     303-415-9701 x222
NWRA, Boulder/CoRA Office             FAX: 303-415-9702
3380 Mitchell Lane                       orion@nwra.com
Boulder, CO 80301                   http://www.nwra.com

PR #974, opened 2015-02-25
Shorewall shorewall-ipset-proto6.conf

ediazrod:
Use ipset proto6 in shorewall. You must follow the rules to enable ipset in you blacklist

if you have a lot of spam (my case) is better use ipset rather than shorewall command line (is my firewall)

stop fail2ban with shorewall on one list of 1000 Ips takes 5 min with ipset in shorewall 10 sec.

Any comments are welcome.

### DIFF
--- a/config/filter.d/shorewall-ipset-proto6.conf
+++ b/config/filter.d/shorewall-ipset-proto6.conf
@@ -1,0 +1,57 @@
+# Fail2Ban configuration file
+#
+# Author: Eduardo Diaz
+#
+# This is for ipset protocol 6 (and hopefully later) (ipset v6.14).
+# for shorewall
+#
+# This requires the program ipset which is normally in package called ipset.
+#
+# IPset was a feature introduced in the linux kernel 2.6.39 and 3.0.0 kernels.
+#
+# If you are running on an older kernel you make need to patch in external
+# modules.
+# Enable to shorewall use a blacklist using iptables creating a file /etc/shorewall/blacklist
+# and adding +f2b-shorewall to this file see details in shorewall documentation blacklist
+# to enable restore you ipset You must set SAVE_IPSETS=Yes in shorewall.conf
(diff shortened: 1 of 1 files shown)


leeclemens:
Disclaimer: I don't have experience with shorewall.

The iptables/ipset equivalent executes both `ipset` and `iptables`, while you are not executing `shorewall` (shorewall action does)...only `ipset`. At the risk of sounding naive, does this work without executing `shorewall`?

Unrelated to the above:
 - "f2b-shorewall" should be "f2b-<name>" so this action can be used by multiple jails without overlap.
 - Sample logs should be provided (See Travis' failure).
 - Should you also execute `ipset destroy` during `actionstop` (see `iptables-ipset-proto6.conf`)?

ediazrod:
yes is based in iptables-ipset-proto6.conf

I dont use f2b-<name> because you  must put the name of ipset in /etc/shorewall/blacklist
cat /etc/shorewall/blacklist
#ADDRESS/SUBNET PROTOCOL PORT
+f2b-shorewall
#LAST LINE -- ADD YOUR ENTRIES BEFORE THIS ONE -- DO NOT REMOVE

in addition shorewall take care of every command creating and destroy and save the ipset you must use shorewall (only need to create the ipset at the first execution (this is why I put the all the ips in the same ipset) I am open to recommendations +1:  this is the first version I want to keep simple

who can I provide the sample logs I don't have a lot of experience in github.?

leeclemens:
Okay, thanks for the clarification. Will this action always block all ports/protos as long as "f2b-shorewall" is present in /etc/shorewall/blacklist? How does that string get put into /etc/shorewall/blacklist?

Filters need to have sample log files....actually...you added this file as though it is a filter (in filter.d), but it looks like it is an action...

ediazrod:
Yes is a action!! My mistake I will remove this and make a new version!!

Issue #973, opened 2015-02-24
CRITICAL Unhandled exception in Fail2Ban


tarampampam:
**In `/var/log/fail2ban.log`:**
```
2015-02-24 16:29:03,190 fail2ban.filter         [13436]: INFO    [nginx-http-auth] Found 178.72.80.156
2015-02-24 16:29:03,192 fail2ban                [13436]: CRITICAL Unhandled exception in Fail2Ban:
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/fail2ban/server/jailthread.py", line 64, in run_with_except_hook
    run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/fail2ban/server/filterpoll.py", line 102, in run
    self.jail.putFailTicket(ticket)
  File "/usr/lib/python2.7/site-packages/fail2ban/server/jail.py", line 193, in putFailTicket
    self.database.addBan(self, ticket)
  File "/usr/lib/python2.7/site-packages/fail2ban/server/database.py", line 57, in wrapper
    return f(self, self._db.cursor(), *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/fail2ban/server/database.py", line 373, in addBan
    "failures": ticket.getAttempt()}))
InterfaceError: Error binding parameter 3 - probably unsupported type.
2015-02-24 16:29:03,404 fail2ban.actions        [13436]: NOTICE  [nginx-http-auth] Ban 178.72.80.156
[...]

PR #975, opened 2015-02-25
BF: binding parameter error (unsupported type) (closes gh-973) ...

sebres:
... by writing json with invalid encoded lines into sqlite database;
especially python < 3.0; try to prevent occurring such errors in the future;

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      system authentication issues
    * fail2ban-regex reads filter file(s) completely, incl. '.local' file etc. (gh-954)
    * firewallcmd-* actions: split output into separate lines for grepping (gh-908)
+   * binding parameter error (unsupported type) by writing json with invalid encoded
+     lines into sqlite database (gh-973), thanks to kot for issue reporting;
 
 - New Features:
    - New filters:
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,8 +38,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      system authentication issues
    * fail2ban-regex reads filter file(s) completely, incl. '.local' file etc. (gh-954)
    * firewallcmd-* actions: split output into separate lines for grepping (gh-908)
-   * binding parameter error (unsupported type) by writing json with invalid encoded
-     lines into sqlite database (gh-973), thanks to kot for issue reporting;
+   * Guard unicode encode/decode issues while storing records in the database.
+     Fixes "binding parameter error (unsupported type)" (gh-973), thanks to kot for reporting
 
 - New Features:
    - New filters:
--- a/fail2ban/server/database.py
+++ b/fail2ban/server/database.py
@@ -45,24 +45,38 @@ if sys.version_info >= (3,):
 			logSys.error('json dumps failed: %s', e)
 			x = '{}'
 		return x
+	def _json_loads_safe(x):
+		try:
+			x = json.loads(x.decode(locale.getpreferredencoding(), 'replace'))
+		except Exception, e: # pragma: no cover
+			logSys.error('json loads failed: %s', e)
+			x = {}
+		return x
 else:
+	def _normalize(x):
+		if isinstance(x, dict):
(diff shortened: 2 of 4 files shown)


ChangeLog
```diff
@@ -38,6 +38,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
      system authentication issues
    * fail2ban-regex reads filter file(s) completely, incl. '.local' file etc. (gh-954)
    * firewallcmd-* actions: split output into separate lines for grepping (gh-908)
+   * binding parameter error (unsupported type) by writing json with invalid encoded
+     lines into sqlite database (gh-973), thanks to kot for issue reporting;
```
yarikoptic: imho it might be difficult to comprehend what is this about and again  What about
"Guard unicode encode/decode issues while storing records in the database. Fixes "binding parameter error (unsupported type)" (gh-973), thanks to kot for reporting"

sebres:
@yarikoptic amend last commit with changelog entry rewritten as suggested.

sebres:
@yarikoptic please wait with merge - I have found another sitiation, I'm realize it now:
ich we have two matches - one is invalid, another valid, but unicode:
`['user "\xd1\xe2\xe5\xf2\xe0"', u'user "\xc3\xa4\xc3\xb6\xc3\xbc\xc3\x9f"']`

sebres:
ready. (what is with travis currently?)

sebres:
@fail2ban/contributors travis failed again with BadIPsActionTest... this time with 'connection refused'.
I think, such functionality as BadIPs, CymRU etc. should be tested as "option" and if it failed, should be logged, but not stop a rest of the test cases. 
I'd rather moved such functionality into separate project, like `f2b-plugins`, then can be also tested separately.

yarikoptic:
There's an option to run only non network tests. But it is the point to fail if test fails... that said we can indeed adopt pandas improved decorator for network tests so it first checks if network "works".

sebres:
@yarikoptic what's about something like skip "if no network / other criteria" instead of failed?

yarikoptic:
Yeah, that is what I meant to skip if there is no network (or "no network tests" mode was explicitly requested). But "no network" check could be specific for the test, depending on which network aspect they need (specific site or dns service) to skip rightfully.

leeclemens:
I don't agree with a "the test is failing, let's ignore the test" approach/mindset. Some build/platforms are passing, so Travis seems to have some network connectivity. I think the real question is: Is badips.com as unreliable in practice as it seems to be from rudimentary Travis build/tests? (Travis is down now, but iirc 3/9 platforms failed...and at least one fails often enough for it to be brought up as an issue.)

sebres:
We meant not "the test is failing, let's ignore the test", but more "the test **CAN'T** be executed anyway - skip it". 
With other words, before executing some test cases, it should check his availability etc.

leeclemens:
I guess I'm just not clear that is what is happening here. Since some Python version's built fine, is it the Travis test server not having a network connection that is causing the failure? Why would it be different based upon Python version? Why **can't** the test be executed (it passes more often than not...)?

sebres:
<blockquote>Why can't the test be executed?</blockquote>Well, badip is a foreign resource - it can be overloaded or totally down, we have no influences on.

leeclemens:
It is possible to test against a foreign resource, it is not an absolute "can not"-type of case. The fact that some of the platforms passed proves that. The failed tests call into question the interface fail2ban has with the external, potentially unreliable interface, so it seems, imho, to be a perfect example of a valid test.

fail2ban/server/database.py
```diff
@@ -37,17 +37,49 @@
...
+			x = {}
+		return x
 else:
-	sqlite3.register_adapter(dict, json.dumps)
-	sqlite3.register_converter("JSON", json.loads)
+	def _normalize(x):
+		if isinstance(x, dict):
+			return dict((_normalize(k), _normalize(v)) for k, v in x.iteritems())
+		elif isinstance(x, list):
+			return [_normalize(element) for element in x]
```
yarikoptic: I will merge it shortly but theoretically here it would be best to reuse the original class in case we/json modules comes up with derivatives of dict/list, thus still matching isinstance check BUT being a different ```_class_```

PR #978, opened 2015-02-26
shorewall-ipset-proto6.conf for shorewall

ediazrod:
Use ipset proto6 in shorewall. You must follow the rules to enable ipset in you blacklist
if you have a lot of spam (my case) is better use ipset rather than shorewall command line (is my firewall)
stop fail2ban with shorewall on one list of 1000 Ips takes 5 min with ipset in shorewall 10 sec.

### DIFF
--- a/config/action.d/shorewall-ipset-proto6.conf
+++ b/config/action.d/shorewall-ipset-proto6.conf
@@ -22,10 +22,9 @@
 #
 # A lot of this depends of the shorewall version at least have the 4.5.5 Version
 #
-# Enable to shorewall use a blacklist using iptables creating a file /etc/shorewall/blrules
-# and adding "DROP     net:+f2b-ssh    all" one for every jail
-# details in shorewall documentation blacklist.
-# to enable restore you ipset You must set SAVE_IPSETS=Yes in shorewall.conf
+# Enable shorewall to use a blacklist using iptables creating a file /etc/shorewall/blrules
+# and adding "DROP     net:+f2b-ssh    all" one for every jail details in shorewall documentation blacklist.
+# To enable restore you ipset You must set SAVE_IPSETS=Yes in shorewall.conf
 # Is importan to read this documentation from shorewall http://shorewall.net/ipsets.html
 #
 # To force create the ipset in the case that somebody delete the ipset create a file
--- a/config/action.d/shorewall-ipset-proto6.conf
+++ b/config/action.d/shorewall-ipset-proto6.conf
@@ -1,0 +1,84 @@
+# Fail2Ban configuration file
+#
+# Author: Eduardo Diaz
+#
+# This is for ipset protocol 6 (and hopefully later) (ipset v6.14).
+# for shorewall
+#
+# Use this setting in jail.conf to modify the name and the max time in every jain
+# action   = shorewall-ipset-proto6[name=SSH, bantime=10000]
+#
+# This requires the program ipset which is normally in package called ipset.
+#
+# IPset was a feature introduced in the linux kernel 2.6.39 and 3.0.0 kernels.
+#
+# The default Shorewall configuration is with "BLACKLISTNEWONLY=Yes" (see
+# file /etc/shorewall/shorewall.conf). This means that when Fail2ban adds a
+# new shorewall rule to ban an IP address, that rule will affect only new
+# connections. So if the attempter goes on trying using the same connection
+# he could even log in. In order to get the same behavior of the iptable
+# action (so that the ban is immediate) the /etc/shorewall/shorewall.conf
+# file should me modified with "BLACKLISTNEWONLY=No".
+#
+# A lot of this depends of the shorewall version at least have the 4.5.5 Version
+#
+# Enable to shorewall use a blacklist using iptables creating a file /etc/shorewall/blrules
+# and adding "DROP     net:+f2b-ssh    all" one for every jail
+# details in shorewall documentation blacklist.
+# to enable restore you ipset You must set SAVE_IPSETS=Yes in shorewall.conf
+# Is importan to read this documentation from shorewall http://shorewall.net/ipsets.html
+#
+# To force create the ipset in the case that somebody delete the ipset create a file
+# /etc/shorewall/initdone and add one line for every ipset (this files are in Perl)
+# take care of add 1; at the end of the file
+# the example is:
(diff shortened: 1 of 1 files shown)


config/action.d/shorewall-ipset-proto6.conf
```diff
@@ -0,0 +1,84 @@
+# Fail2Ban configuration file
+#
+# Author: Eduardo Diaz
+#
+# This is for ipset protocol 6 (and hopefully later) (ipset v6.14).
+# for shorewall
+#
+# Use this setting in jail.conf to modify the name and the max time in every jain
+# action   = shorewall-ipset-proto6[name=SSH, bantime=10000]
```
yarikoptic: why just not to say to set banaction = shorewall-ipset-proto6 ?

config/action.d/shorewall-ipset-proto6.conf
```diff
@@ -0,0 +1,84 @@
...
+#
+# The default Shorewall configuration is with "BLACKLISTNEWONLY=Yes" (see
+# file /etc/shorewall/shorewall.conf). This means that when Fail2ban adds a
+# new shorewall rule to ban an IP address, that rule will affect only new
+# connections. So if the attempter goes on trying using the same connection
+# he could even log in. In order to get the same behavior of the iptable
+# action (so that the ban is immediate) the /etc/shorewall/shorewall.conf
+# file should me modified with "BLACKLISTNEWONLY=No".
+#
+# A lot of this depends of the shorewall version at least have the 4.5.5 Version
```
yarikoptic: does it mean that BLACKLISTNEWONLY was introduced in 4.5.5 or ... ?

config/action.d/shorewall-ipset-proto6.conf
```diff
@@ -0,0 +1,84 @@
...
+#
+# Enable to shorewall use a blacklist using iptables creating a file /etc/shorewall/blrules
```
yarikoptic: Enable shorewall to

config/action.d/shorewall-ipset-proto6.conf
```diff
@@ -0,0 +1,84 @@
...
+# and adding "DROP     net:+f2b-ssh    all" one for every jail
+# details in shorewall documentation blacklist.
+# to enable restore you ipset You must set SAVE_IPSETS=Yes in shorewall.conf
+# Is importan to read this documentation from shorewall http://shorewall.net/ipsets.html
```
yarikoptic: importan[t]   ... and it would be great if above 4 lines were made full sentences
NB thanks for being descriptive BTW!

config/action.d/shorewall-ipset-proto6.conf
```diff
@@ -0,0 +1,84 @@
...
+# /etc/shorewall/initdone and add one line for every ipset (this files are in Perl)
+# take care of add 1; at the end of the file
+# the example is:
+# system("/usr/sbin/ipset -quiet -exist create f2b-ssh hash:ip timeout 600 ");
+# 1;
+#
+# To destroy the ipset in shorewall you must add to the file /etc/shorewall/stopped
+# # One line of every ipset
+# system("/usr/sbin/ipset -quiet destroy f2b-ssh ");
+# 1; # This must go to the end of the file if not shorewall complilation fails
```
yarikoptic: compilation

config/action.d/shorewall-ipset-proto6.conf
```diff
@@ -0,0 +1,84 @@
...
+#
+# A lot of this depends of the shorewall version at least have the 4.5.5 Version
+#
+# Enable to shorewall use a blacklist using iptables creating a file /etc/shorewall/blrules
+# and adding "DROP     net:+f2b-ssh    all" one for every jail
+# details in shorewall documentation blacklist.
+# to enable restore you ipset You must set SAVE_IPSETS=Yes in shorewall.conf
+# Is importan to read this documentation from shorewall http://shorewall.net/ipsets.html
+#
+# To force create the ipset in the case that somebody delete the ipset create a file
```
yarikoptic: why somebody should have deleted the ipset???  are there multiple admins?

yarikoptic:
Thanks for the contribution.  just to make sure -- you do use this action successfully on your boxes?

config/action.d/shorewall-ipset-proto6.conf
ediazrod: you must use a name and a bantime to pass this parameters to the aplication you create two ipset one for every jail

config/action.d/shorewall-ipset-proto6.conf
ediazrod: BLACKLISTNEWONLY is very util, if are yes only ban the new connexions if are in no ban all.

This is importan con avoid for example some DDOS if you bann all the conexions the conexion open to ssh for example will keep open until the system closes the conexion in a big DOS you can keep a big about conexion to close and the system can have problems open more process..

Some time is better only blacklist new only.

config/action.d/shorewall-ipset-proto6.conf
ediazrod: I don't know some time the administrator delete the ipset if the ipset is deleted shorewall don't start, I understand that is a problem of shorewall but anyway

ediazrod:
Yes I am using in my boxes with shorewall. Is feature is more efficient that the shorewall.conf in a big DOS start and stop fail to ban can take 15 minutes to delete all the ips in shorewall.conf with this .conf file using ipset take 1 minute to delete 5000 ips. Is a improvement.

yarikoptic:
then I could only ask to
- merge current state of master
- update Changelog
- add yourself to THANKS

config/action.d/shorewall-ipset-proto6.conf
yarikoptic: correct but! -- you can use string interpolations to pass those parameters borrowing variables values, thus we have
```
action_ = %(banaction)s[name=%(__name__)s, bantime="%(bantime)s", port="%(port)s", protocol="%(protocol)s", chain="%(chain)s"]
```
so it should be enough to say "banaction = shorewall-ipset-proto6".  I will do that change

yarikoptic:
@leeclemens there was a confusion -- proto6 is not about ipv6 , it is about ipset protocols (thus we have already iptables-ipset-proto4.conf iptables-ipset-proto6.conf) thus removed the ipv6 tag

leeclemens:
@yarikoptic you're right, overzealous tagging on my part :)

PR #979, opened 2015-02-26
Flush logs at USR1 signal

tgyurci:
FreeBSD's newsyslog can not run a command at log rotation, so it can not invoke `fail2ban-client flushlogs`, but it can send a configurable signal to a process.


sebres:
__sigUSR1Handler vs __sigUSR1handler ( write "h" in lowercase, python is case sensitive :)

yarikoptic:
just to make sure -- for consistency with other handlers -- should be lower case... I will just merge and fix up. thanks

PR #985, opened 2015-03-06
BF: asyncore.loop poll=True for recent (>=3.4) pythons too

yarikoptic:
should avoid
  File /usr/lib/python3.4/asyncore.py, line 208, in loop
    poll_fun(timeout, map)
  File /usr/lib/python3.4/asyncore.py, line 145, in poll
    r, w, e = select.select(r, w, e, timeout)
OSError: [Errno 9] Bad file descriptor

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,8 +32,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    * filter.d/dovecot.conf
      - match pam_authenticate line from EL7
      - match unknown user line from EL7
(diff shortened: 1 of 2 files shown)


sebres:
Hmm, very strange... As I worked on #918, I thought this python "bug" <a href="http://bugs.python.org/issue10878">issue10878</a> would not be fixed, because "2.7 is closed to enhancements". 
But 3.4?
Could be, at the moment, the poll should be used for any python versions?

yarikoptic:
could well be.  Let's just proceed with poll for now

PR #986, opened 2015-03-06
BF: use daily build of pypy until next one gets released

yarikoptic:


yarikoptic:
issue still persists
https://travis-ci.org/fail2ban/fail2ban/jobs/55317532
so merging

PR #996, opened 2015-03-21
BF/TST: expect ['error'] as a value if no dns module available

yarikoptic:
As many buildbots show:
e.g. http://nipy.bic.berkeley.edu/builders/fail2ban-py2.7-osx-10.8_master/builds/163/steps/shell_1/logs/stdio


yarikoptic:
all good ones passed -- merging

PR #997, opened 2015-03-22
DOC: make a warning for recidive jail to increase dbpurgeage (Closes #964)

### LINKED ISSUE #964: It seems default dbpurgeage have too small interval
Just compare
```
# Values: [ SECONDS ] Default: 86400 (24hours)
dbpurgeage = 86400
```
and
```
[recidive]
logpath = /var/log/fail2ban.log
banaction = iptables-allports
bantime = 604800 ; 1 week
```
Perhaps I misunderstand something, but dbpurgeage needs to be increased, like 1 week + 10 minutes (room for cleanup). We lost those entries from db otherwise.

yarikoptic:


yarikoptic:
no objections -- merging

PR #1001, opened 2015-03-24
Match hacking attempt IP instead of asterisk server IP (closes #1000)

### LINKED ISSUE #1000: asterisk regex bans asterisk server instead of attacker
From mailing list conversation: http://sourceforge.net/p/fail2ban/mailman/message/33603322/

It seems the 'Sending fake auth rejection for device' message is the server, not malicious client.

[2015-03-16 18:46:34] NOTICE[3453] chan_sip.c: hacking attempt
detected 'a.b.c.d'
[2015-03-16 18:46:34] NOTICE[3453] chan_sip.c: Sending fake auth
rejection for device
30001<sip:30001@<asterisk-machine-ip>>;tag=8314aeaa

However, I cannot confirm if the same is for the 'Sending fake auth rejection for user' message, but it seems likely.

leeclemens:
Sending fake auth rejection for (device|user) logs use the asterisk server's public IP (per referenced URLs in previous sample logs file. Replaced the regex with one matching (hacking attempt) log mentioned on mailing list.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -40,6 +40,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    * firewallcmd-* actions: split output into separate lines for grepping (gh-908)
    * Guard unicode encode/decode issues while storing records in the database.
      Fixes "binding parameter error (unsupported type)" (gh-973), thanks to kot for reporting
+   * filter.d/asterisk.conf - Match hacking attempt IP instead of asterisk server IP
 
 - New Features:
    - New filters:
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -40,6 +40,7 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    * firewallcmd-* actions: split output into separate lines for grepping (gh-908)
    * Guard unicode encode/decode issues while storing records in the database.
      Fixes "binding parameter error (unsupported type)" (gh-973), thanks to kot for reporting
+   * filter.d/asterisk.conf - Match hacking attempt IP instead of asterisk server IP
```
yarikoptic: better be more explicit here:  dropped "Sending fake auth rejection" failregex since it incorrectly targets the originating server?  But I would need to better grasp the source/semantic of this msg . 

added catching "hacking attempt detected"

leeclemens: The removed failregex logs the asterisk server's IP address versus the new failregex which logs the remote client/bad guy/IP we want fail2ban to ban.

I wasn't sure of a concise way to reference the bad guy's IP in the scenario, so I used "hacking attempt IP" - granted it's not the best phrase.

I'd say the dropped failregex incorrectly bans the server fail2ban is protecting, in this case the asterisk server itself. I'm not sure 'originating server' is especially clear either.

leeclemens: @yarikoptic  Thoughts on:

* filter.d/asterisk.conf - Dropped "Sending fake auth rejection" failregex since it incorrectly targets the asterisk server
* filter.d/asterisk.conf - Added matching "hacking attempt detected" logs

?

leeclemens:
@yarikoptic Had to merge Changelog, but I updated it using my last response to your comment

yarikoptic:
wow -- we are at the 1000 of issues/PRs -- congrats!  ;)

PR #1004, opened 2015-03-26
Fix of LC_TIME usage, it should be LC_ALL

tsabi:
Hi,
this branch fixes a bug in mail generation: i use linux with hungarian locale, and the time generation generates localed time, what breaks the mail processing.
This branch fixes this with the use of LC_ALL instead of LC_TIME.

example:
```
juno ~ # LC_TIME=C date +"%a, %d %h %Y %T %z"
cs, 26 márc 2015 03:04:22 +0100
juno ~ # LC_ALL=C date +"%a, %d %h %Y %T %z"
Thu, 26 Mar 2015 03:05:42 +0100
```

BR,
tsabi


yarikoptic:
That is strange since it should work (and does):

```
$> date                       
Tue Apr  7 21:48:41 EDT 2015

$> LC_TIME=ru_RU.UTF-8 date 
Вт апр  7 21:48:44 EDT 2015
```

my guess is that you overrode LC_ALL which thus overrides LC_TIME.... what is your output of "locale"?

tsabi:
Yes, i override LC_ALL, because not i am the only one who uses this computer, and the others want to see the results as hungarian when they use sudo.

yarikoptic:
ok -- I am ok to accept it... Just add a ChangeLog entry plz

PR #1006, opened 2015-03-27
ENH: Character detection heuristics for whois output via optional setting in mail-whois*.conf (Closes #1003)

### LINKED ISSUE #1003: Correct MIME type and file extension for attached report file in emails
fail2ban is sometimes sending its report as a file attachment via email.

However, using mail clients such as Thunderbird on Gnu/Linux, the user is forced to store the file to disk an cannot open it directly.

The reason for that is a wrong MIME type:

    Subject: [Fail2Ban] somejail: banned XXX.XXX.XXX.XXX
    User-Agent: Heirloom mailx 12.5 6/20/10
    MIME-Version: 1.0
    Content-Type: application/octet-stream
    Content-Transfer-Encoding: quoted-printable

The reason for the wrong MIME type is either, that it's not explicitly set, or that the mail program cannot determine it because it has no file extension. In Thunderbird, there is no file name visible at all (user has to click on white space), and when opening, he or she is offered the server's name for the file name.

So I would suggest to give the file a name and the extension ".txt". Plus, the content type should be changed to "text/plain" (if not set automatically via server-side mail programm via file extension).

I'm using fail2ban 0.8.6 on up-to-date debian 7. I know this version is rather old. If the issue is already fixed in newer versions, then I would still appreciate a backport to the debian 7 packages.

thomaszbz:
when set by user,
 - detects character set of whois output (which is undefined by RFC 3912) via heuristics of the file command
 - converts whois data to UTF-8 character set with iconv
 - sends the whois output in UTF-8 character set to mail program
 - avoids that heirloom mailx creates binary attachment for input with unknown character set

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -78,7 +78,8 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    * Absorbed DNSUtils.cidr into addr2bin in filter.py, added unittests
    * Added syslogsocket configuration to fail2ban.conf
    * Note in the jail.conf for the recidive jail to increase dbpurgeage (gh-964)
-   * Character detection heuristics for whois output via optional setting in mail-whois*.conf. When set by user,
+   * Character detection heuristics for whois output via optional setting in mail-whois*.conf. Thanks Thomas Mayer
+     When set by user,
      - detects character set of whois output (which is undefined by RFC 3912) via heuristics of the file command
      - converts whois data to UTF-8 character set with iconv
      - sends the whois output in UTF-8 character set to mail program
--- a/ChangeLog
+++ b/ChangeLog
@@ -78,6 +78,11 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    * Absorbed DNSUtils.cidr into addr2bin in filter.py, added unittests
    * Added syslogsocket configuration to fail2ban.conf
    * Note in the jail.conf for the recidive jail to increase dbpurgeage (gh-964)
+   * Character detection heuristics for whois output via optional setting in mail-whois*.conf. When set by user,
+     - detects character set of whois output (which is undefined by RFC 3912) via heuristics of the file command
+     - converts whois data to UTF-8 character set with iconv
+     - sends the whois output in UTF-8 character set to mail program
+     - avoids that heirloom mailx creates binary attachment for input with unknown character set
 
 
 ver. 0.9.1 (2014/10/29) - better, faster, stronger
--- a/config/action.d/mail-whois-common.conf
+++ b/config/action.d/mail-whois-common.conf
@@ -15,9 +15,10 @@ _whois = whois <ip> || echo "missing whois program"
 
 # use heuristics to convert charset of whois output to a target
 # character set before sending it to a mail program
+# make sure you have 'file' and 'iconv' commands installed when opting for that
 _whois_target_charset = UTF-8
 _whois_convert_charset = whois <ip> |
-                         { c=$(cat) ; cs=$(printf %%b "$c" | file -b --mime-encoding -) ; printf %%b "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
+                         { WHOIS_OUTPUT=$(cat) ; WHOIS_CHARSET=$(printf %%b "$WHOIS_OUTPUT" | file -b --mime-encoding -) ; printf %%b "$WHOIS_OUTPUT" | iconv -f $WHOIS_CHARSET -t %(_whois_target_charset)s//TRANSLIT - ; }
 
 # choose between _whois and _whois_convert_charset in mail-whois-common.local
 # or other *.local which include mail-whois-common.conf.
--- a/config/action.d/mail-whois-common.conf
+++ b/config/action.d/mail-whois-common.conf
@@ -15,9 +15,10 @@ _whois = whois <ip> || echo "missing whois program"
 
 # use heuristics to convert charset of whois output to a target
 # character set before sending it to a mail program
+# make sure you have 'file' and 'iconv' commands installed when using this
 _whois_target_charset = UTF-8
 _whois_convert_charset = whois <ip> |
-                         { c=$(cat) ; cs=$(echo -ne "$c" | file -b --mime-encoding -) ; echo -ne "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
+                         { c=$(cat) ; cs=$(echo "$c" | file -b --mime-encoding -) ; echo "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
 
 # choose between _whois and _whois_convert_charset in mail-whois-common.local
 # or other *.local which include mail-whois-common.conf.
--- a/config/action.d/mail-whois-common.conf
+++ b/config/action.d/mail-whois-common.conf
@@ -15,10 +15,9 @@ _whois = whois <ip> || echo "missing whois program"
 
 # use heuristics to convert charset of whois output to a target
 # character set before sending it to a mail program
-# make sure you have 'file' and 'iconv' commands installed when using this
 _whois_target_charset = UTF-8
 _whois_convert_charset = whois <ip> |
-                         { c=$(cat) ; cs=$(echo "$c" | file -b --mime-encoding -) ; echo "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
+                         { c=$(cat) ; cs=$(echo -ne "$c" | file -b --mime-encoding -) ; echo -ne "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
 
 # choose between _whois and _whois_convert_charset in mail-whois-common.local
 # or other *.local which include mail-whois-common.conf.
--- a/config/action.d/mail-whois-common.conf
+++ b/config/action.d/mail-whois-common.conf
@@ -17,7 +17,7 @@ _whois = whois <ip> || echo "missing whois program"
 # character set before sending it to a mail program
 _whois_target_charset = UTF-8
 _whois_convert_charset = whois <ip> |
-                         { c=$(cat) ; cs=$(echo -ne "$c" | file -b --mime-encoding -) ; echo -ne "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
+                         { c=$(cat) ; cs=$(printf %%b "$c" | file -b --mime-encoding -) ; printf %%b "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
 
 # choose between _whois and _whois_convert_charset in mail-whois-common.local
 # or other *.local which include mail-whois-common.conf.
--- a/config/action.d/mail-whois-common.conf
+++ b/config/action.d/mail-whois-common.conf
@@ -20,5 +20,8 @@ _whois_convert_charset = whois <ip> |
                          { c=$(cat) ; cs=$(echo -ne "$c" | file -b --mime-encoding -) ; echo -ne "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
 
 # choose between _whois and _whois_convert_charset in mail-whois-common.local
+# or other *.local which include mail-whois-common.conf.
 _whois_command = %(_whois)s
 #_whois_command = %(_whois_convert_charset)s
+
+[Init]
--- a/config/action.d/mail-whois-common.conf
+++ b/config/action.d/mail-whois-common.conf
@@ -1,0 +1,24 @@
+# Fail2Ban configuration file
+#
+# Common settings for mail actions
+#
+# Users can override the defaults in mail-whois-common.local
+
+[INCLUDES]
+
+# Load customizations if any available
+after = mail-whois-common.local
+
+[DEFAULT]
+#original character set of whois output will be sent to mail program
+_whois = whois <ip> || echo "missing whois program"
+
+# use heuristics to convert charset of whois output to a target
+# character set before sending it to a mail program
+_whois_target_charset = UTF-8
+_whois_convert_charset = whois <ip> |
+                         { c=$(cat) ; cs=$(echo -ne "$c" | file -b --mime-encoding -) ; echo -ne "$c" | iconv -f $cs -t %(_whois_target_charset)s//TRANSLIT - ; }
+
+# choose between _whois and _whois_convert_charset in mail-whois-common.local
+_whois_command = %(_whois)s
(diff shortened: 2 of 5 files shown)


sebres:
- you should add a [Init] section to pass test-cases (and next time you can run fail2ban-testcases before push to test all that local ```cd f2b; PYTHONPATH=. bin/fail2ban-testcases ?-l debug|heavydebug?```);
- the redefinition of `_whois_command` - "choose between _whois and _whois_convert_charset" may be everywhere (in all configs included that, not in mail-whois-common.local only);
- LBNL: don't forget THANKS.

sebres:
> You mean Init before the selection?

No, the init block just should be present inside (possible only in test-cases). Look for example at  `fail2ban/config/action.d/sendmail-common.conf` (another include but for sendmail's configs).

> Which place (file) ...

The place for vars is all right, what I mean: you can extend a comment, something like:
```
# choose between _whois and _whois_convert_charset in mail-whois-common.local 
# or other *.local included this one.
```

> Is it sufficient to run the tests without compiling current master? 

Nothing except python is needed to test it.
Just `cd /path/to/f2b-cloned-repo` and run `bin/fail2ban-testcases`

thomaszbz:
fail2ban-testcases reports 2 errors in current master (c926af1fce4833ee22d74f57f49f810ee002af60 or before).

Adding the empty [Init] section removed a regression from c9b24839e4338059224dcc743eb7f458f850f18e.

thomaszbz:
I did some testing: One of the '-ne' options is in the E-Mail output. That said, it's not interpreted and I can hopefully safely remove it.

thomaszbz:
Build pypy failed - without reason?

sebres:
The last time it's normal (some JIT error, I didn't have a time to look at closer).

szepeviktor:
Will sendmail accept UTF-8 characters?

thomaszbz:
@szepeviktor Basically it could happen any time that fail2ban sends UTF8 WHOIS output to sendmail or mailx, because the WHOIS character set is undefined (originally it was ASCII, but nobody respected that) (read #1003 ).

This PR targets problems when using hairloom-mailx (read #1003 ). It solves the problem that WHOIS output has an undefined character set while hairloom-mailx expects it somehow defined. Plus, the new config needs to be activated by user in his config files. Currently, we don't default to the new behaviour because of new dependencies (file, iconv).

Would you like to check what happens when applying it to a sendmail configuration? Feel free to enhance #1006, but use a new issue/branch/PR, please. You can still start with 923d807 . I did not test it but in general I would like to see a defined character set for WHOIS output for sendmail as well. You can still set a target set other than UTF-8 in the config (but... why should you?).

szepeviktor:
I think everyone will prefer ASCII as the default.

szepeviktor:
One more thing: loglines could also contain non-ASCII characters.

thomaszbz:
@szepeviktor That's true, because ASCII by convention would be the easy solution. However, it does not meet reality as most people on this planet are not Anglo-Americans and want to participate in their local languages and character sets. That must be respected and a technical concept for that is most of the time the use of UTF-8.

I get WHOIS output like

    TELEF�NICA BRASIL S.A (in ISO-8859-1, displayed in UTF-8 terminal)
    บริษัท ฟีนิกซ์ พิซซ่า จำกัด (in UTF-8)

which looks wrong and ugly compared to

    TELEFÔNICA BRASIL S.A (converted to UTF-8)
    บริษัท ฟีนิกซ์ พิซซ่า จำกัด (still in UTF-8)

There is no representation for many (or better: *most*) characters in ASCII. Keep in mind that both examples came directly from whois command. There's no way to change that unless some miracle happens (e.g. UTF-8 becomes standard for WHOIS).

thomaszbz:
@szepeviktor for the log lines: You are right, that could be another issue. I assume that most of the log output is written in UTF-8 already (when systems are configured in UTF-8). In that case, whois output, strings from fail2ban and log output are all in UTF-8. However there is indeed no guarantee for that.

Still, it's up to the administrator to avoid character set mismatch for the moment. Be it converting log files, configuring system and applications to use UTF-8 for logging, ...

Administrators should decide which character set the need for sending mails. The only thing they need to consider is that they should avoid character set mismatch (e.g. ASCII in strings, ISO-8859-1 in log files and UTF-8 for the whois output). For the whois output (which this PR is about), they just need to change UTF-8 to something else in case their system can't use UTF-8. But I think nobody really wants that in 2015.

szepeviktor:
A lot of servers live on the '90-s.

szepeviktor:
For example proftpd needs in `/etc/default/proftpd`
```
export LANG="en_US.UTF-8"
```

thomaszbz:
Again, log file character set is not touched by this PR at all. Feel free to open a new issue for log file character set.

The example you mention basically confirms what I wrote: It's up to the admistrator to care for proper character set definition. Plus, I don't think log file character set is something that needs to be handled by fail2ban. If servers "live in the '90s" because of misconfiguration, so be it.

szepeviktor:
I hope free speech is not *banned* here.

thomaszbz:
@szepeviktor I think the whole discussion you kicked off is just at the wrong place here. At this place, you can discuss if PR #1006 is making it into fail2ban or not. Everything else is off-topic here. If you want to discuss something else, feel free to open a new issue.

I found nothing in your comments which contradicts merging PR #1006.

szepeviktor:
No, I don't contradict merging your PR. I just notice things.

szepeviktor:
@rfc1036 could you help us?

szepeviktor:
@thomaszbz Turns out most whois servers already talk UTF-8: https://github.com/rfc1036/whois/blob/a509f3b1e7a79090d187a5b07d819bbb0ccab1d2/servers_charset_list

szepeviktor:
https://github.com/rfc1036/whois/blob/c9af7b9cb6645f29dfe0d8001db7137f89535226/debian/changelog#L165

thomaszbz:
@szepeviktor "most servers" does not equal "*all* servers". Plus, even if all WHOIS servers spoke UTF-8, you cannot rely on that unless it's standard (via RFC 3912). Please read #1003 for that.

Plus, I'm not sure if there is no character set mismatch in the WHOIS data itself. I get ISO-8859-1 data from DENIC via WHOIS although, after your list, DENIC claims to speak UTF-8.

Basically, this means that you still cannot rely on fixed character sets in WHOIS data, at least for the near future (5 years+).

szepeviktor:
I think @rfc1036 will tell us the same.

szepeviktor:
BTW. I use [GeoIP database](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/sendmail-geoip-lines.conf) which is local opposite to whois.

thomaszbz:
Anyways, RFC 1036 (written in 1987!) is outdated in favour of RFC 5536 and 5537 according to http://fr.wikipedia.org/wiki/RFC_1036 (french has been the language of diplomacy in ~1987 and before). Plus, Usenet and News have not much to do with WHOIS and this PR.

Same holds true for GeoIP which is a different approach with a different featureset.

szepeviktor:
This RFC says WHOIS has no internationalization: https://tools.ietf.org/html/rfc3912#section-4

thomaszbz:
@szepeviktor That's what I wrote and cited on 2015-03-25 in #1003. Nothing new, nothing to be written again here.

rfc1036:
If you use my client (like all current Linux distributions do. Ports are available for every other relevant OS) and export to its environment a UTF-8 locale (en_US.UTF-8 is a safe choice) in $LANG or $LC_CTYPE then you will get UTF-8 for every server whose charset is known to the program.

Servers not in the list, which is reasonably complete and accurate, will return an unspecified charset depending on the object requested and there is nothing you can do about this.
Just running my client without a proper locale configured will return an unspecified charset.
Any other client AFAIK will return an unspecified charset.

Unless you use my client with a UTF-8 locale then you cannot assume that you will get UTF-8 output from a query. Even this may not be enough, but I have been working on this for over 15 years and I am not aware of a more complete solution.

As a general solution I recommend to:
* configure the MUA to expect UTF-8
* run the whois client with LC_CTYPE=en_US.UTF-8
* maybe sanitize any output that is not valid UTF-8

thomaszbz:
@rfc1036: The last step ("maybe sanitize any output that is not valid UTF-8") is exactly what this PR is about.

rfc1036:
But unless you pass a UTF-8 locale to the client you will get suboptimal results: for the most common cases there is no need to guess.

thomaszbz:
I have

    echo $LANG
    en_US.UTF-8

Please read #1003 for examples and further explanation first.

thomaszbz:
$LC_TYPE also seems not to solve this issue:

    # echo $LANG
    en_US.UTF-8
    # echo $LC_TYPE
    en_US.UTF-8
    # whois 179.111.212.221|file -
    /dev/stdin: ISO-8859 text

rfc1036:
But it does. The LACNIC whois server is a good example of why you need a transcoding client since its output is Latin-1.

I will assume that you are either not actually exporting the environment variable(s) or have a somehow broken or more complex locales configuration than the one you are showing here.

thomaszbz:
As far as I know, there's nothing special with my installation. $LANG is defined via debian wheezy's update-locale in /etc/default/locale. Actually, that is nothing special. And I have

    locale -a
    C
    C.UTF-8
    de_DE.utf8
    en_US.utf8
    POSIX


What is *your* output of

    whois 179.111.212.221|file -

? I'm still using the default debian package
    whois --version
    Version 5.1.1.

Is it too old for handling character sets right? Still, we cannot expect something newer to be used by fail2ban.

rfc1036:
whois.lacnic.net was added to the charsets table in 5.1.4.

szepeviktor:
I get very nice text:
```
person:      Administração Rede IP Telesp
```
I think only `whois $IP | iconv -f utf-8 -t utf-8 -c` is enough. It'll drop all non-utf8 characters as @rfc1036 [recommended](maybe sanitize any output that is not valid UTF-8).

thomaszbz:
@szepeviktor rfc1036 recommended to "sanitize any output that is not valid UTF-8" which is different from just dropping something. Again, you did not read what you were asking for. Plus, your "from" (-f) option might not correspond to the input character set in some cases nor is it trying to do so. The heuristics of this PR is handling that ways cleaner.

I don't mind if you play around here if you can't find a better place to write what you do or think. You can also invite others again and talk to them. But please don't expect me to answer your questions if you don't read my answers.

szepeviktor:
OK-OK I am sorry for my comments. I didn't mean to ask a question.

ChangeLog
```diff
@@ -78,6 +78,12 @@ ver. 0.9.2 (2014/XX/XXX) - wanna-be-released
    * Absorbed DNSUtils.cidr into addr2bin in filter.py, added unittests
    * Added syslogsocket configuration to fail2ban.conf
    * Note in the jail.conf for the recidive jail to increase dbpurgeage (gh-964)
+   * Character detection heuristics for whois output via optional setting in mail-whois*.conf. Thanks Thomas Mayer
+     When set by user,
+     - detects character set of whois output (which is undefined by RFC 3912) via heuristics of the file command
+     - converts whois data to UTF-8 character set with iconv
+     - sends the whois output in UTF-8 character set to mail program
+     - avoids that heirloom mailx creates binary attachment for input with unknown character set
```
yarikoptic: "avoids [...] creat[ing]" right?

config/action.d/mail-whois-lines.conf
```diff
@@ -4,6 +4,10 @@
 # Modified-By: Yaroslav Halchenko to include grepping on IP over log files
 #
 
+[INCLUDES]
+
+before = mail-whois-common.conf
```
yarikoptic: what about naming this one "mail-whois-autodetect.conf" and then having "mail-whois-basic.conf" which would simply define ```_whois_command = whois``` so if ran on systems e.g. without iconv and who knows what else (didn't try runing this one on OSX or some other systems yet), then could switch?  and add a comment there above about that?  I am ok to have -autodetect by default ;)

config/action.d/mail-whois-common.conf
```diff
@@ -0,0 +1,28 @@
...
+# character set before sending it to a mail program
+# make sure you have 'file' and 'iconv' commands installed when opting for that
+_whois_target_charset = UTF-8
+_whois_convert_charset = whois <ip> |
+                         { WHOIS_OUTPUT=$(cat) ; WHOIS_CHARSET=$(printf %%b "$WHOIS_OUTPUT" | file -b --mime-encoding -) ; printf %%b "$WHOIS_OUTPUT" | iconv -f $WHOIS_CHARSET -t %(_whois_target_charset)s//TRANSLIT - ; }
+
+# choose between _whois and _whois_convert_charset in mail-whois-common.local
+# or other *.local which include mail-whois-common.conf.
+_whois_command = %(_whois)s
+#_whois_command = %(_whois_convert_charset)s
```
yarikoptic: ah -- rright -- that is where you are choosing between "autodetect" and "not".  Indeed may be it is even better here in this single place!

Could you come up with a basic unittest which would try to exec this _whois_command for some "known to be problematic" domain? then we could verify its operation.

thomaszbz: The whois command needs to be mocked then. We do not want to test whois here, because it depends of the whois version (and a corresponding list of whois servers and their default character set) which character set whois is requesting and most of the time converting to $LANG character set itself (as I learned recently, but I don't rely on that). Basically that means that some example output should be used for a unit test instead of executing whois during the tests.

I'm not sure if I can realize that because it brings me much deeper in the world of fail2ban and python than just the configuration change.

ChangeLog
thomaszbz: I guess the whole construction "avoid so. do sth. is wrong". http://dict.leo.org/#/search=avoid&searchLoc=0&resultOrder=basic&multiwordShowSingle=on

What about "avoids binary being attached by heirloom mailx for input with unknown character set"?

config/action.d/mail-whois-lines.conf
thomaszbz: -autodetect and -basic sounds reasonable (and consistent to the way existing .conf files are structured).

For the moment, I did not default to -autodetect because of new dependencies (file, iconv). That said, please tell me if you still want to default to -autodetect.

ChangeLog
yarikoptic: not a native speaker here ;)  but never saw such a use in my non-native english speaking career ;)

config/action.d/mail-whois-lines.conf
yarikoptic: ok -- let's leave as is since switching between those might be trickier than just overriding a variable with .local file.  I am ok to keep save default, but also ok to make default more powerful and rely on systems having those core tools.

yarikoptic:
ok -- it seems that discussion brew down to that this PR would be of help and should do no harm ;)  So, I will stay positive -- merging now manually. Thanks everyone

PR #1014, opened 2015-04-06
Add missing ignoreregex lines to avoid warnings

opoplawski:


yarikoptic:
ah -- 0.8, right. thanks

PR #1015, opened 2015-04-06
Add missing ignoreregex lines to avoid warnings

opoplawski:
Conflicts:
	config/filter.d/ejabberd-auth.conf


yarikoptic:
confused why travis fails for this one ... will check tomorrow

yarikoptic:
well -- we actually were not "missing" any in the current state of master.  and the one for postfix-sasl overrode proper one thus causing the test to fail... since this is not really anything worth merging -- I will just close.  Thanks anyways though ;)

PR #1024, opened 2015-04-16
updated filter.d/sshd.conf

themanwhosold:
Added line to match sshd auth errors on OpenSuSE systems

### DIFF
--- a/config/filter.d/sshd.conf
+++ b/config/filter.d/sshd.conf
@@ -33,6 +33,7 @@ failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|erro
             ^(?P<__prefix>%(__prefix_line)s)User .+ not allowed because account is locked<SKIPLINES>(?P=__prefix)(?:error: )?Received disconnect from <HOST>: 11: .+ \[preauth\]$
             ^(?P<__prefix>%(__prefix_line)s)Disconnecting: Too many authentication failures for .+? \[preauth\]<SKIPLINES>(?P=__prefix)(?:error: )?Connection closed by <HOST> \[preauth\]$
             ^(?P<__prefix>%(__prefix_line)s)Connection from <HOST> port \d+(?: on \S+ port \d+)?<SKIPLINES>(?P=__prefix)Disconnecting: Too many authentication failures for .+? \[preauth\]$
+             ^%(__prefix_line)spam_unix\(sshd:auth\):\s+authentication failure.*rhost=<HOST>\s.*$
 
 ignoreregex = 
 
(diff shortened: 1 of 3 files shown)


sebres:
Thx for you contribution. 
But by extending a filter expected at least one sample test case for new regexp (see https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/sshd). Otherwise have :boom: in travis. :)

themanwhosold:
Sorry, I'll fix my mistakes.

Stupid question, how do I get the "# failJSON"-Line above the sample test case? Do I have to write it by myself or is it generated by Travis?

sebres:
you should set proper failJSON (time, ip) - will be checked in tests.
you can self test it (local) - to test samples only, just run:
```bash
PYTHONPATH=. bin/fail2ban-testcases FilterSamplesRegex
```

themanwhosold:
Finally fixed my mistakes. Sorry I failed to notice the readme file which describes what to do.

leeclemens:
No worries, the addition is much appreciated. Just a minor typo with "Authentification" :)

config/filter.d/sshd.conf
```diff
@@ -33,6 +33,7 @@ failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|erro
             ^(?P<__prefix>%(__prefix_line)s)User .+ not allowed because account is locked<SKIPLINES>(?P=__prefix)(?:error: )?Received disconnect from <HOST>: 11: .+ \[preauth\]$
             ^(?P<__prefix>%(__prefix_line)s)Disconnecting: Too many authentication failures for .+? \[preauth\]<SKIPLINES>(?P=__prefix)(?:error: )?Connection closed by <HOST> \[preauth\]$
             ^(?P<__prefix>%(__prefix_line)s)Connection from <HOST> port \d+(?: on \S+ port \d+)?<SKIPLINES>(?P=__prefix)Disconnecting: Too many authentication failures for .+? \[preauth\]$
+             ^%(__prefix_line)spam_unix\(sshd:auth\):\s+authentication failure.*rhost=<HOST>\s.*$
```
yarikoptic: There is a problem here that we have ```.*``` on both sides of the ```<HOST>``` and after HOST there is a "user" entry, which would be what user provides... and since first .* is greedy one I am afraid that via "user" field he would be able to inject more than necessary. But anchoring at the end also might not be as efficient so I see only that you need to enter all those previous fields before rhost:
```;\s*logname=\S*\s*uid=\d*\s*euid=\d*\s*tty=\S*\s*ruser=<HOST>.*$```

please try

themanwhosold: Fixed it, thx!

Issue #980, opened 2015-02-27
Handling around iptables should be breakdown safe


sebres:
Just as reminder for me.

To prevent such errors like `iptables: Too many links.` by stop/start after invalid banning / breakdown of f2b:
- (start) create chain only if it not already exists;
- (stop) removing of the chain should take place, also if some rules of it still available (or some rules should be removed if no references available anymore);
- (reload/restart) ensure that changing of parameters in configs (like port by multiport action) does not affect execution of commands like `actionstop` (to replace old rules with old port);
  at the moment because as identifier all parameters are used, it would be unable to remove old iptables chain from INPUT, so leave still one rule with the same name (but old port);
- (any) check a possibility to implement race-safe handling - although we use `-w` option, it is still open to a time-of-check-to-time-of-use race condition (TOCTTOU-RC): 
  * f2b check the rule (and for example if not available should create the chain/rule);
  * some foreign service removes it after check;
  * some further f2b handler (like create) could be not invoked (due to successful check).

  typical scenario showing TOCTTOU-RC (although during each single operation it is locked (`-w`), the whole script may fail if both running simultaneously, there are no common lock/unlock due to missing locking-capabilities in shell):
```bash
# process 1:
iptables -w -C INPUT -p tcp -j f2b-jail &>/dev/null || {
  iptables -w -N f2b-jail || :
  iptables -w -A f2b-jail -j DROP || :
  iptables -w -I INPUT -p tcp -j f2b-jail
}
# process 2:
iptables -w -C INPUT -p tcp -j f2b-jail &>/dev/null && {
  iptables -w -D INPUT -p tcp -j f2b-jail
  iptables -w -X f2b-jail
}
```

sebres:
Possible the same or similar issue occurred in branch "ban-time-incr":
```
2015-02-09 10:42:56,285 fail2ban.action         [7401]: ERROR   iptables -n -L INPUT | grep -q 'f2b-ssh[ \t]' -- returned 1
2015-02-09 10:42:56,285 fail2ban.CommandAction  [7401]: ERROR   Invariant check failed. Trying to restore a sane environment
2015-02-09 10:42:56,389 fail2ban.action         [7401]: ERROR   iptables -D INPUT -p tcp --dport ssh -j f2b-ssh
iptables -F f2b-ssh
[...]

PR #1027, opened 2015-04-19
Add unban script for iptables

kyle95wm:


kyle95wm:
Does this mean you're not merging my pr?

kyle95wm:
I don't understand the whole "checks" thing. All I want is to make it easier for people to un-ban people from IPTABLES

yarikoptic:
but we  have already fail2ban-client unban,  so why  this iptables  specific one?

kyle95wm:
are you sure? I've tried every possible way but my script does what I want. Whats the syntax exactly?

kyle95wm:
Okay nevermind I just used the build from the repo here and I saw there was an unban option looks like the debian package is outdated by a lot. Just one question before we close this, is there a way to uninstall if for whatever reason we need to?

sebres:
possible a "duplicate" issue of #980

yarikoptic:
@sebres -- I don't really see how this one (unban action) relates to 980.
@kyle95wm backports for debians of fail2ban are available from the neurodebian (unannounced -devel carries 0.9. series as well) see e.g http://sourceforge.net/p/fail2ban/mailman/message/32143078/

uninstall -- would depend on how you installed it so can't give you exhaustive answer... but if you just "setup.py install"ed it -- you would need to manually remove   /usr/local/*bin/*fail2ban* and wherever it installed the module under /usr/local/lib somewhere -- just search for fail2ban

sebres:
@yarikoptic I thought that stop of f2b-service resp. unban an IP, had left any leaks in iptables...

PR #1033, opened 2015-04-26
Add drupal-auth filter and jail

leeclemens:


leeclemens:
Don't merge, having issues with travis not updating so prompting a pull request to get tests to run again

config/filter.d/drupal-auth.conf
```diff
@@ -0,0 +1,28 @@
...
+[INCLUDES]
+
+before = common.conf
+
+
+[Definition]
+
+_daemon =
+
+failregex = ^%(__prefix_line)s.+\|\d{10}\|user\|<HOST>\|.+\|.+\|\d\|.?\|Login attempt failed for .+\.$
```
yarikoptic: `<HOST>` is floating inbetween `.+`s -- not good... I smell possible injection especially since we are monitoring syslog . Please make leading `.+` way more specific (pattern for url with no spaces etc)

leeclemens: Thanks for the input, I'll take a look tomorrow and squash these commits. Not sure why Travis was lagging behind (and seems to assume dates are from 10 years ago?).

leeclemens: I think I made it specific enough, still `.+`s after `<HOST>`, but I'm not sure what they may be (e.g. referrer, link). Also found some info that said `base_url` will never have a trailing slash, but can be a subdirectory, so I tweaked a sample log for the subdir (and tested the trailing slash exclusion myself). Ref #1035

yarikoptic: I still see .+  right after  prefix and  before  HOST

leeclemens: Sorry if I made it more confusing, I deleted this branch and created PR #1035 with a revised failregex

yarikoptic:
gotcha
 imho usually it is better to push to the same branch/pr  even if you have to rewrite/ force push

leeclemens:
I'll keep that in mind :) Frustrated with Travis not updating last night and the 10 years thing I just noticed, but I see your point on keeping the same branch/pr

PR #1038, opened 2015-04-29
BF: download/use 2.5.1 release of PyPy to avoid problems with dailies

yarikoptic:


yarikoptic:
travis is happy I am happy

sebres:
we all are happy :smile:

PR #1050, opened 2015-05-10
BF: guarantee order of dbfile to be before dbpurgeage (Closes #1048)

### LINKED ISSUE #1048: dbpurgeage won't be honored when be set in fail2ban.local
Setting dbpurgeage = 691200 in fail2ban.local has no effect.
fail2ban-client get dbpurgeage still respond with the Default value of 86400seconds

yarikoptic:
I think this should fix #1048 but need to run atm so can't check "live". may be later

### DIFF
--- a/fail2ban/client/fail2banreader.py
+++ b/fail2ban/client/fail2banreader.py
@@ -53,17 +53,19 @@ class Fail2banReader(ConfigReader):
 	
 	def convert(self):
 		stream = list()
+		# Ensure logtarget/level set first so any db errors are captured
+		# Also dbfile should be set before dbpurgeage.
+		# So adding order indices into items, to be stripped after sorting, upon return
 		for opt in self.__opts:
 			if opt == "loglevel":
-				stream.append(["set", "loglevel", self.__opts[opt]])
+				stream.append((3, ["set", "loglevel", self.__opts[opt]]))
 			elif opt == "logtarget":
-				stream.append(["set", "logtarget", self.__opts[opt]])
+				stream.append((2, ["set", "logtarget", self.__opts[opt]]))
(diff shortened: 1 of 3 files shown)


sebres:
I have months ago fixed it,
see https://github.com/fail2ban/fail2ban/pull/716/files#diff-be2e644f6b6fe295b29c242518fadb7fR53

Seconds, I've said already several times, purge will never called (only in test case):
https://github.com/fail2ban/fail2ban/search?utf8=%E2%9C%93&q=%22purge%28%29%22&type=Code

But also already fixed in #716

yarikoptic:
ah, cool ,, indeed -- so what about finalizing that #716 so we could just take that solution instead? ;-)

sebres:
> what about finalizing that #716
It's finalized imho long time ago :) But I'm ready to listen arguments why not...

yarikoptic:
I think I left a bunch of comments which weren't yet addressed.  And it is a big change so some of them are quite important (we need to test regressions etc), post-release master needs to be merged as well etc

yarikoptic:
if #716 doesn't get finalized, may be we should extract at least from it the fix for this issue @sebres ?

sebres:
Well, no problem, just the solution is prorated in new observer-module, that a large part of #716.

yarikoptic:
@sebres would you be interested to submit an excerpt from #716 (or whatever is the latest) as a fix for #1048? I guess if you do it, it might minimize conflicts with your branches

Cheers

sebres:
If this one https://github.com/fail2ban/fail2ban/commit/80fb58110f152d24a46453724ca08b0d32aa8eda is a good enough, I can cherry pick it in master.

PR #1051, opened 2015-05-10
Update regex to work with roundcube 1.0.5 and 1.1.1

leeclemens:
Updated regex to work with roundcube 1.0.5 and 1.1.1 'errors' log lines
Added regex for 'userlogins' log (optionally enabled)
Updated jail.conf logpath for roundcube-auth to match default log, 'errors'

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -348,7 +348,7 @@ logpath = %(lighttpd_error_log)s
 [roundcube-auth]
 
 port     = http,https
-logpath  = /var/log/roundcube/userlogins
+logpath  = logpath = %(roundcube_errors_log)s
 
 
 [openwebmail]
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,10 @@ Fail2Ban: Changelog
 ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
 -----------
 
+- IMPORTANT incompatible changes:
+   * filter.d/roundcube-auth.conf
+     - Changed logpath to 'errors' log (was 'userlogins')
+
 - Fixes:
    * reload in interactive mode appends all the jails twice (gh-825)
    * reload server/jail failed if database used (but was not changed) and
@@ -17,6 +21,10 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
      Thanks Anton Shestakov
    * Fix fail2ban-regex not parsing journalmatch correctly from filter config
    * filter.d/asterisk.conf - fix security log support for Asterisk 12+
+   * filter.d/roundcube-auth.conf
+     - Updated regex to work with 'errors' log (v1.0.5)
+     - Added regex to work with 'userlogins' log
+   * jail.conf: Changed default logpath for roundcube-auth
 
 - New Features:
    - New filters:
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,7 +22,7 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
    * Fix fail2ban-regex not parsing journalmatch correctly from filter config
    * filter.d/asterisk.conf - fix security log support for Asterisk 12+
    * filter.d/roundcube-auth.conf
-     - Updated regex to work with 'errors' log (v1.0.5)
+     - Updated regex to work with 'errors' log (1.0.5 and 1.1.1)
      - Added regex to work with 'userlogins' log
    * jail.conf: Changed default logpath for roundcube-auth
 
--- a/config/filter.d/roundcube-auth.conf
+++ b/config/filter.d/roundcube-auth.conf
@@ -13,8 +13,8 @@ before = common.conf
 
 [Definition]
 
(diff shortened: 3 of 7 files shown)


config/jail.conf
```diff
@@ -344,7 +344,7 @@ logpath = %(lighttpd_error_log)s
 [roundcube-auth]
 
 port     = http,https
-logpath  = /var/log/roundcube/userlogins
+logpath  = /var/log/roundcubemail/errors
```
yarikoptic: I guess original path was also present on some systems... so we shouldn't just blindly change it but rather add old value as roundcube_error_log to config/paths-common.conf,  use it here, and then customize for CentOS specifically I guess in config/paths-fedora.conf

leeclemens: Are you referring to the directory /var/log/roundcube being changed to /var/log/roundcubemail? Or the file "userlogins" being changed to "errors"?

yarikoptic: Both I guess - I am referring to the full path

leeclemens: I checked out the Debian package and it does seem to define /var/log/roundcube; so I can understand adding that to config/paths-*. However, the file name was not blindly changed. (Actually I posed the path question to the mailing list specifically regarding other platforms but received no feedback - so that was not done blindly either.)

IMHO, default fail2ban should aim to work with default roundcube. The userlogins file is not enabled by default, while the errors log is.  The userlogins file can be optionally enabled (with CentOS and from what I can tell with Debian as well).

I can explain this in more detail if needed, but my aim was to make default fail2ban work with default roundcube - today it does not (regardless of platform). There are a lot of How-To's out there for getting fail2ban and roundcube to work, I just wanted to finally correct fail2ban.

yarikoptic: On Tue, 26 May 2015, Lee Clemens wrote:
>  > -logpath  = /var/log/roundcube/userlogins
>  > +logpath  = /var/log/roundcubemail/errors

>    I can explain this in more detail if needed, but my aim was to make
>    default fail2ban work with default roundcube - today it does not
>    (regardless of platform). There are a lot of How-To's out there for
>    getting fail2ban and roundcube to work, I just wanted to finally correct
>    fail2ban.

and thanks for that ;)

well -- default one for this one is quite vague beast since

./config/defaults.inc.php:$config['log_dir'] = RCUBE_INSTALL_PATH . 'logs/';


and INSTALL_PATH by default is

define('INSTALL_PATH', dirname($_SERVER['SCRIPT_FILENAME']).'/');

if I see it correctly.... 

But what I meant is that there possibly was a reason why it was pointing to
/var/log/roundcube/userlogins... good to have git and sensible commit messages,
so it stems from using an additional plugin
http://mattrude.com/projects/roundcube-fail2ban-plugin/ to log those failed
attempts.  I guess it is no longer necessary?  If not, then indeed we could
just migrate to a new default (which is imho again somewhat vague) and specify
per each OS that canonical location.


-- 
Yaroslav O. Halchenko, Ph.D.

leeclemens: I did some digging into roundcube's source, it seems logging failed login was added in 2013: https://github.com/roundcube/roundcubemail/commit/060467df9d95be3768dab51ff5dd4e6214ec86a0 - which is tagged as 1.0-rc.

Assuming we don't plan on having fail2ban support pre-1.0 roundcube, the plugin you referenced is not needed.

Roundcube logs are stored within its content directory by roundcube default (/logs). The .deb and .rpm configure `log_dir`. This is from the .deb (not direct configuration, but a clue nonetheless):
```
./postinst:	chown -R www-data:adm /var/log/roundcube
./postinst:	chmod -R 750 /var/log/roundcube
./postrm:	rm -rf /var/log/roundcube
```
I don't have a test Debian box to use, but several points show /var/log/roundcube is used.

On CentOS the defaults.inc.php includes `$config['log_dir'] = '/var/log/roundcubemail/';`.

So, as I mentioned before, it makes sense to use these two directories as variables in paths-*.

However, the optional configuration (`$config['log_logins'] = true;`) would still need to be added manually to enable the 'userlogins' log (instructions for which I also added to the filter config). Whereas the 'errors' log is always present (hence default fail2ban matching default roundcube). That is the crux of my justification for changing the default `logpath` (conceding the directory be a configuration per platform).

Assuming we do change the default `logpath`, how can it be made less vague? Change log says it changed, fail2ban will fail to start after upgrade if the `logpath` is not valid (that is pretty obvious imho). Point being fail2ban won't silently stop working as a result; with proper change control (did it start after you upgraded?) or monitoring (is fail2ban running?) it should be more than evident to an administrator. But if more clarity can be added somewhere specifically, I would be more than happy to.

yarikoptic: ok -- just list this change in Changelog at the very top under 
````
- IMPORTANT incompatible changes:
```
"section"

leeclemens:
Added parameterized path to paths-debian and paths-fedora, should it be set in paths-common as well? (should everything be set to 'something' in paths-common?)

yarikoptic:
I think so
Ideally common should reflect upstream defaults

PR #1058, opened 2015-05-25
add froxlor-auth filter and jail

jomu78:
add filter and jail to ban on failed login attempts into froxlor (www.froxlor.org) installations.

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -408,6 +408,13 @@ port    = 10000
 logpath = %(syslog_authpriv)s
 
 
+[froxlor-auth]
+
+port    = http,https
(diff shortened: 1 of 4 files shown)


config/jail.conf
```diff
@@ -408,6 +408,13 @@ port    = 10000
 logpath = %(syslog_authpriv)s
 
 
+[froxlor-auth]
+
+port    = http,https
+logpath  = %(syslog_authpriv)s
+
+
+
```
yarikoptic: keep just 2 empty lines

yarikoptic: nevermind -- did it after merge ;) thanks!

Issue #168, opened 2013-04-09
Server won't start if socket folder doesn't exist


Neraste:
Hey fail2ban development team!

Using systemd on a openSuse 12.2 system, it seems fail2ban cannot start if there is no ```fail2ban``` folder in ```/var/run```.

So, I have made a small directory creation instruction in ```fail2ban-client``` at line 187:
```python

# Create socket folder if it doesn't exist
try:
        os.mkdir(os.path.dirname(self.__conf["socket"]))
except OSError:
        logSys.info("Socket folder already exists")
```
It is probably in the wrong place and I am not sure it occurs in any Linux distribution, but hope it helps.

Edit: I am sorry I have missed to notice post (https://github.com/fail2ban/fail2ban/issues/72) before posting...

leeclemens:
This seems like the package maintainers' responsibility, which we can certainly add to our "files/" scripts. And #399 may also be closable...which may make this closable as well.

leeclemens:
While maintainer and script should deal with this, imho, distutils/setuputils does now too (thanks to #1068), closing.

leeclemens:
Maybe over zealously closed...

yarikoptic:
I think it is ok to let this issue die indeed, especially with #1068 addition in place, so you weren't over zealous @leeclemens ;)

PR #1068, opened 2015-06-12
BF: 'create' /var/run/fail2ban on systems with /var/run

yarikoptic:
Should overcome problems of some users installing using setup.py install

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,12 @@ if setuptools:
 else:
 	setup_extra = {}
 
+data_files_extra = []
+if os.path.exists('/var/run'):
+	# if we are on the system with /var/run -- we are to use it for having fail2ban/
+	# directory there for socket file etc
+	data_files_extra += [('/var/run/fail2ban', '')]
+
 # Get version number, avoiding importing fail2ban.
 # This is due to tests not functioning for python3 as 2to3 takes place later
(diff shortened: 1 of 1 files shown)


yarikoptic:
if noone sees obvious problems @fail2ban/contributors  ?  I have filed #1089 regarding what caused travis to "Fail"

leeclemens:
Do you have a link to an issue or email where someone experienced and issue this aims to resolve? I don't necessarily see any harm in it, just not sure what problem this specific PR aims to solve.

yarikoptic:
issue would be experienced by anyone who is trying to just install fail2ban with "python setup.py install" I guess, and somewhat relates to #168 which you have just closed ;-)

leeclemens:
#168 is not closed - I did comment on it though, suggesting it may be closable :)

I guess I'm not sure how this PR addresses the issue...does this change create the directory "/var/run/fail2ban" if it does not already exist? i.e. does having a path in data_files automatically create the directory (I lack experience in setuptools stuff like this, I admit).

leeclemens:
Well, hey, if I wasn't so lazy I might see it does create the directory as `root:root` `0755`. Didn't personally test on a system without "/var/run", but seems reasonable to me.

PR #1070, opened 2015-06-13
nginx-noscript.conf working with nginx 1.6.2

pridetechdesignllc:
these attacks are very common. This file works with nginx access logs.

### DIFF
--- a/config/filter.d/nginx-noscript.conf
+++ b/config/filter.d/nginx-noscript.conf
@@ -1,0 +1,11 @@
+# Noscript filter /etc/fail2ban/filter.d/nginx-noscript.conf:
+#
+# Block IPs trying to execute scripts such as .php, .pl, .exe and other funny scripts.
+# 
+# Matches e.g.
+# 192.168.1.1 - - "GET /something.php
+# works with NGINX access logs
+#
+[Definition]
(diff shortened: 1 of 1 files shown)


sebres:
I believe, we already have similar filter, that easy let do the same by redefining or extending of `<block>` parameter.
See https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/nginx-botsearch.conf.
It'll be best to customize this config file. For example, something like separated `<block_noscript>` parameter, included in `<block>`.

BTW. By all new filter, don't forget to add an example test log-file under `fail2ban/tests/files/logs/<filter-name>` (to pass a test cases and travis build).

sebres:
JFYI.

I think, everyone understands that such filters can be really "dangerous". 
I mean for example something like this:
- there's a popular site "mysite.example" where this filter is active;
- someone, possibly comparable popular another "othersite.example" (for example competitor of "mysite"), in content of himself includes a hidden frame with 3 url's to "mysite", matched this filter (any-not-exists.php url, that really not exists - 404). 
- any customer visit the page of "othersite" and after N clicks on the page - makes N*3 failures for the site "mysite".
- as result the site "mysite" will be blocked for this customer (his IP-adress was banned).

I am not against filter like this, just to be careful...

yarikoptic:
yeap -- testcases are a must.

Issue #1072, opened 2015-06-17
Cannot change database when there are jails present


szepeviktor:
Could you explain this message?
```
# modified courier-smtp.conf to courier-smtp.local
$ fail2ban-client reload courier-smtp
ERROR  NOK: ('Cannot change database when there are jails present',)
```
v0.9.1-1

leeclemens:
Had you changed the value of `dbfile` since the last restart/reload?

szepeviktor:
Ofcourse not.

sebres:
It's a bug, will fix it...

sebres:
closed with #1077

szepeviktor:
Thank you.

PR #1077, opened 2015-06-22
diverse small bug fixes with reload

### LINKED ISSUE #825: reload in interactive mode shows some errors
If I start fail2ban-client in interactive mode (-i), I see some errors using reload (still I don't know is it only a info or not)...
I have it in upstream/master and all my branches (was found within #824).

```bash
bin/fail2ban-client -c ... -v -i
```
```
fail2ban> reload
fail2ban> reload
ERROR  NOK: ('pam-generic',)
ERROR  NOK: ('Action iptables-allports already exists',)
```
Just added as issue (to not forget), will assign it to me...

sebres:
closes #825 and #1072

### DIFF
--- a/fail2ban/client/jailsreader.py
+++ b/fail2ban/client/jailsreader.py
@@ -50,6 +50,7 @@ class JailsReader(ConfigReader):
 		return self.__jails
 
 	def read(self):
+		self.__jails = list()
 		return ConfigReader.read(self, "jail")
 
 	def getOptions(self, section=None):
--- a/fail2ban/server/server.py
+++ b/fail2ban/server/server.py
@@ -494,20 +494,24 @@ class Server:
 			return "flushed"
 			
 	def setDatabase(self, filename):
-		if len(self.__jails) == 0:
-			if filename.lower() == "none":
-				self.__db = None
-			else:
-				if Fail2BanDb is not None:
-					self.__db = Fail2BanDb(filename)
-					self.__db.delAllJails()
-				else:
-					logSys.error(
-						"Unable to import fail2ban database module as sqlite "
-						"is not available.")
-		else:
+		# if not changed - nothing to do
+		if self.__db and self.__db.filename == filename:
+			return
+		if not self.__db and filename.lower() == 'none':
+			return
+		if len(self.__jails) != 0:
 			raise RuntimeError(
 				"Cannot change database when there are jails present")
+		if filename.lower() == "none":
+			self.__db = None
+		else:
+			if Fail2BanDb is not None:
+				self.__db = Fail2BanDb(filename)
+				self.__db.delAllJails()
+			else:
+				logSys.error(
+					"Unable to import fail2ban database module as sqlite "
+					"is not available.")
 	
 	def getDatabase(self):
 		return self.__db
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,7 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
 -----------
 
 - Fixes:
+   * reload in interactive mode appends all the jails twice (gh-825)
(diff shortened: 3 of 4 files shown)


fail2ban/server/server.py
```diff
@@ -494,20 +494,24 @@ def flushLogs(self):
...
-					self.__db.delAllJails()
-				else:
-					logSys.error(
-						"Unable to import fail2ban database module as sqlite "
-						"is not available.")
-		else:
+		# if not changed - nothing to do
+		if self.__db and self.__db.filename == filename:
+			return
+		if not self.__db and filename.lower() == 'none':
```
yarikoptic: shouldn't it be `or` here?

fail2ban/client/jailsreader.py
```diff
@@ -50,6 +50,7 @@ def jails(self):
 		return self.__jails
 
 	def read(self):
+		self.__jails = list()
```
yarikoptic: if only there was a test demonstrating all the logic ... ;)

yarikoptic:
uff -- pypy keeps giving us grief... may be we should disable it until real need arises, and objections @fail2ban/contributors ?

fail2ban/client/jailsreader.py
sebres: yep, this is interactive only :(

see #825, example:
fail2ban-client start
fail2ban-client -i
... and make reload twice ...

fail2ban/server/server.py
sebres: no, I do not like that, because of test case (code coverage) for totally different logic...

sebres:
have rebased with testcases; can I merge it now?

fail2ban/client/jailsreader.py
yarikoptic: On Mon, 22 Jun 2015, Serg G. Brester wrote:
>    yep, this is interactive only :(

>    see #825, example:
>    fail2ban-client start
>    fail2ban-client -i
>    ... and make reload twice ...

but the bug is within  JailsReader so 
fail2ban/tests/clientreadertestcase.py
would be a good place to test it (theoretically).

P.S. Eventually someone brave will come in to move code base out of
bin/fail2ban-* into the fail2ban module so it could be tested properly
;)

-- 
Yaroslav O. Halchenko, Ph.D.

fail2ban/server/server.py
yarikoptic: ah right -- there is also a condition for None checking below, missed that

yarikoptic:
merge if tests do not puke (ignore pypy)

fail2ban/client/jailsreader.py
sebres: Sorry, my mistake - there are 2 bugs (2 commits), this test case ist for 1st only. 
The bug in jailsreader that you mean is interactive only, I should think a bit about a test case

leeclemens:
My concern is based on how this happened in the first place, that no test caught this happening (it seems it would occur during any `reload`?) I haven't looked it at much more than that, but seems cause for concern regardless.

leeclemens:
guess I missed my 1 hour window to comment :)

sebres:
@leeclemens 
#825 bug (2nd commit) never take place by normally using of fail2ban-client (because __jails in client is always empty after each start of it) however not in interactive mode.
#1072 bug (1st commit) occurs only by reload single jail, never by completely reload (because then f2b goes down, f2b goes up).

szepeviktor:
You seem like you want to add more tests.

leeclemens:
@szepeviktor The issue you reported (#1072) should have, IHMO, been caught by automated tests, although I haven't reviewed the commits very closely yet. I'm basing that on general coding principle and upon what I know of the issue (which includes information you provided).

PR #1082, opened 2015-06-26
Fix pep8 w6 - Deprecation warnings

leeclemens:


yarikoptic:
Thanks!  BTW if you run into similar refactorings later, let's better use '%', e.g. instead of ```"blah" + repr(bleh) + ``` use ```"blah %r ..." % (bleh, ...)```

leeclemens:
Use of `repr` was a direct replacement for the backticks, but '%' would have been a good step forward from that. Maybe replacing all string concatenations is in the future :)

yarikoptic:
On Fri, 03 Jul 2015, Lee Clemens wrote:

>    Use of repr was a direct replacement for the backticks, but '%' would have
>    been a good step forward from that. Maybe replacing all string
>    concatenations is in the future :)

yeap, that was a comment for the future ;)

PR #1085, opened 2015-06-30
Updated CF action

szepeviktor:
1. Store authentication data in jail.local
1. Allow multiple CF accounts
1. Using CF names of variables

### DIFF
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -52,7 +52,8 @@ actionunban = curl -s -o /dev/null https://www.cloudflare.com/api_json.html -d '
 
 [Init]
 
-# This action needs to be set up in jail.local or in jail.d/somefile.conf
+# If you like to use this action with mailing whois lines, you could use the composite action
+# action_cf_mwl predefined in jail.conf, just define in your jail:
 #
 # action = %(action_cf_mwl)s
 # # Your CF account e-mail
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -1,10 +1,23 @@
 #
 # Author: Mike Rushton
 #
-# Referenced from from http://www.normyee.net/blog/2012/02/02/adding-cloudflare-support-to-fail2ban by NORM YEE
+# IMPORTANT
 #
-# To get your Cloudflare API key: https://www.cloudflare.com/my-account
+# Please set jail.local's permission to 640 because it contains your CF API key.
 #
+# This action depends on curl.
+# Referenced from http://www.normyee.net/blog/2012/02/02/adding-cloudflare-support-to-fail2ban by NORM YEE
+#
+# To get your CloudFlare API Key: https://www.cloudflare.com/a/account/my-account
+#
+# This action needs to be set up in jail.local
+#
+# action = %(action_cf_mwl)s
+# # Your CF account e-mail
+# cfemail  = 
+# # Your CF API Key
+# cfapikey = 
+
 
 [Definition]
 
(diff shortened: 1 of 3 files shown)


config/action.d/cloudflare.conf
```diff
@@ -34,7 +36,7 @@ actioncheck =
 #          <time>  unix timestamp of the ban time
 # Values:  CMD
 #
-actionban = curl https://www.cloudflare.com/api_json.html -d 'a=ban' -d 'tkn=<cftoken>' -d 'email=<cfuser>' -d 'key=<ip>'
+actionban = curl -s -o /dev/null https://www.cloudflare.com/api_json.html -d 'a=ban' -d 'tkn=<cfapikey>' -d 'email=<cfemail>' -d 'key=<ip>'
```
szepeviktor: Where does action output go??

yarikoptic: I believe might simply appear in the terminal where fail2ban-server is running (which is usually none, unless ran manually), thus just nowhere ;)

yarikoptic:
what is the purpose of this pull request??? you renamed some variables (thus breaking existing deployments), removed [Init] section (which is needed)...

szepeviktor:
1. Store authentication data in jail.local
1. Allow multiple CF accounts
1. Using CF names of variables

yarikoptic:
1. or in jail.d/somefile.conf
2. logical regardless, just have multiple  actions per jail or multiple jails
3. once again -- why bother renaming?

once again -- all the arguments must be listed in [Init] 

So -- just limit your

szepeviktor:
> once again -- why bother renaming?

Because CF call these email and API on its admin interface, not user and token.
I've left the variable names in the action untouched, but used the CF nomenclature in jail.conf.

config/action.d/cloudflare.conf
```diff
@@ -43,13 +48,18 @@ actionban = curl https://www.cloudflare.com/api_json.html -d 'a=ban' -d 'tkn=<cf
...
-
+actionunban = curl -s -o /dev/null https://www.cloudflare.com/api_json.html -d 'a=nul' -d 'tkn=<cftoken>' -d 'email=<cfuser>' -d 'key=<ip>'
 
 [Init]
 
-# Default Cloudflare API token 
-cftoken = 
+# This action needs to be set up in jail.local or in jail.d/somefile.conf
+#
+# action = %(action_cf_mwl)s
```
yarikoptic: this would be an action with mail etc... please give complete example which would just use
```
action = cloudflare[cfuser="%(cfemail)s", cftoken="%(cfapikey)s"]
```

szepeviktor: Here it is. Down below:
https://github.com/fail2ban/fail2ban/pull/1085/files#diff-04e245e7ddfd46056df5e124cd8d5c11R179

yarikoptic: I know that you have defined it!  But I meant that people do not have to use this composite action, and it is better to document here how THIS action alone could be used (although it is somewhat obvious and other actions do not dive into this gory descriptions per each action).  But ok -- for this one, then expand

```This action needs to be set up in jail.local or in jail.d/somefile.conf```
into
```If you like to use this action with mailing whois lines, you could use predefined in jail.conf
    action_cf_mwl composite action, just define in your jail:```

smth like that...

szepeviktor:
> although it is somewhat obvious and other actions do not dive into this gory descriptions per each action

Quoted.

leeclemens:
Quoting a portion of someone's conversational comment on a commit and the commit's message itself are different things.  But I'll let @yarikoptic weigh in if he intended that portion of his comment to be your commit message. IMHO, it does not conform to any google:"good git commit message" guidelines.

yarikoptic:
and the changelog entry (we got new composite action in jail.conf etc) please! ;)

yarikoptic:
>    Quoting a portion of someone's conversational comment on a commit and the
>    commit's message itself are different things. But I'll let @yarikoptic
>    weigh in if he intended that portion of his comment to be your commit
>    message. IMHO, it does not conform to any google:"good git commit message"
>    guidelines.

as long as the conversation public, I don't mind ;)  I think I do it
from time to time to provide more background, usually pointing url to
specific discussion/comments.  But yes - commit msg itself (especially
leading summary line) should be a concise description of action/effect
not a reference.

szepeviktor:
Is it OK to merge?

yarikoptic:
changelog entry is still to-do

szepeviktor:
Let's skip the `Thanks ...` part.

yarikoptic:
uff... merging and will fix up changelog later  :-P

PR #1088, opened 2015-07-03
Add specific errors for client socket access (closes 595)

leeclemens:
Add os.access() calls and specific error messages regarding socket access

### DIFF
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -171,13 +171,33 @@ class Fail2banClient:
 					streamRet = False
 			except socket.error:
 				if showRet:
-					logSys.error("Unable to contact server. Is it running?")
+					self.__logSocketError()
 				return False
 			except Exception, e:
 				if showRet:
 					logSys.error(e)
 				return False
 		return streamRet
+
+	def __logSocketError(self):
+		try:
+			if os.access(self.__conf["socket"], os.F_OK):
+				# This doesn't check if path is a socket,
+				#  but socket.error should be raised
+				if os.access(self.__conf["socket"], os.W_OK):
+					# Permissions look good, but socket.error was raised
+					logSys.error("Unable to contact server. Is it running?")
+				else:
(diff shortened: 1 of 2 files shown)


bin/fail2ban-client
```diff
@@ -144,32 +144,48 @@ class Fail2banClient:
...
-		for c in cmd:
-			beautifier.setInputCmd(c)
-			try:
-				client = CSocket(self.__conf["socket"])
-				ret = client.send(c)
-				if ret[0] == 0:
-					logSys.debug("OK : " + `ret[1]`)
+		streamRet = False
+		if self.__hasSocketAccess(showRet):
+			beautifier = Beautifier()
```
yarikoptic: imho overindentation makes code more difficult to parse... why not
```python
if not self.__hasSocketAccess(showRet):
    return False
# continue the main logic
```
?

sebres:
Because `fail2ban-client` can be started multiple times, I believe, the checking `__hasSocketAccess` in error case only, would be much better solution.
I mean for example in `except` scope. Please, makes f2b not slower, than it already is.

bin/fail2ban-client
sebres: I think, it should be at all not executed, before any error occured... (keyword: performance).

leeclemens:
@sebres I see your point, and I initially thought about adding all of that logic into the `except socket.error` block. However, I tend to shy away from any more logic (potentially creating an error/exception) in an `except` block than necessary.

bin/fail2ban-client
leeclemens: You're right, for the wrong reason (imho) :)   I made my change thinking __processCmd "processed a command"...negating the loop of potentially multiple commands...any of which could fail or succeed independently and during which the socket could vanish

leeclemens:
Okay, new logic only occurs if socket.error was raised and showRet==True. Wrapped in its own try block, so it should be safe.

sebres:
:+1:

PR #1093, opened 2015-07-04
Fix PEP8 E701, E703 and E712

leeclemens:


leeclemens:
Coverage decreased since I split an uncovered line into two lines (one covered and one *still* not covered).

Issue #1089, opened 2015-07-03
we need fixtures for network based tests (use vcr.py or alike??)


yarikoptic:
otherwise now some travis builds fail due to badips tests getting stuck etc... annoying and not helpful

leeclemens:
I see pypy fail pretty frequently, re-running usually passes. When I check the log, it's usually testSampleRegexsASSP:
```python
testSampleRegexsASSP (fail2ban.tests.samplestestcase.FilterSamplesRegex) ... 2015-07-04 17:39:15,560 Loading configs for filter.d/assp under config 
2015-07-04 17:39:15,561   Loading files: ['config/filter.d/assp.conf']
2015-07-04 17:39:15,561   Loading files: ['config/filter.d/assp.conf']
2015-07-04 17:39:15,562   Loading files: ['config/filter.d/assp.conf']
RPython traceback:
  File "rpython_rlib_rsre_rsre_core.c", line 19736, in portal_34
  File "rpython_jit_metainterp_warmstate.c", line 57801, in maybe_compile_and_run__star_4_6
  File "rpython_jit_metainterp_warmstate_1.c", line 61704, in execute_assembler__star_2_9
  File "rpython_jit_metainterp_compile.c", line 20452, in ResumeGuardDescr_handle_fail
  File "rpython_jit_metainterp_compile.c", line 23769, in ResumeGuardDescr__trace_and_compile_from_bridge
  File "rpython_jit_metainterp_pyjitpl_5.c", line 13520, in MetaInterp_handle_guard_failure
  File "rpython_jit_metainterp_pyjitpl_5.c", line 27585, in MetaInterp__handle_guard_failure
  File "rpython_jit_metainterp_pyjitpl.c", line 7211, in MetaInterp_interpret
  File "rpython_jit_metainterp_pyjitpl.c", line 12542, in MetaInterp__interpret
  File "rpython_jit_metainterp_pyjitpl.c", line 20716, in MIFrame_run_one_step
  File "rpython_jit_metainterp_pyjitpl_1.c", line 55719, in MIFrame_opimpl_jit_merge_point
  File "rpython_jit_metainterp_pyjitpl_2.c", line 37517, in MetaInterp_reached_loop_header
  File "rpython_jit_metainterp_pyjitpl_3.c", line 16391, in MetaInterp_compile_loop
  File "rpython_jit_metainterp_compile.c", line 16879, in compile_retrace
[...]

PR #1095, opened 2015-07-04
temporarily disabled pypy in .travis.yml

leeclemens:
See #1089 - non-reproducable failures occur frequently


leeclemens:
Merging per https://github.com/fail2ban/fail2ban/issues/1089#issuecomment-118540141 one line comment change, and travis tests pass (except the expected pypy test, which has been excluded).

PR #1096, opened 2015-07-05
Fix error logging - not enough arguments (tuple is 1 arg, need 2)

leeclemens:


leeclemens:
```2015-07-05 14:32:33,631 Found no accessible config files for 'd' under /tmp/f2b-tempdI5EyU```

Was

```TypeError: not enough arguments for format string```

PR #1097, opened 2015-07-05
Multiple .travis.yml updates

leeclemens:
Restructuring, run coveralls for each env, less conditionals

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,6 +5,4 @@ source =
     config
     fail2ban
 omit =
-    /usr/*
-    /home/travis/virtualenv/*
     fail2ban/server/filtersystemd.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,10 @@
 
 [run]
 branch = True
-omit = /usr*
+source =
+    config
+    fail2ban
+omit =
+    /usr/*
+    /home/travis/virtualenv/*
+    fail2ban/server/filtersystemd.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,5 +4,8 @@ branch = True
 source =
     config
     fail2ban
-omit =
-    fail2ban/server/filtersystemd.py
+
+[report]
+exclude_lines =
+    pragma: no cover
+    pragma: systemd no cover
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,47 @@
 # vim ft=yaml
 # travis-ci.org definition for Fail2Ban build
+# https://travis-ci.org/fail2ban/fail2ban/
 language: python
 python:
-  - "2.6"
-  - "2.7"
-  - "3.2"
-  - "3.3"
-  - "3.4"
-#  - "pypy" # disabled for now due to frequent unreproducable failures # gh-1089
-  - "pypy3"
+  - 2.6
+  - 2.7
+  - pypy
+  - 3.2
+  - 3.3
+  - 3.4
+  - pypy3
 before_install:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then travis_retry sudo apt-get update -qq; fi
+  - travis_retry sudo apt-get update -qq
 install:
-  - travis_retry pip install pyinotify
+  # Install Python packages we'll use later
+  #   coverage
+  - travis_retry pip install coverage
+  #   coveralls
+  - travis_retry pip install coveralls
+  #   dnspython or dnspython3
   - if [[ $TRAVIS_PYTHON_VERSION == 2* || $TRAVIS_PYTHON_VERSION == 'pypy' ]]; then travis_retry pip install dnspython; fi
   - if [[ $TRAVIS_PYTHON_VERSION == 3* || $TRAVIS_PYTHON_VERSION == 'pypy3' ]]; then travis_retry pip install dnspython3; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then travis_retry sudo apt-get install -qq python-gamin; cp /usr/share/pyshared/gamin.py /usr/lib/pyshared/python2.7/_gamin.so $VIRTUAL_ENV/lib/python2.7/site-packages/; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then cd ..; travis_retry pip install -q coveralls; cd -; fi
-  # overcome buggy pypy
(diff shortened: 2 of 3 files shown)


leeclemens:
Well, coverage/coveralls is interesting...

yarikoptic:
note how bad your coverage became after you stop using our custom config for it...  I don't know how coveralls  reacts when multiple things submit coverage fir the same  revision,  might be taking the last one ( non deterministic reports % then to come)?

leeclemens:
I thought it used `.coveragerc`, so `cp -v .travis_coveragerc .coveragerc` would work. (I didn't get coverage info from my fork's executions.)  I didn't see --rcfile as an option when I executed coverage locally.

The percentage did drift around (ended higher than I saw it) while the tests executed. So is a bit of a  misrepresentation of the actual lack of coverage. However, there is code that only executes with Python 3 (e.g.) which was never checked for coverage (used to be only 2.7).

The lack of coverage I saw decreased the most was in setup.py. But I'll have to look into this phenomenon more closely.

leeclemens:
I think it is deterministic, there are a number of other projects which show multiple jobs per commit/PR on coveralls.io. I don't know the underpinnings enough, but I'd guess not having a previous build to view affected the drop. Subsequent runs have been reasonable (-1,-2%) and which I would expect for them being checked for the first time.

leeclemens:
I gave coveralls access to my repo, hopefully I can test things out easier that way.

yarikoptic:
theoretically there should be no drop in coverage with this change, right?   But I see a problem with submissions for python3 tests -- we build and test out of source for python3 so coverage for "fail2ban" reports huge loss... may be for now we should submit only for python2's.  Now that we support only pyhon >= 2.6 it might become feasible to make codebase friendly to bother 2 and 3 without out of source rebuilding. Then submitting coverage for python3 would become sensible

leeclemens:
Correct, I didn't expect a drop (unless minor for previously uncovered lines due to Python version?).

Strangely, the sidebar shows:
```91.98% COVERED
12043 OF 21475 
RELEVANT LINES COVERED```

Yes, python3 builds show the decreased coverage. So it seems the reported "covered" percentage is really "covered by at least one build" and they are merged that way (I'm just now playing with pylint and coveralls, so bare with me :) )  Since README.md does include >= Python 3.2 (and should say PyPy3 since we have Travis test it), I'm concerned about our unittest implementation if Python 3 is skipping over 95% of the code.

Running coverage with Python3 locally (embarrassed I hadn't done this before), I get a lot of lines like:
> fail2ban/protocol   NotPython: Couldn't parse '/blah/fail2ban/fail2ban/protocol.py' as Python source: 'Missing parentheses in call to 'print'' at line 145

So it seems the tests themselves don't support python3? (And we didn't notice, since Travis only executed coverage/coveralls for Python 2.7). I don't like that :/

Should we be running `bin/fail2ban-testcases`, instead of `setup.py test`

yarikoptic:
On Mon, 06 Jul 2015, Lee Clemens wrote:

>    Correct, I didn't expect a drop (unless minor for previously uncovered
>    lines due to Python version?).

>    Strangely, the sidebar shows:
>    91.98% COVERED
>    12043 OF 21475
>    RELEVANT LINES COVERED

>    Yes, python3 builds show the decreased coverage. So it seems the reported
>    "covered" percentage is really "covered by at least one build" and they
>    are merged that way (I'm just now playing with pylint and coveralls, so
>    bare with me :) ) Since README.md does include >= Python 3.2 (and should
>    say PyPy3 since we have Travis test it), I'm concerned about our unittest
>    implementation if Python 3 is skipping over 95% of the code.

brr -- is it?  I think there is mixup... if you are to use
fail2ban-testcases you need to run fail2ban-2to3 first.  That was the
purpose of enabling  "setup.py test" so it does conversion (outside of
the source) and then runs tests on the converted base. So you get

> python3 setup.py test
...
Ran 279 tests in 97.789s

OK (skipped=1)


>    Running coverage with Python3 locally (embarrassed I hadn't done this
>    before), I get a lot of lines like:

>      fail2ban/protocol NotPython: Couldn't parse
>      '/blah/fail2ban/fail2ban/protocol.py' as Python source: 'Missing
>      parentheses in call to 'print'' at line 145

once again -- needs to go through  fail2ban-2to3 first

>    So it seems the tests themselves don't support python3? (And we didn't
>    notice, since Travis only executed coverage/coveralls for Python 2.7). I
>    don't like that :/

none of the codebase supports python3 natively

>    Should we be running bin/fail2ban-testcases, instead of setup.py test

that change was done in 55810a3c30a745d5b099b40c42d0446b492ea995 to harmonize
running of the tests... but indeed such way coverage gets hurt...  anyways --
coverage reports wouldn't be easy to digest on "converted" code base even if we
revert back to use fail2ban-testcases.  But in general I would not mind.  So
give it a shot again I guess, just don't forget about fail2ban-2to3.  Check
that commit on how it was before

-- 
Yaroslav O. Halchenko, Ph.D.

leeclemens:
You're right; I missed `fail2ban-2to3` in my local tests, thanks!

Now that we only support >=2.6, we should be able to support python3 natively. So ideally fail2ban-2to3 will eventually not make any changes (and be removed), right?

And thanks for the commit reference! I've been testing out some different approaches (kind of time consuming waiting on Travis each time).

---
From the best I can tell, we may have never been installing fail2ban using any Python but the system's (2.7). I tried `sudo -E`, but `which python` still doesn't use the virtual environment's Python (eventhough `$PATH` is retained (with or without `-E`, actually). I removed the shebang from `setup.py` as well.

`echo $PATH`, `sudo echo $PATH` and `sudo -E echo $PATH` produce the same output (which has /home/travis/virtualenv/python2.7.9/bin: first).

However:

> `which python`
> /home/travis/virtualenv/python2.7.9/bin/python
> `sudo which python`
> /usr/bin/python
> `sudo -E which python`
> /usr/bin/python

https://travis-ci.org/leeclemens/fail2ban/jobs/69917595

Travis bug? (Couldn't find much documentation about sudo, and never describes this weirdness.) Regardless, I address this in what will be a new squashed commit for .travis.yml.

yarikoptic:
` fail2ban-2to3 will eventually not make any changes (and be removed), right?`

correct! if we succeed making the code both 2.6,2.7 and >=3.2 compatible which might get challenging (easier if we drop 2.6 as well ;) )

`we may have never been installing fail2ban using any Python but the system's (2.7)`

nice catch!  Try with ```sudo `which python```` ?

leeclemens:
Glad we agree :)   I'd love to drop 2.6 too, but CentOS ships with it (and not sure what other distros). Since "Python 2.6 incorporates new features and syntax from 3.0 while remaining compatible with existing code by not removing older features or syntax.", we should be alright enough :)

Regarding `which python`...that's the weirdness...I now save the path to python (the virtual env / bin/python) and use that after sudo, currently just for `setup.py install`. The backticks don't seem to be correct: https://travis-ci.org/leeclemens/fail2ban/jobs/69938154 ?

PR #1099, opened 2015-07-06
performance fix: minimizes connection overhead (close only at communication end)

sebres:
minimizes connection overhead, using same socket by multiple commands without close it (ex.: `start` sends several hundreds commands at once)

### DIFF
--- a/fail2ban/tests/sockettestcase.py
+++ b/fail2ban/tests/sockettestcase.py
@@ -63,6 +63,11 @@ class Socket(unittest.TestCase):
 		testMessage = ["A", "test", "message"]
 		self.assertEqual(client.send(testMessage), testMessage)
 
+		# test close message
+		client.close()
+		# 2nd close does nothing
+		client.close()
+
 		self.server.stop()
 		serverThread.join(1)
 		self.assertFalse(os.path.exists(self.sock_name))
--- a/fail2ban/tests/sockettestcase.py
+++ b/fail2ban/tests/sockettestcase.py
@@ -25,11 +25,13 @@ __copyright__ = "Copyright (c) 2013 Steven Hiscocks"
 __license__ = "GPL"
 
 import os
+import sys
 import tempfile
 import threading
 import time
 import unittest
 
+from .. import protocol
 from ..server.asyncserver import AsyncServer, AsyncServerException
 from ..client.csocket import CSocket
 
@@ -88,3 +90,17 @@ class Socket(unittest.TestCase):
 		self.server.stop()
 		serverThread.join(1)
 		self.assertFalse(os.path.exists(self.sock_name))
+
+
+class ClientMisc(unittest.TestCase):
+
+	def testPrintFormattedAndWiki(self):
+		# redirect stdout to devnull
+		saved_stdout = sys.stdout
+		sys.stdout = open(os.devnull, 'w')
+		try:
+			protocol.printFormatted()
+			protocol.printWiki()
+		finally:
+			# restore stdout
+			sys.stdout = saved_stdout
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,8 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
      - Added regex to work with 'userlogins' log
    * action.d/sendmail*.conf - use LC_ALL (superseeding LC_TIME) to override 
      locale on systems with customized LC_ALL
+   * performance fix: minimizes connection overhead, close socket only at
+     communication end (gh-1099)
 
 - New Features:
    * New filters:
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -153,30 +153,36 @@ class Fail2banClient:
 		return self.__processCmd([["ping"]], False)
 
 	def __processCmd(self, cmd, showRet = True):
-		beautifier = Beautifier()
-		streamRet = True
-		for c in cmd:
-			beautifier.setInputCmd(c)
-			try:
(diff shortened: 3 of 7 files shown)


sebres:
during review of #1088 I noticed large overhead (as described in the subject) - the socket will be created/closed each time we send a command, also if a list of commands will be sended (example: start or reload).

sebres:
@fail2ban/developers I've found twice a workaround for obsolete python versions:
```python
# b"" causes SyntaxError in python <= 2.5, so below implements equivalent
```
If we are >= 2.6, then I could fix it at the same time.

sebres:
@yarikoptic do you know why travis builds twice this pull request? Possible because I've pushed a branch direct to upstream (not to my clone)?

yarikoptic:
@sebres never saw such travis behavior, but yeah -- I guess that is because you pushed it direct upstream (please don't ;) )

yarikoptic:
and yeah -- we could strip <2.6 compatibility code, just don't go to wild with changes in API ;)

yarikoptic:
Looks reasonable to me

leeclemens:
Would it make sense to define those strings (bytes) somewhere shared? They aren't really used the same, but it seems like a little unnecessary duplication (and related to this PR)

sebres:
@leeclemens made.
now, wait for travis before merge.

yarikoptic:
Don't forget about changelog entry in enhancements

leeclemens:
@sebres Thanks, I like it (and initial changes too). Since you now have a branch on the main repo, we should be alright to merge via github's website, right? Then you can clean it up without harm to the fail2ban remote?

sebres:
@leeclemens a branch is a branch, also if it will be fast forwarded, regardless of whether in upstream or in my clone... Or did you mean something else?

leeclemens:
Unrelated, but perhaps a reason not to use "/" in a branch's name :) (I'm guilty too.)
@sebres of course, of course. Looking at fail2ban's branches, I now see "_sebres/min-connect-overhead" which I think is where the push/travis inconsistency is from - although maybe got fixed up now (I no longer see the two tests)? Usually a feature branch wouldn't be created in the upstream repo. So long as github's automated merge button functionality works as may be expected, I don't see any issue. That was the crux of my concern (I know how github's merge button reacts with a fork's feature branch...not with a branch 'locally').

leeclemens:
I'm about to give up on coverage/coveralls...it says it decreased 0.37% to 91.6, but the link shows two files increasing to 87.78 and 96.88. The rest of the files remained the same (i.e. coverage not changed). How does some aggregate decrease when all of its members either increased or remained the same?

sebres:
Well, about the coverage: I've a bad luck - I have removed a lot of "good" lines in the meaning of coveralls, so the absolute percentage goes down. As punishment I choose something to increment it again :smile:

sebres:
Hmm, very strange - coverage of `fail2ban/protocol.py` was increased from 26% to 100%, but whole increase is minimal...
https://coveralls.io/builds/2988649/source?filename=fail2ban%2Fprotocol.py (new)
https://coveralls.io/builds/2986301/source?filename=fail2ban%2Fprotocol.py (old)
But what is really very weird - this file is not in the list "Changed", "Source Changed" or "Coverage Changed" (simple nowhere):
https://coveralls.io/builds/2988649

fail2ban/tests/sockettestcase.py
```diff
@@ -83,3 +90,17 @@ def testSocketForce(self):
...
+	def testPrintFormattedAndWiki(self):
+		# redirect stdout to devnull
+		saved_stdout = sys.stdout
+		sys.stdout = open(os.devnull, 'w')
+		try:
+			protocol.printFormatted()
+			protocol.printWiki()
+		finally:
+			# restore stdout
+			sys.stdout = saved_stdout
```
yarikoptic: FWIW, someone should adopt smth like our lightweight but thorough mocking for testing stdout/stderr in datalad:  https://github.com/datalad/datalad/blob/master/datalad/utils.py#L213  (and probably swallow_logs for a pair). Becomes as neat as

```python
    with swallow_outputs() as cmo:
        with swallow_logs(new_level=logging.INFO) as cml:
            ret = runner.run(cmd, log_stderr=False)
            eq_(cmo.err.rstrip(), "stderr-Message should not be logged")
            eq_(cml.out, "")
```
don't remember if they are fully 2.6 compatible

sebres: 3 lines of code? as wrapper? OK with try-finally are 5 lines. 
But anyway, if make this, then as decorator...

yarikoptic: here you just do smoke testing, no printouts checked anyhow etc.  with a context manager you can assess either what was printed is actually correct etc.  So far for testing logging we had custom class with fixtures to grab logged output etc, context manager allows in concise manner to surround specific piece of code.

sebres: Agree, your example looks very interesting also.
But here:
- it is not really needed (still not, because I'll rather cover a real f2b-client code, that ATM almost completely missing a coverage);
- something like that https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/servertestcase.py#L930, as temporary replacement to `print` or channel `stdout` would be nice also, sufficient enough and much easier and clearer: 
I mean such "wrapper" like:
```python
	lines = []
	sys.stdout = AnyMemChannelTestClass(lambda *args: lines.append(args))
	...
	self.assertEqual(lines, testLines):
```

PR #1101, opened 2015-07-07
HEAD is a popular break-in attempt type

szepeviktor:
Based on my logs.


yarikoptic:
sample Log line for testing  and changelog please

sebres:
can you do the same for nginx also (all of botsearch-common)?
https://github.com/fail2ban/fail2ban/search?utf8=%E2%9C%93&q=botsearch-common

leeclemens:
May I suggest "Add HEAD method verb to apache-badbots" as a commit message?
http://chris.beams.io/posts/git-commit/

szepeviktor:
@leeclemens Excuse me for my previous commit message.

leeclemens:
@szepeviktor no worries, just trying to help out

szepeviktor:
Something confuses me:
- apache-badbots.conf filters access log
- apache-botsearch.conf filters error log
- nginx-botsearch.conf filters access log
- there is **no** nginx-badbots.conf

leeclemens:
Perhaps unrelated, but seems we never tested the POST regex before. Almost like any | in a regex should get it's own line..from a testing perspective, would hate to see it performance wise (ever regex conditional split out). Point being, the tests never caught that we didn't have a sample POST logline.

szepeviktor:
> Point being, the tests never caught that we didn't have a sample POST logline.

I hope you don't test the tests.

leeclemens:
@szepeviktor I'm not sure what your general objection with software testing is, but that opinion does not add any value here (or anywhere).

szepeviktor:
> Point being, the tests never caught that we didn't have a sample POST logline.

This sounded for me as you test tests. Which is something I've never heard of.
Excuse me for generating this *issues*.

szepeviktor:
As for this PR: all green. Is it clear to merge?

leeclemens:
@szepeviktor The automated tests check that each regex has a corresponding log line provided and vica versa. With using A|B in a regex (as GET|POST), the tests currently do not check that an applicable POST login is made available as long as a GET is (the log line matches).

Ignoring deficiencies in testing only leads to more issues - and those deficiencies should subsequently be addressed rather than ignored.  I hope you can agree to that. If not, see my previous comment re software testing in general. IMHO, ignoring a bug in a test is worse than ignoring a bug in the code itself.

---
Since this was merged while I was typing the above response...I have to ask:

> - IMPORTANT incompatible changes:
>    * filter.d/roundcube-auth.conf
>      - Changed logpath to 'errors' log (was 'userlogins')
>    * filter.d/apache-badbots.conf, filter.d/nginx-botsearch.conf - add
>      HEAD method verb

Was this change really  "IMPORTANT incompatible changes" ? (If you want to ignore the software-testing conversation (which I admitted as potentially off topic) that's one thing. But Incompatible?)

Even if the section placement makes sense, the prefix of "-" isn't on a new line (let along properly indented).

szepeviktor:
You're right. I've overlooked the position in the Changleog.
Could you move my changelog item?

yarikoptic:
On Tue, 07 Jul 2015, Viktor Szépe wrote:

>    You're right. I've overlooked the position in the Changleog.
>    Could you move my changelog item?

done

PR #1104, opened 2015-07-08
Travis and coverage related changes

leeclemens:


leeclemens:
thoughts/questions/concerns @fail2ban/developers ? We are not currently using Travis to install with Python 3, which may be the most concerning fix included here.

leeclemens:
And @fail2ban/contributors :)

setup.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/python
```
yarikoptic: why to remove shebang? setup.py works very nicely as an executable script for many purposes.  And it must have a shebang if executable (as it is now)

leeclemens: I'm not sure what purpose it serves (do you execute setup.py as a script - not as `python setup.py install`)? But it does have executable bit set, and actually didn't solve the problem I uncovered with the wrong `python` executing it - so I put it back in. Ended up moving to `pip install` instead, since that retains metadata for people to uninstall (if installed from source, which we commonly recommend).

yarikoptic: On Thu, 09 Jul 2015, Lee Clemens wrote:
>  > @@ -1,4 +1,3 @@
>  > -#!/usr/bin/python

>    I'm not sure what purpose it serves (do you execute setup.py as a script -
>    not as python setup.py install)?

yes, ./setup.py  install   or ./setup.py test
it is relatively common to have it executable with a shebang

> But it does have executable bit set, and
>    actually didn't solve the problem I uncovered with the wrong python
>    executing it - so I put it back in. Ended up moving to pip install
>    instead, since that retains metadata for people to uninstall (if installed
>    from source, which we commonly recommend).

if you need to executed it with custom/other python, then you indeed
just do "custom/python setup.py", but if default is ok -- ./setup.py
should work as well... so let's just keep it as is for nowbeing.

-- 
Yaroslav O. Halchenko, Ph.D.

szepeviktor: Could you add the execution bit too?

leeclemens: I can't recall any documentation for a Python project saying to run `./setup.py install`, but I already put the shebang back in :)

yarikoptic: On July 9, 2015 6:53:11 PM EDT, Lee Clemens <notifications@github.com> wrote:
>> @@ -1,4 +1,3 @@
>> -#!/usr/bin/python
>
>I can't recall any documentation for a Python project saying to run
>`./setup.py install`, 

Because wouldn't work on Windows :-) our another system where shebang wouldn't point to installed python


> but I already put the shebang back in :)

 Yeay!

szepeviktor: Actually it is easy to use shebang on Windows.
http://whitescreen.nicolaas.net/programming/windows-shebangs
It's like a polyfill.
I rather develop in mcedit on a server.

yarikoptic: That will be New Year wish for all of you guys to never have to do "shebang.bat" and related "magic" of that caliber! ;-)

PR #1112, opened 2015-07-10
Added pass2allow (knocking with fail2ban)

szepeviktor:
Includes a filter and an action.
Example jail opens the FTP port on HTTP auth knocking.


szepeviktor:
I've missed the filter tests.
Please help me on action tests.

szepeviktor:
We have a serious [Travis failure](https://travis-ci.org/fail2ban/fail2ban/jobs/70392036#L430).

leeclemens:
Every filter needs at least one sample log. "No sample log file available for 'apache-pass' filter"

Please checkout [DEVELOP] (https://github.com/leeclemens/fail2ban/blob/master/DEVELOP) and [FILTERS] (https://github.com/leeclemens/fail2ban/blob/master/FILTERS)

szepeviktor:
Sorry. I know I've missed the filter test.

config/action.d/allow-iptables-multiport.conf
```diff
@@ -0,0 +1,59 @@
...
+before = iptables-common.conf
+
+[Definition]
+
+# Option:  actionstart
+# Notes.:  command executed once at the start of Fail2Ban.
+# Values:  CMD
+#
+actionstart = iptables -N f2b-<name>
+              iptables -A f2b-<name> -j <blocktype>
```
yarikoptic: What about instead of redefining many/every iptables-* action we generalize existing ones, i.e. add "returntype" variable defaulting to "RETURN" in -commons and adjusting all iptables-*conf to use it.

Then for 'allow' action it would just be needed to provide (for the pass2allow jail) for the iptables-* action of choice "blocktype=RETURN" (btw, why not ACCEPT?) and "returntype=REJECT" ?

szepeviktor: > add "returntype" variable defaulting to "RETURN" in -commons

I don't dare to touch other filters.



> why not ACCEPT?

`# Notes: ACCEPT skips other chains`

It is necessary to be able to block "pass2allow"-ed IP-s.

yarikoptic: On Sat, 11 Jul 2015, Viktor Szépe wrote:
>  > +actionstart = iptables -N f2b-<name>
>  > +              iptables -A f2b-<name> -j <blocktype>

>      add "returntype" variable defaulting to "RETURN" in -commons

>    I don't dare to touch other filters.

be bold/be brave! ;)  (now you have my blessing)

>      why not ACCEPT?
>    # Notes: ACCEPT skips other chains
>    It is necessary to be able to block "pass2allow"-ed IP-s.

ah right -- here the target/default rule is to reject... smart and nice

-- 
Yaroslav O. Halchenko, Ph.D.

leeclemens: I think `allowtype` should be in iptables-commons.conf as well. It feels a little like duplication, but would allow the use of this addition and not have to override `blocktype` (potential for accidentally overriding it globally, destroying the effectiveness of every other jail).

config/jail.conf
```diff
@@ -767,3 +767,16 @@ port = 2222
...
 maxretry = 1
+
+[pass2allow]
+# allow FTP traffic after successful HTTP auth
+enabled   = false
+filter    = apache-pass
+banaction = allow-iptables-multiport
+# access log of the website with HTTP auth
+logpath   = /var/log/apache2/access.log
+port      = ftp,ftp-data,ftps,ftps-data
```
leeclemens: If this will default to ftp, does it make sense for -ftp to be in the jail name?

config/jail.conf
```diff
@@ -767,3 +767,16 @@ port = 2222
...
 logpath  = /var/lib/portsentry/portsentry.history
 maxretry = 1
+
+[pass2allow]
+# allow FTP traffic after successful HTTP auth
+enabled   = false
+filter    = apache-pass
+banaction = allow-iptables-multiport
+# access log of the website with HTTP auth
+logpath   = /var/log/apache2/access.log
```
leeclemens: logpath should be keeping with the other default jail definitions (defined in distro-specific paths). Should be easy to add access, since error logs are already setup this way (assuming access logs aren't already defined).

szepeviktor:
Have I messed up everything?

config/jail.conf
```diff
@@ -767,3 +767,18 @@ port = 2222
...
 logpath  = /var/lib/portsentry/portsentry.history
 maxretry = 1
+
+[pass2allow]
+# allow FTP traffic after successful HTTP auth
+enabled    = false
+filter     = apache-pass
+banaction  = iptables-multiport
+blocktype  = RETURN
+returntype = DROP
```
szepeviktor: I've chose DROP to hide the protected services even more.

config/action.d/allow-iptables-multiport.conf
leeclemens: Does `RETURN` work?

Thinking out loud:

User gets a 200 from apache, this jail adds the user to the f2b-pass2allow chain for ftp.
User opens a connection to ftp, INPUT > f2b-pass2allow -j RETURN > ftp REJECT.  Assuming of course your INPUT looks like INPUT > ftp > REJECT or falls defaults to a REJECT/DROP.

config/action.d/allow-iptables-multiport.conf
```diff
@@ -0,0 +1,59 @@
...
+#
+actioncheck = iptables -n -L <chain> | grep -q 'f2b-<name>[ \t]'
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionban = iptables -I f2b-<name> 1 -s <ip> -j <allowtype>
```
leeclemens: Does RETURN work here?

Thinking out loud:

User gets a 200 from apache, this jail adds the user to the f2b-pass2allow chain for ftp.
User opens a connection to ftp, INPUT > f2b-pass2allow -j RETURN > ftp REJECT. Assuming of course your INPUT looks like INPUT > ftp > REJECT or falls defaults to a REJECT/DROP.

szepeviktor: I've left in `allowtype`. Sorry.

szepeviktor: I need RETURN to block the user in other jails.

leeclemens: So...I'm not sure I understand the use-case. The purpose of this is for a user to hit a URL successfully, in order to be granted access to FTP ports, correct?   If your default policy is ACCEPT...everyone already has access....right?

config/jail.conf
```diff
@@ -767,3 +767,18 @@ port = 2222
...
+# access log of the website with HTTP auth
+logpath    = /var/log/apache2/access.log
```
leeclemens: logpath should be keeping with the other default jail definitions (defined in distro-specific paths). Should be easy to add access, since error logs are already setup this way (assuming access logs aren't already defined).

config/jail.conf
```diff
@@ -767,3 +767,18 @@ port = 2222
 enabled  = false
 logpath  = /var/lib/portsentry/portsentry.history
 maxretry = 1
+
+[pass2allow]
+# allow FTP traffic after successful HTTP auth
+enabled    = false
+filter     = apache-pass
+banaction  = iptables-multiport
```
leeclemens: This is already the global default, so I think it can be removed

config/jail.conf
```diff
@@ -767,3 +767,18 @@ port = 2222
 enabled  = false
 logpath  = /var/lib/portsentry/portsentry.history
 maxretry = 1
+
+[pass2allow]
+# allow FTP traffic after successful HTTP auth
+enabled    = false
```
leeclemens: This is already the global default, so I think it can be removed

szepeviktor:
Remark: pass2allow could be used to open service tools like phpMyAdmin or https://code.google.com/p/phpmemcacheadmin/

config/jail.conf
szepeviktor: Maybe it is easier to edit an already existing line to `true`.

config/jail.conf
szepeviktor: There is only `apache_access_log` which is **all** access logs. In this case we need only one.

config/action.d/allow-iptables-multiport.conf
leeclemens: Well, you just pushed a new commit with those changes...sure.

config/jail.conf
szepeviktor: OK

config/jail.conf
leeclemens: default fail2ban does not have any jails enabled, why would you default this jail to be enabled?

config/jail.conf
leeclemens: You've provided a distro specific path.  What's wrong with using apache_access_log ?

szepeviktor: `apache_access_log` will watch *all* Apache log. In case of pass2allow you should let in users authenticating on one specific site.

config/jail.conf
szepeviktor: I've added `enabled    = false` which disables pass2allow.

szepeviktor: > Maybe it is easier to edit an already existing line to = true.

I've meant the end-user who may enable this jail.

config/action.d/allow-iptables-multiport.conf
szepeviktor: pass2allow has `actionstart =` DROP all FTP* traffic.

config/jail.conf
szepeviktor: Added a comment: 
`this pass2allow example allows FTP traffic after successful HTTP authentication`

config/jail.conf
szepeviktor: `[pass2allow-ftp]`

config/jail.conf
leeclemens: I'm still not sure I see the value in restricting this default jail to debian, and even further to only one access log.   If you have apache setup with authentication, any other site should yield a 404 or 401...so it wouldn't grant access.  More complicated setups (site-per-service using the same /knocking path) should know better to configure fail2ban properly.

config/jail.conf
leeclemens: `enabled = false` is the global default, as I already mentioned. Therefore, it is unnecessary to set it here.

config/jail.conf
szepeviktor: OK. Changed.
`logpath    = %(apache_error_log)s`

config/jail.conf
szepeviktor: Removed.
Thank you for your comments.
Do you have anything else? May I commit these?

config/jail.conf
leeclemens: error log?

config/jail.conf
szepeviktor: ANd this test.
```
# failJSON: { "time": "2013-06-27T11:55:44", "match": true , "host": "192.0.2.12" }
192.0.2.12 - user1 [27/Jun/2013:11:55:44 +0000] "GET /knocking/ HTTP/1.1" 200 266 "http://domain.net/hello-world/" "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:40.0) Gecko/20100101 Firefox/40.0"
```

config/jail.conf
szepeviktor: A typo.
Thanks.

leeclemens: Come on man...you edited your comment from access log to error log after I pointed out the discrepancy....

ChangeLog
```diff
@@ -34,6 +34,9 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
 - New Features:
    * New filters:
      - froxlor-auth  Thanks Joern Muehlencord
+   * New type of operation:
```
leeclemens: New jail?

config/jail.conf
```diff
@@ -767,3 +767,16 @@ port = 2222
...
+
+[pass2allow-ftp]
+# this pass2allow example allows FTP traffic after successful HTTP authentication on a certain website
+filter     = apache-pass
+blocktype  = RETURN
+returntype = DROP
+# access log of the website with HTTP auth
+logpath    = %(apache_access_log)s
+port       = ftp,ftp-data,ftps,ftps-data
+protocol   = tcp
```
leeclemens: This is the default value, shouldn't need to redefine it here.

szepeviktor:
The time difference in the Travis output is exactly 2 hours.
Please help in setting the timezone.

fail2ban/tests/files/logs/apache-pass
```diff
@@ -0,0 +1,2 @@
+# failJSON: { "time": "2013-06-27T11:55:44", "match": true , "host": "192.0.2.12" }
+192.0.2.12 - user1 [27/Jun/2013:11:55:44 +0000] "GET /knocking/ HTTP/1.1" 200 266 "http://domain.net/hello-world/" "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:40.0) Gecko/20100101 Firefox/40.0"
```
sebres: Just don't use for the test cases any TZ, also not GMT/"+0000" (that will forces f2b to use the same TZ as within json by converting it). You will test here not the time zone / format, but RE / rules ...
```
# failJSON: { "time": "2013-06-27T11:55:44", ...}
... [2013-06-27 11:55:44] ...
```

szepeviktor:
@sebres Thank you!
I've updated my last commit.

config/jail.conf
```diff
@@ -767,3 +767,15 @@ port = 2222
 enabled  = false
 logpath  = /var/lib/portsentry/portsentry.history
 maxretry = 1
+
+[pass2allow-ftp]
+# this pass2allow example allows FTP traffic after successful HTTP authentication
+filter     = apache-pass
+blocktype  = RETURN
+returntype = DROP
+# access log of the website with HTTP auth
```
yarikoptic: I think we need to spell out the action here and explicitly state that "knocking_url" needs to be customized per each deployment, so people do not stick with common default, allowing attacker to try it hoping that fail2ban is forbidding them access due to need to knock

szepeviktor:
Here it is.

yarikoptic:
not yet -- probably do keep original default in, since tests logline relies on it.

yarikoptic:
Looks good to me, unless someone of us two would merge or express further comments for improvement, I will merge within 24h ;)
Thanks and Cheers!

szepeviktor:
I am very glad that it is finished.

leeclemens:
Did you mean to define `knocking_url` in both the jail and the filter?

szepeviktor:
@yarikoptic told me to keep both
https://github.com/fail2ban/fail2ban/pull/1112#issuecomment-120969314

config/jail.conf
```diff
@@ -767,3 +767,17 @@ port = 2222
...
 maxretry = 1
+
+[pass2allow-ftp]
+# this pass2allow example allows FTP traffic after successful HTTP authentication
+port         = ftp,ftp-data,ftps,ftps-data
+filter       = apache-pass
+# access log of the website with HTTP auth
+logpath      = %(apache_access_log)s
+# knocking URL needs to be customized per each deployment
+knocking_url = /secret-knocking-url
```
yarikoptic: ah -- actually now that you brought discussion back... I think I was wrong and mixed up filter and action definitions.  In action definitions we can pass those custom parameters into the action, but we never did that for filters... can we do smth like

```
filter = apache-pass[knocking_url=/secret-knocking-url]
```
please check... I don't think that defining it as a variable within jail would pass it in, but I could be wrong... neither not sure if above would work... if neither works -- just adjust above to state that they need e.g. ```filters.d/apache-pass.local``` defining that variable

sorry for the confusion and somewhat vague request here -- don't have time atm to dig in and investigate either we support those

szepeviktor:
I think the definition in the jail was not passed in because the test went OK.
And the test log line does not contain `/secret-knocking-url` but `/knocking/`.

szepeviktor:
If I do
```
filter = apache-pass[knocking_url=/secret-knocking-url]
````
The test will fail. Could you advise?

**Comment by leeclemens** [comment]
I'd have to test it myself or dig into the code, but to try to point you in the right direction...the action needs the value, iirc.  I'm not sure if the action inheritts from the jail...just try removing one of the declarations and see what happens, then remove the other...use fail2ban-client -d and debug logging to see what is happening.

**Comment by szepeviktor** [comment]
`knocking_url = /secret-knocking-url` does **not** get passed through but `apache-pass[...]` does.

**Comment by szepeviktor** [comment]
And it overwrites `knocking_url = ` in the filter!

**Comment by leeclemens** [comment]
There are three places you can/are defining it. the jail, the filter config and passing it into the filter. Like I said previously, I think it should only be defined in one place, if possible.

**Comment by szepeviktor** [comment]
Is is coincidence to have a botnet attack while I contributing? botnet

**Comment by leeclemens** [comment]
If defining it in filter works and is overridden with the syntax @yarikoptic mentioned (filter = blah[a=b]), then I'd suggest just defining it in the filter and no where else. Anyone can then override it by adding it to the `filter = ` in their jail.

Your last commit https://github.com/fail2ban/fail2ban/commit/dc09d11ad73b965d83a5c92844835420a92c36b3, defines it AND overrides it in the sample jail. Seems unnecessary, imho.

**Comment by szepeviktor** [comment]
> So I leave that in for the test.

What about the test??

**Comment by szepeviktor** [comment]
What is the problem in Travis?
https://travis-ci.org/fail2ban/fail2ban/jobs/70917679#L5969-L5985

**Comment by leeclemens** [comment]
Not the root cause, but I think it will resolve your problem:
> Your last commit dc09d11, defines it AND overrides it in the sample jail. Seems unnecessary, imho.

Did you try just defining `knocking_url` in the filter's .conf and not in the jail or setting it in the jail's declaration?

**Comment by szepeviktor** [comment]
I've tried all combinations and tested with `fail2ban-client -d`.
I think the tests ignore the jail config. That is why I've left `knocking_url` definition in the filter.

**Comment by szepeviktor** [comment]
@yarikoptic How to fix the tests?

**Comment by leeclemens** [comment]
> I've tried all combinations and tested with fail2ban-client -d.

Can you provide your relevant observations?

**Comment by szepeviktor** [comment]
- Defining `knocking_url` in the jail does not get passed on
- Defining `knocking_url` in the filter is OK
- Defining `knocking_url` in the jail's `filter = apache-pass[...]` is also OK *and* it overwrites the value in the filter

**Comment by szepeviktor** [comment]
Now I think - again - that the tests ignore and doesn't understand the jail config:
```ini
filter       = apache-pass[knocking_url=/secret-knocking-url]
```

leeclemens:
Again, without looking into it myself, the variables for `filter` may not work the same as `action` :/

If you declare `knocking_url` in the filter's config, it can still be overridden in a .local file. Does that (and only that) work? You're saying it is ignored when set in the jail, so remove it. And the kwarg like declaration of the filter is causing the tests to fail (and I couldn't find another example using that syntax).

szepeviktor:
> If you declare knocking_url in the filter's config, it can still be overridden in a .local file

Yes does work.

> You're saying it is ignored when set in the jail, so remove it.

But that is ignored **only by the tests**. fail2ban-server reads it in properly.

szepeviktor:
All green. Please merge.

leeclemens:
Okay, so the jail's value is passed into the filter? (The tests don't test jails, but the filters, iirc.)  I think we should have it sorted out then, but your new comment of using "a kwarg-like argument" in the filter caused the test to fail...We shouldn't have comments suggesting to do something that causes the tests to fail, right?

szepeviktor:
> Okay, so the jail's value is passed into the filter?

No. Only the kwarg-like declaration in jail is passed into the filter. A simple `var = value` does not get passed in.

szepeviktor:
Please merge it.

yarikoptic:
so if you say ```filter       = apache-pass[knocking_url="/secret"]``` it gets into effect right?  then we could have

```
filter = apache-pass[knocking_url=%(knocking_url)s]
knocking= /secret
```
and jail variable would get passed via string interpolations, while allowing to override directly within "jail.local" the value for knocking

szepeviktor:
> jail variable would get passed via string interpolations, while allowing to override directly within "jail.local" the value for knocking

Yes. You only need to enhance the tests to understand kwarg-like declaration.

leeclemens:
Why do the tests need to be enhanced? If fail2ban doesn't "understand", the tests surely wouldn't - and shouldn't.

https://github.com/fail2ban/fail2ban/pull/1112#issuecomment-121285055

> Defining `knocking_url` in the filter is OK
> Defining `knocking_url` in the jail's `filter = apache-pass[...]` is also OK and it overwrites the value in the filter

szepeviktor:
> If fail2ban doesn't "understand", the tests surely wouldn't - and shouldn't.

fail2ban-server understands kwargs, only tests do not.

leeclemens:
Like I've said, the tests don't test the default jails...

szepeviktor:
But parse jail.conf:
```
AssertionError: Failed to read filter:apache-pass[knocking_url=/secret-knocking-url]
```

leeclemens:
So in your testing, does `filter = apache-pass[knocking_url=/secret-knocking-url]` work, or not? Pass tests or not? You seem to be providing contradictory information.

szepeviktor:
In fail2ban-server: it works OK.
In test: tests cannot parse kwarg-like declaration: Travis fails.

leeclemens:
So that's bad...like I said.. Have you since tested your changes and used apache-pass.local to override it (as your comment suggested doing)?

szepeviktor:
Yes. It work OK in `.local`.

config/jail.conf
```diff
@@ -767,3 +767,16 @@ port = 2222
 enabled  = false
 logpath  = /var/lib/portsentry/portsentry.history
 maxretry = 1
+
+[pass2allow-ftp]
+# this pass2allow example allows FTP traffic after successful HTTP authentication
+port         = ftp,ftp-data,ftps,ftps-data
+# knocking URL needs to be customized in apache-pass.local
```
leeclemens: needs to be, or can be?

szepeviktor: "needs", a common URL is one step closer to a successful attack as @yarikoptic told me.

leeclemens: "needs to be" or "can be". In direct reference to your comment in jail.conf...this very line which I am commenting on in github...which says "knocking URL needs to be customized"...although a default is already provided...so an override CAN be provided IF the user CHOOSES. So does it "need" to be customized? Or "can" it be customized?

I have no idea why you mention @yarikoptic here.

szepeviktor:
> I think we need to spell out the action here and explicitly state that "knocking_url" needs to be customized per each deployment, so people do not stick with common default, allowing attacker to try it hoping that fail2ban is forbidding them access due to need to knock

He wrote me in https://github.com/fail2ban/fail2ban/pull/1112#discussion_r34471127

config/filter.d/apache-pass.conf
```diff
@@ -0,0 +1,20 @@
...
+
+[Definition]
+
+failregex = ^<HOST> - \w+ \[\] "GET <knocking_url> HTTP/1\.[01]" 200 \d+ ".*" "[^-].*"$
+
+ignoreregex =
+
+[Init]
+
+knocking_url = /knocking/
```
leeclemens: But you DO define it: https://github.com/fail2ban/fail2ban/pull/1112#issuecomment-121317700

You can't have it both ways...take issue with a comment about defining it and then define it, saying you weren't supposed to define it....while defining it anyway....

szepeviktor: I've left it in there because the tests needs its value.

leeclemens: What if it is in the `[Definition]` section?

szepeviktor: I don't know. I am rookie contributor. Sorry.

leeclemens: That's great! I'm trying to help you....you can test locally by running setup.py test or bin/fail2ban-testcases....If you move the declaration of `knocking_url` to `[Definition]`, I'm wondering if it will be better?

szepeviktor:
:smiley: I've found out on my own that the command is:
`bin/fail2ban-testcases -l debug testSampleRegexsAPACHE-PASS`
But the test fails when the declaration is in `[Definition]`.

yarikoptic:
IIRC those should be initialized in `[Init]` section, otherwise they would be not in effect

config/jail.conf
yarikoptic: I am confirming that I have pushed @szepeviktor to that frontier (now hard to find among all the comments ;) ).  It "needs to be" so users don't just enable it with default knocking string, which would open possibility for knowledgeable attackers to exploit it.

config/jail.conf
```diff
@@ -767,3 +767,16 @@ port = 2222
...
+filter       = apache-pass
```
yarikoptic: BTW, what happened with my suggestion (or I email it and it never posted?) to have
```
filter = apache-pass[knocking_url="%(knocking_url)s"]
# TODO:  Define your secret knocking_url in jail.local
knocking_url = 
```
which would use string interpolation to make that jail definition variable available among filter arguments?

szepeviktor:
Thank you for merging.
I hope this PR won't generate hundreds of issues.

leeclemens:
:confounded: I hope no one will merge this until all of the concerns have been addressed. https://github.com/fail2ban/fail2ban/pull/1112#discussion_r34602290 https://github.com/fail2ban/fail2ban/pull/1112#discussion_r34612167

This merge does change the most used, afaik, action...so let's be sure, okay?

yarikoptic:
"`knocking_url` to `[Definition]`" is not going to work I think.  @szepeviktor please try indeed with the one I suggested in https://github.com/fail2ban/fail2ban/pull/1112#discussion_r34612167

szepeviktor:
fail2ban-server work OK with
`filter = apache-pass[knocking_url="%(knocking_url)s"]`
but the test system is not able to parse it:
https://github.com/fail2ban/fail2ban/blob/7fc93cee37c9a8de81f67757ccc27509117ced5a/fail2ban/tests/clientreadertestcase.py#L505
@leeclemens told me not to use this kwarg-like declaration if it fails the test.

yarikoptic:
gotcha -- I guess then we will go with it as it is in this shape.  @leeclemens -- any objections? I think it should be safe to include this change even though it indeed touches the most common action definition(s)

szepeviktor:
As I've stopped contributing:
botnet-stopped

leeclemens:
A little polish based on https://github.com/fail2ban/fail2ban/pull/1112/files#r34595010 and it looks good.
Btw, when I said
> What if it is in the [Definition] section?

and

> If you move the declaration of knocking_url to [Definition], I'm wondering if it will be better?

That was not saying you should or should not put it there...just that you should test it and respond :)

config/jail.conf
leeclemens: Okay, just saw this sneak in here...I agree, leaving it further disabled/undefined would be safer.

leeclemens:
Just saw I missed https://github.com/fail2ban/fail2ban/pull/1112/files#r34611848, are we defining `knocking_url` at all?

config/jail.conf
leeclemens: If it NEEDS to be, then I'm not sure we should be still defining it in the filter's `[Init]` section, please correct me if I'm missing something. Otherwise, I'm not sure we can leave it undefined and have the tests pass?

config/jail.conf
leeclemens: iirc, `apache-pass[knocking_url="%(knocking_url)s"]` failed the tests...but it was literal and unquoted in the last commit I recall seeing.

leeclemens: We'd have to make sure the undefined isn't interpreted as an empty string, and subsequently "GET /" or something similar - that may end up matching even more...

leeclemens:
https://github.com/fail2ban/fail2ban/pull/1112#issuecomment-121430465 r*2?

yarikoptic:
May be because today from beginning I am slow, but I am now fully lost in references in comments.  github comments do not scale if we don't start pruning them somehow.  Regarding why we are defining it at all -- e.g. to have tests pass.  Indeed it would have been better if we fixed up the tests and allowed that string interpolation method definition in the jail.  But imho it could be some next PR ;)

leeclemens:
Inclined to agree, it's gotten hard to follow, even more so with comments on outdated commits.

Honestly I hadn't looked into it and thought it may be due to quoting...but are you saying the test didn't match the filter name because of the kwarg bits/string interpolation part that followeed it? If so, yikes... (side note: string interpolation as in string formatting? or the pseudo-dict/kwargs [] appended to the filter/action names? I think we're talking about the same thing?).

yarikoptic:
ah, no -- not even that level of sophistication ;)... in my case I was talking about even earlier test failure -- if we leave out any default value  in the filter's `[Init]` then sample logline would need to be "fixed up" as well to really non-sensical one (anyone going to the website's root would get the door open ;))

yarikoptic:
So merge, and let's go from there ;-) if anywhere ;)

leeclemens:
> anyone going to the website's root would get the door open

Yeah, that's what I said :) ....https://github.com/fail2ban/fail2ban/pull/1112#discussion_r34636728

yarikoptic:
so we agreed it shouldn't be empty by default ;)  we don't have means atm to specify that parameter must not be empty atm, so I don't see anything else we could do.  Merge? ;)

szepeviktor:
Yesss!!

leeclemens:
So, `knocking_url` is defined in the filter, but the comment in the jail says it NEEDS to be customized in filter.local. It still doesn't seem as though it NEEDS to be overridden, as it is already defined?

Defaulting it in the filter and re-wording the comment in the jail to point the user to the right place ("knocking_url can be overridden in....") to override it seems to be the best alternative at this point.

Concerns about the same url for multiple sites can be mitigated by using a different access log for sites and different jails looking at those separate access logs.

yarikoptic:
I am a bit lost where is the problem... so is this wording something like what you would want to see

```
# knocking_url variable must be overridden to some secret value in filter.d/apache-pass.local
```
?

we wouldn't be able to mitigate "multiple sites" issues since we can't define them per jail (well -- we can via that string interpolation but we don't demonstrate that since tests weren't ready for it)

leeclemens:
the variable is already defined, so it doesn't really need to be overridden at all (if you're using the default). So why does it NEED to be overridden, or MUST be overridden?  It's semantic, I admit, but read any RFC and the definitions of MUST and MAY are clear. We may not have formal definitions, but the comment is written in English - so why not write it in proper English?

leeclemens:
character

szepeviktor:
Thank you.

sebres:
:+1: good job!

Issue #1122, opened 2015-07-18
Enhancement Request: Allow fail2ban to peacefully coexist with other heavy iptables users by adding -w to all iptables commands


davidfavor:
In the action.d files iptables-multiport.conf + iptables-allports.conf the iptables commands are run without the -w option.

On machines with constant iptables command sequences running from many different sources, this can either cause one of fail2ban's iptables commands to fail or some other iptables command to fail.

By adding -w to all iptables commands, this ensures every iptables command fail2ban executes will run to completion as expected.

sebres:
Executing of each jail action is mutex ptotected, so this should obtain an exclusive lock at launch of action (iptables also). 
I should check the same will work cross-over jail, but I think it could be nevertheless added as option (via iptables-common or something else). 
So thanks for a hint.

davidfavor:
I see many other files using iptables also.

What I'm seeing is many errors from iptables + my guess is it's because iptables in fail2ban are run without -w so I'm changing all these to use -w + will run + see what occurs.

davidfavor:
Good idea about adding to iptables-common.

yarikoptic:
On Sat, 18 Jul 2015, David Favor wrote:

>    Good idea about adding to iptables-common.

to me as well  BUT we would need to verify first that any reasonably
supported version of iptables has -w.  If not -- state explicitly when
it was introduced so people on older systems could adjust

sebres:
I meant, it should be optional.

yarikoptic:
On Sat, 18 Jul 2015, Serg G. Brester wrote:

>    I meant, it should be optional.

ie not enabled by default, right?
what could be cons of default operation with -w?

sebres:
No, default enabled, but as you already said, if unsupported (an error occurred) can be disabled in jail.local [default].

leeclemens:
Added to iptables:
http://git.netfilter.org/iptables/commit/?id=93587a04d0f2511e108bbc4d87a8b9d28a5c5dd8

Likely included in 2013-Aug-08: iptables-1.4.20
Phil Oester:
    ip[6]tables: Add locking to prevent concurrent instances

yarikoptic:
cool -- thanks for tracking it down @leeclemens !
in Debian only wheezy and squeeze (lts) caries 1.4.14 and ubuntu precise is at 1.4.12.  It would be ok with me to enable it by default while allowing for quick change to have it removed.  I will send a PR shortly

[...]

PR #1127, opened 2015-07-24
WIP ENH Add <lockingopt> (Close: #1122) and <iptables> to define the iptables call

yarikoptic:
This was initially to address #1122 to specify -w for all iptables calls.  But since it would require change in all iptables rules, I have decided that it might be better to immediately parametrize iptables call itself.  Then it might help those who e.g. come up with wrappers around iptables to switch between ipv4 and ipv6 -- they would need just to override that one.

WIP as for "Work in Progress" -- this one wasn't tested or even tried, please don't merge unless we are sure it is functioning as it should

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,13 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
 - IMPORTANT incompatible changes:
    * filter.d/roundcube-auth.conf
      - Changed logpath to 'errors' log (was 'userlogins')
+   * action.d/iptables-common.conf
+     - All calls to iptables command now use -w switch introduced in
+       iptables 1.4.20 (some distribution could have patched their
+       earlier base version as well) to provide this locking mechanism
+       useful under heavy load to avoid contesting on iptables calls.
+       If you need to disable, define 'action.d/iptables-common.local'
+       with empty value for 'lockingopt' in `[Init]` section.
 
 - Fixes:
    * reload in interactive mode appends all the jails twice (gh-825)
--- a/config/action.d/iptables-allports.conf
+++ b/config/action.d/iptables-allports.conf
@@ -17,23 +17,23 @@ before = iptables-common.conf
 # Notes.:  command executed once at the start of Fail2Ban.
 # Values:  CMD
 #
-actionstart = iptables -N f2b-<name>
-              iptables -A f2b-<name> -j <returntype>
-              iptables -I <chain> -p <protocol> -j f2b-<name>
+actionstart = <iptables> -N f2b-<name>
+              <iptables> -A f2b-<name> -j <returntype>
+              <iptables> -I <chain> -p <protocol> -j f2b-<name>
 
 # Option:  actionstop
 # Notes.:  command executed once at the end of Fail2Ban
 # Values:  CMD
 #
-actionstop = iptables -D <chain> -p <protocol> -j f2b-<name>
-             iptables -F f2b-<name>
-             iptables -X f2b-<name>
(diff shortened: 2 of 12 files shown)


leeclemens:
> 2015-07-24 11:45:29,205 fail2ban.CommandAction  [30472]: ERROR   properties contain self referencing definitions and cannot be resolved

> Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/fail2ban/server/actions.py", line 221, in run
    action.start()
  File "/usr/local/lib/python2.7/dist-packages/fail2ban/server/action.py", line 278, in start
    raise RuntimeError("Error starting action")
RuntimeError: Error starting action

```
64c64
< iptables = <iptables> <lockingopt>
---
> iptables = iptables <lockingopt>
```

> 2015-07-24 11:50:58,367 fail2ban.action         [31488]: DEBUG   iptables -w -N f2b-asterisk
iptables -w -A f2b-asterisk -j RETURN
iptables -w -I INPUT -p tcp -m multiport --dports 5060,5061 -j f2b-asterisk

> Chain f2b-asterisk (1 references)
target     prot opt source               destination
RETURN     all  --  0.0.0.0/0            0.0.0.0/0

leeclemens:
And ChangeLog ;)

yarikoptic:
@leeclemens ChangeLog is in.  Anyone else from @fail2ban/contributors spot any other possible recipe for a disaster? ;)

ChangeLog
```diff
@@ -12,6 +12,13 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
 - IMPORTANT incompatible changes:
    * filter.d/roundcube-auth.conf
      - Changed logpath to 'errors' log (was 'userlogins')
+   * action.d/iptables-common.conf
+     - All calls to iptables command now use -w switch introduced in
+       iptables 1.4.20 (some distribution could have patched their
+       earlier base version as well) to provide this locking mechanism
+       useful under heavy load to avoid contesting on iptables calls.
```
szepeviktor: Does  contesting mean *race condition*?

leeclemens: Sort of...not really?
`-w` forces iptables to wait for the lock, where it would have otherwise exited if unable to immediately obtain it. Since proper sequence isn't necessarily important (e.g. banning two IPs at the same time), it's not quite a race condition.

yarikoptic: On Sun, 26 Jul 2015, Lee Clemens wrote:

>  > @@ -12,6 +12,13 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
>  >  - IMPORTANT incompatible changes:
>  >     * filter.d/roundcube-auth.conf
>  >       - Changed logpath to 'errors' log (was 'userlogins')
>  > +   * action.d/iptables-common.conf
>  > +     - All calls to iptables command now use -w switch introduced in
>  > +       iptables 1.4.20 (some distribution could have patched their
>  > +       earlier base version as well) to provide this locking mechanism
>  > +       useful under heavy load to avoid contesting on iptables calls.

>    Sort of...not really?
>    -w forces iptables to wait for the lock, where it would have otherwise
>    exited if unable to immediately obtain it. Since proper sequence isn't
>    necessarily important (e.g. banning two IPs at the same time), it's not
>    quite a race condition.

yeah...  we used to (before threads got locking fixed) have issues
on startup whenever two (or more) iptables commands were ran in
parallel.  I guess this could have fixed that too ;)


-- 
Yaroslav O. Halchenko, Ph.D.

szepeviktor: Thanks.

leeclemens: just for iptables, but besides the point.

PR #1130, opened 2015-07-24
Remove literal "TODO" from method's name

leeclemens:
Also changed expected log message, since this test hasn't been executed, possibly ever


yarikoptic:
it was prefixed with TODO since it lead to unresolved failure due to unhandled misspecification at that time.  Sure thing we could now "test" that it indeed crashes with generic Python exception message -- no harm in that I guess.

leeclemens:
When the test failed, I figured it was "disabled" with the prefix intentionally...still seemed odd, especially with no comment explaining its purpose (why not comment it out, delete it). Thanks for merging!

PR #1134, opened 2015-07-26
WiP BF: kill children of hanging processes as well (Close #1129)

### LINKED ISSUE #1129: whois lookup used in DoS attacks against Fail2Ban
Remote admins can use whois redirects to a server that accepts the connection and hangs on to it forever.  If using a whois client that doesn't time out by default, it causes fail2ban to not proceed, and intruders can continue brute-forcing the server.

This is currently observed in the wild.

A workaround for this until someone fixes the whois callouts in fail2ban is to replace the whois binary with (or for EL, make the whois symlink point to) a script that calls the real whois with a timeout.
I use this:

#!/bin/sh
timeout 8 /usr/bin/jwhois "$@"

yarikoptic:
do not merge yet -- needs tests and haven't tested it yet at all yet

came out a bit longish BUT should be easy to absorb into a patch for previous releases -- amount of conflicting changes should be minimal if any.

### DIFF
--- a/fail2ban/helpers.py
+++ b/fail2ban/helpers.py
@@ -25,6 +25,9 @@ import os
 import traceback
 import re
 import logging
+import psutil
+import signal
+import time
 
 
 def formatExceptionInfo():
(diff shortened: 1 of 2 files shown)


yarikoptic:
Crap -- forgot that psutil is not in standard library, need to recode that part for children collection... later today unless someone beats me to it with a PR against my PR (just against bf-timeout-nested-commands branch in my yarikoptic clone)

yarikoptic:
scrap this overengineered solution (See https://github.com/fail2ban/fail2ban/issues/1129#issuecomment-125047780)

PR #1135, opened 2015-07-27
Limit the number of log lines in *-lines.conf actions

szepeviktor:
1000 lines are around 200-300kB at max.
Is someone expecting more than 1000 lines in an email? Thus is it necessary to tell about this in the changelog?

### DIFF
--- a/config/action.d/mail-whois-lines.conf
+++ b/config/action.d/mail-whois-lines.conf
@@ -42,7 +42,7 @@ actionban = printf %%b "Hi,\n
             Here is more information about <ip>:\n
             `whois <ip> || echo missing whois program`\n\n
             Lines containing IP:<ip> in <logpath>\n
-            `grep -E '(^|[^0-9])<ip>([^0-9]|$)' <logpath>`\n\n
+            `grep -E <grepopts> '(^|[^0-9])<ip>([^0-9]|$)' <logpath>`\n\n
             Regards,\n
             Fail2Ban"|mail -s "[Fail2Ban] <name>: banned <ip> from  `uname -n`" <dest>
 
@@ -67,3 +67,7 @@ dest = root
 # Path to the log files which contain relevant lines for the abuser IP
 #
 logpath = /dev/null
+
+# Number of log lines to include in the email
+#
+grepopts = -m 1000
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,8 @@ ver. 0.9.3 (2015/XX/XXX) - wanna-be-released
 - IMPORTANT incompatible changes:
    * filter.d/roundcube-auth.conf
      - Changed logpath to 'errors' log (was 'userlogins')
+   * mail-whois-lines, sendmail-geoip-lines and sendmail-whois-lines
+     actions include only the first 1000 log lines in the email
 
 - Fixes:
    * reload in interactive mode appends all the jails twice (gh-825)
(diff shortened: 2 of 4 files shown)


yarikoptic:
FWIW checked: -m is available in grep on older centos (6.6) and OSX I have access to, so adding it feasible.  But "-m -1" does output all hits on my debian but outputs nothing on OSX (so we can't use it as an option to "give me all entries while using -m option")

BUT not sure "how": although sounds like a reasonable feature but I guess we need to make it an option and actually by default make it very large -- since it would just silently limit number of listed entries -- might upset may be someone  e.g. automatically processing those emails etc.

So -- we would need "grepopts" defined in those actions and defaulting to empty  OR making it `-m 1000` and adding a FAT warning in the ChangeLog

szepeviktor:
Here it is.

yarikoptic:
Thanks -- Looks good to me. Any other feedback from @fail2ban/contributors  ?

config/action.d/mail-whois-lines.conf
```diff
@@ -42,7 +42,7 @@ actionban = printf %%b "Hi,\n
             Here is more information about <ip>:\n
             `whois <ip> || echo missing whois program`\n\n
             Lines containing IP:<ip> in <logpath>\n
-            `grep -E '(^|[^0-9])<ip>([^0-9]|$)' <logpath>`\n\n
+            `grep -E <grepopts> '(^|[^0-9])<ip>([^0-9]|$)' <logpath>`\n\n
```
yarikoptic: eh... when will we get our "actions tests" going? ;)
insert <grepopts> before -E, since now it would constitute the regexp!

szepeviktor: Isn't it substituted as
`grep -E -m 1000 '...'` or - when empty - `grep -E  '...'`?

szepeviktor: Would placing it before -E help?
`grep "" -E '...'`

szepeviktor: Maybe `grepopts = -E -m 1000`?

yarikoptic: My bad!  I thought that -E serves as -e, i.e. requires the actual extended regexp as its argument.  But it doesn't -- just a flag....  so it is fine as is.  I would probably keep -E still within the actual definition since it is (is it? ;)   ) mandatory for coded there regexp.

yarikoptic:
So with my confusion being resolved, looks all good to me.  Any other feedback from @fail2ban/contributors  ?

config/action.d/mail-whois-lines.conf
szepeviktor: > (is it? ;) ) mandatory

Yes, `|` and `(` otherwise would need escaping.

yarikoptic: On Sun, 26 Jul 2015, Viktor Szépe wrote:
>  > @@ -42,7 +42,7 @@ actionban = printf %%b "Hi,\n
>  >              Here is more information about <ip>:\n
>  >              `whois <ip> || echo missing whois program`\n\n
>  >              Lines containing IP:<ip> in <logpath>\n
>  > -            `grep -E '(^|[^0-9])<ip>([^0-9]|$)' <logpath>`\n\n
>  > +            `grep -E <grepopts> '(^|[^0-9])<ip>([^0-9]|$)' <logpath>`\n\n

>      (is it? ;) ) mandatory
>    Yes, | and ( otherwise would need escaping.

THAT IS RIGHT! ;)

PR #1136, opened 2015-07-27
WiP BF: kill the entire process group upon timeout (Close #1129)

### LINKED ISSUE #1129: whois lookup used in DoS attacks against Fail2Ban

yarikoptic:
Requires also establishing a new process group for a child
process, which changes previous behavior

~~TODO/DONE: test~~

### DIFF
--- a/fail2ban/server/action.py
+++ b/fail2ban/server/action.py
@@ -565,7 +565,9 @@ class CommandAction(ActionBase):
 			stderr = tempfile.TemporaryFile(suffix=".stderr", prefix="fai2ban_")
 			try:
 				popen = subprocess.Popen(
-					realCmd, stdout=stdout, stderr=stderr, shell=True)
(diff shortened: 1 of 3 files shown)


yarikoptic:
@arth1 how does this look like? ;) thanks again for pointing to killpg

yarikoptic:
could ok -- tests pass and no objections, so -- merging!

Issue #876, opened 2014-12-01
PEP8 compliance


craneworks:
Today is so much action I thought I could also open a new ticket. ;)

I would like to see a stop in developing new features and clean up the code to be in compliance with PEP8. 

Why should this be done?
ATM most people are having their own coding styles which results in funny stuff like `var1+<tab><tab><tab>var2<space>=var3`. This is just one example but I found really weird looking stuff.
So the benefits would be:
- the code is getting cleaner
- easier to submit patches from developers
- some strange code may get killed
- the code is getting a full review

On the opposite...
- this will result in a lot of work. A modern IDE (even vim can handle stuff like this with the right plugins) will reduce the amount of work. I fixed most of the PEP8 errors in just a few days.
- features, patches and so on have to be rebased before we can add them. But this is only a one time job.

I think the best way would  be starting by killing each PEP8 Error one by one instead of fixing it file by file.

leeclemens:
I agree, overall, but the issue isn't that IDE's are smarter (even vim) but that github doesn't use blame to ignore whitespace. This creates confusion when looking at old commits where functionality was changed but recent whitespace-only updates were made.

rumpelsepp:
Just append `?w=1` to the URLs and everything is fine:
https://github.com/blog/967-github-secrets

So IMO that's not really a reason.

leeclemens:
@rumpelsepp thanks for that info!

leeclemens:
My previous PEP8 related PRs were met with sarcastic remarks from other fail2ban developers.

sebres:
I have too many large branches, so toooo many conflicts on active development.
In which case the effort (and my cost) does not fit to the profit.
I think, the time was simply out of line.
Sorry if I was sometimes too hard.

yarikoptic:
The verdict was/is/will be:   changes (either new code, or lines touched in changes) should generally aim to comply with PEP8. Exclusions are:

- we are using tabs instead of spaces for indentation
- code lines are allowed to slightly exceed a norm of 79 BUT should caution you about that there might be a need in refactoring.

For everyone I would advise to watch/listen to https://www.youtube.com/watch?v=wf-BqAjZb8M from the father of PEP8.  With that I will close this issue since it can't otherwise be "fixed" at once.

PR #1143, opened 2015-07-31
PEP8 violations

pheanex:
I noticed there were quite a few pep8 violations, including:
* Trailing spaces and tabs
* Depracted backticks (repr())
* Redundant parenthesis
* Unnecessary semicolons
* Inconsistent indentation
* Too many/few newlines/whitespaces
* Simplifyable assignments/comparisons
* Replaceable tests for membership with "not in/is"
* Unused imports

I don't know if you are interested in such a cleanup => Just take a look at it
FYI: I squashed all commits together for better reviewability, but if you like just some of them just let me know

### DIFF
--- a/config/filter.d/ignorecommands/apache-fakegooglebot
+++ b/config/filter.d/ignorecommands/apache-fakegooglebot
@@ -6,20 +6,22 @@
 #
 import sys
 
+
 def process_args(argv):
     if len(argv) != 2:
-       sys.stderr.write("Please provide a single IP as an argument. Got: %s\n"
+        sys.stderr.write("Please provide a single IP as an argument. Got: %s\n"
                         % (argv[1:]))
-       sys.exit(2)
+        sys.exit(2)
 
     ip = argv[1]
 
     from fail2ban.server.filter import DNSUtils
     if not DNSUtils.isValidIP(ip):
-       sys.stderr.write("Argument must be a single valid IP. Got: %s\n"
+        sys.stderr.write("Argument must be a single valid IP. Got: %s\n"
                         % ip)
-       sys.exit(3)
+        sys.exit(3)
     return ip
+
 
 def is_googlebot(ip):
     import re
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -25,12 +25,9 @@ __license__ = "GPL"
 import getopt
 import logging
 import os
-import pickle
-import re
 import shlex
 import signal
 import socket
-import string
 import sys
 import time
 
@@ -47,6 +44,7 @@ logSys = getLogger("fail2ban")
 ##
 #
 # @todo This class needs cleanup.
+
 
 class Fail2banClient:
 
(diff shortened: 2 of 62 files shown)


yarikoptic:
Hi, Thanks for taking time!  As always for such ones I am not sure since some folks (and partially me) would get then really upset with all the merge conflicts they would get for some perspective branches etc, right @sebres ?
Note also the failed tests ("is not" not "not is" ;))

sebres:
Thx for your effort, but please read, what the law lays down: 
https://github.com/fail2ban/fail2ban/issues/876#issuecomment-122382586
Thus closed.

sebres:
Oh, he was already itself present :)

config/filter.d/ignorecommands/apache-fakegooglebot
```diff
@@ -6,21 +6,23 @@
 #
 import sys
 
+
 def process_args(argv):
     if len(argv) != 2:
-       sys.stderr.write("Please provide a single IP as an argument. Got: %s\n"
+        sys.stderr.write("Please provide a single IP as an argument. Got: %s\n"
```
yarikoptic: this doesn't look right

yarikoptic:
@sebres -- you don't have much of changes "staged" for bin/* scripts, right?  so it might actually make sense to pick those PEP8 fixes (within bin/*) + may be pruning trailing spaces within header docstrings (unlikely to conflict), and unused imports pruning (but should be done with care -- we don't have 100% code coverage yet!)

sebres:
And what about "law"? :)

I am always an opponent of such changes (only to be PEP compliant). 
It's of course a best way to be developer of the year (or should I say commiter of ther year), but in my reality does not matter (not even a little bit).

Especially, all the guys who have a fork (or just many custom local changes) be really amazed hereafter, and it's a best way to annoy all them. 
And, imho, satisfaction is very important in the open source world.

My conclusion: let us do it really only there, where it would be needed or as your already said in new code, or lines touched in changes.

Imho, <s>paranoiacs or people who read glossy papers</s> people who really need it, can make it PEP-compliant simply with a small script or just merge together with this or similar branches.
By the way, I'm not against a PEP-compliant branch parallel to master.

yarikoptic:
There is no "law" -- only common wisdom!  As for forks and branches -- that is why I have asked explicitly about those changes I listed -- they will not conflict with your changes (to any major extent), right?  If so we could get that done, and then whenever we move bin/* functionality under fail2ban it will come at once with better compliance with PEP8.  But it should be done before or after not during the move, so git could track the renames correctly.

An once again -- I am not treating PEP8 as a bible which we must conform, but having PEP8-cleaner code is better for long term maintenance, that is why there were those decisions to do it whenever it doesn't have "conflicts".

Making a compliant branch would not be feasible to maintain, thus wasteful.

sebres:
Yes, but who says there are no people in the world that it will not affect?

I know an another common wisdom - don't touch a "shit", then it does not stink also (otherwise known also as Never Touch A Running System resp. NTRS). And of course I touch it also all the time, but not for such things like PEP.

You are the boss :fearful: (But I stand still to my conclusion above :beers:).

pheanex:
So do you want some of them for just the bin/ directory?
Maybe just the not so critical ones?

PR #1146, opened 2015-08-02
renamed <NAME> to correct <ACT> in protocol

pheanex:
Thats better right?


yarikoptic:
just woke up,   change looked right  to me -  why closed PR?

pheanex:
forgot about the man page file ;-D -> so reopened other PR

PR #1147, opened 2015-08-02
renamed <NAME> to correct <ACT> in protocol

pheanex:
That's more correct?


yarikoptic:
ah ;)
1. manpages are autogenerated anyways, so sufficient to fix the code (only better with the manpage though ;) )
2. yo udon't need to close/open new PR -- just push more commits to the same branch and that will update PR with new changes

pheanex:
1. shouldn't it then be removed from the repo?
2. i did test that, but it would not update, so id closed and opened a new one ;-)

yarikoptic:
1 we keep them to  not demand having help2man  for installation
2  it should have  :-)  May be you pushed to another branch

yarikoptic:
2  next time keep it in that state and point  me to it,  we will get it figured out

pheanex:
2. sure, i might have forgot or pushed somewhere else, though ;-)

PR #1148, opened 2015-08-02
Documentation: explicit enumeration of backend and usedns options

pheanex:
Analogous to
["set logtarget <TARGET>", "sets logging target to <TARGET>. Can be STDOUT, STDERR, SYSLOG or a file"].

### DIFF
--- a/fail2ban/protocol.py
+++ b/fail2ban/protocol.py
@@ -63,7 +63,7 @@ protocol = [
 ["set dbpurgeage <SECONDS>", "sets the max age in <SECONDS> that history of bans will be kept"], 
 ["get dbpurgeage", "gets the max age in seconds that history of bans will be kept"], 
 ['', "JAIL CONTROL", ""],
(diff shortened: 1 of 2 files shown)


yarikoptic:
although not sure if not too long

pheanex:
It's as long as the mentioned "logtarget".
I found it quite helpful to have a short overview on possible options, maybe others too?
Consitency -> enumerate options everytime or never?
Its also the same with "usedns".

yarikoptic:
Well, with some, like backend, available values in the end depend on available backends, e.g. pyinotify will not be a good value if no pyinotify is available. 
I agree though that explicit enumeration where not prohibitively long would be a good thing :-)

pheanex:
They are possible options and also mentioned in jail.conf.
Mentioning and explaining those options there => enumerationg available options in help?
Nevertheless i'm just looking for the consistency. (Which would be?)

PR #1149, opened 2015-08-02
bug-fix: set correct bash completion options for fail2ban-regex

pheanex:
fail2ban-regex does not have the same parameters like fail2ban-client or fail2ban-server.
The bash-completion script offered the same options for all of them though.

### DIFF
--- a/files/bash-completion
+++ b/files/bash-completion
@@ -31,7 +31,12 @@ __fail2ban_jail_action_methods () {
 
 _fail2ban () {
     local cur prev words cword
-    _init_completion || return 
+    _init_completion || return
+
+    if [[ "$1" == *"fail2ban-regex" ]] && [[ "$cur" == "-"* ]];then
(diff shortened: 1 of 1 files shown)


files/bash-completion
```diff
@@ -31,7 +31,12 @@ __fail2ban_jail_action_methods () {
 
 _fail2ban () {
     local cur prev words cword
-    _init_completion || return 
+    _init_completion || return
+
+    if [[ "$1" == *"fail2ban-regex" ]] && [[ "$cur" == "-"* ]];then
```
yarikoptic: with this change it doesn't suggest any file for the first argument `<LOG>` and would only complete options (but that also did it before).  So what gets fixed really? ;)

PS sorry for the delay

pheanex: tbh i don't know anymore (too long ago).
Surely id made sense back then => just close the PR ;-)

Issue #1106, opened 2015-07-09
RFC: Drop support of Python 2.6


leeclemens:
[PEP 361] (https://www.python.org/dev/peps/pep-0361/) states:
> With its release on October 29, 2013, all official support for Python 2.6 has ended.  Python 2.6 is no longer being maintained for any purpose.

From [Porting Python 2 Code to Python 3] (https://docs.python.org/3/howto/pyporting.html):
> But you should aim for only supporting Python 2.7. Python 2.6 is no longer supported and thus is not receiving bugfixes.

leeclemens:
Regarding performance, efficiency with a common code base, [PEP 469 - migrating-from-python-3-to-the-common-subset-with-python-2-7] (http://legacy.python.org/dev/peps/pep-0469/#migrating-from-python-3-to-the-common-subset-with-python-2-7):
> Python 2.7 compatibility is often sufficient

sebres:
I want to save support of 2.6 so long as possible (think about all weak hardware like routers etc., that will not get python 2.7). They are not so different with 2.7 (2.x and 3.x - absolutely yes, but not 2.6 and 2.7).
Doesn't matter to me, that 2.6 is no longer supported from maintainers and what any PEPs will say about that.
@fail2ban/developers The next developer, that agree with me, simply closes this ticket. :)

leeclemens:
@sebres How many users run fail2ban on routers? Where Python 2.7 is not an option?

The difference between 2.6 and 2.7 boils down to being able to ultimately support Python 3 out of the same code base (at least without a lot of branching/performance impact - you like performance :) ). We claim to support Python 3 and we know that's where everything is heading...why not embrace it? Plus, if any low-level issue pops up in Python 2.6, we'll get no support from Python, it's simply not supported by anyone.

I created this Issue to open a discussion (RFC), so hopefully it's not just a '-2 vote and close' as suggested (Especially given some of the Issues I've seen lately that have been open for 2-3 years.)

yarikoptic:
[...]

PR #1161, opened 2015-08-18
Detect and ban an IP adress in a subnet

Eaque:
I realized that I had a high number of failed connection attempts through ssh. Those attemps didn't come from totally different IP adresses but from several groups of adresses into different subnets.

Thus, this feature is an option in the jail configuration which adds the ability to detect an IP in a subnet, consider it as a fail, and ban the whole subnet when the maximum number of attempts is reached.

For example, if the option subnetmask is set to 24, failed connections with the following IPs : 
* 1.2.3.4
* 1.2.3.5
* 1.2.3.6

will result as 3 fails for the subnet 1.2.3.0/24, and if the maxretry is set to 3, the subnet 1.2.3.0/24 will be ban.


yarikoptic:
The issue of blocking subnets came up multiple times in our tentative ipv6 support discussions.  Although a different case here, I guess it would bear somewhat the same importance, but making it just an option per jail might indeed be worthwhile, so thanks for bringing it up.  @fail2ban/developers and @fail2ban/contributors -- what do you think? ;)

.pydevproject
```diff
@@ -1,12 +1,11 @@
...
-<pydev_project>
-<pydev_property name="org.python.pydev.PYTHON_PROJECT_VERSION">python 2.3</pydev_property>
+<?eclipse-pydev version="1.0"?><pydev_project>
+<pydev_property name="org.python.pydev.PYTHON_PROJECT_VERSION">python 2.7</pydev_property>
 <pydev_pathproperty name="org.python.pydev.PROJECT_SOURCE_PATH">
-<path>/fail2ban-0.8/client</path>
-<path>/fail2ban-0.8/server</path>
-<path>/fail2ban-0.8/testcases</path>
-<path>/fail2ban-0.8</path>
+<path>/fail2ban-unstable/client</path>
```
yarikoptic: heh -- didn't know even that this file exists... as you can see -- doesn't generalize well across users (varying paths) /platforms (varying python versions, although now we are pretty much with 2.7 but still -- moving target) so I will simply remove it

fail2ban/server/dnsutils.py
```diff
@@ -0,0 +1,124 @@
...
+# This class contains only static methods used to handle DNS and IP
+# addresses.
+
+import re, socket, struct
+from ..helpers import getLogger
+
+# Gets the instance of the logger.
+logSys = getLogger(__name__)
+
+class DNSUtils:
```
yarikoptic: if we are to do this RFing we better than get rid of useless DNSUtils class and just keep those  functions (well -- staticmethods ;) ) at the module level

fail2ban/server/failmanager.py
```diff
@@ -43,6 +44,33 @@ def __init__(self):
...
+		
+	def setSubnetMask(self, value):
+		try:
+			self.__lock.acquire()
+			if "." in value: # 255.255.255.0 style mask
+				maskparts = value.split(".")
+				if len(maskparts) != 4:
+					raise ValueError("subnetmask must be valid")
+				for part in maskparts:
+					if int(part) < 0 or int(part) > 255:
```
yarikoptic: wouldn't `if not (0 <= int(part) <= 255)` be more pythonic and save us mighty second `int(part)` evaluation? ;)

fail2ban/server/failmanager.py
```diff
@@ -43,6 +44,33 @@ def __init__(self):
...
 		self.__failTotal = 0
+		self.__subnetMask = None
+		
+	def setSubnetMask(self, value):
+		try:
+			self.__lock.acquire()
+			if "." in value: # 255.255.255.0 style mask
+				maskparts = value.split(".")
+				if len(maskparts) != 4:
+					raise ValueError("subnetmask must be valid")
```
yarikoptic: I prefer to give as informative error message as possible, e.g. here  `"subnetmask %s is not valid since doesn't conform 255.255.255.255 style mask" % value`

fail2ban/server/failmanager.py
```diff
@@ -43,6 +44,33 @@ def __init__(self):
...
+				for part in maskparts:
+					if int(part) < 0 or int(part) > 255:
+						raise ValueError("subnetmask must be valid")
```
yarikoptic: same here -- spit out that "part" into the msg

fail2ban/server/failmanager.py
```diff
@@ -43,6 +44,33 @@ def __init__(self):
...
+				self.__subnetMask = len(re.search("(?<=b)1+", bin(DNSUtils.addr2bin(value))).group())
```
yarikoptic: should or shouldn't we issue a warning whenever someone provides obscure netmask (e.g. by typo??) e.g. '254.255.255' which wouldn't resolve to binary leading with 1s and trailing with 0s?
imho we should and we should fail with a bang so smth like
```python
len(re.match("0b(1+)(0*)$", bin(DNSUtils.addr2bin(value)).group(1))
```
could be the pattern to look for while separating re.match and if None -- puking error?

fail2ban/server/failmanager.py
```diff
@@ -43,6 +44,33 @@ def __init__(self):
...
+			else:
+				intValue = int(value)
+				if intValue < 1 or intValue > 32:
```
yarikoptic: gy -- I want to see the user which would set his netmask option to 1 (may be military would want to go to full lockdown upon the first sight of trouble? ;) )

fail2ban/server/failmanager.py
yarikoptic: ah apparently it is already a pattern used in inIgnoreIPList  -- so we thus need a function with proper unit-test for it

fail2ban/server/filter.py
```diff
@@ -357,7 +375,7 @@ def getIgnoreIP(self):
 	# @param ip IP address
 	# @return True if IP address is in ignore list
 
-	def inIgnoreIPList(self, ip, log_ignore=False):
+	def inIgnoreIPList(self, ip, subnetmask=None, log_ignore=False):
```
yarikoptic: unit-testing for cases with subnetmask should be provided

fail2ban/server/failmanager.py
```diff
@@ -92,6 +120,8 @@ def addFailure(self, ticket):
 			ip = ticket.getIP()
 			unixTime = ticket.getTime()
 			matches = ticket.getMatches()
+			if self.__subnetMask is not None:
+				ip = DNSUtils.bin2addr(DNSUtils.addr2bin(ip, self.__subnetMask)) + "/" + str(self.__subnetMask)
```
yarikoptic: and we would need to unit-test that banning with netmasking actually works.

yarikoptic:
also docs would need to be adjusted (man/jail.conf.5) and jail.conf get sample setting I guess?

.pydevproject
leeclemens: Why is this file in vcs? seems outdated, at best, considering min python version at least

leeclemens: @yarikoptic very much 2.6..considering https://github.com/fail2ban/fail2ban/issues/1106

yarikoptic: > > -<path>/fail2ban-0.8</path>
> > +<path>/fail2ban-unstable/client</path>

> Why is this file in vcs? seems outdated, at best, considering min python
> version at least

yeap yeap... so I removed it now altogether

ramon-garcia:
ping

VampireSilence:
Is someone still working on this?

jameskimmel:
Is this still being worked on? 
Seems like IPv6 makes the issue even worse. 
Some residential customers already get /48 subnets. Even if we only block /64 subnets, a single customer could attack from 65,536 subnets.

PR #1168, opened 2015-08-25
Ispconfig-auth rule

AlexanderSk:
Adds support for ispconfig banning


AlexanderSk:
I retested the jail and i can see it can actually see the failed attempts when issuing
journalmatch =
backend  = polling

Fail2Ban v0.9.2
Centos 7 64bit

PR #1172, opened 2015-08-28
Update datetemplate.py to add spanish month names

darues:
In order to correctly detect the spanish months it's mandatory to add their names in the dictionary table, as should be done with the rest of the supported languages (http://lh.2xlibre.net/values/abmon/)


yarikoptic:
d'oh - didn't mention (although did think -- didn't we get rid of this already?) that it was against 0.8 branch... oh well... the fact is that not sure if there would be any next 0.8 release ;) (beyond a tag on github at some point)

PR #1177, opened 2015-09-02
Create htaccess.conf

denics:
In our configuration I cannot use IPTables because all of our traffic come from a firewall that I do not control, with the firewall IP. This solution is not the most elegant, as htaccess slow down the server, but it is better than having attackers...

### DIFF
--- a/config/action.d/htaccess.conf
+++ b/config/action.d/htaccess.conf
@@ -35,7 +35,7 @@ actioncheck =
 # Values:  CMD
 # You can have multiple virtual hosts
 #
-actionban = printf %%b "Deny from <ip>\n" >> "<htaccess1>"
+actionban = printf %%b "Deny from <ip>\n" >> "<htaccess>"
 #	printf %%b "Deny from <ip>\n" >> "<htaccess2>"
 
 # Option:  actionunban
@@ -46,8 +46,8 @@ actionban = printf %%b "Deny from <ip>\n" >> "<htaccess1>"
 #          <time>  unix timestamp of the ban time
 # Values:  CMD
 #
-actionunban = sed -i '/Deny from <ip>$/d' "<htaccess1>"
-	sed -i '/Deny from <ip>$/d' "<htaccess2>"
+actionunban = sed -i '/Deny from <ip>$/d' "<htaccess>"
+#	sed -i '/Deny from <ip>$/d' "<htaccess2>"
 
 [Init]
 
@@ -55,5 +55,5 @@ actionunban = sed -i '/Deny from <ip>$/d' "<htaccess1>"
 # Notes.:  htaccess file path.
 # Values:  STR Default /var/www/.htaccess
 #
-htaccess1 = /var/www/.htaccess
+htaccess = /var/www/.htaccess
 #htaccess2 = /var/www/site2/.htaccess
--- a/config/action.d/htaccess.conf
+++ b/config/action.d/htaccess.conf
@@ -1,9 +1,8 @@
 # Fail2Ban configuration file
-#
-# Author: Denis Pitzalis
-#
-# $Revision$
-#
+# htaccess
+# You can ban IPs using the Deny directive of Apache in your htaccess. 
+# You should add, at the very end of the .htaccess  "Order Deny,Allow"
+# This action will append the "deny" rule at the end of the file.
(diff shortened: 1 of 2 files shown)


config/action.d/htaccess.conf
```diff
@@ -0,0 +1,66 @@
+# Fail2Ban configuration file
+#
+# Author: Denis Pitzalis
+#
+# $Revision$
```
yarikoptic: no need for those ... we should strip them all eventually -- remnant from SVN

config/action.d/htaccess.conf
```diff
@@ -0,0 +1,66 @@
...
+
+[Init]
+
+# Option:  localhost
+# Notes.:  htaccess file path.
+# Values:  STR Default /var/www/drupal-git/src/.htaccess
+#
+htaccess1 = /var/www/site1/.htaccess
+htaccess2 = /var/www/site2/.htaccess
+htaccess3 = /var/www/site3/.htaccess
```
yarikoptic: remove lines for htaccess3, comment all out for htaccess2, make default match the actual default value for htaccess1 one way or another, add yourself to THANKS, add changelog entry

config/action.d/htaccess.conf
```diff
@@ -0,0 +1,66 @@
...
+# Option:  actionunban
+# Notes.:  command executed when unbanning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    <ip>  IP address
+#          <failures>  number of failures
+#          <time>  unix timestamp of the ban time
+# Values:  CMD
+#
+actionunban = IP=<ip> &&
+	sed -i "/Deny from $IP/d" <htaccess1>
```
yarikoptic: must be anchored at the end ```/Deny from $IP$$/```.  Just in case -- place `<htaccess1>` within `""`?

yarikoptic: and not sure if worthwhile creating `IP=<ip>` -- just plug `<ip>` within sed line, and then can use `'` and thus a single `$` for anchoring

config/action.d/htaccess.conf
```diff
@@ -0,0 +1,66 @@
+# Fail2Ban configuration file
+#
```
yarikoptic: Please give a brief description for the action here and hints on what apache settings must be inplace to make it work

config/action.d/htaccess.conf
```diff
@@ -0,0 +1,59 @@
+# Fail2Ban configuration file
+# htaccess
+# You can ban IPs using the Deny directive of Apache in your htaccess. 
+# You should add, at the very end of the .htaccess  "Order Deny,Allow"
```
yarikoptic: so shouldn't there be 
```
actionstart = grep "^Order " <htaccess1> || echo "Order Deny,Allow" >> <htaccess1>
```

config/action.d/htaccess.conf
```diff
@@ -0,0 +1,59 @@
...
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    <ip>  IP address
+#          <file>  htaccess
+# Values:  CMD
+# You can have multiple virtual hosts
+#
+actionban = printf %%b "Deny from <ip>\n" >> "<htaccess1>"
+#	printf %%b "Deny from <ip>\n" >> "<htaccess2>"
```
yarikoptic: On a sober thought ;):
- rename htaccess to htaccess
- remove any htaccess2

since we can simply configure multiple actions per jail, thus allowing arbitrary number of htaccess files without need to modify this action

denics: Yep, right!

yarikoptic:
so @denics -- interested to finish up ?  there should be only a single htaccess file entry and there is an outstanding comment to address

denics:
yep, sorry! I was travelling. I get back to this asap.

yarikoptic:
@denics still interested to pursue this PR?

denics:
Hi @yarikoptic sure, sorry. In the meantime, the system has been running for the last 4 months on our servers and it works as good as expected.

sebres:
see #1811

PR #1179, opened 2015-09-04
Update ChangeLog

denics:
added myself


yarikoptic:
gy -- I meant to be directly added to #1177, but that is ok -- I will merge this and #1178 whenever #1177 gets in

PR #1184, opened 2015-09-08
Bugfix for dnsToIp resolver for fqdn with large list of IPs;

### LINKED ISSUE #1164: fail2ban crash with some fqdn
Hi,

When a attack come from a fqdn who has lot of ip-address we find fail2ban stuck. He don't really crash but until we remove the line inside the log file, fail2ban do nothing.

The fqdn is unassigned.infomir.com.ua, when you try a nslookup on this name you got...1362 ip-address.

Regards.

sebres:
closes #1164

### DIFF
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -855,8 +855,9 @@ class DNSUtils:
 		""" Convert a DNS into an IP address using the Python socket module.
 			Thanks to Kevin Drapel.
 		"""
+		# retrieve ip (todo: use AF_INET6 for IPv6)
 		try:
-			return set(socket.gethostbyname_ex(dns)[2])
+			return set([i[4][0] for i in socket.getaddrinfo(dns, None, socket.AF_INET, 0, socket.IPPROTO_TCP)])
 		except socket.error, e:
 			logSys.warning("Unable to find a corresponding IP address for %s: %s"
 						% (dns, e))
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
 
 - Fixes:
    * roundcube-auth jail typo for logpath
+   * Fix dnsToIp resolver for fqdn with large list of IPs (gh-1164)
 
 - New Features:
 


fail2ban/server/filter.py
```diff
@@ -855,8 +855,9 @@ def dnsToIp(dns):
 		""" Convert a DNS into an IP address using the Python socket module.
 			Thanks to Kevin Drapel.
 		"""
+		# retrieve ip (todo: use AF_INET6 for IPv6)
 		try:
-			return set(socket.gethostbyname_ex(dns)[2])
+			return set([i[4][0] for i in socket.getaddrinfo(dns, None, socket.AF_INET, 0, socket.IPPROTO_TCP)])
```
yarikoptic: is there a  point to override "wide" defaults for the getaddrinfo's socktype and proto (last two parameters) ?

yarikoptic: i.e. wouldn't we might miss some entries?

but why fail2ban was getting stuck? socket.gethostbyname_ex was getting stuck for me it just pukes:
```
In [2]: socket.gethostbyname_ex('unassigned.infomir.com.ua')
---------------------------------------------------------------------------
herror                                    Traceback (most recent call last)
<ipython-input-2-8e5a8f862f26> in <module>()
----> 1 socket.gethostbyname_ex('unassigned.infomir.com.ua')

herror: [Errno 1] Unknown host
```
so should have returned an empty list and nothing indeed would have been banned (but shouldn't get stuck).  Now it would return 1000s of IPs and ban ALL of them... not sure which behaviour is better (and we do recommend to not rely on names but rather log IPs), but the question -- what /why was "stuck"??

sebres: short I don't know, but I had a time ago the same issue...

sebres: and as regards the defaults: sometimes that will prevent to retrieve the same adresses twice (kind of a `distinct`, I know that the `set` does also, but ...) - if you don't like it, I can remove it.

yarikoptic: > > -                     return set(socket.gethostbyname_ex(dns)[2])
> > +                     return set([i[4][0] for i in socket.getaddrinfo(dns, None, socket.AF_INET, 0, socket.IPPROTO_TCP)])

> and as regards the defaults: sometimes that will prevent to retrieve the same
> adresses twice (kind of a distinct, I know that the set does also, but ...) -
> if you don't like it, I can remove it.

Indeed let's remove the last two for now. 
Also please add above it a comment e.g.

# socket.AF_INET limits to IPv4, remove for IPv6 support

yarikoptic: as for unknown origin of the stall -- I will comment on the original post I guess

yarikoptic: i.e. before we know what was going on, I am not sure if we could/should claim that issue was fixed

sebres: well, since then I had used it (already a long time) - I had no once this issue.

yarikoptic:
ok, I guess we would never get necessary detail to figure out original case... and probably previous behaviour of failing DNS whenever name associated with too many IPs was a side-effect feature, but this implementation is "the right thing" so -- let's go with it. thanks

PR #1188, opened 2015-09-13
RF(TST): self.assertTrue(self._is_logged()) -> self.assertLogged

yarikoptic:
and corresponding one for not + support for multiple entries at once,
and failure message listing actual log to ease troubleshooting


yarikoptic:
no comments/objections from @fail2ban/contributors ?

PR #1191, opened 2015-09-16
ENH: new date pattern with year after day (not after entire entry)

yarikoptic:
e.g. Sun Jan 23 2005 21:59:59.011
ref: http://bugs.debian.org/798923


yarikoptic:
checks passed, no objections expressed -- win-win I guess

PR #1203, opened 2015-09-26
fail2ban-testcases-all hang 

maxbeth:
Hello,
the urlopen has no timeout, therefore it hang if a firewall block outgoing traffic.
Regards.


yarikoptic:
awesome!  thanks

PR #1205, opened 2015-10-02
Add drupal-botsearch filter and jail

cdubz:
Piggybacking off @leeclemens #1035, this filter detects excessive "page not found" errors (entry point snooping) in syslog entries for Drupal installations.


sebres:
We have already several botsearch-filters (apache, nginx), etc. I don't think, we need special filters like this for drupal or any other CMS, but... @fail2ban/contributors Any comments?

BTW. See https://github.com/fail2ban/fail2ban/pull/1070#issuecomment-111754915 and https://github.com/fail2ban/fail2ban/pull/1070#issuecomment-113232258.

cdubz:
I actually rather agree that you don't want a bunch of CMS-specific filters. I just had to create this one because Drupal captures these 404 errors so I can't use the standard Apache error log with the botsearch filter.

sebres:
But you can surely use an include configuration `botsearch-common.conf` where the botsearch-regexp logic was outsourced.

szepeviktor:
I have a load of *entry point snooping* for WordPress:
https://github.com/szepeviktor/wordpress-fail2ban
For example: https://github.com/szepeviktor/wordpress-fail2ban/blob/master/block-bad-requests/wp-fail2ban-bad-request-instant.inc.php#L263-L269

szepeviktor:
Actually HTTP/404 is not meant for Fail2ban and it is very dangerous, it need *maintenance*.
My Fail2ban 404 `maxretry` is 10 minutes.

cdubz:
Both fair points. Do you think it's worth a new PR with ```botsearch-common.conf``` integrated or should this just not be accepted since it's CMS-specific and HTTP/404?

szepeviktor:
I think Fail2ban is slow for monitoring access logs.
Maybe it will change in the future.
I use Fail2ban only for small logs.

sebres:
> a new PR with botsearch-common.conf integrated

It would be nice... but you don't need a new PR. Just amend commit in your branch with botsearch-common included (github will do all in current PR).
I mean, amend hereafter you have added an include in `config/filter.d/drupal-404.conf` and have replaced your regexp with something using `<block>` substitution, for example like here https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/nginx-botsearch.conf#L11.

But I am still not convinced, that we need it really (as part of the f2b) and as long as other @fail2ban/contributors keep silent ... :)

> botsearch monitors Apache access logs.

It has little to do with including of botsearch-common itself - you configure a `logpath` in jail.conf/local. In this sense the `botsearch-common.conf` is a **central** collection of regular expressions only and possibly more **safe** as simple 404 detection, because will be already maintained by community.

cdubz:
> you don't need a new PR. Just amend commit in your branch

Ah nice, I'm learning... (:

I'll make the changes and commit (probably tomorrow) either way. Thanks for the help!

leeclemens:
On 10/02/2015 08:43 AM, Viktor Szépe wrote:
> I think Fail2ban is slow for monitoring access logs.

Application X's use/purpose of the log file does not influence fail2ban's performance. Frequency of writes play a far more significant role.

> Maybe it will change in the future.

Any performance related pull requests are welcome, I'm sure.

> I use Fail2ban only for small logs.

Ignoring initialization (and even that has been addressed a while ago), the size of the log file is of little consequence.
Also, Drupal can log to SYSLOG; you can easily configure SYSLOG to log Drupal to its own dedicated file.

szepeviktor:
OK-OK. Excuse me for arguing.

leeclemens:
@cdubz Examples of risks related to blocking based on 404's is favicon.ico or any other URI that is automatically requested by a browser (possibly including /apple-app-site-association) or broken links from your same or other sites, all of which may be requested with no malicious intent. Just some more detailed background info :)

szepeviktor:
> any other URI that is automatically requested by a browser

https://github.com/szepeviktor/wordpress-plugin-construction/blob/master/shared-hosting-aid/README.md#root-files

cdubz:
Thanks all for the pointers and info. I'll just leave this here and let you make the call on whether or not it should be merged in (:

yarikoptic:
So I feel that conclusion was not to add this filter since it might cause DoS for legitimate users in various scenarios, thus I am closing it for now at least.  If I misunderstood, please reopen etc

leeclemens:
I think this PR was updated (as requested in https://github.com/fail2ban/fail2ban/pull/1205#issuecomment-145069540) to use botsearch-common.conf, which looks for explicitly defined paths (which are also used by the [apache|nginx]-botsearch filters).

Drupal uses a specific "access log" format, just as both Apache and nginx both do (and there are separate filters for).

Perhaps renaming to drupal-botsearch.conf would help making it more identifiable and fit with the current apache and nginx focused filters using botsearch?

If the conclusion is to not include this filter on the basis that it is CMS dependent, that's entirely different and open to discussion.

Note:
The best info I could find quickly was organization permissions, which indicate the OP cannot reopen this Issue, @yarikoptic

yarikoptic:
since discussion is ongoing -- reopening.  My concern was either this one could be triggered on a normal drupal installation without malicious intents?  Many sites might be 404-ing in general.  When 404 for scripts which must not be there (e.g. searching for admin pannels etc) it is one, but banning for any 404 seems to be a recipe for DoS in many cases.

cdubz:
As @leeclemens notes, my second commit (988c7bb) addressed the non-malicious 404 concerns by taking advantage of the general ```botsearch-common.conf``` rules that look for specific malicious patterns. I agree that renaming to ```drupal-botsearch.conf``` is wise and will push another commit when I get a chance.

The debate about whether CMS-specific rules like this should be included at all "rages" on (:

yarikoptic:
Cool then. Don't forget about changeling and thanks files :-)

PR #1206, opened 2015-10-02
ssh.conf: Fix disconnect "Auth fail" matching

kevinoid:
The regex for matching against the "Auth fail" disconnect log message from sshd does not match against current versions of OpenSSH.  OpenSSH 5.9 introduced privilege separation of the pre-auth process, which included [logging through monitor.c](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/monitor.c.diff?r1=1.113&r2=1.114) which adds " [preauth]" to the end of each message and causes the log level to be prepended to each message.

It also fails to match against clients which send a disconnect message with a description that is either empty or includes a space, since this is the content in the log message after the disconnect code, per [packet.c:1785](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/packet.c?annotate=1.215), which was matched by `\S+`.  Although I have not observed this yet, I couldn't find anything which would preclude it in [RFC 4253](https://tools.ietf.org/html/rfc4253#section-11.1) and since the message is attacker-controlled it provides a way to avoid getting banned.

This PR fixes both issues.  I hope you find it useful.

Thanks for considering,
Kevin

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,8 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
    * Fix dnsToIp resolver for fqdn with large list of IPs (gh-1164)
    * filter.d/apache-badbots.conf
      - Updated useragent string regex adding escape for `+`
+   * filter.d/sshd.conf
+     - Updated "Auth fail" regex for OpenSSH 5.9 and later
    * Treat failed and killed execution of commands identically (only
      different log messages), which addresses different behavior on different
      exit codes of dash and bash (gh-1155)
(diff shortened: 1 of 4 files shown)


sebres:
Thx... looks not bad (except `.*` in regexp, but I think it will stay, if it logs really so now).

ChangeLog and THANKS entries?

kevinoid:
Great, glad it looks good (excepting the `.*`, which seems unavoidable to me, since it matches against an attacker-controlled string).

I've added a commit to update the ChangeLog and THANKS files.  Let me know how it looks and if there's anything else that needs work.

sebres:
as well as old and new tests passed, so merged now.
Thx @kevinoid

kevinoid:
My pleasure.  Thanks @sebres!

Issue #289, opened 2013-07-13
Not able to detect password-less SSH login attempts


moneytoo:
I'm using openssh-server-5.3p1-84.1.el6 and have disabled password login. Login attempts of non existing users are detected and correctly banned (test from Android client):

```
Jul 13 18:43:05 mdop sshd[4848]: Invalid user user from 89.24.13.192
Jul 13 18:43:05 mdop sshd[4849]: input_userauth_request: invalid user user
Jul 13 18:43:05 mdop sshd[4849]: Received disconnect from 89.24.13.192: 3: com.jcraft.jsch.JSchException: Auth fail
```

However attempts with valid user are not detected.

```
Jul 13 18:44:28 mdop sshd[4931]: Received disconnect from 89.24.13.192: 3: com.jcraft.jsch.JSchException: Auth fail
```

or

```
Jul 13 19:59:19 mdop sshd[5489]: Received disconnect from 194.228.20.113: 14: No supported authentication methods available
```

It might seem unnecessary to block SSH connections if it uses key certs but I understand that fail2ban helps avoiding unnecessary load/traffic. I also have setup default action `iptables-allports` where I expect that intruders trying SSH logins first are blocked from from other running services.
These messages are already mentioned in #175. Isn't there any workaround for this? Different log level or something?

EDIT: I had to write all this to realize that (at least on Centos which I use) there are all attempts correctly listed in /var/log/audit/audit.log (with appropriate host and timestamp). I guess I have to check if there's any filter for that.

```
type=USER_LOGIN msg=audit(1373740007.039:1197): user pid=5567 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:sshd_t:s0-s0:c0.c1023 msg='op=login acct="root" exe="/usr/sbin/sshd" hostname=? addr=89.24.13.224 terminal=ssh res=failed'
```

PR #1209, opened 2015-10-05
sshd-aggressive (new ssh rules added (gh-864) and code review...)

### LINKED ISSUE #864: ssh new rule to be added
The following attack is not detected by fail2ban version 0.9.1 (gentoo)

 Nov 24 23:46:39 weurope sshd[32686]: SSH: Server;Ltype: Version;Remote: 121.199.22.83-1780;Protocol: 2.0;Client: libssh2_1.4.3
Nov 24 23:46:40 weurope sshd[32686]: SSH: Server;Ltype: Kex;Remote: 121.199.22.83-1780;Enc: aes128-ctr;MAC: hmac-sha1;Comp: none [preauth]
Nov 24 23:46:41 weurope sshd[32686]: SSH: Server;Ltype: Authname;Remote: 121.199.22.83-1780;Name: root [preauth]
Nov 24 23:46:43 weurope sshd[32686]: fatal: Read from socket failed: Connection reset by peer [preauth]

To detect it I did add in sshd.conf the following:
^.*Remote: <HOST>-.*\n.*fatal: Read from socket failed: Connection reset by peer \[preauth\]$
This needs to be changed to a proper/better regexp.

### LINKED ISSUE #1117: I'd like to see a rule blocking ssh attempts that fail to negotiate a key exchange
I followed the recommendations at https://stribika.github.io/2015/01/04/secure-secure-shell.html and now I see lots of inbound connection attempts that never even make it to login attempts, because the client is using ssh connection parameters that are not compatible.  Therefore, my logs are full of such connection attempts.

Here's some example lines:

```
Jul 16 01:19:55 bminton sshd[22440]: Connection from 43.229.53.67 port 39678 on 192.168.1.9 port 22
Jul 16 01:20:00 bminton sshd[22440]: fatal: Unable to negotiate a key exchange method [preauth]
```

I'd like to request the sshd filter have that line added.

thanks

### LINKED ISSUE #1545: fail2ban/config/filter.d/sshd.conf: match "Unable to negotiate with <ip> port <port>: no matching cipher found. Their offer: <ciphers>,none [preauth]"
Stupid bots are annoying. Log example: `sept  12 13:03:29 conversionready sshd[45]: Unable to negotiate with 31.168.172.145 port 55419: no matching cipher found. Their offer: aes256-cbc,rijndael-cbc@lysator.liu.se,aes192-cbc,aes128-cbc,arcfour128,arcfour,3des-cbc,none [preauth]`

sebres:
test all config-regexp, that contains greedy catch-all before `<HOST>`, that is hard-anchored at end or precise sub expression after `<HOST>`;

new ssh rule(s) added:
- Connection reset by peer (multi-line line during authorization process);
- No supported authentication methods available;
- No matching cipher found;
- No matching key exchange method found;

Single line and multi-line expression optimized, added optional prefixes and suffix (logged from several ssh versions);
Closes #864
Closes #1545
Closes #1117
yoh edited1: just `ed HOST

### DIFF
--- a/config/filter.d/suhosin.conf
+++ b/config/filter.d/suhosin.conf
@@ -17,7 +17,7 @@ _daemon = (?:lighttpd|suhosin)
 
 _lighttpd_prefix = (?:\(mod_fastcgi\.c\.\d+\) FastCGI-stderr:\s)
 
-failregex = ^%(__prefix_line)s%(_lighttpd_prefix)s?ALERT - .* \(attacker '<HOST>', file '.*'(?:, line \d+)?\)$
+failregex = ^%(__prefix_line)s%(_lighttpd_prefix)s?ALERT - .*? \(attacker '<HOST>', file '.*'(?:, line \d+)?\)$
 
 ignoreregex = 
 
--- a/config/filter.d/suhosin.conf
+++ b/config/filter.d/suhosin.conf
@@ -17,7 +17,7 @@ _daemon = (?:lighttpd|suhosin)
 
 _lighttpd_prefix = (?:\(mod_fastcgi\.c\.\d+\) FastCGI-stderr:\s)
 
-failregex = ^%(__prefix_line)s%(_lighttpd_prefix)s?ALERT - .*? \(attacker '<HOST>', file '.*'(?:, line \d+)?\)$
+failregex = ^%(__prefix_line)s%(_lighttpd_prefix)s?ALERT - .*? \(attacker '<HOST>', file '[^']*'(?:, line \d+)?\)$
 
 ignoreregex = 
 
--- a/fail2ban/tests/samplestestcase.py
+++ b/fail2ban/tests/samplestestcase.py
@@ -103,68 +103,75 @@ def testSampleRegexsFactory(name, basedir):
 			os.path.isfile(os.path.join(TEST_FILES_DIR, "logs", name)),
 			"No sample log file available for '%s' filter" % name)
 
-		logFile = fileinput.FileInput(
-			os.path.join(TEST_FILES_DIR, "logs", name))
-
-		# test regexp contains greedy catch-all before <HOST>, that is
-		# not hard-anchored at end or has not precise sub expression after <HOST>:
-		for fr in self.filter.getFailRegex():
-			if RE_WRONG_GREED.search(fr): #pragma: no cover
-				raise AssertionError("Following regexp of \"%s\" contains greedy catch-all before <HOST>, "
-					"that is not hard-anchored at end or has not precise sub expression after <HOST>:\n%s" %
-					(name, str(fr).replace(RE_HOST, '<HOST>')))
-
 		regexsUsed = set()
-		for line in logFile:
-			jsonREMatch = re.match("^# ?failJSON:(.+)$", line)
-			if jsonREMatch:
-				try:
-					faildata = json.loads(jsonREMatch.group(1))
-				except ValueError as e:
-					raise ValueError("%s: %s:%i" %
-						(e, logFile.filename(), logFile.filelineno()))
-				line = next(logFile)
-			elif line.startswith("#") or not line.strip():
-				continue
-			else:
-				faildata = {}
+		filenames = [name]
+		i = 0
+		while i < len(filenames):
+			filename = filenames[i]; i += 1;
+			logFile = fileinput.FileInput(os.path.join(TEST_FILES_DIR, "logs",
+				filename))
 
-			ret = self.filter.processLine(
-				line, returnRawHost=True, checkAllRegex=True)[1]
-			if not ret:
-				# Check line is flagged as none match
-				self.assertFalse(faildata.get('match', True),
-					 "Line not matched when should have: %s:%i %r" %
-					(logFile.filename(), logFile.filelineno(), line))
-			elif ret:
-				# Check line is flagged to match
-				self.assertTrue(faildata.get('match', False),
-					"Line matched when shouldn't have: %s:%i %r" %
-					(logFile.filename(), logFile.filelineno(), line))
-				self.assertEqual(len(ret), 1, "Multiple regexs matched %r - %s:%i" %
-								 (map(lambda x: x[0], ret),logFile.filename(), logFile.filelineno()))
+			# test regexp contains greedy catch-all before <HOST>, that is
+			# not hard-anchored at end or has not precise sub expression after <HOST>:
+			for fr in self.filter.getFailRegex():
+				if RE_WRONG_GREED.search(fr): #pragma: no cover
+					raise AssertionError("Following regexp of \"%s\" contains greedy catch-all before <HOST>, "
+						"that is not hard-anchored at end or has not precise sub expression after <HOST>:\n%s" %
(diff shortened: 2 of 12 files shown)


sebres:
Old as well as new test cases passed.
@yarikoptic, @fail2ban/developers : some counter-arguments against this PR?
Otherwise I will add changelog entries and merge.

yarikoptic:
on vacation atm, gimme few days since this one is the main filter and change mixes optimizations and extensions and not easy to review while on the phone ;)

sebres:
@yarikoptic no problem, have a nice holiday...

config/filter.d/suhosin.conf
```diff
@@ -17,7 +17,7 @@ _daemon = (?:lighttpd|suhosin)
 
 _lighttpd_prefix = (?:\(mod_fastcgi\.c\.\d+\) FastCGI-stderr:\s)
 
-failregex = ^%(__prefix_line)s%(_lighttpd_prefix)s?ALERT - .* \(attacker '<HOST>', file '.*'(?:, line \d+)?\)$
+failregex = ^%(__prefix_line)s%(_lighttpd_prefix)s?ALERT - .*? \(attacker '<HOST>', file '.*'(?:, line \d+)?\)$
```
yarikoptic: any particular reason for making first '.*' non-greedy here? as well not really related to what PR targets really, correct?

fail2ban/tests/samplestestcase.py
```diff
@@ -31,12 +31,22 @@
...
 
+# regexp to test greedy catch-all should be not-greedy:
+RE_HOST = Regex('<HOST>').getRegex()
+RE_WRONG_GREED = re.compile(r'\.[+\*](?!\?).*' + re.escape(RE_HOST) + r'.*(?:\.[+\*].*|[^\$])$')
+# test asserts:
+assert RE_WRONG_GREED.search('greedy .* test' + RE_HOST + ' test not hard-anchored')
+assert RE_WRONG_GREED.search('greedy .+ test' + RE_HOST + ' test vary .* anchored$')
+assert not RE_WRONG_GREED.search('greedy .* test' + RE_HOST + ' test no catch-all, hard-anchored$')
+assert not RE_WRONG_GREED.search('non-greedy .*? test' + RE_HOST + ' test not hard-anchored')
+assert not RE_WRONG_GREED.search('non-greedy .+? test' + RE_HOST + ' test vary catch-all .* anchored$')
```
yarikoptic: why not to embed it into a proper test method somewhere?  I understand that it is more of testing that non-greedy matching still functions as assumed, but it is just yet another test -- why making it here at the module level asserts?

config/filter.d/suhosin.conf
sebres: just now this kind of error will be checked in samplestestcase (otherwise test case `testFilter` will fail on this regexp)

fail2ban/tests/samplestestcase.py
sebres: I prefer that test of variable is intact is near the place to variable self; But if you necessarily want that as testcase, I can move it.

yarikoptic: I see no reason for variable(s) to be more special than e.g. functions. but it somewhat occludes code, so better indeed be moved into a test case. also failure here would ruin entire module not importable, so all tests would "fail" although possibly being correct.

sebres: done (moved)

sebres:
Ready. Amend commited with changelog and moved asserts for regexp (as test case now).
(Failed in IP2Name - DNS bug for googlebot).

sebres:
Rebased over 61ac48170308da590d317cac7e0b846461a86352

yarikoptic:
NB that is why I don't like reviewing mix of RF and NF... 

one thing at a time: new rule "No supported authentication methods available" -- we have discussed it briefly in #864 without decision to include it and as Dan rightfully commented in https://github.com/fail2ban/fail2ban/issues/289#issuecomment-20929052  it indeed could be triggered in legitimate case.  So it should not be a part of sshd filter which intended to catch illegit attempts to login which might eventually succeed.  Those which intended to drain resources are at best intended for ddos one, but even there, if it could be triggered by a legit user doing no malicious stuff, it should not be a part of any filter in my opinion.

yarikoptic:
probably the same about "Connection reset by peer ..." -- I am just not sure if it couldn't be triggered in a legit case for some reason (unreliable connection etc) - nothing in that regex directly points to malicious attempt to login.  I would hate to go on a trip and be unable to login simply because I am on some slow and jiggery (if there is such a word) connection (and I had those quite often) connection.  So this PR should concentrate on RF part of it.  For the new added cases we might want a separate filter then, some users might want to enable if they feel like it... we could call it   sshd-aggressive or sshd-paranoid or smth along those lines but shouldn't be a part of the default sshd filter.

For that we might want to add sshd-common.conf to be included by both of those where those common pref/suff variables move into to be reused by both filters (and probably sshd-ddos as well I guess?)

also that new `__suff` with preauth was added to all lines now, although many of those aren't logs for preauth stage ... what for?  better not to dilute regexes without necessity

sebres:
> "No supported authentication methods available" ... could be triggered in legitimate case.

Many things could be triggered in legitimate case (ex.: http 401 first time before really authorized), but it should reach maxretry...

> that new __suff with preauth was added to all lines now, although many of those aren't logs for preauth stage

See https://github.com/fail2ban/fail2ban/pull/1206/files#diff-56cef6a5cfe6e39eceea88d750652307R30, it was also without suffix before OpenSSH 5.9.

I like the idea with `sshd-aggressive`, will try to do it.

yarikoptic:
should we wait for sshd-aggressive? ;-)

sebres:
Yes we schould :)

yarikoptic:
Oh  @sebres, "I can wait so good"  ;-)

sebres:
Well, the time is finite >o)

sebres:
> I can wait so good ;-)

the wait was worth it :)

PS too many changes in sshd (and basically in master) since this PR, so I've rebased it now

sebres:
I've extended ddos (aggressive) filter with 2 other regex (closes #1545 and #1117) - no matching cipher  resp. no matching key exchange method found

TimothyGu:
Is there anything blocking this PR from being applied? I am currently using a packaged version of fail2ban so I was wondering if I could backport this change manually to my configuration.

sebres:
> Is there anything blocking this PR from being applied?

From my side there is nothing, @yarikoptic what would you say?

The changes are not so big if compared ignoring whitespaces (because of indentation change) - https://github.com/fail2ban/fail2ban/pull/1209/files?w=1

dimon222:
Trying to do the same thing @TimothyGu is doing, but for whatever reason I can't get it to work. Still getting this spam in my SSH logs " fatal: Read from socket failed: Connection reset by peer [preauth]" without any activity from fail2ban. Isn't this regex change supposed to mitigate it?

sebres:
> Isn't this regex change supposed to mitigate it?

Yes, the last rule in sshd-ddos (or sshd-aggressive) does this

sebres:
rebased on new master

sebres:
Closes #1652

Issue #1211, opened 2015-10-06
mysqld-auth not working due to format change


justinjameshayes:
I was having problems with mysql 5.6 error logs and found this:

http://dev.mysql.com/doc/relnotes/mysql/5.6/en/news-5-6-9.html

"mysqld now writes dates to the error log in ISO (YYYY-MM-DD hh:mm:ss) format. It also includes its process ID following the date. Thanks to Davi Arnaut for the patch. (Bug #56240, Bug #11763523)"

So there really should have two separate filters

This is what I'm using for the new format:
failregex = ^%(__prefix_line)s(\d{4}-\d{2}-\d{2}\s?\d{2}:\d{2}:\d{2} )?\s?\d+\s\[Warning\] Access denied for user '\w+'@'<HOST>' (to database '[^']*'|\(using password: (YES|NO)\))*\s*$

sebres:
Can you provide an example for that log entry (include it between \`\`\` ... \`\`\` for correct representation in github)?

noramtkane:
Sorry, I created the issue while logged into the wrong github account.

Here is the new format:
```2015-10-07 06:09:42 5907 [Warning] Access denied for user 'root'@'218.104.143.56' (using password: YES)```

and here is the old one for comparison:
```151007  6:13:33 [Warning] Access denied for user 'root'@'180.97.28.11' (using password: YES)```

sebres:
Fixed in #1212, Thanks to @justinjameshayes (@noramtkane)

zchellpy:
Hai, this regex cant match my maria db log, this is my log

2018-12-26  0:31:35 140502474295040 [Warning] Access denied for user 'admin'@'192.168.1.85' (using password: YES)

My sql log have 2 spasi after date. Thanks u for helping.

PR #1212, opened 2015-10-07
mysqld-auth: Updated "Access denied ..." regex for MySQL 5.6 and later

sebres:
closes gh-1211


sebres:
old as well as new test cases passed, so will merge after travis ready

yarikoptic:
Thanks
In general though give at least 24h for others to get a chance to review/merge. Ideally PRs of one developers should be merged by another developer ( unless none follows-up within those 24h). How about that?

PR #1213, opened 2015-10-08
Make sshd jail default to systemd on Fedora

opoplawski:
A lot of nice work was done a while back to allow distros to configure the default logpaths, e.g. 
~~~~~
logpath = %(sshd_log)s
~~~~~
I'd like to be able to do the same for configuring fail2ban to use the journal by default for certain jails without having to maintain a cumbersome patch.  This is an attempt to start doing so.  Thoughts?

See also https://bugzilla.redhat.com/show_bug.cgi?id=1269226

### DIFF
--- a/config/paths-debian.conf
+++ b/config/paths-debian.conf
@@ -9,6 +9,8 @@ after  = paths-overrides.local
 
 [DEFAULT]
 
+sshd_backed = auto
+
 syslog_mail = /var/log/mail.log
 
 syslog_mail_warn = /var/log/mail.warn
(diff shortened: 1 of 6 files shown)


config/paths-debian.conf
```diff
@@ -9,6 +9,8 @@ after  = paths-overrides.local
 
 [DEFAULT]
 
+sshd_backed = auto
```
yarikoptic: typo (`backe[n]d`) here and in the other files, which rightfully caught up by tests on travis.
+ please add a changelog in Enhancements?
otherwise sounds reasonable ;)

PR #1215, opened 2015-10-08
files: Strip trailing whitespace from files

paulmenzel:
Run the command `StripWhitespace` from the [Vim Better Whitespace
Plugin](https://github.com/ntpeters/vim-better-whitespace).

### DIFF
--- a/files/bash-completion
+++ b/files/bash-completion
@@ -31,7 +31,7 @@ __fail2ban_jail_action_methods () {
 
 _fail2ban () {
     local cur prev words cword
-    _init_completion || return 
+    _init_completion || return
 
     case $prev in
         -V|--version|-h|--help)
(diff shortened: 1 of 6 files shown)


sebres:
Thanks, but ... 
If that would be a source code I would have said: see verdict of @yarikoptic about things like this: https://github.com/fail2ban/fail2ban/issues/876#issuecomment-122382586
For all files affected from this PR, I see no problem, but I stay neutral and will say @fail2ban/developers should decide...

yarikoptic:
this one is concerned primarily with side shell scripts and docs in which I don't expect any conflicts with existing  forks, so should be ok to merge as to me ;) @sebres do you see being anyhow affecting your  forks? if not -- merge in! ;)

sebres:
aye-aye!

PR #1219, opened 2015-10-10
Updated route.conf to clear warnings

agentmoller001:
Does not throw warnings when starting/restarting by adding three lines of code.


yarikoptic:
Thanks!

PR #1220, opened 2015-10-10
Banning complete IP subnet /24 attacks

agentmoller001:
Using the iptables-multiport-24.conf file will help defend against attacks from an entire subnet.

### DIFF
--- a/config/action.d/iptables-multiport-24
+++ b/config/action.d/iptables-multiport-24
@@ -1,0 +1,74 @@
+# Fail2Ban configuration file
+#
+# Author: Cyril Jaquier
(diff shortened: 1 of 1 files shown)


yarikoptic:
Please have a look our better tryhttps://github.com/fail2ban/fail2ban/pull/1161

sebres:
I believe we've already enough iptables-actions (that BTW could be easy extended with option `subnet`, e. g. to use something like this without creating of additional action-file:
```INI
# jail.local:
[my-strict-jail]
banaction = iptables-multiport[subnet="/24"]`
```

PR #1223, opened 2015-10-17
Implementing fail2ban with OpenHAB home automation software

1technophile:
With the help of Serg G. Brester who gave me the date pattern I wrote these two filters so as to enable to ban IP that fails authenticate to openHAB, either by web interface or rest api.

If you think it can be interesting for others.


sebres:
Looks good to me, but:
- ChangeLog entry (Enhancements)?
- I don't think we should add such rare using jails to `jail.conf`, @yarikoptic what do you think? 
Not at all? As comment? (I know per default all jails are disabled, but nevertheless - each such enhancement makes config reader to work slower and slower :). If we will decide to not include such jails, many other rare jails like this should be commented/removed also.

config/filter.d/openhab.conf
```diff
@@ -0,0 +1,15 @@
+# Openhab brute force auth filter: /etc/fail2ban/filter.d/openhab.conf:
+#
+# Block IPs trying to auth wp openhab
+#
+# Matches e.g.
+# 12.34.33.22 -  -  [26/sept./2015:18:04:43 +0200] "GET /openhab.app HTTP/1.1" 401 1382 
+#
+[Definition] 
+failregex = ^<HOST> -  - \s*\[\]\s* "GET /openhab.app HTTP/1.1" 401 1382.*$
```
yarikoptic: It seems that it should be possible to simply merge both regular expressions (path1|path2) and this avoid duplication etc

config/jail.conf
```diff
@@ -306,6 +306,18 @@ port    = http,https
...
+filter = openhab
+action = iptables-allports[name=NoAuthFailures]
+logpath = /opt/openhab/logs/request.log
+
+[openhab-auth-rest]
+
+filter = openhab-rest
+action = iptables-allports[name=NoAuthFailures]
+logpath = /opt/openhab/logs/request.log
+
```
yarikoptic: With merged failregex would be possible to have just a single jail.

yarikoptic:
Hm... How slow is slower? In general I would just keep adding them since allows for easy immediate use by just enabling them. As long as loading it is not ridiculously long I would keep it growing, and when it gets really slow - into speeding it up ;-)

sebres:
Well, it's already >100x accelerated, since I have implemented a lazy includes / config sharing inside config reader, but... 
I still prefer to rename `jail.conf` to `jail.conf.org` on my machines to hold it small, and use `jail.local` alone.

config/filter.d/openhab.conf
1technophile: Yes I think so, I will do that

config/filter.d/openhab.conf
```diff
@@ -0,0 +1,16 @@
+# Openhab brute force auth filter: /etc/fail2ban/filter.d/openhab.conf:
+#
+# Block IPs trying to auth openhab by web or rest api
+#
+# Matches e.g.
+# 12.34.33.22 -  -  [26/sept./2015:18:04:43 +0200] "GET /openhab.app HTTP/1.1" 401 1382 
+# 175.18.15.10 -  -  [02/sept./2015:00:11:31 +0200] "GET /rest/bindings HTTP/1.1" 401 1384
+
+[Definition] 
+failregex = ^<HOST> -  - \s*\[\]\s* "GET /.*HTTP/1.1" 401 (1382|1384)
```
sebres: two things:
- please do not use not hard-anchored at end (`$`) regexp together with greedy catch-all (`.*`)
- I believe the last numbers are variable (size of 401-page? url-length? can be changed for future versions), don't match it
- don't restrict request method (GET, POST,...) and protocol (HTTP/1.1), may be a weak spot later;

I think this regexp would be better (not tested):
```regexp
^<HOST>\s+-\s+-\s+\[\]\s+"[A-Z]+ .*" 401 \d+\s*$
```

1technophile: It's working and it's integrated, thanks

sebres:
Rebased to 1 commit + changelog and thanks entries added
Thx @1technophile

PR #1225, opened 2015-10-19
Add *_backend options for services to allow distros to set the default backend

opoplawski:
Add *_backend options for services to allow distros to set the default backend per service.
Set default to systemd for Fedora as appropriate.


config/jail.conf
```diff
@@ -491,18 +501,21 @@ logpath  = /root/path/to/assp/logs/maillog.txt
...
 port     = smtp,465,submission
 logpath  = %(postfix_log)s
+backend  = %(postfix_backend)s
 
 
 [postfix-rbl]
 
 port     = smtp,465,submission
-logpath  = %(syslog_mail)s
+logpath  = %(postfix_mail)s
```
yarikoptic: there is no postfix_mail -- postfix_log

config/paths-common.conf
```diff
@@ -8,8 +8,10 @@ after  = paths-overrides.local
 [DEFAULT]
 
 sshd_log = %(syslog_authpriv)s
+sshd_backend = auto
```
yarikoptic: may be it is worth introducing `default_backend = auto` and then reusing that one as default for all so with a single change all could be changed, instead of per each one

ChangeLog
```diff
@@ -34,6 +34,8 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      (Thanks M. Maraun)
    * Added check against atacker's Googlebot PTR fake records
      (Thanks Pablo Rodriguez Fernandez)
+   * Add *_backend options for services to allow distros to set the default
+     backend per service, set default to systemd for Fedora as appropriate
```
yarikoptic: add changelog for the postfix_rbl path change

opoplawski:
Ping?

yarikoptic:
Looks good to me. @fail2ban/developers?

sebres:
:+1:

Issue #1194, opened 2015-09-23
apache-fakegooglebot - stdout: ''


frollic:
I get these in the fail2ban.log

```
2015-09-23 01:37:30,065 fail2ban.action         [1661]: ERROR   /etc/fail2ban/filter.d/ignorecommands/apache-fakegooglebot 5.39.222.253 -- stdout: ''
2015-09-23 01:37:30,066 fail2ban.action         [1661]: ERROR   /etc/fail2ban/filter.d/ignorecommands/apache-fakegooglebot 5.39.222.253 -- stderr: ''
2015-09-23 01:37:30,066 fail2ban.action         [1661]: ERROR   /etc/fail2ban/filter.d/ignorecommands/apache-fakegooglebot 5.39.222.253 -- returned 1
2015-09-23 01:37:30,462 fail2ban.actions        [1661]: NOTICE  [apache-fakegooglebot] Ban 5.39.222.253
```
matching entry in access_log is
```
5.39.222.253 - - [23/Sep/2015:01:37:29 +0200] "GET / HTTP/1.0" 200 49 "-" "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"
```
fail2ban comes from FC22:
fail2ban-0.9.3-1.fc22.noarch
fail2ban-server-0.9.3-1.fc22.noarch
fail2ban-sendmail-0.9.3-1.fc22.noarch
fail2ban-firewalld-0.9.3-1.fc22.noarch

sebres:
This script can return 1 (without stdout/stderr) if host resolved for the IP adress does not match `crawl-.*\.googlebot\.com` (see https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/ignorecommands/apache-fakegooglebot#L29).
I think, it can be expected situation, but at the moment I don't know how the logging of this "error" can be prevented.

@yarikoptic do you know by chance, it was always that way, or approximately it is result of #1155 (resp. #1187)?

starrychloe:
Still happening.

```
2016-09-02 21:26:30,774 fail2ban.actions        [1341]: NOTICE  [apache-fakegooglebot] 139.59.31.211 already banned
2016-09-02 21:26:30,879 fail2ban.action         [1341]: ERROR   /etc/fail2ban/filter.d/ignorecommands/apache-fakegooglebot 139.59.31.211 -- stdout: ''
[...]

PR #1116, opened 2015-07-15
Collection of all kinds of changes, fixes, refactorings [Was: F2B performance branch]

sebres:
- huge increasing of fail2ban performance and especially test-cases performance (see #1109);
- some features and extensions were partially cherry picked from #716;
- prepared to easy merge or rebase later a newest extended version of 'ban-time-incr';
- closes #1048, closes #1050 (but still partially, don't fix "purge will never called, currently still test case only", fixed in #716).

PR #1229, opened 2015-10-22
Enhancements on the ignorecommand execution

pablorf-dev:
- ignorecommand is executed just before banning, instead of every time an ip is found
- Fail of ignorecommand is not logged with ERROR level, but INFO

### DIFF
--- a/fail2ban/server/action.py
+++ b/fail2ban/server/action.py
@@ -532,7 +532,7 @@ class CommandAction(ActionBase):
 		return self.executeCmd(realCmd, self.timeout)
 
 	@staticmethod
-	def executeCmd(realCmd, timeout=60):
+	def executeCmd(realCmd, timeout=60, testCmd=False):
 		"""Executes a command.
 
 		Parameters
@@ -541,6 +541,8 @@ class CommandAction(ActionBase):
 			The command to execute.
 		timeout : int
 			The time out in seconds for the command.
+		testCmd : bool
+			If the command is used as a test; return nonzero is not an error.
 
 		Returns
 		-------
@@ -591,7 +593,7 @@ class CommandAction(ActionBase):
 			_cmd_lock.release()
 
 		std_level = retcode == 0 and logging.DEBUG or logging.ERROR
-		if std_level >= logSys.getEffectiveLevel():
+		if std_level >= logSys.getEffectiveLevel() and not testCmd:
 			stdout.seek(0)
 			logSys.log(std_level, "%s -- stdout: %r" % (realCmd, stdout.read()))
 			stderr.seek(0)
@@ -611,7 +613,8 @@ class CommandAction(ActionBase):
 				(realCmd, signame.get(sigcode, "signal %i" % sigcode), retcode))
 		else:
 			msg = _RETCODE_HINTS.get(retcode, None)
-			logSys.error("%s -- returned %i" % (realCmd, retcode))
+			std_level = logging.INFO if testCmd else logging.ERROR
+			logSys.log(std_level, "%s -- returned %i" % (realCmd, retcode))
 			if msg:
 				logSys.info("HINT on %i: %s"
 							% (retcode, msg % locals()))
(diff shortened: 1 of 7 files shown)


fail2ban/server/action.py
```diff
@@ -532,7 +532,7 @@ def _processCmd(self, cmd, aInfo = None):
 		return self.executeCmd(realCmd, self.timeout)
 
 	@staticmethod
-	def executeCmd(realCmd, timeout=60):
+	def executeCmd(realCmd, timeout=60, testCmd=False):
```
sebres: this is not practicable in such a way to outwit the tests...

sebres:
BTW 1) test cases will failed (`AttributeError: 'DummyJail' object has no attribute 'filter'`) so long as dummy jail does not get test filter object.
2) Because testIgnoreCommand failed also, something fundamental is wrong in this PR, imho.

sebres:
Seems to met the "error" `AttributeError: 'DummyJail' object has no attribute 'filter'`, while I wrote test cases for my observer module.
I have "repaired" or avoided it either in #716 (or in #1116).

fail2ban/server/action.py
pablorf-dev: Why? What is the problem?

sebres: Because some branch lines will be not executed during test cases, but in production (and possible makes boom).
Counter question: why you will exclude seek/log for the test case?

pablorf-dev: testCmd is not addressed to test environment but when a command is used as a test condition to something; in this case, with ignorecommand, which returns a nonzero or zero depending on which IP should be ignored or not.

Using this var, you are telling executeCmd to do not log the nonzero return as an ERROR, because it is expected. This closes #1194 .

sebres: Well, understood. Then me confused name of the parameters (why this parameter called testCmd?).
I will make a new branch above your one. Wait a little bit.

sebres:
I don't like the solution with "ignorecommand is executed just before banning, instead of every time an ip is found".
If I understood it correct, you just want not execute a command for each failure made from IP (because possible slow, too many calls, etc.).

Why I don't like it just before banning? Because:
- the entry will be still added through FailManager as failure ticket (what is not preferable resp. even not reasonable in some f2b-forks);
- the IP will be recognized as "IP to be ignored" much later as the failure actually occurred (but that can be important);
- the same can be made using small short-alive caching added in `inIgnoreIPList` (I have made such cache for DNS-resolver inside #1116);

fail2ban/server/action.py
pablorf-dev: I used this name to expose that this command is a test on a condition; may be a better name?

pablorf-dev:
I understand your cons, but my primary pro is the reduction of the server load; using a cache will only reduce the time spent resolving names, but ignorecommand could not resolve names.

sebres:
> using a cache will only reduce the time spent resolving names, but ignorecommand could not resolve names.

Perhaps you did not understand - new cache over result of ignorecommand.

pablorf-dev:
What I meant is that ignorecommand used may be a command that may not perform any name resolution, e.g., a grep for an IP.

sebres:
For better understanding the code:
```python
 	def inIgnoreIPList(self, ip, log_ignore=False):
+		# cached?:
+		v = CACHE_IgnoreIP.get(ip)
+		if v is not None: 
+			return v
...
 		if self.__ignoreCommand:
 			command = CommandAction.replaceTag(self.__ignoreCommand, { 'ip': ip } )
 			logSys.debug('ignore command: ' + command)
			v = CommandAction.executeCmd(command)
...
+		# save in cache:
+		CACHE_IgnoreIP.set(ip, v)
		return v
```

That's it.

pablorf-dev:
Ok, now I got it. However, I have some cons:
- Conceptually ignorecommand is not a list (this is a minor con)
- Still ignorecommand is executed when not needed, e.g., cache has expired and that IP access is still below defined limits

In any case, having such cache would be nice.

sebres:
not sure we should close it, may be usefull in some cases.

sebres:
closed by #2176

PR #1232, opened 2015-10-27
Mac OS Screen Sharing filter

sibrcode:
Adds filtering of system.log for authentication attempts against the Screen Sharing service where a user account & password are used. Password only VNC authentication is not covered here, and is not disclosed in the system.log.


sebres:
Thx for the PR.
You need add a sample log test file for 'screensharingd' filter `fail2ban/tests/files/logs/screensharingd` to pass the test cases (at least one line matched and one not matched).

And don't forget changelog and thanks files :)

config/filter.d/screensharingd.conf
```diff
@@ -0,0 +1,33 @@
...
+_daemon = screensharingd
+
+# Option:  failregex
+# Notes.:  regex to match the password failures messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
+#          be used for standard IP/hostname matching and is only an alias for
+#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
+# Values:  TEXT
+#
+failregex = ^.+ screensharingd.+: Authentication: FAILED :: User Name: .+ :: Viewer Address: <HOST> :: .*$
```
sebres: Please do not use not hard-anchored at end (`.*$`) regexp together with greedy catch-all (`.*`). 
Especially dangerous is it if the greedy catch-all take place before <HOST> - it is a potential weak spot resp. even vulnerability.
Simple way to fix it is to change it to non-greedy catch-all (`.+?`) but it makes search of match slower.
Better would be to state more precisely your regular expression as exact as possible resp. to resolve a not hard-anchored end (`.*` before `$`) in this regexp.

sibrcode:
Hi Serg,

Thanks. I've added some additional commits to address the issues you mentioned:
* use a literal instead of wildcard at end of regex
* added in sample of log file results
* also added in logencoding statement to avoid encoding warnings

sebres:
Well, still contains some errors:
- sample log file should have the same name as filter;
- sample log file should contain especial json structure (before each line should be matched and not matched), see another in this directory for example https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/openhab#L2
- you've forgot a changelog enhancement entry.

You can test it self before pushing to github, use:
```
cd /path/to/f2b
PYTHONPATH=. bin/fail2ban-testcases SampleRegexs
```

config/filter.d/screensharingd.conf
```diff
@@ -0,0 +1,33 @@
...
+_daemon = screensharingd
+
+# Option:  failregex
+# Notes.:  regex to match the password failures messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
+#          be used for standard IP/hostname matching and is only an alias for
+#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
+# Values:  TEXT
+#
+failregex = ^.+ screensharingd.+: Authentication: FAILED :: User Name: .+ :: Viewer Address: <HOST> :: Type: DH$
```
sebres: expression `^.+` makes no sense;
you should not restrict type to DH (`\w+` is absolutely sufficient)

Below the regexp that should work (untested)
```
failregex = ^%(__prefix_line)sAuthentication: FAILED :: User Name: .+ :: Viewer Address: <HOST> :: Type: \w+$
```

sibrcode: I admittedly wasn't sure how __prefix worked, and wasn't expecting that it would do the right thing in this case. But running it does work as expected, and matches the result of my previous regex, so I'll definitely change it to match.

Old version:
 ```fail2ban-regex /var/log/system.log  /etc/fail2ban/filter.d/screensharingd.conf 

Running tests
=============

Use   failregex filter file : screensharingd, basedir: /etc/fail2ban
Use         log file : /var/log/system.log
Use         encoding : UTF-8


Results
=======

Failregex: 22 total
|   1) [22] ^.+ screensharingd.+: Authentication: FAILED :: User Name: .+ :: Viewer Address: <HOST> :: Type: DH$
`-

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [3820] (?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?
|  [6] Year(?P<_sep>[-/.])Month(?P=_sep)Day 24hour:Minute:Second(?:,Microseconds)?
`-

Lines: 4013 lines, 0 ignored, 22 matched, 3991 missed [processed in 0.28 sec]
Missed line(s): too many to print.  Use --print-all-missed to print all 3991 lines```

New Version:

```fail2ban-regex /var/log/system.log  /etc/fail2ban/filter.d/screensharingd.conf 

Running tests
=============

Use   failregex filter file : screensharingd, basedir: /etc/fail2ban
Use         log file : /var/log/system.log
Use         encoding : UTF-8


Results
=======

Failregex: 22 total
|   1) [22] ^\s*(<[^.]+\.[^.]+>)?\s*(?:\S+ )?(?:kernel: \[ *\d+\.\d+\] )?(?:@vserver_\S+ )?(?:(?:\[\d+\])?:\s+[\[\(]?screensharingd(?:\(\S+\))?[\]\)]?:?|[\[\(]?screensharingd(?:\(\S+\))?[\]\)]?:?(?:\[\d+\])?:?)?\s(?:\[ID \d+ \S+\])?\s*Authentication: FAILED :: User Name: .+ :: Viewer Address: <HOST> :: Type: \w+$
`-

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [3820] (?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?
|  [6] Year(?P<_sep>[-/.])Month(?P=_sep)Day 24hour:Minute:Second(?:,Microseconds)?
`-

Lines: 4013 lines, 0 ignored, 22 matched, 3991 missed [processed in 0.30 sec]
Missed line(s): too many to print.  Use --print-all-missed to print all 3991 lines```

**Comment by sebres** [comment]
please rename sample log file to filter name

File: config/filter.d/screensharingd.conf (hunk 2 above)
Comment (sebres) [inline config/filter.d/screensharingd.conf]: __prefix_line contains platform-independent standard prefix-expressions for syslog-lines, like hostname, daemon name, pid etc.
That will be automatically set from fail2ban in accordance with your distribution. 

And as I already said `^.+` makes no sense and can be removed (to parse anchored from end of line only) or even replaced with prefix_line.

**Comment by sibrcode** [comment]
How should I fix the test errors from the logpath statement?

```logpath  = %(system_log)s```

Since it is using a value defined in paths-osx.conf, it appears to me it will always fail b/c OSX paths will never be included in the test environment.

**Comment by sebres** [comment]
I don't know in which log file the messages will be written... and `system_log` as variable does not exists (I don't think the filtering of whole system.log is a good idea at all - just too many messages).

For the list of internal variables see corresponding file `paths-*.conf`.
For example for debian it is https://github.com/fail2ban/fail2ban/blob/master/config/paths-debian.conf

Otherwise just define your one without substitute a variable - use fixed path...
```
logpath = /var/log/screensharingd
# or even
logpath = /var/log/syslog
```

**Comment by sibrcode** [comment]
I defined system_log in the **paths-osx.conf** file, so I suspect the unit tests are never loading this. It works for me since I've switched out the **paths-debian.conf* for the OSX one, but I'll just switch to a fixed path now.

Unfortunately, the system.log is the only location to find the screensharingd login info, (secure.log gone since 10.8). It could be in the ASL(OpenBSM) logs, but these are in a binary format.

File: fail2ban/tests/files/logs/screensharingd [hunk 3]
```diff
@@ -0,0 +1,18 @@
+# NOTE: dates here include years -- this is not the typical configuration for the system.log
+# file on Mac OS. However, without it the test routines will use 2004 as the year and matches will not pass.
+#
+# failJSON: { "match": false }
+Oct 27 2015 09:24:46 test1.beezwax.net screensharingd[1170]: Authentication: SUCCEEDED :: User Name: simon :: Viewer Address: 192.168.5.247 :: Type: DH
+# failJSON: { "match": false }
```
sebres: the next 3 messages have no bearing on this filter... please remove it

config/paths-osx.conf
```diff
@@ -25,3 +25,5 @@ syslog_authpriv = /var/log/secure.log
 
 #syslog_local0 =
 
+# Default Mac OS log location for syslog output.
```
sebres: this changes are not required if you used direct specified `logpath` in `jail.conf`

config/jail.conf
```diff
@@ -786,3 +786,12 @@ returntype   = DROP
 bantime      = 3600
 maxretry     = 1
 findtime     = 1
+
+[screensharing]
+# For Mac OS Screen Sharing Service (VNC)
+# Requires the 'before' statement in the [INCLUDE] section to include paths-osx.conf
```
sebres: please remove it (because not required if you used direct specified `logpath` in `jail.conf`)

sebres:
Except these 3 peculiarities (I wrote above) it looks good to me.
I can merge it tomorrow if @fail2ban/contributors  are not against

szepeviktor:
Nice to have more filters.

config/filter.d/screensharingd.conf
```diff
@@ -0,0 +1,33 @@
+# Fail2Ban configuration file
+#
+# Author: Simon Brown
+#
+# $Revision: 1 $
```
yarikoptic: Those are old from svn, do not bother ;-)

fail2ban/tests/files/logs/screensharingd
yarikoptic: Well, they are negative hits but indeed might be too far to be practically useful here?

config/paths-osx.conf
sibrcode: Hmm, since the unit tests appear to not load it I guess there is no point to it then.

sibrcode:
Any chance of getting this pull request merged? I don't think there are any outstanding issues left.

yarikoptic:
OK I will merge in a second offline to resolve conflicts and some post merge tuneups

PR #1236, opened 2015-10-29
New filter `nginx-limit-req` 

sebres:
used to ban hosts, that were failed through nginx by limit request processing rate (ngx_http_limit_req_module)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,11 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
    * Fix jail.conf.5 man's section (gh-1226)
 
 - New Features:
+   * New filters:
+     - openhab - domotic software authentication failure with the
+       rest api and web interface (gh-1223)
+     - nginx-limit-req - ban hosts, that were failed through nginx by limit
+       request processing rate (ngx_http_limit_req_module)
 
 - Enhancements:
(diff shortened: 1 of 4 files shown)


sebres:
Since v. 0.7.21, nginx has a new module `ngx_http_limit_req_module`, used to limit the request processing rate per a defined key, in particular, the processing rate of requests coming from a single IP address.

Long time had it as 'failregex', but because too many community demands and many issues for example with wrong regular expression, now added it as revised version of filter and new jail.

sebres:
Example configuration:
```nginx
  http {
    ...
    limit_req_zone $binary_remote_addr zone=lr_zone:10m rate=1r/s;
    ...
    # http, server, or location:
    location @backend {
      limit_req zone=lr_zone burst=1 nodelay;
      ...
    }
    ...
  }
  ...
```

JimOct:
Excellent. A great way to do this would be limiting requests from PHP Documents only, otherwise you'd limit all requests (including CSS).

sebres:
Sometimes is also needed to limit static requests (ddos etc), but I have intentionally wrote it in example nginx configuration `location @backend`, to paint the usage of limit_req for dynamic locations (PHP etc).

gagomap:
Thanks,

Issue #1216, opened 2015-10-09
Specific banaction in jail.conf overides default in jail.local


stewardle:
Where banaction for specific jails (eg pam-generic) is defined in jail.conf it must be redefined for those jails in jail.local.
I expected the default banaction in jail.local to take precedence.

sebres:
It's not a bug - is a feature :) 
So works config reader mechanism, that the *.local completely overwrites all of *.conf, but read of defaults occurs in each file separately, before it will be merged together, so the defaults are not interested after each file was read.

If this logic will be changed, other things can be broken.
I need some info to think about - can you provide your situation as example?

stewardle:
On Fri, 09 Oct 2015 03:50:00 -0700
"Serg G. Brester" <notifications@github.com> wrote:

> It's not a bug - is a feature :) 
> So works config reader mechanism, that the *.local completely overwrites all of *.conf, but read of defaults occurs in each file separately, before it will be merged together, so the defaults are not interested after each file was read.
> 
> If this logic will be changed, other things can be broken.
> I need some info to think about - can you provide your situation as example?
> 

I'd updated fail2ban from 8.10 to 9.3 on Mac OS X.

In action.d have a modified version of pf.conf to use Murus firewall anchors.

In jail.local banaction was only specified for default and jail recidive:
	[DEFAULT]
	...
	banaction = pf-murus
	...
	[pam-generic]
	enabled  = true
	action = %(action_mw)s
	logpath  = /var/log/system.log

	[postfix-sasl]
	...
	[recidive]
	enabled  = true
	banaction = pf-murus[pftable=recidive]
	...

On starting fail2ban it tries to use banaction = iptables-multiport for jail pam-generic.

Easily fixed in jail.local
	[pam-generic]
	enabled  = true
	banaction = pf-murus

But I did not expect to have to redefine banaction for pam-generic after setting default.

Steve

sebres:
Can you additionaly provide a dump output of
```bash
fail2ban-client -vvd
```

stewardle:
Attached fail2ban-client dumps and jail.local for current (working) and original configurations and jail.conf (which is as installed).

sebres:
Attached where?

stewardle:
Attached to email I sent but not showing here. Tried via web but kept getting 
"Unfortunately, we don’t support that file type" or "Something went really wrong and we can't process that file" with or without .txt suffix.
Trying dropbox:
[...]

PR #1238, opened 2015-10-30
Fixed directly defined banaction for allports jails like pam-generic, recidive, etc

sebres:
Fixed directly defined banaction for allports jails like pam-generic, recidive, etc with new default variable `banaction_allports` (+ man entries for both variables added);

closes gh-1216

### DIFF
--- a/man/jail.conf.5
+++ b/man/jail.conf.5
@@ -146,6 +146,12 @@ Ensure syslog or the program that generates the log file isn't configured to com
 .B logencoding
 encoding of log files used for decoding. Default value of "auto" uses current system locale.
 .TP
+.B banaction
+default banning action (iptables-multiport) for all jails specified in the \fI[DEFAULT]\fR section.
+.TP
+.B banaction_allports
+default allports banning action (iptables-allports) for some jails like "pam-generic" or "recidive", specified in the \fI[DEFAULT]\fR section.
+.TP
 .B action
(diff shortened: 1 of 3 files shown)


man/jail.conf.5
```diff
@@ -146,6 +146,12 @@ Ensure syslog or the program that generates the log file isn't configured to com
 .B logencoding
 encoding of log files used for decoding. Default value of "auto" uses current system locale.
 .TP
+.B banaction
+default banning action (iptables-multiport) for all jails specified in the \fI[DEFAULT]\fR section.
```
yarikoptic: Default section doesn't have jails. Jails are all other sections and default one just provides defaults

sebres: Yes, but I meant: the banning action (for all jails) specified in the `[DEFAULT]` section.
Not the jails in default section. My bad english :(

yarikoptic: It could also be specified per jail. Again, DEFAULTS are just defaults. So no need to state where it is since it can be anywhere

sebres: Yes I know, but point of this commit resp. of this variables to specify it central in [DEFAULT] section of `jail.local`, just to not specify it per jail.
Also one place for all "normal" jails and one for "allports" jails...

yarikoptic: Yes. But my comment is about documentation.  Now it is misleading and needs to be fixed up.

sebres: > Now it is misleading

Now it is available (before it was at all not documented) ... but I've corrected my "mistake" in 576719198805063c1c3b30a7cb31f7beedd7c2f8.

PR #1239, opened 2015-11-01
Add journalmatch to postfix-rbl filter

rumpelsepp:


yarikoptic:
Although obvious one, did you check that such change is sufficient and jail works as should?
Please add changelog entry

yarikoptic:
@rumpelsepp bzz bzz

PR #1240, opened 2015-11-01
ENH: .mailmap file to bring some names together for git shortlog -sn

yarikoptic:
@fail2ban/developers  and @fail2ban/contributors  did I miss anyone else with multiple git commit identities?

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,5 @@
+Lee Clemens <java@leeclemens.net>
+Serg G. Brester <info@sebres.de>
+Serg G. Brester <serg.brester@sebres.de>
+Serg G. Brester <sergey.brester@W7-DEHBG0189.wincor-nixdorf.com>
+Viktor Szépe <viktor@szepe.net>


.mailmap
```diff
@@ -0,0 +1,5 @@
+Lee Clemens <java@leeclemens.net>
+Serg G. Brester <info@sebres.de>
+Serg G. Brester <serg.brester@sebres.de>
+Serg G. Brester <sergey.brester@W7-DEHBG0189.wincor-nixdorf.com>
```
sebres: please remove this one, because not real e-mail address (I don't know at all where I have used this one!)

yarikoptic: that was actually the point of having this since without it you would double as 
```
$> git shortlog -sn | grep -i bre
    75	Serg G. Brester
     1	sebres
```
and since that is how you got into the git history -- it will now be there forever and ever! ;)

yarikoptic:
ok -- no further objections, thus merging

szepeviktor:
Thanks.

yarikoptic:
you are most welcome!!! ;)

PR #1241, opened 2015-11-02
New interpolation feature for definition config readers - `<known/parameter>`

sebres:
New interpolation feature for definition config readers - `<known/parameter>`, as extension to interpolation `%(known/parameter)s`, that does not works for filter and action init parameters;

Example:
```
# filter.d/test.conf:
[Init]
test.method = GET
baduseragents = IE|wget
[Definition]
failregex = ^%(__prefix_line)\s+"<test.method>"\s+test\s+regexp\s+-\s+useragent=(?:<baduseragents>)

# jail.local:
[test]
# use filter "test", overwrite method to "POST" and extend known bad agents with "badagent":
filter = test[test.method=POST, baduseragents="badagent|<known/baduseragents>"]
```

Test cases and documentation extended.
Test case repaired for new syntax usage - the `jail.conf` file may contain now a definition for filter with options, like `filter = test[option=...]`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,13 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      with new default variable `banaction_allports` (gh-1216)
 
 - New Features:
+   * New interpolation feature for definition config readers - `<known/parameter>`
+     (means last known init definition of filters or actions with name `parameter`).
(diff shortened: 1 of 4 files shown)


yarikoptic:
merging and will introduce minor tune ups (recommended in prev version of this PR)

Issue #875, opened 2014-12-01
DOCS for known/ introduced by https://github.com/fail2ban/fail2ban/pull/868


yarikoptic:
Need to extend man/jail.conf.5 and probably elsewhere ;)

yarikoptic:
@sebres  would you mind to add some documentation on this feature you have contributed? who would describe it better than the author? ;)

yarikoptic:
was done in 94cffece12dc2ff0efd6a8b6761b079eb9491c68

Issue #712, opened 2014-05-01
MJ12bot - try to block 


arekstuka:
Hi All 

This bot MJ12bot banging my server, (i have lot of "connection's" from this bot), i want block it by fail2ban, please take a look on log below and help me with fileregex

i did already this (but not working)

http://www.howtoforge.com/forums/archive/index.php/t-61404.html

log from my server: 

blog.mywebsite.com:80 83.149.126.98 - - [27/Apr/2014:08:20:45 +0200] "GET /robots.txt HTTP/1.0" 404 2091 "-" "Mozilla/5.0 (compatible; MJ12bot/v1.4.5; http://www.majestic12.co.uk/bot.php?+)"

sasha2002:
You need to put entire line :
Mozilla/5.0 (compatible; MJ12bot/v1.4.5; http://www.majestic12.co.uk/bot.php?+)
this is regexp :
Mozilla/5\.0 \(compatible; MJ12bot/v1\.4\.5; http\://www\.majestic12\.co\.uk/bot\.php\?+\)

gagomap:
No, it don't work. I still can't block MJ12bot by fail2ban

sasha2002:
@gagomap what is your browser agent in logs?
Do you monitor logs?

gagomap:
Yes, i monitor logs everyday. I can block them by using nginx server block, but i can't do it with fail2ban.

sasha2002:
can you paste browser agent from logs and your fail2ban config to some site like pastebin

gagomap:
Yes, this is my logs:
http://pastebin.com/XGyhsNmd

I use apache-badbots filter. 

And this is my jail.local
```bash
[nginx-badbots]

enabled  = true
port     = http,https
filter   = apache-badbots
logpath  = /var/log/nginx/*access.log
maxretry = 1
bantime  = 86400
```

sasha2002:
Can you email me at admin@sasha2002.net ?

sebres:
apache-badbots over nginx access log-file?
They have different formats, and according to a different `failregex`.
Atm for nginx we have only botsearch filter. 
To use it with nginx-log, you can customize apache-badbots, implements your own nginx-badbots or just rewrite `failregex` in your nginx-badbots jail.
[...]

PR #1242, opened 2015-11-02
Http badbots

sebres:
new apache/nginx jail and filter http-badbots - (previously apache-badbots) common filter to ban bad-bot hosts for apache/nginx (catches known spambots and software alike);
See gh-712

Based on #1241 because init definition of this filter can be extended using own regex, example:
```
[nginx-badbots]
filter = http-badbots[badbots="SomeBadBot|<known/badbots>"]
```
or
```
[nginx-badbots]
filter = http-badbots[badbots=ThisBadBotOnly, badbotscustom=ThisBadBotOnly]
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,8 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
        rest api and web interface (gh-1223)
      - nginx-limit-req - ban hosts, that were failed through nginx by limit
        request processing rate (ngx_http_limit_req_module)
+     - http-badbots - (previously apache-badbots) revised common filter to ban 
+       bad-bot hosts for apache/nginx (catches known spambots and software alike)
 
 - Enhancements:
    * Do not rotate empty log files
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,13 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      with new default variable `banaction_allports` (gh-1216)
 
 - New Features:
+   * New interpolation feature for definition config readers - `<known/parameter>`
+     (means last known init definition of filters or actions with name `parameter`).
+     This interpolation makes possible to extend a parameters of stock filter or 
+     action directly in jail inside jail.local file, without creating a separately
+     filter.d/*.local file.
+     As extension to interpolation `%(known/parameter)s`, that does not works for
+     filter and action init parameters
    * New filters:
      - openhab - domotic software authentication failure with the
        rest api and web interface (gh-1223)
--- a/config/filter.d/http-badbots.conf
+++ b/config/filter.d/http-badbots.conf
@@ -3,14 +3,25 @@
 # Regexp to catch known spambots and software alike. Please verify
 # that it is your intent to block IPs which were driven by
 # above mentioned bots.
+#
+# You can redefine or extend both parameters "badbots" and "badbotscustom" 
+# in your jails [apache-badbots] or [nginx-badbots]:
+#
+# - to redefine (totally overwrite) it using your own expression only:
+#     filter = http-badbots[badbots=BadBot0, badbotscustom=BadBot1|BadBot2]
+#
+# - to expand it with your own expression use <known/param> syntax:
+#     filter = http-badbots[badbots=BadBot0|<known/badbots>]
 
-
-[Definition]
+[Init]
 
 badbotscustom = EmailCollector|WebEMailExtrac|TrackBack/1\.02|sogou music spider
 badbots = Atomic_Email_Hunter/4\.0|atSpider/1\.0|autoemailspider|bwh3_user_agent|China Local Browse 2\.6|ContactBot/0\.2|ContentSmartz|DataCha0s/2\.0|DBrowse 1\.4b|DBrowse 1\.4d|Demo Bot DOT 16b|Demo Bot Z 16b|DSurf15a 01|DSurf15a 71|DSurf15a 81|DSurf15a VA|EBrowse 1\.4b|Educate Search VxB|EmailSiphon|EmailSpider|EmailWolf 1\.00|ESurf15a 15|ExtractorPro|Franklin Locator 1\.8|FSurf15a 01|Full Web Bot 0416B|Full Web Bot 0516B|Full Web Bot 2816B|Guestbook Auto Submitter|Industry Program 1\.0\.x|ISC Systems iRc Search 2\.1|IUPUI Research Bot v 1\.9a|LARBIN-EXPERIMENTAL \(efp@gmx\.net\)|LetsCrawl\.com/1\.0 \+http\://letscrawl\.com/|Lincoln State Web Browser|LMQueueBot/0\.2|LWP\:\:Simple/5\.803|Mac Finder 1\.0\.xx|MFC Foundation Class Library 4\.0|Microsoft URL Control - 6\.00\.8xxx|Missauga Locate 1\.0\.0|Missigua Locator 1\.9|Missouri College Browse|Mizzu Labs 2\.2|Mo College 1\.9|MVAClient|Mozilla/2\.0 \(compatible; NEWT ActiveX; Win32\)|Mozilla/3\.0 \(compatible; Indy Library\)|Mozilla/3\.0 \(compatible; scan4mail \(advanced version\) http\://www\.peterspages\.net/?scan4mail\)|Mozilla/4\.0 \(compatible; Advanced Email Extractor v2\.xx\)|Mozilla/4\.0 \(compatible; Iplexx Spider/1\.0 http\://www\.iplexx\.at\)|Mozilla/4\.0 \(compatible; MSIE 5\.0; Windows NT; DigExt; DTS Agent|Mozilla/4\.0 efp@gmx\.net|Mozilla/5\.0 \(Version\: xxxx Type\:xx\)|NameOfAgent \(CMS Spider\)|NASA Search 1\.0|Nsauditor/1\.x|PBrowse 1\.4b|PEval 1\.4b|Poirot|Port Huron Labs|Production Bot 0116B|Production Bot 2016B|Production Bot DOT 3016B|Program Shareware 1\.0\.2|PSurf15a 11|PSurf15a 51|PSurf15a VA|psycheclone|RSurf15a 41|RSurf15a 51|RSurf15a 81|searchbot admin@google\.com|ShablastBot 1\.0|snap\.com beta crawler v0|Snapbot/1\.0|Snapbot/1\.0 \(Snap Shots&#44; \+http\://www\.snap\.com\)|sogou develop spider|Sogou Orion spider/3\.0\(\+http\://www\.sogou\.com/docs/help/webmasters\.htm#07\)|sogou spider|Sogou web spider/3\.0\(\+http\://www\.sogou\.com/docs/help/webmasters\.htm#07\)|sohu agent|SSurf15a 11 |TSurf15a 11|Under the Rainbow 2\.2|User-Agent\: Mozilla/4\.0 \(compatible; MSIE 6\.0; Windows NT 5\.1\)|VadixBot|WebVulnCrawl\.unknown/1\.0 libwww-perl/5\.803|Wells Search II|WEP Search 00
 
-failregex = ^<HOST> -.*"(GET|POST|HEAD).*HTTP.*"(?:%(badbots)s|%(badbotscustom)s)"$
+
+[Definition]
+
+failregex = ^<HOST> - \S+ \[\] "([A-Z]+) \S+ HTTP[^"]+" \d+ \d+ "[^"]*" "(?:<badbots>|<badbotscustom>)"$
 
 ignoreregex =
 
@@ -19,3 +30,7 @@ ignoreregex =
 # Generated on Thu Nov  7 14:23:35 PST 2013 by files/gen_badbots.
 #
 # Author: Yaroslav Halchenko
+#
+# sebres:
+#   Since we use %(known/...)s and <known/...> syntax, the parameter "badbotscustom" no more 
+#   needed, but still used here for backwards compatibility.
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -253,6 +253,8 @@ logpath  = %(apache_error_log)s
 [apache-badbots]
 # Ban hosts which agent identifies spammer robots crawling the web
 # for email addresses. The mail outputs are buffered.
+# See filter.d/http-badbots for details
+filter   = http-badbots
 port     = http,https
 logpath  = %(apache_access_log)s
 bantime  = 172800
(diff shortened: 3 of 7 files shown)


yarikoptic:
So is this one extension to previously contributed handling of 'known' within config files? (and somewhat of a fix for #875)?

fail2ban/client/configreader.py
```diff
@@ -285,8 +285,10 @@ def getOptions(self, pOpts):
 		
 		if self.has_section("Init"):
 			for opt in self.options("Init"):
+				v = self.get("Init", opt)
```
yarikoptic: please use more descriptive names, e.g. "init_opt_value" here

fail2ban/tests/clientreadertestcase.py
```diff
@@ -311,9 +315,8 @@ def testConvert(self):
 		# is unreliable
 		self.assertEqual(sorted(filterReader.convert()), sorted(output))
 
-		filterReader = FilterReader(
-			"testcase01", "testcase01", {'maxlines': "5"})
-		filterReader.setBaseDir(TEST_FILES_DIR)
+		filterReader = FilterReader("testcase01", "testcase01", {'maxlines': "5"},
+		  share_config=self.__share_cfg, basedir=TEST_FILES_DIR)
```
yarikoptic: just another stylistic whining from me,  previously it was more obvious where args where etc, so why not just to extend on that i.e. (just typoing -- could have typos etc, do not cut/paste ;-))

```python
filterReader = FilterReader(
    "testcase01", "testcase01", {'maxlines': "5"},
   share_config=self.__share_cfg, basedir=TEST_FILES_DIR)
```

fail2ban/tests/clientreadertestcase.py
```diff
@@ -331,16 +334,34 @@ def testFilterReaderSubstitionDefault(self):
 
 	def testFilterReaderSubstitionSet(self):
 		output = [['set', 'jailname', 'addfailregex', 'to=sour@example.com fromip=<IP>']]
-		filterReader = FilterReader('substition', "jailname", {'honeypot': 'sour@example.com'})
-		filterReader.setBaseDir(TEST_FILES_DIR)
+		filterReader = FilterReader('substition', "jailname", {'honeypot': 'sour@example.com'},
+		  share_config=self.__share_cfg, basedir=TEST_FILES_DIR)
```
yarikoptic: same here -- new line after `FilterReader(` with args indented correspondingly

ChangeLog
```diff
@@ -26,11 +26,20 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      with new default variable `banaction_allports` (gh-1216)
 
 - New Features:
+   * New interpolation feature for definition config readers - `<known/parameter>`
+     (means last known init definition of filters or actions with name `parameter`).
+     This interpolation makes possible to extend a parameters of stock filter or
```
yarikoptic: minor typos with those nasty articles: "extend parameters of a stock filer or an action"

ChangeLog
```diff
@@ -26,11 +26,20 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      with new default variable `banaction_allports` (gh-1216)
 
 - New Features:
+   * New interpolation feature for definition config readers - `<known/parameter>`
+     (means last known init definition of filters or actions with name `parameter`).
+     This interpolation makes possible to extend a parameters of stock filter or 
+     action directly in jail inside jail.local file, without creating a separately
+     filter.d/*.local file.
+     As extension to interpolation `%(known/parameter)s`, that does not works for
+     filter and action init parameters
```
yarikoptic: that "did not" since it does work now (i.e. after these changes) correct?

config/jail.conf
```diff
@@ -253,6 +253,8 @@ logpath  = %(apache_error_log)s
 [apache-badbots]
 # Ban hosts which agent identifies spammer robots crawling the web
 # for email addresses. The mail outputs are buffered.
+# See filter.d/http-badbots for details
+filter   = http-badbots
```
yarikoptic: happen someone did get that apache-badbots.local, they would enjoy a regression here... But since this RF is benefitial then may be it would just suffice an "Incompatible changes" moving introduction of http-badbots on top of the changelog pointing to this RF?

man/jail.conf.5
```diff
@@ -89,13 +89,33 @@ indicates that the specified file is to be parsed before the current file.
 indicates that the specified file is to be parsed after the current file.
 .RE
 
-Using Python "string interpolation" mechanisms, other definitions are allowed and can later be used within other definitions as %(name)s. For example.
+Using Python "string interpolation" mechanisms, other definitions are allowed and can later be used within other definitions as %(name)s. 
+Additionaly fail2ban has an extended interpolation feature named \fB%(known/parameter)s\fR (means last known option with name \fBparameter\fR). This interpolation makes possible to extend a stock filter or jail regexp in .local file (opposite to simply set failregex/ignoreregex that overwrites it). For example.
```
yarikoptic: very minor -- no need for period after "For example" since example follows

man/jail.conf.5
```diff
@@ -89,13 +89,33 @@ indicates that the specified file is to be parsed before the current file.
...
 
 .RS
+.nf
 baduseragents = IE|wget
+failregex = %(known/failregex)s
+            useragent=%(baduseragents)s
```
yarikoptic: hold on -- doesn't this add a second failregex which is in its entirety just "useragent=%(baduseragents)s"????

man/jail.conf.5
```diff
@@ -253,7 +273,7 @@ The maximum period of time in seconds that a command can executed, before being
 Commands specified in the [Definition] section are executed through a system shell so shell redirection and process control is allowed. The commands should
 return 0, otherwise error would be logged.  Moreover if \fBactioncheck\fR exits with non-0 status, it is taken as indication that firewall status has changed and fail2ban needs to reinitialize itself (i.e. issue \fBactionstop\fR and \fBactionstart\fR commands).  
 Tags are enclosed in <>.  All the elements of [Init] are tags that are replaced in all action commands.  Tags can be added by the
-\fBfail2ban-client\fR using the "set <JAIL> action <ACT>" command. \fB<br>\fR is a tag that is always a new line (\\n).
+\fBfail2ban-client\fR using the "set <JAIL> action <ACT>" command. \fB<br>\fR is a tag that is always a new line (\\n).
```
yarikoptic: unnecessary change with trailing spaces, or am I missing smth?

man/jail.conf.5
```diff
@@ -327,6 +347,8 @@ Also, special values of \fIEpoch\fR (UNIX Timestamp), \fITAI64N\fR and \fIISO860
 \fBjournalmatch\fR
 specifies the systemd journal match used to filter the journal entries. See \fBjournalctl(1)\fR and \fBsystemd.journal-fields(7)\fR for matches syntax and more details on special journal fields. This option is only valid for the \fIsystemd\fR backend.
 .PP
+Similar to actions [Init] section enables filter-specific settings. All parameters specified in [Init] section can be redefined or extended in \fIjail.conf/jail.local\fR.
```
yarikoptic: "settings, all"

fail2ban/tests/clientreadertestcase.py
sebres: this one is intention! otherwise setBaseDir will never covered!

fail2ban/client/configreader.py
sebres: `v` will be use in 2 lines below. Please let such easy code fixing not as a hand brake for such PR's again. I mean you have apparently no problem with the `opt` above...

fail2ban/tests/clientreadertestcase.py
sebres: that is another case for setBaseDir coverage also (look inside it, that branches off 2 times (if self._cfg ... else ...)

ChangeLog
sebres: :+1:

ChangeLog
sebres: no, you should distinguish between `%(known/parameter)s` and `<known/parameter>`:
- `%(known/parameter)s` does not works for filter and action init parameters (definitions only, will be substituted inside Includes/JailReader/JailsReader;
- `<known/parameter>` exactly new extension for filter and action init tags;

config/jail.conf
sebres: :+1:

man/jail.conf.5
sebres: ?

man/jail.conf.5
sebres: No, this is extension of documentation for `%(known/parameter)s`, because was never documented (closes #875). 
For `<known/parameter>` is one below...

man/jail.conf.5
sebres: :+1:

man/jail.conf.5
sebres: I don't mind.

fail2ban/client/configreader.py
yarikoptic: OK, I will just do it myself after the merge if that is so much effort

ChangeLog
yarikoptic: gotcha -- thanks.  Then just a typo "does not work[s]" ;)  Also last sentence might better be worded as "It could be considered an extension to ..."

config/filter.d/http-badbots.conf
```diff
@@ -3,14 +3,25 @@
...
+#
+# - to redefine (totally overwrite) it using your own expression only:
+#     filter = http-badbots[badbots=BadBot0, badbotscustom=BadBot1|BadBot2]
+#
+# - to expand it with your own expression use <known/param> syntax:
+#     filter = http-badbots[badbots=BadBot0|<known/badbots>]
 
-
-[Definition]
+[Init]
```
yarikoptic: If these are Init now, please move that section down in the file for consistency, as Init defined at the end in the other filters

fail2ban/tests/clientreadertestcase.py
yarikoptic: I was talking only about formatting of the FilterReader  line(s).

fail2ban/tests/clientreadertestcase.py
yarikoptic: again -- my comment is just about formatting.  If argument list expands to more than 1 line, best would be to start listing all arguments from the new line so it is easier to read the code, imho.

man/jail.conf.5
yarikoptic: That line ends with "For example." which is not a complete sentence ;)  make it "For example" so it flows in the following example

man/jail.conf.5
yarikoptic: I am talking about the fact that manpage here looks like

```
 For example.

              baduseragents = IE|wget
                          useragent=%(baduseragents)s
```
giving new failregex line, not just an addition to the `%(known/failregex)s`

yarikoptic: I guess my whining is about a choosen example... usually folks would not be able to extend failregex this way since those usually are anchored on front/end.  May be better example would be to extend the same baduseragents here thus making clear(er) analogy to below example?

man/jail.conf.5
yarikoptic: great -- make it happen then and my whining comments would automagically disappear! ;)

man/jail.conf.5
sebres: But `%(known/failregex)s` should be used only in this way if possible, because `%(known/failregex)s` may contain a list of expressions (multi-line).
So if we have today in filter:
```
failregex = .* agent=test
```
And someone writes in jail
```
failregex = %(known/failregex)s|.* useragent=%(baduseragents)s
```
it will work, but ...
but tomorrow someone extend failregex with another expression (as list)
```
failregex = .* agent=test
            .* something_else=test
```
the expression in `jail` will be misleading.

Therefore I don't like an example, where the parameter `%(known/failregex)s` will be knowingly used wrong.

config/filter.d/http-badbots.conf
sebres: What has me always bothered...

sebres: And without moving of Definition/Init all the parameters are as where they were in apache-badbots (same places).
It is also better to read the parameters in this order (more consistently).

sebres:
amend b0f3dfcde4c3d10cd6bf732b0728a565e2e055ea

config/filter.d/http-badbots.conf
yarikoptic: "places" (not just line numbers) are completely different -- Definition and Init have different.  Either it is better to have Init before Definition could be left for another issue/PR -- but at this point, as long as you initiate Init section of the config, it better be consistent with other files, i.e. at the end

man/jail.conf.5
yarikoptic: agree!  I think we are converging!
but this example would be a "not so good" example showing adding a really weak (not anchored anywhere) new failregex... either you would need to extend it with some anchoring to make more realistic, or make it all more schematic, or, as I suggested, demo this on extending baduseragents (wouldn't it work as desired?)

config/filter.d/http-badbots.conf
sebres: I meant the places of the parameters, My git-dif looks like this (github dif is just too bad):
```diff
- [Definition]
+ [Init] 
  badbotscustom = EmailCollector ...
  badbots = Atomic_Email_Hunter ... 
+ [Definition]
  failregex = ... (?:%(badbots)s|%(badbotscustom)s) ...
  ignoreregex = 
```
If I do what you want, all this will be moved.

yarikoptic: yes, it will be all moved.   Are we in pursuit of minimum diff here? ;-)

man/jail.conf.5
sebres: I will try to make a better example...

config/filter.d/http-badbots.conf
sebres: Why not if under circumstances the order of sections has no meaning? :)
And you theoretically agree with me that the [Init] should take place (in all filter) above the [Definition] (another PR).

yarikoptic:
@sebres  I guess this one needs master merged to account for already introduced `known/` handling... and then refreshed review?

sebres:
And I fear, I will need to do a little fix inside...

yarikoptic:
should this be let RiP @sebres ?

sebres:
Rebased and resolved conflicts (all changes on known-mimic has seemed being already implemented in other branches, so only filter and changelog going here). I must think about it a bit, let us safe this firstly for a while (I'll block it right now). 
However I don't like such weak filters at all, especially filters for accesslog ([due to parasitic log-traffic](https://github.com/fail2ban/fail2ban/wiki/Best-practice#reduce-parasitic-log-traffic)), so I am almost for RIP of all them, but this PR can be considered as an example for creation of generic common filters. Probably I normalize all that apache/nginx stuff  we have to a single filter.

Issue #1055, opened 2015-05-18
sshd jail: ignore/reset failed attempts when publickey has been accepted


blueyed:
It would be nice, if "Accepted publickey ... from $IP" would unban that IP
again, and not consider it for the recidive filter, too.

I've managed to blacklist myself, because I've not had the ssh-agent setup
properly and use autossh etc and/or dismissed the popup asking for the
password/passphrase.

This might also happen when normal authentication succeeds, not just for
"Accepted publickey".

My current workaround for this might be to automatically add my current IP to
the whitelist - although that would be easier with something like "whitelist
for X minutes" to automatically get it purged again.  Is this worth a new
feature request?

From the logs:

auth.log:May 18 03:53:20 HN sshd[101638]: Failed password for user from X.X.X.X port 59214 ssh2
auth.log:May 18 03:53:20 HN sshd[101638]: Failed password for user from X.X.X.X port 59214 ssh2
auth.log:May 18 03:53:20 HN sshd[101638]: Connection closed by X.X.X.X [preauth]
auth.log:May 18 03:53:25 HN sshd[101700]: Accepted publickey for user from X.X.X.X port 59222 ssh2
auth.log:May 18 03:53:25 HN sshd[101702]: Received disconnect from X.X.X.X: 11: disconnected by user
auth.log:May 18 03:53:36 HN sshd[101807]: Accepted publickey for user from X.X.X.X port 59225 ssh2
auth.log:May 18 03:54:03 HN sshd[101809]: Received disconnect from X.X.X.X: 11: disconnected by user
auth.log:May 18 03:54:13 HN sshd[101887]: Accepted publickey for user from X.X.X.X port 59235 ssh2
auth.log:May 18 03:54:31 HN sshd[101930]: Accepted publickey for user from X.X.X.X port 49400 ssh2
auth.log:May 18 03:59:35 HN sshd[101889]: Received disconnect from X.X.X.X: 11: disconnected by user
auth.log:May 18 03:59:45 HN sshd[101932]: Received disconnect from X.X.X.X: 11: disconnected by user
auth.log:May 18 03:59:45 HN sshd[112301]: Failed password for user from X.X.X.X port 59378 ssh2
auth.log:May 18 03:59:45 HN sshd[112301]: Failed password for user from X.X.X.X port 59378 ssh2
auth.log:May 18 03:59:45 HN sshd[112301]: Connection closed by X.X.X.X [preauth]
[...]

PR #1243, opened 2015-11-02
Regex to reset counter for IP

tomgreen66:
A little branch to reset IP counter as discussed in #1055 - hopefully fairly sensible approach.


fail2ban/server/filter.py
```diff
@@ -161,6 +163,43 @@ def getIgnoreRegex(self):
...
+	# and thus can be used for many purporse.
+	# @param value the regular expression
+
+	def addResetRegex(self, value):
+		try:
+			regex = ResetRegex(value)
+			self.__resetRegex.append(regex)
+			if "\n" in regex.getRegex() and not self.getMaxLines() > 1:
+				logSys.warning(
+					"Mutliline regex set for jail '%s' "
```
yarikoptic: just spotted a little typo in "Multiline"

yarikoptic:
I guess it wouldn't be complete without some resetregex'es defined e.g. for sshd filter.  Or in general there should be no general ignoreregex whitelisting and users would specify specific lines matching their logins etc?
But that let me thinking what would be safer by default, if in unlikely even that one of my hosts get compromised and attacker sees (eg. via simple ps dump) to which hosts I connect and starts trying to compromise them.  If I have a succesful login with this feature it would whitelist my originating IP allowing attacker to continue his evil activities...  somewhat sucked out of the finger use-case, I agree... just a thought

sebres:
I don't like the name `resetregex`, just because:
- it should be an antonym for `failregex`;
- in some future extensions it possibly does more as just reset;
- for example in #716, it should additionally do 2 things: sound the alarm (notify admin) if the IP was previously very-very bad, otherwise mark an IP as good (remove it from table of bad ips);

Al that sounds for something like `successregex` or `gainregex` or some other really antonym for `failregex`.

fail2ban/server/filter.py
tomgreen66: Shall I fix it in addFailRegex as well? - I just copied the exact text but didnt spot that.

tomgreen66:
Thanks for the comments - in reply:
@yarikoptic - agree there are probably some edge cases but I think the positives outweight the negatives - will users of fail2ban find some even worse workaround to stop banning users who have logged in okay but due to authentication mechanism triggers ban (such as ignoring the ip all the time)?  Maybe add a warning to documentation about the use of this new option?
@sebres - I started with successregex (hence branch name) but kept on spelling successregex wrong (to many repetitive letters and also just too long to type) and also thought it wasnt quite the opposite of failregex (i.e. I wasnt whitelisting an ip if I got an accept, just resetting the counter).  Maybe change the name in the future when the use-case requires it?

sebres:
> change the name in the future

Bad idea, very bad. If that is once released it would be an incompatible change.

yarikoptic:
@tomgreen66 interested to push it forward?  I think it would be valuable to have this feature but it would require a bit more of work.  as far as I see TODOs:

- indeed decide now on the name.  Since others are imperative ('fail', 'ignore') this one should be as well.  'reset' is indeed a bit ambiguous, but may be 'succeed', 'clear', 'forget'...? (after writing below with 'succeed', I came up with 'forget' so will use 'forgetregex' for now, but if better ideas -- I have no strong feeling for it) 
- at least sshd should get its entry within this PR.  This will help to "properly" resolve #1263
- we should extend failJSON specs used within `fail2ban/tests/files/logs` and support in the code to specify what regex it should match (now matches failregex) (or we could define forgetJSON, instead of adding some key to specify which regex to match, this one would assume the forgetJSON)
- possibly make this "forget failures upon successful login" an option for each jail (with default setting to be decided), so paranoid me, could disable it to prevent scenarios as the one I have depicted
Sounds like a plan?

tomgreen66:
Hi - sorry for the radio silence had some rather big personal circumstances crop up so did not have time to push this forward.  Happy to work on this now if people would find it useful.  Especially thanks for the input in December it is appreciated so can work on those points.

sebres:
Welcome back :)

BTW, as regards the controversial name of the new regexp - in my opinion, this could be called `gainregex`, also to be exact so long (aligned) in the config files...
```
gainregex = ^%(__prefix_line)s\s*Accepted (publickey|password) for \S+ from <HOST>
failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|error) for .* from <HOST>( via \S+)?\s*$
```
What do you think?

mc0e:
The problem in https://github.com/fail2ban/fail2ban/pull/1243#issuecomment-153189711 may be more serious than has been been recognised.  A user with a valid logon on one account could use this to reset the counter while brute forcing another account.

sebres:
> A user with a valid logon on one account could use this to reset the counter while brute forcing another account.

Well:
- 0.10th branch operates with failure-id (can be host, user-id or even both) and groups retrieved from failregex will be saved in the ticket, so it can and should precise differentiate between accounts.<br/>I had not yet merged this PR with 0.10 to touch this little bit more (and take a look about all pros and cons);
- in v.0.9 as long as not implemented, we can just avoid to specify `gainregex` for such jails, where cross login may be suggested;

erlkonig:
I think I'd have gone with "succregex" or "goodregex", but that doesn't really matter.  A lot of sysadmins are suffering because the passwd-based pam module's error message can't be squashed even if the user authenticates to LDAP in the same pass.  The problem of wanting to mark a IP as "good" instead of an IP/USER pair is a good point - you'd think you'd need a way to have a <USER> in the regexps or something, but don't forget that the fail/succ lines can be linked by the PID involved.  Take this example:
```
Jan  9 19:46:46 yggdrasil sshd[9763]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=99.58.203.19  user=fuzzhead Jan  9 19:46:46 yggdrasil sshd[9763]: Accepted password for fuzzhead from 99.58.203.19 port 52593 ssh2
```
Sure, we could try to try to track tuples of (99.58.203.19, fuzzhead), but why not look at the linking 9763 instead?  That approach would, I think, eliminate the security issue of using successes to clear the IP address periodically while brute forcing another account.  Go ahead and grumble about PID wrap, of course..

giosh94mhz:
👍
Same problem for SMTP SASL module which use multiple auth method in sequence. E.g. DIGEST-MD5 fails and than LOGIN method works.

durnin:
Hey, is this PR dead? Nobody else finds this useful? Would work great for some type of logs like app logs. For instance I have some variable ip's trying to reach a site each 25m and trying to enter a pass and log in, let's say it's slow brute force but still could harm you. It's clearly a bot. I've been manually banning this ip's but more appear each day with the same behavior. I was thinking I could make em fail on a sign in error and then reset if they succeed later (so that normal users don't get banned for failing a couple of times). I could do this at the application level, but looks like something could be done with fail2ban. Maybe I'm missing something and it could be done already, but I'm not an expert.

sebres:
> Hey, is this PR dead? Nobody else finds this useful?

No.
But the largest problem of this PR is the impossibility to take the username into account.
Some user with a existing logon on exactly one account could misuse this to reset the counter for its IP, while brute-forcing another account.

Such user-recognition was admittedly introduced now in v.10, but:
- [ ] it should be taken into account during the failure reseting (extend this PR with "prohibit reset if username/logon-params on success login does not equal a username/logon-params of the invalid logins")
- [ ] several filters should get this optional tag in many failregex (I mean `<F-USER>...</F-USER>`, for example see https://github.com/fail2ban/fail2ban/blob/0.10/config/filter.d/sshd.conf#L31)
- [ ] user-less failregex, contains some logon-params (e. g. key instead of user), should be extended to get this logon-params as `<F-LOGIN>...</F-LOGIN>`, that should be also considered by the reseting.
- [ ] several filters should get `gainregex` (or howsoever it's called) with same possibility as `failregex` to provide login information (tags `<F-USER>` resp. `<F-LOGIN>`)
- [ ] the conflicts should be resolved (since this PR is a little bit outdated).

> Maybe I'm missing something and it could be done already, but I'm not an expert.

Currently the resolving of many mistakes (wrong recognized failures) is possible only via increasing of `maxretry` resp. decreasing of `findtime` parameters.

durnin:
Thanks for your reply @sebres , I guess I'll workaround my issue in the app layer, but I'll be watching this PR!

sebres:
> I have some variable ip's trying to reach a site each 25m and trying to enter a pass and log in...

@durnin You can try another PR #1460 (that will be soon a part of v.0.10 release). 
In your case just allow increment and set small initial bantime (that grows fast for known bad IPs):
```bash
[jail]
# incremental ban-time:
bantime.increment = true
# factor for the increase (default 1):
bantime.factor = 5
# max ban-time 10 week:
bantime.maxtime = 10w
# default (initial) ban-time (10 seconds):
bantime = 10s
# 3 failures in 3 hours:
maxretry = 3
findtime = 3h
```
If some IP will made 3 failures in 3 hours, it will be banned firstly only for the 10 seconds (but it will be known as bad now).
If they make then further failures, 2nd time they go banned for 25 minutes, corresponds to 
`bantime * math.exp(float(bancount+1)*factor) / math.exp(1*factor)` which corresponds to 
`10 * math.exp(float(2)*5)/math.exp(1*5)` by factor 5.
Third time it will be already 60 hours, etc.

Instead of the factor and exponent-formula you can set own multipliers:
```bash
# for bantime=60 the multipliers are minutes and equal: 1 min, 5 min, 30 min, 1 hour, 5 hour, 12 hour, 1 day, 2 day
bantime.multipliers = 1 5 30 60 300 720 1440 2880
# for bantime=10 the multipliers are one-fifth of minutes (so for the same increment):
bantime.multipliers = 5 25 150 300 1500 3600 7200 14400
```

durnin:
That's a nice solution! I may just wait for the PR to be integrated into v0.10 if you say it's coming out soon. Also, is there a way of unmark an ip as being "bad"? Just curious because it may happen. Thanks @sebres !

sebres:
> Also, is there a way of unmark an ip as being "bad"?

Manually unbanning does this (by removing from sqlite-database).
```bash
# everywhere (in all jails and database, >= v.0.10):
?sudo? fail2ban-client unban <IP> ... <IP> 
# per jail (>= v.0.9)
?sudo? fail2ban-client set <JAIL> unbanip <IP>
```

treck:
As i currently ran into the problem, that users delete accounts on the server, but their clients still try to connect to those resulting in the client getting blocked bei fail2ban i would appreciate more some kind of counterpart counter for successfull logins. With that clients fetching multiple accounts would not getting blocked for one single failing account.

helmo:
+1 for implementing something like this.

I was thinking about using a successregex to relax the settings for such IP's, e.g. increasing the maxretry value dynamically when it matches. That would allow a strict default while giving clients a bit more slack for PEBKAC cases.

I've currently implemented this with an additional custom action that looks up the IP to block and directly unblocks on finding a successful login.
To avoid endless attempts I specifically do not add this to the recidive jail. There I have an action that mails me with a warning which client is being blocked.

shamefulCake1:
>A user with a valid logon on one account could use this to reset the counter while brute forcing another account.

Sorry, I don't think that this issue is an issue that fail2ban should be trying to solve, because this "closes" one attack vector but opens another one. 

Indeed, the user that has an account gains an ability to brute-force other users, true. However, an Evil user that does _not_ have an account gains an ability to _block legitimate users_ with accounts by being in the same subnet and getting a NATted to the same public ip.
I would say that this is way more harmful, because it attains service deterioration for legitimate users without even breaking anywhere. 

And if a user who has an account is bruting other accounts on the same machine, what stops her from brute-forceing over "su user2"?

Not even mentioning that never finding out that you gave an account to a brute-forcer is just ... when the medicine is worse than the illness.

sebres:
> However, an Evil user that does not have an account gains an ability to block legitimate users with accounts by being in the same subnet and getting a NATted to the same public ip.

How resetting the counter may help the evil user to block someone behind the same public IP? It is basically vice versa situation. I guess you don't understand the issue - the resetting of counter may cause that a ban never happens, however and where it should. So the ability to reset counter regardless the user needs to be "deactivated".

> And if a user who has an account is bruting other accounts on the same machine, what stops her from brute-forceing over "su user2"?

To do `su user2` he must be sudoer, in that case the issue is almost to neglect, because the host will have much worse problems. 
The issue with reset counter and switching user is orthogonal and has basically nothing to do with scenarios you're trying to describe.

shamefulCake1:
> How resetting the counter may help the evil user to block someone behind the same public IP? It is basically vice versa situation. I guess you don't understand the issue - the resetting of counter may cause that a ban never happens, however and where it should. So the ability to reset counter regardless the user needs to be "deactivated".

You did not read my comment properly. When this suggestion is NOT implemented, as it RIGHT NOW, an evil person from the same subnet starts brute-forcing my IP, and blocks ME, because we share the same IP.

This happens all the time, because in a big company there is always some idiot who forgets to properly add keys to the ssh server and ends up doing multiple failed attempts with autossh. And due to CGNAT, whole districts are NAT'ed to the same IP.
 
> To do `su user2` he must be sudoer, 

This is absolutely FALSE.

>in that case the issue is almost to neglect, because the host will have much worse problems. The issue with reset counter >and switching user is orthogonal and has basically nothing to do with scenarios you're trying to describe.

But this is exactly the excuse presented in this issue as an insurmountable obstacle. If a user has an account on the same server, he can do 'su user2'. No need to have `sudo`, I have no idea where such a blatant misconception is coming from.

shamefulCake1:
> This feature would not help, because to avoid that with "reset counter" feature, you'd need to connect to the host in the same frequency than evil host - the reset takes place only if a successful connect will happen in between

Ideally it wouldn't be just a "reset" it would be adding that IP to the list of "ips from which good people connect", and would whitelist them for a long time.

>It doesn't work if ban already happen before, because "evil person" is unable to connect successfully. If "evil person" is able to connect successfully, why shall he generate failures (to produce a ban) at all?

In the case of "before", indeed, this wouldn't work, but that means that the legitimate user is visiting the subnet of the evil user, which is, well, not very typical.

> Let alone it looks like a really special case (sharing the same IP with evildoers).

This is not at all special in 2024, with corporate NAT and CGNAT we are sharing IPs with thousands of people.

> > This is absolutely FALSE.
> 
> What is false? They need to be root (know root password or be sudoer, or know the user password)... Anyway it is definitely orthogonal issue and has nothing to do, neither with reset counter, nor with fail2ban as is. The fact that evildoer is able to invoke `su` on some host just opens the door on that host and it is the issue of admin of this host.

You don't need to know the password to brute-force the "su" command. SSH is limited in brute force speed by the tcp connection delay, which is tens of milliseconds, whereas brute-forceing "su" is limited only by the CPU speed.

> > But this is exactly the excuse presented in this issue as an insurmountable obstacle. If a user has an account on the same server, he can do 'su user2'. No need to have sudo
> 
> What you're trying to describe is calling "horizontal privilege escalation" and need to be definitely avoided for any system, because it is simply a vulnerability... In this case fail2ban would also not help to protect against that, sorry. Let alone the issue of horizontal privilege escalation has nothing to do with this PR and fail2ban at all. But again - it needs to be root (or know the password) to be able to invoke `su`. If someone (with evil intentions) can act as another person or even escalate to root on the host, you have completely different problems there. Anyway, if the user is already connected to host, fail2ban mostly bypass anything what user does (one have to create special filter to find the errors during a successful connect - there is no stock filters monitoring active ssh connection) - fail2ban only protects before the connect succeeds and mostly ignores successfully established connections. I guess you simply don't know what you're talking about, so please stop to flood here in this PR (create your own discussion if you want).

I never said anything about filtering successful logins. In 2024 we are sharing (dynamic) IPs with evildoers all the time, so this feature is both necessary and critical. 

You can say "this introduces unnecessary complexity", which is a valid argument. But "one who has an account can reset the counter" is not a valid argument.

benoitg:
> Let alone it looks like a really special case (sharing the same IP with evildoers).

No, I want to emphasise that as @shamefulCake1 this happens really regularly!  Not everyone works remotely in 2024, and all it takes is someone else in the company trying too many time to login unsuccessfully on a machine in a datacenter (possibly on another service on the same machine) to lock everyone else out...

sebres:
> Ideally it wouldn't be just a "reset" it would be adding that IP to the list of "ips from which good people connect"
> You don't need to know the password to brute-force the "su" command. 

Indirectly you contradict yourselves with this 2 sentences, don't you?
But first is something about to ignore (avoid ban), another is to avoid bruteforce by su command. Both have nothing to do with each others and particularly with this PR.

Again:
- this PR is about totally different concerns and the approach (and a restriction) is clearly described.
- the issue with horizontal privilege escalation cannot be avoided with such a reset (vice versa it'd be more evil), basically it is completely orthogonal thing.

A brutforce of su (and other similar attack vectors) expecting different approaches, e. g. new filters, ignorecommand and especially different action - the action shall reject and/or ban a user (not its IP), probably notify admin about a bruteforce possibility, etc.
Please don't put everything in single pot, the resulting vinaigrette will be uneatable.

> all it takes is someone in the company trying too many time to login unsuccessfully on a machine in a datacenter (possibly on another service on the same machine) to lock everyone else out...

Sure, however they are not "evildoers" directly - they are then rather unintentional attempts. But the approaches are totally different - then you have to ban users (or IP/user tuples) and not the IPs... Or probably their originated IPs (the IPs behind the NAT, corporate firewalls, proxy, VPN etc). And the IPs or DNS of VPN/proxy/NAT are mostly whitelisted too (`ignoreip` or `ignorecommand`)... It is the way how the IPS/IDS basically do in corporate networks. 
Something like a reset is not only completely wrong here, it could additionally open much more doors for an intruders.

benoitg:
> Sure, however they are not "evildoers" directly - they are then rather unintentional attempts. 

If these people are likely to exist (and several of us claim to have repeatedly run into them), and likely to _accidentally_ DOS the service for the people most likely to need it (their coworkers), their motivation or competence is a rather moot point.

> But the approaches are totally different - then you have to ban users (or IP/user tuples) and not the IPs... 

Ideally, yes.  But I presume it will take a very long time for the default regexes to support extracting the user information even once this is fully supported.  

Dynamically setting the threshold higher as has been suggested higher in the discussion seems a reasonable compromise for now, and likely still useful down the line for services that for whatever reason do not happen to have the username of login attempts easily accessible.

> Or probably their originated IPs (the IPs behind the NAT, corporate firewalls, proxy, VPN etc). And the IPs or DNS of VPN/proxy/NAT are mostly whitelisted too (`ignoreip` or `ignorecommand`)... It is the way how the IPS/IDS basically do in corporate networks.
> Something like a reset is not only completely wrong here, it could additionally open much more doors for an intruders.

I didn't want to get pulled into the technical discussion because I don't have time to contribute code.  While network security isn't my actual specialty, I actually quite agree with @sebres you that a reset (throwing away the correlation) is not the right solution!  [helmo](https://github.com/helmo)'s solution seems much more sensible.

It won't fix @treck 's case, which while accidental is an actual automated attack (albeit with the same password over and over, but unless every application logs a hash of it, which is unlikely, there is no way for fail2ban to distinguish it from a real brute force attack on a specific user).  But it will practically manage most of the "actual humans trying to remember their password 10 times and blocking access to the service for everyone" problem. 

fail2ban's core assumption is that there is a correlation between an IP and a bad actor's next attempt to remote brute force a credential. 

The practical problems here is that in some contexts (like an office) there is also a very strong correlation between false positives and the people negatively impacted by said false positive.  And in that **same** context there is a negative correlation with the likelihood of bad actors running a deliberate attack from the same IP (it is off course possible, but less likely.  And such an attack isn't distributed so thershold can be higher and still practically block it). 

All that (IMHO) warrants different security-convenience tradeoffs and is worth managing.

What I am arguing it that it's not:

1. "few users" as you put it
2. A practically avoidable problem by network configuration as @shamefulCake1 seem to imply in the quote above.   I mean theoretically it is, but for many it's in practice  very much not.  A lot of service providers do not and cannot control the configuration of the corporate networks which connect to the services they manage (and there may be quite a few even for a single client).

Can we focus on what would be the proper solution in a fail2ban context and how it can be best implemented?   Disabling fail2ban does not improve security either.  

And it is what can happen unless like several people in this discussion you have the time to write a custom solution or propose patches.  And while fail2ban is hardly perfect security it's in practice a lot better than nothing, if only because it reduces noise from untargetted automated attacks.

Can we just find a path forward?

Chealer:
Thanks Thomas, but I agree this is not an ideal solution.

@mc0e wrote:
> A user with a valid logon on one account could use this to reset the counter while brute forcing another account.
> 
> Ideally a successful login should only negate the effect of failures to log in for the same account. Guess that makes it a harder problem.

Your concern is most valid, but I disagree that what you describe is _ideal_. Typos can occur both in the password and in the username. If I managed to connect 10 times as "chealer", erroneous attempts to connect as "chealler" should (in general) not penalize me as much.

Similarly, many people have several accounts on a host. If I regularly connect successfully as _chealer_, the target host should be more permissive if I then try to connect as _testdev4_ mistakenly entering the password for _testdev3_ or _chealer_.

Fail2Ban is heuristic; as Benoit wrote, it's always a matter of tradeoffs. But in general, if address _x_ connected to the target as _chealer_ hundreds of times, it should:
* get a lot more attempts to connect as chealer
* get more attempts to connect as any other account.

Fully resetting an address on success is suboptimal. Ticket #4199 suggests smarter solutions/mitigations for this issue, such as whitelisting.

sebres:
> Your concern is most valid, but I disagree that what you describe is ideal. 

Nothing is ideal here. In the security world by the decision between "ban someone accidentally" and "open a vector to avoid ban" one would almost always opt for the former.

> Ticket #4199 suggests smarter solutions/mitigations for this issue, such as whitelisting.

Sorry, but it does not. It is even not a raw idea.

By the way, please stop add link to your page on every comment (you have a place for that on your profile).

Chealer:
@sebres wrote:
> > Your concern is most valid, but I disagree that what you describe is ideal.
> 
> Nothing is ideal here.

Eh, did you not see my citation? @mc0e wrote (emphasis mine):
>> **Ideal**ly a successful login should only negate the effect of failures to log in for the same account.

> In the security world by the decision between "ban someone accidentally" and "open a vector to avoid ban" one would almost always opt for the former.

There is no "security world"; there is 1 world, where we cannot choose between security or functionality. We always need a balance between both of these ideals.
 
> > Ticket #4199 suggests smarter solutions/mitigations for this issue, such as whitelisting.
> 
> Sorry, but it does not. It is even not a raw idea.

What?

> By the way, please stop add link to your page on every comment (you have a place for that on your profile).

Which page are you talking about?

Issue #1248, opened 2015-11-09
fail2ban-regex stops working on invalid (wrong encoded) character


svennd:
I believe I found a bug in the regex of sshd jail (centos 6.7, fail2ban 0.9.3)

The following user attack breaks the regex
```
[root@main ~]# cat /var/log/secure | grep "llinco"
Nov  8 00:16:12 main sshd[32547]: Invalid user llinco\361ir from 189.20.50.98
Nov  8 00:16:12 main sshd[32548]: input_userauth_request: invalid user llinco\361ir
Nov  8 00:16:12 main sshd[32547]: pam_succeed_if(sshd:auth): error retrieving information about user llinco�ir
Nov  8 00:16:14 main sshd[32547]: Failed password for invalid user llinco\361ir from 189.20.50.98 port 57025 ssh2
```
```
Nov  8 00:16:12 main sshd[32547]: Invalid user llinco\361ir from 189.20.50.98
Nov  8 00:16:12 main sshd[32548]: input_userauth_request: invalid user llinco\361ir
Nov  8 00:16:12 main sshd[32547]: pam_succeed_if(sshd:auth): error retrieving information about user llinco�ir
Nov  8 00:16:14 main sshd[32547]: Failed password for invalid user llinco\361ir from 189.20.50.98 port 57025 ssh2
```

fail2ban-regex output
```
[root@main ~]# fail2ban-regex /var/log/secure /etc/fail2ban/filter.d/sshd.conf 

Running tests
=============

Use   failregex filter file : sshd, basedir: /etc/fail2ban
Use         maxlines : 10
Use         log file : /var/log/secure
Use         encoding : UTF-8

Traceback (most recent call last):
  File "/usr/bin/fail2ban-regex", line 589, in <module>
    fail2banRegex.process(test_lines)
  File "/usr/bin/fail2ban-regex", line 381, in process
    line_datetimestripped, ret = fail2banRegex.testRegex(line)
  File "/usr/bin/fail2ban-regex", line 340, in testRegex
    line, ret = self._filter.processLine(line, date, checkAllRegex=True)
  File "/usr/lib/python2.6/site-packages/fail2ban/server/filter.py", line 417, in processLine
    tupleLine, date, returnRawHost, checkAllRegex)
  File "/usr/lib/python2.6/site-packages/fail2ban/server/filter.py", line 497, in findFailure
    failRegex.search(self.__lineBuffer)
[...]

PR #1249, opened 2015-11-09
fixed fail2ban-regex reads invalid character (in sense of given encoding) ...

sebres:
continuing to process line ignoring invalid characters (still has no test cases);
filter test cases added for same issue inside fail2ban-server / fail2ban-testcases;
closes gh-1248

### DIFF
--- a/bin/fail2ban-regex
+++ b/bin/fail2ban-regex
@@ -84,8 +84,14 @@ def file_lines_gen(hdlr):
 		try:
 			line = line.decode(fail2banRegex.encoding, 'strict')
 		except UnicodeDecodeError:
-			if sys.version_info >= (3,): # Python 3 must be decoded
-				line = line.decode(fail2banRegex.encoding, 'ignore')
+			logSys.warning(
+				"Error decoding line from '%s' with '%s'."
+				" Consider setting logencoding=utf-8 (or another appropriate"
+				" encoding) for this jail. Continuing"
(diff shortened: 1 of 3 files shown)


bin/fail2ban-regex
```diff
@@ -84,8 +84,14 @@ def file_lines_gen(hdlr):
...
 			line = line.decode(fail2banRegex.encoding, 'strict')
 		except UnicodeDecodeError:
-			if sys.version_info >= (3,): # Python 3 must be decoded
-				line = line.decode(fail2banRegex.encoding, 'ignore')
+			logSys.warning(
+				"Error decoding line from '%s' with '%s'."
+				" Consider setting logencoding=utf-8 (or another appropriate"
+				" encoding) for this jail. Continuing"
+				" to process line ignoring invalid characters: %r" %
+				('<LOG>', fail2banRegex.encoding, line))
```
yarikoptic: and `<LOG>` here in the message to signal that line came from some log file?

I wonder why we conditioned it previously only to be ran for python3 -- any idea?

yarikoptic:
passes for all Pythons so I guess it is safe even though conditioning on python3 was removed.  So +1 for merging

bin/fail2ban-regex
sebres: > and \<LOG\> here in the message to signal that line came from some log file?

Just to hold it more similar to filter.py, where this bug was already fixed time ago.

PR #1250, opened 2015-11-10
fail2ban-regex coverage

sebres:
fail2ban-regex moved to the client + test cases for initial coverage added;
small fixes (debuggexURL on wrong encoded chars, etc.)
based on #1249 (because was not covered at all)

Initially `fail2ban-regex` was renamed into `client/fail2banregex.py`, see 0877d662287bb74cb403c0159841dbbafda28c08.

All other changes are relevant to the coverage.

### DIFF
--- a/bin/fail2ban-regex
+++ b/bin/fail2ban-regex
@@ -1,0 +1,34 @@
+#!/usr/bin/python
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: t -*-
+# vi: set ft=python sts=4 ts=4 sw=4 noet :
+#
+# This file is part of Fail2Ban.
+#
+# Fail2Ban is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# Fail2Ban is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Fail2Ban; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Fail2Ban  reads log file that contains password failure report
+and bans the corresponding IP addresses using firewall rules.
+
+This tools can test regular expressions for "fail2ban".
+
+"""
+
+__author__ = "Fail2Ban Developers"
+__copyright__ = "Copyright (c) 2004-2008 Cyril Jaquier, 2012-2014 Yaroslav Halchenko"
+__license__ = "GPL"
+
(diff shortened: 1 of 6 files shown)


bin/fail2ban-regex
```diff
@@ -29,569 +29,6 @@ __author__ = "Fail2Ban Developers"
 __copyright__ = "Copyright (c) 2004-2008 Cyril Jaquier, 2012-2014 Yaroslav Halchenko"
```
yarikoptic: Why it dropped executable permission?

sebres:
BTW: I think, because some files from `bin` are involved, the coveralls calculates invalid whole coverage.
Because as coverage changed I see only 2 files - https://coveralls.io/builds/4103648
And another two, that so far were never covered, are 96% and 100% now (from 0%):
<table>
<tr><th>Coverage</th><th>File</th><th>Lines</th><th>Relevant</th><th>Covered</th><th>Missed</th><th>Hits/Line</th></tr>
<tr><td>96.01</td><td>fail2ban/client/fail2banregex.py	</td><td>597</td><td>326</td><td>313</td><td>55</td><td>1.0</td></tr>
<tr><td>100.0</td><td>fail2ban/tests/fail2banregextestcase.py</td><td>181</td><td>90</td><td>90</td><td>0</td><td>1.0</td></tr>
</table>

bin/fail2ban-regex
sebres: Sure?

yarikoptic: That what it says above
bin/fail2ban-regex 100755 → 100644

sebres: Hm, really.
```
diff --git a/bin/fail2ban-regex b/bin/fail2ban-regex
old mode 100755
new mode 100644
index fdbfcb7..584c1ea
--- a/bin/fail2ban-regex
+++ b/bin/fail2ban-regex
```
I will try to restore it...

sebres: done.

fail2ban/client/fail2banregex.py
```diff
@@ -0,0 +1,599 @@
...
+
+		self.process(test_lines)
+
+		if not self.printStats():
+			return False
+
+		return True
+
+
+def exec_command_line(): # pragma: no cover
```
yarikoptic: Just a note -- now there is no reason why this one can't/shouldn't be tested really.  It will require a bit of mocking (e.g. using mock) but really doable.

I will merge and then remove this no cover statement

yarikoptic:
cool, thanks!

fail2ban/client/fail2banregex.py
sebres: this was just an initial coverage :) I will make further ...
And you don't need remove a "no cover" statement, because we can see it all grayed in coveralls ...

sebres:
Back to coverage: no idea why the coverage still remains under 92% after this PR resp. as #1249 was merged.
In reality - it was increased here (see my https://github.com/fail2ban/fail2ban/pull/1250#issuecomment-155404634 above).
@yarikoptic any idea?

fail2ban/client/fail2banregex.py
yarikoptic: FWIW I did remove those and pushed directly.

PR #1251, opened 2015-11-12
Added PartOf to service file so f2b restarts when deps do

### LINKED ISSUE #1005: Rules in firewallcmd-* filters don't survive firewalld reload/restart
Rules for firewalld don't have `--permanent` option, so they won't survive firewalld reload (eg. via `--reload`) or restart (which happens after updates). This is really annoying, because, for example, in case of firewallcmd-ipset filter firewalld reload/restart (eg. after system update) deletes all references to ipset sets, but fail2ban will silently continue adding new banned IPs to them, even if it has no effect.

I came up with possible solution (at least for firewallcmd-ipset filter) which is to modify `firewallcmd-ipset.conf` Definition section from
```
[Definition]
actionstart = ipset create fail2ban-<name> hash:ip timeout <bantime>
              firewall-cmd --direct --add-rule ipv4 filter <chain> 0 -p <protocol> -m multiport --dports <port> -m set --match-set fail2ban-<name> src -j <blocktype>

actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -p <protocol> -m multiport --dports <port> -m set --match-set fail2ban-<name> src -j <blocktype>
             ipset flush fail2ban-<name>
             ipset destroy fail2ban-<name>
```
to
```
[Definition]
actionstart = ipset create fail2ban-<name> hash:ip timeout <bantime>
              firewall-cmd --direct --add-rule ipv4 filter <chain> 0 -p <protocol> -m multiport --dports <port> -m set --match-set fail2ban-<name> src -j <blocktype>
              firewall-cmd --permanent  --direct --add-rule ipv4 filter <chain> 0 -p <protocol> -m multiport --dports <port> -m set --match-set fail2ban-<name> src -j <blocktype>

actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -p <protocol> -m multiport --dports <port> -m set --match-set fail2ban-<name> src -j <blocktype>
             firewall-cmd --permanent --direct --remove-rule ipv4 filter <chain> 0 -p <protocol> -m multiport -dports <port> -m set --match-set fail2ban-<name> src -j <blocktype>
             ipset flush fail2ban-<name>
             ipset destroy fail2ban-<name>
```
where the second firewall-cmd call with the `--permanent` option will ensure that the rules persist through restarts/reloads. (The first call without `--permanent` has immediate effect, the second one works only after reload/restart).

Another possible solution would be to use only calls with `--permanent` and after fail2ban start/stop just use `firewall-cmd --reload` to load/unload all added rules.

I hope someone will come up with better solutions.

fastest963:
If `iptables` restarts the jails aren't re-created automatically and the rules are not re-added until fail2ban restarts. Adding `PartOf`[1] solves this by restarting `fail2ban` whenever `iptables` is restarted. I added `firewalld` to the `PartOf` command since that might resolve #1005 but I'm not sure since we don't use that.

If there's a better way to achieve this, please let me know.

[1] http://www.freedesktop.org/software/systemd/man/systemd.unit.html#PartOf=

### DIFF
--- a/files/fail2ban.service
+++ b/files/fail2ban.service
@@ -2,6 +2,7 @@
 Description=Fail2Ban Service
 Documentation=man:fail2ban(1)
 After=network.target iptables.service firewalld.service
+PartOf=iptables.service firewalld.service
 
 [Service]
 Type=forking


fastest963:
Any update on this?

sebres:
I believe that should be a thing of distribution maintainers resp. part of setup dependent on platform resp. services availability.
Secondly I can imagine a usage of fail2ban without any firewalls (just jails, that will blocked using software internal mechanisms). In this case it is not so good to create such relationships.
@fail2ban/contributors any ideas?

yarikoptic:
I could sworn that I have posted a comment but apparently doing from phone has failed again.  Although indeed fail2ban could be used without any firewall, most common scenario is with it.  And since we do provide this file, we better indeed have it fixed.  Let's indeed wait for one more day for @fail2ban/developers and contributors to provide feedback on possible cons, and if not -- we will merge.  But we will also need a changelog entry in enhancements I guess

opoplawski:
FYI - this appears to have caused some issues  (systemctl restart firewalld fails), see:
http://thread.gmane.org/gmane.comp.sysutils.systemd.devel/35934

PR #1258, opened 2015-11-22
Add 'Sender address rejected: Domain not found' Postfix failregex

aleksandrs-ledovskis:


yarikoptic:
If no objections, I think it is safe to include this one

PR #1262, opened 2015-11-28
Small fixes for Gentoo initd script

sarneaud:
These fixes are pretty pedantic, but they do simplify the script a
little.

* Checking the existence of a file/directory before creating/deleting
  it adds complexity and raciness.  There are better options.
* mkdir -p does the job of making sure a directory exists.  (It only
  fails if there's a filesystem error or something.)
* Likewise, rm -f doesn't fail if the file doesn't exist.
* rm -r isn't neccessary because the socket shouldn't be a directory.
  (If it is for some reason, that should be an error.)

### DIFF
--- a/files/gentoo-initd
+++ b/files/gentoo-initd
@@ -30,14 +30,10 @@ depend() {
 
 start() {
 	ebegin "Starting fail2ban"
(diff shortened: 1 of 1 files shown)


sebres:
Looks good, merge, Thx.

PR #1264, opened 2015-11-29
Added filter and jail for murmur/mumble-server.

rbrownwsws:


config/filter.d/murmur.conf
```diff
@@ -0,0 +1,21 @@
...
+
+before = common.conf
+
+
+[Definition]
+
+_daemon = murmurd
+
+failregex = Rejected connection from <HOST>:\d+: Invalid server password$
+            Rejected connection from <HOST>:\d+: Wrong certificate or password for existing user$
```
yarikoptic: since I am not familiar with that beast and thus not sure if any "user" input could be logged as in the line, what if we try to anchor it on both ends:
```
-
-failregex = Rejected connection from <HOST>:\d+: Invalid server password$
-            Rejected connection from <HOST>:\d+: Wrong certificate or password for existing user$
+_prefix = <W>[\n\s]*(\.\d{3})?\s+\d* => <\d*:test\(-1\)>\s+Rejected connection from <HOST>:\d+:
+failregex = ^%(_prefix)s Invalid server password$
+            ^%(_prefix)s Wrong certificate or password for existing user$
```
this patch should make it work... a bit obscure but passes the fail2ban-regex testing. `(\.\d{3})?` is there to match possibly left unmatched milliseconds, `[\n\s]` is because time gets stripped but during matching both preceding and following it parts of line merged  with '\n'

yarikoptic: if works for you -- great -- add it in, add a ChangeLog and we will get it merged

rbrownwsws: As a bit of background "murmur" is a server for the "mumble" voice chat software:
 - http://mumble.info
 - https://github.com/mumble-voip/mumble/tree/master/src/murmur

I admit I'm not an expert on murmur but I've had a quick look through through the source code and I cannot see a way for a user to get anything into the log that would trip up my original regex.

The only avenue for abuse I can see would be malicious users that already have access to the server trying to register new users or channels that include "Rejected connection from *\<VICTIM\>* ..." in their name but two things prevent this from actually working:
 - The default validation regexes for user and channel names would not allow you to create a user/channel with a name like that.
 - The log messages always wrap the names in brackets and/or add their ID as a suffix so the failregex being anchored by the end is sufficient to stop it from matching.


On another note the string "test" in my sample logs is actually the username that is attempting to authenticate.  This means that if you think anchoring at both ends is needed "test" will need to be replaced in your regex with either the default username validation regex:
```
[-=\w\[\]\{\}\(\)\@\|\.]+
```
or just allow anything in case the default username validation regex is changed by a server admin:
```
.*
```
to get something like this:

```
 _usernameregex = [-=\w\[\]\{\}\(\)\@\|\.]+

 _prefix = <W>[\n\s]*(\.\d{3})?\s+\d+ => <\d+:%(_usernameregex)s\(-1\)> Rejected connection from <HOST>:\d+:

 failregex = ^%(_prefix)s Invalid server password$
             ^%(_prefix)s Wrong certificate or password for existing user$
```

For my own use I am fine with just anchoring the failregex by the end but if you tell me your preferences I'll make the changes in my fork.
 - **anchoring:** end only / start & end
 - **username matching:** strict / loose

yarikoptic:
Thank you @rbrownwsws for checking all the details!  quick reply -- Let's proceed as follows
* anchroing:  start&end
* username (THANKS for spotting missed by me test):  just use `[^>]+` which I guess would be strict enough ;)

Issue #1265, opened 2015-12-02
Filter performance degrade on high number of log files


kshetragia:
I have web hosting with > 300 domains and every domain has a couple log files which i want to monitor by f2b.

A jail config:
>[apache-badbot]
>enabled  = false
>filter   = apache-badbots
>action   = iptables-multiport[name=BadBots, port="http,https"]
>logpath  = /var/www/vhosts/*/logs/*access_log
>           /var/log/httpd/*access_log
>bantime  = 172800
>maxretry = 1

I'm trying to enable jails and see multiple strings in fail2ban.log:
>2015-12-02 15:25:25,546 fail2ban.filter         [19153]: INFO    Added logfile = /var/www/vhosts/example280.com/logs/proxy_access_log
>2015-12-02 15:25:27,495 fail2ban.filter         [19153]: INFO    Added logfile = /var/www/vhosts/example280.com/logs/access_log

But the enable jails process goes to performance degrade and holds after 580-590 added logfiles.

After the brief research i've found the possible root cause in [server/filter.py](https://github.com/fail2ban/fail2ban/blob/master/fail2ban/server/filter.py) well known as [Shlemiel algorithm](https://en.wikipedia.org/wiki/Joel_Spolsky#Schlemiel_the_Painter.27s_algorithm)

```python
self.__logPath = []

def addLogPath(self, path, tail = False):
        if self.containsLogPath(path):
            ...........
        else:
            ...........
            self.__logPath.append(container)

def containsLogPath(self, path):
        for log in self.__logPath:
            if log.getFileName() == path:
                return True
        return False
```

Please review all f2b source code and fix this. I sure this is not a single place with such algorithm. As far as i see there are follow functions with the same problem:
```python
   def delLogPath(self, path):
   def status(self, flavor="basic"):
   def getFileContainer(self, path):
```

sebres:
Firstly the log files will be added ONCE by starting of jail...

Secondly I see here no problem and the piece of code above will work even up to 5000 logs without great degradation... (see an example below).
Of course degrade that, and of course we can rewrite it as ordered dict or something else. BUT by monitoring of 5000 files you will get another problems as this small few "not pretty" piece of code.
Started by initial searching of failure inside this couple of log file, including monitoring of all this files in single thread. 
And if you use such backends as `polling` (instead of for example `pyinotify`) you lose twice and triple.

Also - no, "the possible root" of the issue is in monitoring of really large files inside a single thread (jail)!

[...]

PR #1266, opened 2015-12-02
small improvement for better handling of many log files (gh-1265)

sebres:
the FileFilter contains now a (ordered) dict of the log files (instead of list), as discussed in gh-1265

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -59,8 +59,9 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      rest api and web interface (gh-1223)
    * Add *_backend options for services to allow distros to set the default
      backend per service, set default to systemd for Fedora as appropriate
-   * small improvement for better handling of many log files (gh-1265)
-     Thanks @kshetragia
+   * Performance improvements while monitoring large number of files (gh-1265).
+     Use associative array (dict) for monitored log files to speed up lookup 
+     operations. Thanks @kshetragia
 
 ver. 0.9.3 (2015/08/01) - lets-all-stay-friends
 ----------
--- a/ChangeLog
+++ b/ChangeLog
@@ -59,6 +59,8 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      rest api and web interface (gh-1223)
    * Add *_backend options for services to allow distros to set the default
      backend per service, set default to systemd for Fedora as appropriate
+   * small improvement for better handling of many log files (gh-1265)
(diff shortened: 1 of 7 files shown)


yarikoptic:
Thanks... Let me just do some benchmarking when I get to the laptop... Still not clear to me if there would be any benefit from ordered dict

sebres:
> benchmarking

Don't forget that a writing resp. modifying of this list (add/del) will occur comparatively rarer as reading (get/iter).

fail2ban/server/filter.py
```diff
@@ -38,6 +38,11 @@
 from .action import CommandAction
 from ..helpers import getLogger
 
+try:
+	from collections import OrderedDict
+except ImportError:
+	OrderedDict = dict
+
```
yarikoptic: http://nbviewer.ipython.org/url/www.onerussian.com/tmp/dics-vs-ordereddict.ipynb
so why should we use slower OrderedDict?

ChangeLog
```diff
@@ -59,6 +59,8 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      rest api and web interface (gh-1223)
    * Add *_backend options for services to allow distros to set the default
      backend per service, set default to systemd for Fedora as appropriate
+   * small improvement for better handling of many log files (gh-1265)
+     Thanks @kshetragia
```
yarikoptic: gr... I guess github didn't post my comment from the phone :-(
would be better IMHO to provide a bit more specific than "small improvement" description, e.g.

"Use associative array (dict) for monitored log files to speed up lookup operations. Provides performance improvements while monitoring large number of files ..."

kshetragia:
The issue #1265 was already closed so i'll write here.
Thank you for the fast answer and provided solution.

 I agree with you these issues in filter.py are not a root cause of problem and i mentioned ones as POSSIBLE root cause. But problem is real. I've tried to split f2b jail to four ones and pyinotify was used, but f2b performance still is not enough for big amount of log files.

I try to enable these four jails and f2b client process do not stuck anymore but --enable-jails is still very very slow and takes about 10 minutes.

Please investigate further. I can provide all additional information.

fail2ban/server/filter.py
yarikoptic: so agreed that no OrderedDict here? ;)

sebres:
all wishes are done, test cases passed, so merge now...

yarikoptic:
Awesome! Thanks!

PR #1272, opened 2015-12-14
ENH+BF: Provide <fail2ban-agent-string> tag

### LINKED ISSUE #1271: In action "blocklist_de.conf" is hardcoded user-agent "fail2ban v0.8.12"
Version 0.9.2 and current dev code

In file
/etc/fail2ban/action.d/blocklist_de.conf
is the line
```
actionban = curl --fail --data-urlencode 'server=<email>' --data 'apikey=<apikey>' --data 'service=<service>' --data 'ip=<ip>' --data-urlencode 'logs=<matches>' --data 'format=text' --user-agent "fail2ban v0.8.12" "https://www.blocklist.de/en/httpreports.html"
```
The user-agent setting is wrong.
I would suggest to determine the fail2ban version on runtime and use that value instead of hardcoded version, since, like in this case, it will probably be wrong in the future.

yarikoptic:
Was crafted to resolve #1271 but actually not sure now if is worth this approach over "let's execute fail2ban-client version" and define that variable for string interpolations in e.g. actions/common.conf

### DIFF
--- a/fail2ban/version.py
+++ b/fail2ban/version.py
@@ -25,3 +25,5 @@ __copyright__ = "Copyright (c) 2004 Cyril Jaquier, 2005-2015 Yaroslav Halchenko,
 __license__ = "GPL-v2+"
 
 version = "0.9.3.dev"
+
+agent_string = "Fail2Ban %s" % version
--- a/fail2ban/version.py
+++ b/fail2ban/version.py
@@ -26,4 +26,5 @@ __license__ = "GPL-v2+"
 
 version = "0.9.3.dev"
 
-agent_string = "Fail2Ban %s" % version
+# Format https://tools.ietf.org/html/rfc7231#section-5.5.3
+agent_string = "Fail2Ban/%s" % version
--- a/config/action.d/badips.conf
+++ b/config/action.d/badips.conf
@@ -10,7 +10,7 @@
 
 [Definition]
 
-actionban = curl --fail  --user-agent "fail2ban v0.8.12" http://www.badips.com/add/<category>/<ip>
+actionban = curl --fail  --user-agent "<fail2ban-agent-string>" http://www.badips.com/add/<category>/<ip>
 
 [Init]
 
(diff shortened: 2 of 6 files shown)


fail2ban/version.py
```diff
@@ -25,3 +25,5 @@
 __license__ = "GPL-v2+"
 
 version = "0.9.3.dev"
+
+agent_string = "Fail2Ban %s" % version
```
ranvis: Perhaps we can use "Fail2Ban/%s" to confirm [RFC 7231 5.5.3](https://tools.ietf.org/html/rfc7231#section-5.5.3)?

yarikoptic: On Fri, 18 Dec 2015, SATŌ Kentarō wrote:

> In [1]fail2ban/version.py:

> > @@ -25,3 +25,5 @@
> >  __license__ = "GPL-v2+"

> >  version = "0.9.3.dev"
> > +
> > +agent_string = "Fail2Ban %s" % version

> Perhaps we can use "Fail2Ban/%s" to confirm [2]RFC 7231 5.5.3?
Great idea ;)   pushed update

yarikoptic:
@fail2ban/developers , @fail2ban/contributors -- any feedback?

sebres:
I don't like it - too many changes (and conflicts :)) for one little thing (a new parameter used in action), that can be also easy added somewhere in configreader (ActionReader or JailReader)...
Will try to check it now...

yarikoptic:
I don't quite like it as well, but not for "too many changes" of which I see not that many ;-)  Will wait then for your version

sebres:
You are welcome: https://github.com/fail2ban/fail2ban/compare/master...sebres:bf-fail2ban-version

The whole changes are in  `fail2ban/client/jailreader.py` (and its usage in configs).
But because we had at all no test cases for `action_blocklist_de` and `action_badips` the test case grows a little bit...

Should I make a PR?

sebres:
Ah yeah, by the way, badips definition in jail.conf had an interpolation bug, because `%(name)s` not available:
```dif
-action_badips = badips.py[category="%(name)s", banaction="%(banaction)s"]
+action_badips = badips.py[category="%(__name__)s", banaction="%(banaction)s", agent="%(fail2ban_agent)s"]
```

yarikoptic:
PR would be much better for testing and review. Cheers!

PR #1273, opened 2015-12-14
ENH: log at heavydebug level what actually we are matching for failregex

yarikoptic:
Just found it useful while troubleshooting using fail2ban-regex


yarikoptic:
ok -- no objections, merging

PR #1286, opened 2015-12-30
ENH: harmonize jail.conf + 1 more test that passed bantime is non-degenerate and int

yarikoptic:
- there is no need for maxretry ~= 5 in each jail
- separate jails for different services with 2 empty lines
- no need for explicit filter specification if matches jail name
- renamed jail introduced in #1232 to match filter and thus remove explicit filter name

also had worries that bantime with ';' comment might be screwed in Python 3 so added a unittest for that  (didn't run locally -- let travis test it)

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -240,7 +240,6 @@ backend  = %(dropbear_backend)s
 
 port     = ssh
 logpath  = %(auditd_log)s
-maxretry = 5
 
 
 #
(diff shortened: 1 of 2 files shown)


yarikoptic:
@fail2ban/developers @fail2ban/contributors -- no objections?

sebres:
My bantime is already long time not more integer (you know - 60m, 1d, etc.)... :smile: 
But OK.

yarikoptic:
@sebres yeah -- that is a lovely feature we haven't merged yet since it was never "walking alone".   I would happy to get it in if it was possible to review it alone

yarikoptic:
merging with @sebres blessing ;-)

szepeviktor:
OK

PR #1287, opened 2015-12-30
ENH+BF: Provide `fail2ban_agent` and `fail2ban_version` config options

### LINKED ISSUE #1271: In action "blocklist_de.conf" is hardcoded user-agent "fail2ban v0.8.12"

sebres:
Provides fail2ban version to jail (as interpolation variable during parse of jail.conf);
BF: use `fail2ban_agent` as user-agent in actions badips, blocklist_de, etc. (closes #1271, closes #1272)

### DIFF
--- a/config/action.d/badips.conf
+++ b/config/action.d/badips.conf
@@ -10,7 +10,7 @@
 
 [Definition]
 
-actionban = curl --fail  --user-agent "fail2ban v0.8.12" http://www.badips.com/add/<category>/<ip>
+actionban = curl --fail  --user-agent "<agent>" http://www.badips.com/add/<category>/<ip>
 
 [Init]
 
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -32,7 +32,6 @@ else:
 	from urllib import urlencode
 
 from fail2ban.server.actions import ActionBase
-from fail2ban.version import version as f2bVersion
 
 
 class BadIPsAction(ActionBase):
@@ -86,10 +85,10 @@ class BadIPsAction(ActionBase):
 		return Request(url, headers={'User-Agent': self.agent}, **argv)
 
 	def __init__(self, jail, name, category, score=3, age="24h", key=None,
-		banaction=None, bancategory=None, bankey=None, updateperiod=900, agent=None):
+		banaction=None, bancategory=None, bankey=None, updateperiod=900, agent="Fail2Ban"):
 		super(BadIPsAction, self).__init__(jail, name)
 
-		self.agent = agent if agent is not None else ("Fail2Ban/%s" % f2bVersion)
+		self.agent = agent
 		self.category = category
 		self.score = score
 		self.age = age
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -21,7 +21,6 @@ import sys
 if sys.version_info < (2, 7):
 	raise ImportError("badips.py action requires Python >= 2.7")
 import json
-from functools import partial
 import threading
 import logging
 if sys.version_info >= (3, ):
@@ -72,6 +71,9 @@ class BadIPsAction(ActionBase):
 	updateperiod : int, optional
 		Time in seconds between updating bad IPs blacklist.
 		Default 900 (15 minutes)
+	agent : str, optional
+		User agent transmitted to server.
+		Default `Fail2Ban/ver.`
 
 	Raises
 	------
@@ -80,13 +82,14 @@ class BadIPsAction(ActionBase):
 	"""
 
 	_badips = "http://www.badips.com"
-	_Request = partial(
-		Request, headers={'User-Agent': "Fail2Ban %s" % f2bVersion})
+	def _Request(self, url, **argv):
+		return Request(url, headers={'User-Agent': self.agent}, **argv)
 
 	def __init__(self, jail, name, category, score=3, age="24h", key=None,
-		banaction=None, bancategory=None, bankey=None, updateperiod=900):
+		banaction=None, bancategory=None, bankey=None, updateperiod=900, agent=None):
 		super(BadIPsAction, self).__init__(jail, name)
 
+		self.agent = agent if agent is not None else ("Fail2Ban/%s" % f2bVersion)
 		self.category = category
 		self.score = score
 		self.age = age
--- a/config/action.d/blocklist_de.conf
+++ b/config/action.d/blocklist_de.conf
@@ -54,7 +54,7 @@ actioncheck =
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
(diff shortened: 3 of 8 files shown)


fail2ban/client/jailreader.py
```diff
@@ -108,6 +109,10 @@ def getOptions(self):
 				["string", "filter", ""],
 				["string", "action", ""]]
 
+		# Before interpolation (subtitution) add static options always available as default:
```
yarikoptic: Minor typo I think: substitution

config/action.d/blocklist_de.conf
```diff
@@ -54,7 +54,7 @@ actioncheck =
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-actionban = curl --fail --data-urlencode 'server=<email>' --data 'apikey=<apikey>' --data 'service=<service>' --data 'ip=<ip>' --data-urlencode 'logs=<matches>' --data 'format=text' --user-agent "fail2ban v0.8.12" "https://www.blocklist.de/en/httpreports.html"
+actionban = curl --fail --data-urlencode 'server=<email>' --data 'apikey=<apikey>' --data 'service=<service>' --data 'ip=<ip>' --data-urlencode 'logs=<matches>' --data 'format=text' --user-agent "<agent>" "https://www.blocklist.de/en/httpreports.html"
```
yarikoptic: There were other actions iirc as well

yarikoptic:
Besides comments let's go with your version since less of runtime impact

fail2ban/client/jailreader.py
sebres: done

config/action.d/blocklist_de.conf
sebres: already found 2 other, will amend now

sebres:
Amend done, besides blocklist_de, we have 2 badips: one reporting only (badips.conf), other reporting and banning - badips.py (I don't know which would be still runnable / better, but in jail.conf will be offered both now), action mynetwatchman and test cases extended also. 

May be added later to other action, used version resp. user agent, without necessarily extending of the test cases (we want don't overdo =).

sebres:
We have no test cases for badips.py corresponding resp. together with jail.conf parameters, so I found currently a small bug - agent will be transmitted as parameter to BadIPsAction from jail.conf now, so added as optional parameter + small code review in this respect (correct default user-agent with `/` instead of space)

config/action.d/badips.py
```diff
@@ -80,13 +82,14 @@ class BadIPsAction(ActionBase):
...
-		Request, headers={'User-Agent': "Fail2Ban %s" % f2bVersion})
+	def _Request(self, url, **argv):
+		return Request(url, headers={'User-Agent': self.agent}, **argv)
 
 	def __init__(self, jail, name, category, score=3, age="24h", key=None,
-		banaction=None, bancategory=None, bankey=None, updateperiod=900):
+		banaction=None, bancategory=None, bankey=None, updateperiod=900, agent=None):
 		super(BadIPsAction, self).__init__(jail, name)
 
+		self.agent = agent if agent is not None else ("Fail2Ban/%s" % f2bVersion)
```
yarikoptic: Which reveals the duplication... Why not to define fail2ban_agent along with the version, to be also reused here?

config/action.d/badips.conf
```diff
@@ -10,7 +10,7 @@
 
 [Definition]
 
-actionban = curl --fail  --user-agent "fail2ban v0.8.12" http://www.badips.com/add/<category>/<ip>
+actionban = curl --fail  --user-agent "<agent>" http://www.badips.com/add/<category>/<ip>
```
yarikoptic: Recent changes below reminded me that for any if those action parameters we are providing defaults in the same file.

config/action.d/badips.py
```diff
@@ -80,13 +82,14 @@ class BadIPsAction(ActionBase):
 	"""
 
 	_badips = "http://www.badips.com"
-	_Request = partial(
-		Request, headers={'User-Agent': "Fail2Ban %s" % f2bVersion})
+	def _Request(self, url, **argv):
+		return Request(url, headers={'User-Agent': self.agent}, **argv)
```
yarikoptic: Could you elaborate a bit more why partial want sufficient? 
Also if indentation is right (of just effect of tabs?) this def is just within __init__

config/action.d/badips.conf
yarikoptic: given two notes above (duplication and using action options needing defaults).  May be we should just add those (`fail2ban_version/_agent`) as interpolation defaults in our config reader, document that in jail.conf, and use them here while defining alongside in version (as I did iirc) for reuse in the code?

config/action.d/badips.py
sebres: I tend to completely remove setting version in badips.py, because now will be set in new jail.conf.

config/action.d/badips.conf
sebres: Well I don't like to use agent additionally to version in config reader:
- version is sufficient enough (agent can be and will be set in jail.conf);
- needed in some jails only (that will be not used from all the people);
- actions configs will not acquire this setting direct (added in jailreader currently), so it should be pass as parameter to corresponding action, and adding it in all readers will over-complicate a little bit, imho.

config/action.d/badips.py
sebres: Well, I saw that was used quasi as lambda (older py-versions?), so I've replaced it with def (to clarify the code and to remove an unnecessary module import);
What should be wrong with indentation? I see nothing here... Both (new and old version) set the member `_Request` in scope of class.

yarikoptic: sorry -- got confused by the positioning of `__doc__` here so diff looked like the body of `__init__` to me (despite it below ;) )

config/action.d/badips.conf
yarikoptic: Still not convinced but ok -- let it be this way.  Add a Changelog entry and I will merge

sebres:
merged (reset point 69aa1feac087bafae266949ca3111f827f8ec837, if not desired)

PR #1288, opened 2015-12-31
Added HAProxy HTTP Auth filter

jmoeser:
Added a filter for matching HTTP 401 errors in logs produced by HAProxy when it's managing HTTP Authentication.


config/filter.d/haproxy-http-auth.conf
```diff
@@ -0,0 +1,37 @@
...
+_daemon = haproxy
+
+# Option:  failregex
+# Notes.:  regex to match the password failures messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
+#          be used for standard IP/hostname matching and is only an alias for
+#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
+# Values:  TEXT
+#
+failregex = ^%(__prefix_line)s<HOST>.*NOSRV.*401
```
yarikoptic: if I got it right -- this would result in a ban if URLs for some reason match NOSRV.*401 so self-DoS... not sure if we want that upon our users

So I guess regex should be tuned up so that path to NOSRV and 401 lacks .* thus its gets more stringently anchored upfront.

jmoeser: The URLs appear towards the end of the line in the log, after the 401. Like in this sample line, `GET /`

`Nov 14 22:45:11 test haproxy[760]: 192.168.33.1:58430 [14/Nov/2015:22:45:11.608] main main/<NOSRV> -1/-1/-1/-1/0 401 248 - - PR-- 0/0/0/0/0 0/0 "GET / HTTP/1.1"`

If a user named their backend server NOSRV it's possible this regex would mistakenly match it.

I could more stringently match the area between NOSRV and the 401 (and more strictly match the NOSRV by including the <> chevrons as well). This would change the regex to this:

`^%(__prefix_line)s<HOST>.*<NOSRV> -1/-1/-1/-1/\+*\d* 401`

As discussed in the [documentation](https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#8.2.3), the Tq, Tw, Tc and Tr values should always be -1 if the connection is interrupted (for instance by authentication failure) before being passed to a backend. The Tt value however will be variable as it counts the total session duration. The Tt value can also be prepended with a + sign if the option logasap is used. 

I'll update the PR with this regex if you agree.

yarikoptic: >    I'll update the PR with this regex if you agree.

I think this one might work ;)

thanks

yarikoptic:
on its way as #1310

Issue #1118, opened 2015-07-16
Add support for nftables


mrueg:
It would be great if fail2ban could add support for nftables. [1]
Maybe it is even possible to replace all iptables calls with a corresponding nftables call making all actions switchable between both implementations?

[1] http://netfilter.org/projects/nftables/

leeclemens:
Interesting, just glossed over the main page...do you have any info about what distros ship it? or replace {ip,ip6}tables with it? Seems to be geared for backwards compatibility, should fail2ban even care which underlying libs/infrastructure are used?

mrueg:
I don't know which distributions ship it by default, but it's available for many distributions:
https://packages.debian.org/search?keywords=nftables
https://admin.fedoraproject.org/pkgdb/package/nftables/
https://wiki.gentoo.org/wiki/Nftables
https://wiki.archlinux.org/index.php/Nftables

pfactum:
+1 for the feature request.

I've tried to adapt existing iptables fail2ban configs, but the obstacle is that nftables doesn't (currently, as they state in docs) allow removing specific rule using rule definition (only using handle number). I guess I may workaround it with some script that converts rule into handle, but that is cumbersome.

am1github:
@pfactum 
Any news?

I try to port iptables-allports.conf to nftables.conf in Gentoo.

actionban/actioncheck seems working fine, but actionunban needs to remove rules over handles - that a kind of difficult (workaround)

pfactum:
@ambiente1
No news, I've ended up with writing script that parses handles, but that is not the proper solution, so I've given up.

ambiente1:
@pfactum 
Okay, looking into the documentation of nftables, they wrote:
>### Note: There are plans to support rule deletion by passing:
>% nft delete rule filter output ip saddr 192.168.1.1 counter
>
>**but this is not yet implemented. So you'll have to use the handle to delete rules until that feature is implemented.**

[Nftables WIKI - Simple rule management - Removing rules](http://wiki.nftables.org/wiki-nftables/index.php/Simple_rule_management#Removing_rules)

ambiente1:
Ok tested many things on Gentoo using nftables and find out a working solution for me. :+1:

```
# Fail2Ban configuration file
#
# Block IPs over nftables chain
#
# Author: Alexander Meinhardt <hello@ambiente.one>
#
#

[Definition]

# Option:  actionstart
# Notes.:  command executed once at the start of Fail2Ban.
[...]

PR #1292, opened 2016-01-03
Add nftables actions

albel727:
This aims to fulfill #1118 by adding `nftables-allports` and `nftables-multiport` actions, with `nftables-common.conf` containing some common options. 

Those are far from ideal, but should be a good start. Among the deficiencies:

1. Since in nftables creation of a separate chain isn't required for one-rule-bans-set-of-ips functionality, `returntype` variable is not supported. In this regard proposed actions are more like `iptables-ipset*.conf`, than `iptables-multiport/allports`. The consequence is that one can't turn them into "whitelist" actions, like one can with iptables ones.

2. Syntax for port ranges and chain names in nftables are different from iptables. Namely, port ranges are using dash, and chain names are normally lowercase.

3. Since nftables allows to delete rules only by rule handles, as opposed to iptables way of deleting by repeated rule specification with -D option, proposed actions rely on grep for extracting numeric handles from nftables output.

4. In nftables, tables are initially empty and don't even contain an input chain. Proposed actions blindly rely on existence of `input` chain in `filter` table in `ip` family, or existence of any other chain that is specified to be used via configuration. I think it's beyond the scope of actions to create an input chain with desired priority options and whatnot. This is a job of administrator.

5. Protocol "all" is not supported, only "tcp" and "udp". Should be relatively easily fixable with shell conditionals for `iptables-allports`, but I'm lazy.

Unfortunately `1.` and `2.` prevent nftables actions from being a drop-in replacement for iptables ones, unless more option processing is introduced.

One can use the actions by putting following lines into `jail.local`:
```bash
# chain variable needs to be overridden in jail.local,
# as the uppercase `chain = INPUT` declaration in jail.conf
# shadows proper lowercase declaration in nftables-common.conf
chain = input

[sshd]
banaction = nftables-multiport
enabled = true

[recidive]
banaction = nftables-allports
enabled = true
```

### DIFF
--- a/config/action.d/nftables-common.conf
+++ b/config/action.d/nftables-common.conf
@@ -22,13 +22,13 @@ after = nftables-common.local
 # Notes.:  additional expressions for nftables filter rule
 # Values:  nftables expressions
 #
-nftables_mode = <protocol> dport { <port> }
+nftables_mode = <protocol> dport \{ <port> \}
 
 # Option:  actionstart
 # Notes.:  command executed once at the start of Fail2Ban.
 # Values:  CMD
 #
-actionstart = <nftables> add set filter f2b-<name> { type ipv4_addr\; }
+actionstart = <nftables> add set filter f2b-<name> \{ type ipv4_addr\; \}
               <nftables> insert rule filter <chain> %(nftables_mode)s ip saddr @f2b-<name> <blocktype>
 
 # Option:  actionstop
@@ -50,7 +50,7 @@ actioncheck = <nftables> list chain filter <chain> | grep -q '@f2b-<name>[ \t]'
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-actionban = <nftables> add element filter f2b-<name> { <ip> }
+actionban = <nftables> add element filter f2b-<name> \{ <ip> \}
 
 # Option:  actionunban
 # Notes.:  command executed when unbanning an IP. Take care that the
@@ -58,7 +58,7 @@ actionban = <nftables> add element filter f2b-<name> { <ip> }
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-actionunban = <nftables> delete element filter f2b-<name> { <ip> }
+actionunban = <nftables> delete element filter f2b-<name> \{ <ip> \}
 
 [Init]
 
--- a/config/action.d/nftables-common.conf
+++ b/config/action.d/nftables-common.conf
@@ -1,0 +1,50 @@
+# Fail2Ban configuration file
+#
+# Author: Daniel Black
+# Modified: Alexander Belykh <albel727@ngs.ru>
+#                       adapted for nftables
+#
+# This is a included configuration file and includes the definitions for the nftables
+# used in all nftables based actions by default.
+#
+# The user can override the defaults in nftables-common.local
+
+[INCLUDES]
+
+after = nftables-common.local
+
+[Init]
+
+# Option:  chain
+# Notes    specifies the nftables chain to which the Fail2Ban rules should be
+#          added
+# Values:  STRING  Default: input
+chain = input
+
+# Default name of the filtering set
+#
+name = default
+
+# Option:  port
+# Notes.:  specifies port to monitor
+# Values:  [ NUM | STRING ]  Default:
+#
+port = ssh
+
+# Option:  protocol
+# Notes.:  internally used by config reader for interpolations.
+# Values:  [ tcp | udp ] Default: tcp
+#
+protocol = tcp
+
+# Option:  blocktype
+# Note:    This is what the action does with rules. This can be any jump target
+#          as per the nftables man page (section 8). Common values are drop
+#          reject, reject with icmp type host-unreachable
+# Values:  STRING
+blocktype = reject
+
+# Option:  nftables
+# Notes.:  Actual command to be executed, including common to all calls options
+# Values:  STRING
+nftables = nft
--- a/config/action.d/nftables-common.conf
+++ b/config/action.d/nftables-common.conf
@@ -31,11 +31,15 @@ nftables_mode = <protocol> dport \{ <port> \}
 actionstart = <nftables> add set <nftables_family> <nftables_table> f2b-<name> \{ type <nftables_type>\; \}
               <nftables> insert rule <nftables_family> <nftables_table> <chain> %(nftables_mode)s ip saddr @f2b-<name> <blocktype>
 
+_nft_list = <nftables> --handle --numeric list chain <nftables_family> <nftables_table> <chain>
+_nft_get_handle_id = grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'
+
 # Option:  actionstop
 # Notes.:  command executed once at the end of Fail2Ban
 # Values:  CMD
 #
-actionstop = HANDLE_ID=$(<nftables> --handle --numeric list chain <nftables_family> <nftables_table> <chain> | grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'); <nftables> delete rule <nftables_family> <nftables_table> <chain> $HANDLE_ID
+actionstop = HANDLE_ID=$(%(_nft_list)s | %(_nft_get_handle_id)s)
+             <nftables> delete rule <nftables_family> <nftables_table> <chain> $HANDLE_ID
              <nftables> delete set <nftables_family> <nftables_table> f2b-<name>
 
 # Option:  actioncheck
--- a/config/action.d/nftables-common.conf
+++ b/config/action.d/nftables-common.conf
@@ -37,6 +37,7 @@ actionstart = <nftables> add set filter f2b-<name> { type ipv4_addr\; }
 #
 actionstop = HANDLE_ID=$(<nftables> --handle --numeric list chain filter <chain> | grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'); <nftables> delete rule filter <chain> $HANDLE_ID
              <nftables> delete set filter f2b-<name>
+
 # Option:  actioncheck
 # Notes.:  command executed once before each actionban command
 # Values:  CMD
--- a/config/action.d/nftables-common.conf
+++ b/config/action.d/nftables-common.conf
@@ -1,6 +1,9 @@
 # Fail2Ban configuration file
 #
 # Author: Daniel Black
+# Author: Cyril Jaquier
+# Modified: Yaroslav O. Halchenko <debian@onerussian.com>
+# 			made active on all ports from original iptables.conf
 # Modified: Alexander Belykh <albel727@ngs.ru>
 #                       adapted for nftables
 #
@@ -12,6 +15,49 @@
 [INCLUDES]
 
 after = nftables-common.local
+
+[Definition]
+
+# Option:  nftables_mode
+# Notes.:  additional expressions for nftables filter rule
+# Values:  nftables expressions
+#
+nftables_mode = <protocol> dport { <port> }
+
+# Option:  actionstart
+# Notes.:  command executed once at the start of Fail2Ban.
+# Values:  CMD
+#
+actionstart = <nftables> add set filter f2b-<name> { type ipv4_addr\; }
+              <nftables> insert rule filter <chain> %(nftables_mode)s ip saddr @f2b-<name> <blocktype>
+
+# Option:  actionstop
+# Notes.:  command executed once at the end of Fail2Ban
+# Values:  CMD
+#
+actionstop = HANDLE_ID=$(<nftables> --handle --numeric list chain filter <chain> | grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'); <nftables> delete rule filter <chain> $HANDLE_ID
+             <nftables> delete set filter f2b-<name>
+# Option:  actioncheck
+# Notes.:  command executed once before each actionban command
+# Values:  CMD
+#
+actioncheck = <nftables> list chain filter <chain> | grep -q '@f2b-<name>[ \t]'
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionban = <nftables> add element filter f2b-<name> { <ip> }
+
+# Option:  actionunban
+# Notes.:  command executed when unbanning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionunban = <nftables> delete element filter f2b-<name> { <ip> }
 
 [Init]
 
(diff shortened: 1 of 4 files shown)


sebres:
Looks good, but several comments with conspicuity:

- all parameters that are equal (like `actionban`, `actionunban`, etc) can be placed in `[Definition]` (or `[Default]`) block of common include `config/action.d/nftables-common.conf`;
- If the whole difference between files `nftables-allports` and `nftables-multiport` is `dport { <port> }` only, then all parameters can be moved into common file, and this difference specified as substitution, for example as something like `nftables_mode`, see example below...
```
## in nftables-multiport.conf:
nftables_mode = dport { <port> }

## in nftables-common.conf:
actionstart = <nftables> add set filter f2b-<name> { type ipv4_addr\; }
              <nftables> insert rule filter <chain> <protocol> %(nftables_mode)s ip saddr @f2b-<name> <blocktype>
```
- don't forget changelog entry (and thanks :);

albel727:
@sebres How about that? Good thing there are python interpolations, though I'm quite unsure how `%()s` and `<>` substitutions differ, and what's exact difference between `[Definition]` and `[Init]` sections. 

If there were conditional operators or something, I could've emulated `protocol = "all"` functionality for `nftables-allports` action by setting `nftables_mode` to empty string. Maybe even automatically lowercase table names and replaced `:` to `-` in `port` to support iptables syntax. Wonder if actions are templated with jinja2 and if I could manage something with it.

sebres:
> How about that?

Looks much better... imho, but can be optimized further =) see below...

> how %()s and <> substitutions differ

- `%(par)s` - an interpolation parameter defined somewhere in includes resp. in the same file for the same section or within default section.
- `<par>` - either an interpolation runtime parameter (substituted through execution not by read config, like <ip>, <HOST>, etc.) or pre-subtituted parameter defined within `[Init]` section resp. as parameter given for action/filter direct within `jail.conf/jail.local` (replaced by jailreader):
```
action = someaction[par="test"]
```

So for example, in our case (nftables) we can even define all this as single file (nftables.conf) without common, just using `<nf-mode>` as parameter substituation and defining `nf-mode = <protocol> dport { <port> }` in `[Init]` section of `nftables.conf` as default for multiport. Then we can do something like following code, for example defined in default of jail.conf/jail.local to automatically switching to nftables for all (enabled) jails: 
```
banaction = nftables
banaction_allports = nftables[nf-mode="ip protocol <protocol>"]
```

albel727:
@sebres 
I see. Well, I thought about changing mode via action parameter and leaving only one file, but I'm not sure people would immediately understand what to pass and that it can be passed at all. After all, `iptables-multiport` too differs from `iptables-allports` only by mere `-m multiport --dports <port>` addition to the filtering rule, but we still see it as two different files + common.conf. They didn't even extract all the common action stuff, like we did. I guess, it's either due to legacy or for convenience, or both. I'm also afraid that making nftables action a single file, it becomes even less of a drop-in replacement for iptables.

I'm also not sure if we can rely on recursive replacement of runtime parameter <protocol> inside nf-mode in `nftables[nf-mode="ip protocol <protocol>"]`. Will the substitution happen, and if yes, from what scope <protocol> will be picked, and whether it will depend on the order of assignment, e.g.
`nftables[nf-mode="ip protocol <protocol>", protocol="udp"]` differ from
`nftables[protocol="udp", nf-mode="ip protocol <protocol>"]`.

Having tested it, I found out that neither works. `<protocol>` has to be set on a separate line before calling the action in order to be reflected in it. It also appears that if a parameter is passed, then other params, like "name", have to be passed explicitly too, or e.g. all jails will use "f2b-default" name. I therefore don't think it would be quite usable.

Finally I wonder if we also can do something about the necessity to explicitly override `chain = input` in jail.local. Maybe we should just rename the parameter to something like nftables_chain, so that it won't conflict with iptables and `chain = INPUT` declaration in jail.conf?

Here is the attempt to make single nftables.conf, in case I'm missing something. [nftables.txt](https://github.com/fail2ban/fail2ban/files/78946/nftables.txt)

sebres:
> `iptables-multiport` too differs from `iptables-allports` only by mere `-m multiport --dports <port>` addition to the filtering rule, but ...

Yes, but this actions were written as our config reader was still not so powerful as it is today.

> Will the substitution happen

Yes
> from what scope will be picked, and whether it will depend on the order of assignment

Order doesn't matter, but ...

> Having tested it, I found out that neither works

Sorry, my mistake (I use a long time very different jail.local as our stock jail.conf). So it will not work, because banaction will be already interpolated like here:
```
# The simplest action to take: ban only
action_ = %(banaction)s[name=%(__name__)s, bantime="%(bantime)s", port="%(port)s", protocol="%(protocol)s", chain="%(chain)s"]
```
And currently we don't support supplying of parameters twice `[...][...]`. Resp. the jail.conf should be a little bit modified. 
Firstly we will need test cases that cover this situation (that we do not have currently)

I need to think a bit about this...
@yarikoptic, @fail2ban/developers Any ideas resp. doubts are welcome

config/action.d/nftables-common.conf
```diff
@@ -0,0 +1,96 @@
...
+actionstart = <nftables> add set filter f2b-<name> { type ipv4_addr\; }
+              <nftables> insert rule filter <chain> %(nftables_mode)s ip saddr @f2b-<name> <blocktype>
+
+# Option:  actionstop
+# Notes.:  command executed once at the end of Fail2Ban
+# Values:  CMD
+#
+actionstop = HANDLE_ID=$(<nftables> --handle --numeric list chain filter <chain> | grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'); <nftables> delete rule filter <chain> $HANDLE_ID
+             <nftables> delete set filter f2b-<name>
+# Option:  actioncheck
```
yarikoptic: minor -- empty line before please

config/action.d/nftables-common.conf
```diff
@@ -0,0 +1,96 @@
...
+# Notes.:  additional expressions for nftables filter rule
+# Values:  nftables expressions
+#
+nftables_mode = <protocol> dport { <port> }
+
+# Option:  actionstart
+# Notes.:  command executed once at the start of Fail2Ban.
+# Values:  CMD
+#
+actionstart = <nftables> add set filter f2b-<name> { type ipv4_addr\; }
```
yarikoptic: shouldn't `{` and `}` be escaped as well?  those are used in posix shell

yarikoptic: actually seems to work in bash and dash but not zsh (with escapes works everywhere)... and by "works" I mean just that it doesn't fail in syntax, since probably I do not really use nftables yet and still iptables, I just get:
```
# nft add set filter f2b-test \{ type ipv4_addr\; \}
<cmdline>:1:1-43: Error: Could not process rule: Table 'filter' does not exist
add set filter f2b-test { type ipv4_addr; }
```
is that a table we might need to setup or a normal system should have it?

yarikoptic: I am yet to learn so much about this new beast... but FWIW in Debian default configuration either doesn't have some generic "filter" but has "inet filter" (not yet even sure if inet used as a family or just a part of two word identifier here, or both):
```
> head /etc/nftables.conf  # that is the default configuration
#!/usr/sbin/nft -f

flush ruleset

table inet filter {
	chain input {
...
```
```
> nft list table inet filter
table inet filter {
	chain input {
		type filter hook input priority 0; policy accept;
		iif lo accept 
		ct state established,related accept 
		ip6 nexthdr ipv6-icmp icmpv6 type { nd-router-advert, nd-neighbor-solicit, nd-neighbor-advert} accept 
		counter packets 11 bytes 1314 drop 
	}
}
```
so I need "nft add set inet filter f2b-test..."

but if that is the truth of life in nft world, we might need an additional parameter identifying that `filter` to manipulate

yarikoptic: and would we need to explicitly limit to ipv4 here?

config/action.d/nftables-common.conf
```diff
@@ -0,0 +1,96 @@
...
+              <nftables> insert rule filter <chain> %(nftables_mode)s ip saddr @f2b-<name> <blocktype>
+
+# Option:  actionstop
+# Notes.:  command executed once at the end of Fail2Ban
+# Values:  CMD
+#
+actionstop = HANDLE_ID=$(<nftables> --handle --numeric list chain filter <chain> | grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'); <nftables> delete rule filter <chain> $HANDLE_ID
```
yarikoptic: this is bloody pity if with nft we need this kind of contraption to clean things up after ourselves :-/  great to learn about "-o" for grep though! ;)
if there is no way around somehow, we should at least split this into some helper variables and use string interpolations, e.g.

```
_nft_list = <nftables> --handle --numeric list chain filter <chain>
_get_handle_id = grep -m1 'ip saddr @f2b-<name> <blocktype> # handle' | grep -oe ' handle [0-9]*'

actionstop = HANDLE_ID=$(%(_nft_list)s | %(_get_handle_id)s); <nftables> delete rule filter <chain> $HANDLE_ID
```
so it fits the screen somewhat better

yarikoptic:
as for above discussion... I guess there would be no way to override via 'banaction' unless we implemented double [][] handling of parameters for actions.  IMHO for the beginning we could happily live while requiring full `action` definition if nftables is to be used.

albel727:
@yarikoptic Wow, great comments! Let's address them.

>empty line before please

Done.

>shouldn't { and } be escaped as well?

Worked in bash, and I didn't think anyone would use other shells for scripting anyway. But I have escaped all braces anyway, keeps working for me, at least.

>is that a table we might need to setup or a normal system should have it?

I'm not really sure about that one.

Thing is, in nftables tables and chains don't pre-exist, and their names don't mean a thing on their own. Hierarchy goes as follows: `[address_family] <table_name> <chain_name>`. With `address_family` being one of `ip`, `ip6`, `inet` (a separate convenience family that can accept both ip and ip6 addresses), etc. Table and chain names are arbitrary, and the only thing that distinguishes, an "input" chain from an "output" one, is that `type filter hook input priority 0;` declaration in it, which tells nftables that this particular chain in this particular table should accept input packets for filtering.

I would expect in a normal system you would create things yourself. Your system pre-creates an `input` chain in a `filter` table in the convenience `inet` address family. On my Gentoo, with nftables-0.5, I don't have `/etc/nftables.conf`, but rather a whole directory `/etc/nftables/` where there are example configurations for filters, nats, bridges and whatnot. I'm clearly supposed to execute whichever I like myself. And then there are startup/shutdown services, store and restore nftables contents and so help to keep chosen configuration across reboots.

```bash
# ls /etc/nftables/
bridge-filter  inet-filter  ipv4-filter  ipv4-mangle  ipv4-nat  ipv6-filter  ipv6-mangle  ipv6-nat
```

>Debian default configuration either doesn't have some generic "filter" but has "inet filter" (not yet even sure if inet used as a family or just a part of two word identifier here, or both)

It's not really a "generic" `filter`. It's just that in absence of address family specifier the default of `ip` family is assumed. Naturally, I have a `filter` table in `ip` family preconfigured. You have a `filter` table in `inet` family instead.

```bash
# nft list tables
table ip filter
```

So I would say you're supposed to create tables and chains on your own and only tell fail2ban which chain to use. 

>we might need an additional parameter identifying that filter to manipulate

I was going to add a "family" parameter before you raised this topic, but now I've also decided to add a "table" parameter as well, in case someone will be unconventional enough to use "darth_vader" table instead of "filter" one. So now there are "nftables_family" and "nftables_table" parameters in `nftables-common.conf`. In fact they're always used in pair throughout the action, and I'm not sure what use there can be for us in family parameter alone, so we might as well leave only "nftables_table" parameter which would contain, e.g. "inet filter" in your case. What do you think?

>and would we need to explicitly limit to ipv4 here?

I'm afraid we do. Unlike convenience `inet` address family, there's no convenience `inet_type` type that could be used to create a set that would contain ipv4 and ipv6 addresses alike. There's only distinct `ipv4_type` and `ipv6_type` types. I didn't think it was a problem anyway until fail2ban begins to support ipv6, but now that you mentioned it, I have added "nftables_type" parameter too.

>this is bloody pity if with nft we need this kind of contraption to clean things up after ourselves

Indeed. IIRC, nftables' author promised to introduce iptables-like "removal by rule contents" eventually, but we'll have to keep the handle approach for backward compatibility for a long time anyway.

>if there is no way around somehow, we should at least split this into some helper variables and use string interpolations

I did as advised, but I'm not sure if I'm supposed to add some comments to those, like for other parameters, and if yes then which. Right now I just stuck the helper variables before actionstop.

sebres:
:+1:

yarikoptic:
@albel727 thanks for the detailed answer!!! I will get through it later today.  Meanwhile let me just ask the question which nagged me yesterday -- if we can create arbitrary tables, why not to create one per each fail2ban jail -- then we could (may be) just kill those at once without that lengthy dance with --handles?

sebres:
@yarikoptic did you mean something like this?
```
nftables_table = f2b-<name>
```
Question would be what a sense do chains in nft, resp. how differently work two solution (by N jails):
```
1 table(s) with ever N chain(s)
N table(s) with ever 1 chain(s)
```
Unfortunately have currently no mashine with nft readily to hand to test it (handling, memory and performance)...

BTW: for first resp. current variant, we could rewrite it as `nftables_table = fail2ban`, because "filter" would be to commonly

albel727:
@yarikoptic 
>if we can create arbitrary tables, why not to create one per each fail2ban jail -- then we could (may be) just kill those at once without that lengthy dance with --handles?

I'm afraid only the nftables authors themselves could answer this question with confidence.

Firstly, I'm not sure about semantics of two chains, e.g. what would it mean if in one table a packet is accepted, and then rejected another. Will another chain even see the packet if the first one accepts it, etc? Don't even want to investigate right now, to be honest, as it's clearly a non-typical, even if possible, setup.

Secondly, aggravating that lack of understanding there's the issue of undefined order of calling input chains. That priority number in `type filter hook input priority 0;` declaration is supposed to define that order, but there are reports that as of now it does nothing.

Lastly, I'm not quite sure if two input chains in separate tables would be as performant as a rule in an existing input chain. Without working priority setting it becomes a total mess, as banned packets may be allowed to pass full length of checks in other input chains before ever reaching ours, which is clearly bad for performance. With working priority setting, we would somehow have to ensure that our chain has the biggest priority - which would mean either hardcoding the biggest integer or using the same table-grepping hacks to figure out existing priorities and doing +1. And even then there's a question if additional input chains bring any overhead.

I honestly don't think grepping is such a big of a hack to warrant going at such lengths to avoid it. Other actions are happily using sed and whatnot. If we don't want to call grep, we might rather code this action in python and parse handle out manually instead.

yarikoptic:
yeah -- if priorities business is indeed this uncertain than not even worth looking into performance impact

Just for my own info:  is priority 0 the highest/processed first or -1 (gets converted to highest unsigned 32bit int according to documentation)? ;)

albel727:
>is priority 0 the highest/processed first 

I only read the same documentation as you, and it has that incomprehensible "highest number is the highest priority" as if that tells something about processing order really. Well, a quick test appears to suggest that highest number means the chain will be the first to receive the packet. I'm not sure if priorities are really like that, as I haven't conducted enough tests to rule out the possibility that what I observed was simply a result of undefined priority order which only seemed to favor higher priority chains. It also appears that semantics of multiple base chains are "accept = pass to the next base chain in order; everything else = stop processing", which was quite expected.

Nevertheless I wouldn't use multiple chains on my systems. I also think that allowing user to specify pre-existing chain to add our filtering rule to offers more fine-grained control for positioning fail2ban's influence, than a simple, even if customizable, chain priority number, which would be the only meaningful thing to control in separate table setup.

albel727:
So, it's been 20 days. Is there something to improve still for it to be merged?

sebres:
will merge, imho further improvements can be made later

albel727:
Great. Thanks, everyone.

PR #1295, opened 2016-01-09
The sender option is ignored by some actions

obounaim:
the actions action_mw, action_mwl and action_cf_mwl ignore the sender option when sending a notification email. This may cause the email to be rejected by the destination or the smarthost in some configuration.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -29,6 +29,7 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
    * Use postfix_log logpath for postfix-rbl jail
    * filters.d/postfix.conf - add 'Sender address rejected: Domain not found' failregex
    * use `fail2ban_agent` as user-agent in actions badips, blocklist_de, etc (gh-1271)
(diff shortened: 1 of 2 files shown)


yarikoptic:
looks good to me. thanks!

mitchellkrogza:
I have noted when using mail and not sendmail as the default mta and I use either action_mw or action_mwl that no sender address is able to be utilised by mail-whois-lines.com or mail-whois.conf because the header "from" is not included in the mail string.

Currently has
```
actionstart = printf %%b "Hi,\n
              The jail <name> has been started successfully.\n
              Regards,\n
              Fail2Ban"|mail -s "[Fail2Ban] <name>: started on `uname -n`" <dest>

```
Should be:
```
actionstart = printf %%b "Hi,\n
              The jail <name> has been started successfully.\n
              Regards,\n
              Fail2Ban"|mail -s "[Fail2Ban] <name>: started on `uname -n`" <dest> -a "From: <sendername> <sender>"


```

PR #1310, opened 2016-01-28
Pr 1288: slight tune up to #1288

yarikoptic:
sending as a PR for testing


yarikoptic:
So no objections to merge it @fail2ban/developers?

PR #1311, opened 2016-01-28
gentoo-initd: do not hide useful output

sbraz:
Gentoo has been applying a patch to work around this issue: https://bugs.gentoo.org/show_bug.cgi?id=536320
It would be nice if we could drop it.


sebres:
:+1: 
btw, @fail2ban/contributors any suggestions, why py3.3 all the time failed (timeouted) on bad_ips test case?

yarikoptic:
I would say that it was just a connection fluke.  Since gentoo ppl say they wanted it this way -- we better keep it that way for gentoo ;)

PR #1312, opened 2016-01-28
Remove compression and count from logrotate

sbraz:
Initially reported at https://bugs.gentoo.org/show_bug.cgi?id=549856
As the reporter mentioned, the `count` parameter means logs are deleted regardless of the global logrotate config; the compression setting should also inherit the global defaults.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,8 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
    * filters.d/postfix.conf - add 'Sender address rejected: Domain not found' failregex
    * use `fail2ban_agent` as user-agent in actions badips, blocklist_de, etc (gh-1271)
    * Fix ignoring the sender option by action_mw, action_mwl and action_c_mwl
+   * Remove compression and rotation count from logrotate (inherit them from
+     the global logrotate config)
 
 - New Features:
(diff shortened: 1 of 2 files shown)


yarikoptic:
sounds sensible to me (I will let this issue age a bit to hear other @fail2ban/developers  and @fail2ban/contributors )... I wonder though why so many logrotate configs on my debian system has those specified explicitly...

szepeviktor:
Sometime it is critical to find a 4 months old mail server log line or PayPal IPN in Apache's log.

yarikoptic:
On Thu, 28 Jan 2016, Viktor Szépe wrote:

>    Sometime it is critical to find a 4 months old mail server log line or
>    PayPal IPN in Apache's log.

sure, but the question is more of either Fail2Ban should inherit
settings from the main logrotate configuration or have its own "opinion"
on the defaults ;)

szepeviktor:
Let's say the max bantime is 1 week ([recidive]). [Logrotate's default](https://github.com/logrotate/logrotate/blob/dd139ef33b163b7f5252c9db3ed4a39586008719/examples/logrotate-default) is 4×weeks, so I would merge this PR.

yarikoptic:
merged manually to resolve the conflict and pushed. thanks

PR #1317, opened 2016-02-05
Add a regex into postfix.conf filter

Turgon37:
The new regexp is able to detect bad formatted SMTP EHLO command

### DIFF
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -15,6 +15,7 @@ _daemon = postfix/(submission/)?smtp(d|s)
 failregex = ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 554 5\.7\.1 .*$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.7\.1 Client host rejected: cannot find your hostname, (\[\S*\]); from=<\S*> to=<\S+> proto=ESMTP helo=<\S*>$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.7\.1 : Helo command rejected: Host not found; from=<> to=<> proto=ESMTP helo= *$
+            ^%(__prefix_line)sNOQUEUE: reject: EHLO from \S+\[<HOST>\]: 504 5\.5\.2 <\S+>: Helo command rejected: need fully-qualified hostname; proto=SMTP helo=<\S+> *$
             ^%(__prefix_line)sNOQUEUE: reject: VRFY from \S+\[<HOST>\]: 550 5\.1\.1 .*$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.1\.8 <\S*>: Sender address rejected: Domain not found; from=<\S*> to=<\S+> proto=ESMTP helo=<\S*>$
             ^%(__prefix_line)simproper command pipelining after \S+ from [^[]*\[<HOST>\]:?$
--- a/ChangeLog
+++ b/ChangeLog
@@ -79,6 +79,7 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
      .service file -- would reload fail2ban if those services are restarted
    * Provides new default `fail2ban_version` and interpolation variable
      `fail2ban_agent` in jail.conf
+   * Enhance filter 'postfix' to ban incoming SMTP client with no fqdn hostname
 
 
 ver. 0.9.3 (2015/08/01) - lets-all-stay-friends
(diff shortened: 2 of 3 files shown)


sebres:
Thx for PR,
to pass all test cases you should extend https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/postfix with corresponding log lines (matched your regexp) and failJSON record.
Don't forget changelog (and thanks :) entry also.

config/filter.d/postfix.conf
```diff
@@ -15,6 +15,7 @@ _daemon = postfix/(submission/)?smtp(d|s)
 failregex = ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 554 5\.7\.1 .*$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.7\.1 Client host rejected: cannot find your hostname, (\[\S*\]); from=<\S*> to=<\S+> proto=ESMTP helo=<\S*>$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.7\.1 : Helo command rejected: Host not found; from=<> to=<> proto=ESMTP helo= *$
+            ^%(__prefix_line)sNOQUEUE: reject: EHLO from \S+\[<HOST>\]: 504 5\.5\.2 <\S+>: Helo command rejected: need fully-qualified hostname; proto=SMTP helo=<\S+> *$
```
sebres: I think, because regexp is ahead anchored, this shorter line would be good enough und more robust:
```
^%(__prefix_line)sNOQUEUE: reject: EHLO from \S+\[<HOST>\]: 504 5\.5\.2 <\S+>: Helo command rejected: need fully-qualified hostname;
```

Turgon37: You are right, I've tested your smaller regexp on my server and it work successfully.

Turgon37:
I'm not full accustomed with pull request process.
In order to make change you asked me, should I create a new pull request that replace the current one ?
In which section of changelog should I add my contribution ? I think to 'Enhancements'

sebres:
- no new PR, please. You should just checkout (or switch to) your branch `patch-1` localy, if you will commit (possibly with amend) new changes in this branch, you can push all the changes it to your remote origin (`git push -f ...`). All changes will be "attached" this PR automatically.
Before you commit/push you can test fail2ban using following from fail2ban current directory:
```
PYTHONPATH=. bin/fail2ban-testcases -n
```
- enhancements - yes

Turgon37:
My too small experience with git have created some really bad commit, but the test pass (at 50 percent).
To prevent that this bad commit line of being integrated in main fail2ban's commit line would you like me to create a new clean pull request ?

sebres:
Don't worry, it is git... I can combine resp. squash all your commits to single one by merge/rebase.
The rest looks good to me (I don't know why the covarage was decreased, but nothing was changed, just ignore this).

Thx.
Will merge it tomorrow if nobody agains it...

config/filter.d/postfix.conf
```diff
@@ -15,6 +15,7 @@ _daemon = postfix/(submission/)?smtp(d|s)
 failregex = ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 554 5\.7\.1 .*$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.7\.1 Client host rejected: cannot find your hostname, (\[\S*\]); from=<\S*> to=<\S+> proto=ESMTP helo=<\S*>$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.7\.1 : Helo command rejected: Host not found; from=<> to=<> proto=ESMTP helo= *$
+            ^%(__prefix_line)sNOQUEUE: reject: EHLO from \S+\[<HOST>\]: 504 5\.5\.2 <\S+>: Helo command rejected: need fully-qualified hostname;
```
yarikoptic: Could n't it be anchored at the end as well roi stay in line with the other regexeses and to satisfy never ending paranoia?

yarikoptic:
Coverage decreased probably since new lines were added, so overall % dropped :-)

config/filter.d/postfix.conf
Turgon37: With SMTP I assume that if a 'smtp client' does not have a full domain name it can not be a valid server. All smtp agent of people in my organisation are able to use my postfix SMTP relay with valid credentials that cause the SMTP daemon to bypass EHLO verification.
It's that why I think this rule is justified to limit spam.
For example : Three days ago, a malicious agent have made my postfix bad by trying to introduce itself with a bad SMTP ehlo string. And of course that had cause some thousand of postmaster mails.

If you don't want to check the format of the EHLO command you can remove the postfix configuration option from main configuration file. In this case, this new fail2ban regex will be useless and will not impact the performance (or a little).
But for the system's administrators that want to be as 'strict' as me this regexp will prevent some malicious behaviour.

sebres: > to satisfy never ending paranoia...
To satisfy it, we should review all the regexp, and rewrite the biggest part of it.
Imho, this shorter regexp (if long enough) is much more robust, so it should satisfy the paranoia rather :) In addition, it is also faster.
Do not misunderstand, I'm not against long regexp at all. Just think as programmer - the shorter, the less vulnerable.

sebres:
> Coverage decreased probably since new lines were added, so overall % dropped 

It would be such a thing, if the code would modified. Because coverals counts code rows only and if you click to details, it says: "coverage remained the same at 91.583%".

sebres:
Squashed (merged) to single commit b5a07741c82acdb9448584a0d39d3df09d28e162
Close.

PR #1309, opened 2016-01-28
Update asterisk.conf

3eBoP:
Changed regex for "Call from ...". Sometimes extension can have a plus symbol (+) because they can be phone number.

PR #1319, opened 2016-02-08
Asterisk regexp fix (replacement for Update asterisk.conf gh-1309)

sebres:
replacement, closes #1309

### DIFF
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -19,7 +19,7 @@ iso8601 = \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+[+-]\d{4}
 log_prefix= (?:NOTICE|SECURITY)%(__pid_re)s:?(?:\[C-[\da-f]*\])? \S+:\d*( in \w+:)?
 
 failregex = ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Registration from '[^']*' failed for '<HOST>(:\d+)?' - (Wrong password|Username/auth name mismatch|No matching peer found|Not a local domain|Device does not match ACL|Peer is not supposed to register|ACL error \(permit/deny\)|Not a local domain)$
-            ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Call from '[^']*' \(<HOST>:\d+\) to extension '\d+' rejected because extension not found in context 'default'\.$
+            ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Call from '[^']*' \(<HOST>:\d+\) to extension '[\d+]+' rejected because extension not found in context 'default'\.$
             ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Host <HOST> failed to authenticate as '[^']*'$
             ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s No registration for peer '[^']*' \(from <HOST>\)$
             ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Host <HOST> failed MD5 authentication for '[^']*' \([^)]+\)$
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -19,7 +19,7 @@ iso8601 = \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+[+-]\d{4}
 log_prefix= (?:NOTICE|SECURITY)%(__pid_re)s:?(?:\[C-[\da-f]*\])? \S+:\d*( in \w+:)?
 
 failregex = ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Registration from '[^']*' failed for '<HOST>(:\d+)?' - (Wrong password|Username/auth name mismatch|No matching peer found|Not a local domain|Device does not match ACL|Peer is not supposed to register|ACL error \(permit/deny\)|Not a local domain)$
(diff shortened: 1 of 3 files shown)


config/filter.d/asterisk.conf
```diff
@@ -19,7 +19,7 @@ iso8601 = \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+[+-]\d{4}
 log_prefix= (?:NOTICE|SECURITY)%(__pid_re)s:?(?:\[C-[\da-f]*\])? \S+:\d*( in \w+:)?
 
 failregex = ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Registration from '[^']*' failed for '<HOST>(:\d+)?' - (Wrong password|Username/auth name mismatch|No matching peer found|Not a local domain|Device does not match ACL|Peer is not supposed to register|ACL error \(permit/deny\)|Not a local domain)$
-            ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Call from '[^']*' \(<HOST>:\d+\) to extension '\d+' rejected because extension not found in context 'default'\.$
+            ^(%(__prefix_line)s|\[\]\s*)%(log_prefix)s Call from '[^']*' \(<HOST>:\d+\) to extension '[^']*' rejected because extension not found in context
```
yarikoptic: Why not to keep trailing anchor with a similar non-'* glob?

sebres: Why we should do that? The first non-'* glob will be used to match rest of the string, that contains information used to identify an event, namely "rejected because extension not found...". The second part is not required, because under no circumstances contributes to the quality of the expression.
The expression should be long enough to identify (resp. definitely match) the event. And it should be very precise before `<HOST>` (what is also the case). Because it does all this, no matter what is even more. 

Still again, the shorter the expression, the few vulnerable.
Think about possible extensions of log-entries by maintainer in featured versions (added more details at end of log-entry), by similar strings described the same event (conditional details at end of log-entry), etc. 
Not to mention about performance.

yarikoptic: because
0. it was anchored at the end before
1. more stringent expression is better/less vulnerable.  Sure thing longer one has higher chance of 'bugs' to creep in.  But with both ends anchored it becomes more robust to such misspecifications.  Minimizing the chance of false positives is our goal.
2. more stringent expression better captures our assumptions on the situation.  Possible extensions to the message might mean different things. again -- we need no false positives, we can allow for false negatives and adjust it later
3. performance might actually get better happen regexp matching engine starts matching from the end and immediately finds a string to reject

sebres: 0\. no matter (because the info at end has no meaning) and it is already anchored at the begin
1 and 2. No, because in since of current failure, all this messages are the same (just failure X):
```
1.2.3.4 failure because reason X
1.2.3.4 failure because reason X, protocol=tcp
1.2.3.4 failure because reason X, whatever
```
3\. that may be possible by real PCRE (finite automata like NFA, TCL, thompson etc.) not for regexp engine of python...
Here is an example of profiler:
```
   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.521    2.521 misctestcase.py:175(_testAnchorPosLeftRight)
        1    0.000    0.000    2.204    2.204 misctestcase.py:170(_testAnchorPosLeft)
        1    0.000    0.000    1.326    1.326 misctestcase.py:172(_testAnchorNegLeft)
        1    0.000    0.000    1.326    1.326 misctestcase.py:177(_testAnchorNegLeftRight)
```

```python
class TestsUtilsTest(unittest.TestCase):

	l = [
		"NOTICE[3477][C-000003bb] chan_sip.c: Call from '' (1.2.3.4:10836) to extension '++441772285411' rejected because extension not found in context 'default'.",
		"NOTICE[3477][C-000003bb] chan_sip.c: just something else",
	]
	r = [
		r"^\w+\[\d+\]\[[\w-]+\] chan_sip.c: Call from '[^']*' \(<HOST>:\d+\) to extension '[^']*' rejected because extension not found",
		r"^\w+\[\d+\]\[[\w-]+\] chan_sip.c: Call from '[^']*' \(<HOST>:\d+\) to extension '[^']*' rejected because extension not found in context '[^']*'\.$",
	]

	def _testAnchor(self, r, l):
		r = r.replace("<HOST>", "(?:::f{4,6}:)?(?P<host>[\w\-.^_]*\w)")
		r = re.compile(r, re.MULTILINE)
		for i in xrange(1000000):
			r.search(l+'\n')
		return r.search(l+'\n')

	def _testAnchorPosLeft(self):
		self.assertTrue( self._testAnchor(TestsUtilsTest.r[0], TestsUtilsTest.l[0]))
	def _testAnchorNegLeft(self):
		self.assertFalse(self._testAnchor(TestsUtilsTest.r[0], TestsUtilsTest.l[1]))

	def _testAnchorPosLeftRight(self):
		self.assertTrue( self._testAnchor(TestsUtilsTest.r[1], TestsUtilsTest.l[0]))
	def _testAnchorNegLeftRight(self):
		self.assertFalse(self._testAnchor(TestsUtilsTest.r[1], TestsUtilsTest.l[1]))

	def testAnchor(self):
		self._testAnchorPosLeft()
		self._testAnchorPosLeftRight()
		self._testAnchorNegLeft()
		self._testAnchorNegLeftRight()
```

yarikoptic: thanks for the analysis!... paranoid me would have still preferred anchoring at the end, but "whatever" - if we get bitten I will just blame on you ;)  I will merge now

PR #1331, opened 2016-02-23
Add support for matching postfix multi-instance daemon names by default

whyscream:
Postfix supports multi instance since 2.6. These instances are configured by default to use the instance name in logging. Naming is restricted to `postfix-<something>`. This PR adds support for this scheme to all postfix filters.


yarikoptic:
thanks

PR #1333, opened 2016-02-24
mysqld: failregex fixed (accepts different log level, more secure expression now)

### LINKED ISSUE #1332: mysqld-auth.conf not compatible with mysql 5.7.11
### Environment:

Fedora 23 Server Edition, Fail2Ban installed via OS

### The issue:

it seems that mysqld-auth conf regex is not compatible with the mysql 5.7.11 error log

log sample

> 2016-02-24T15:26:18.237955Z 6 [Note] Access denied for user 'root'@'localhost' (using password: YES)
as noticed:
1. separator between date time using "T"
2. there's leading numeric after time
3. using [Note] instead of [Warning]

#### Steps to reproduce
1. enabled the mysqld-auth filter
2. emulate fail login in mysql
3. fail2ban-client status mysqld-auth not adding any ip address

#### Expected behavior
fail2ban should add new ip address

sebres:
closes #1332 

Pro rata already "fixed" in #1212 (for issue #1211). 
Here taken into account a different log level: `Note` vs. `Warning`

### DIFF
--- a/config/filter.d/mysqld-auth.conf
+++ b/config/filter.d/mysqld-auth.conf
@@ -17,7 +17,7 @@ before = common.conf
 
 _daemon = mysqld
 
-failregex = ^%(__prefix_line)s(?:\d+ |\d{6} \s?\d{1,2}:\d{2}:\d{2} )?\[Warning\] Access denied for user '\w+'@'<HOST>' (to database '[^']*'|\(using password: (YES|NO)\))*\s*$
+failregex = ^%(__prefix_line)s(?:\d+ |\d{6} \s?\d{1,2}:\d{2}:\d{2} )?\[\w+\] Access denied for user '[^']+'@'<HOST>' 
 
 ignoreregex = 
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,7 +15,7 @@ ver. 0.9.4 (2015/XX/XXX) - wanna-be-released
    * filter.d/apache-badbots.conf
      - Updated useragent string regex adding escape for `+`
    * filter.d/mysqld-auth.conf
-     - Updated "Access denied ..." regex for MySQL 5.6 and later (gh-1211)
+     - Updated "Access denied ..." regex for MySQL 5.6 and later (gh-1211, gh-1332)
    * filter.d/sshd.conf
      - Updated "Auth fail" regex for OpenSSH 5.9 and later
    * Treat failed and killed execution of commands identically (only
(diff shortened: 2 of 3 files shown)


config/filter.d/mysqld-auth.conf
```diff
@@ -17,7 +17,7 @@ before = common.conf
 
 _daemon = mysqld
 
-failregex = ^%(__prefix_line)s(?:\d+ |\d{6} \s?\d{1,2}:\d{2}:\d{2} )?\[Warning\] Access denied for user '\w+'@'<HOST>' (to database '[^']*'|\(using password: (YES|NO)\))*\s*$
+failregex = ^%(__prefix_line)s(?:\d+ |\d{6} \s?\d{1,2}:\d{2}:\d{2} )?\[\w+\] Access denied for user '[^']+'@'<HOST>'
```
yarikoptic: Why to  strip trailing anchor of it didn't hurt anyone?

sebres: > didn't hurt anyone

still not (for example the "warning" was also not bothered until today)...
We have already time talk about the subject "trailing anchor", or not? :)

By building of regexp I follow always minimalist principle. And all standing after `Access denied for user ...<HOST>` bears no meaning for decision on ban... (is only a possible disorder for future changes, any extensions of authentication methods or still not yet occurring cases, that are unknown for us, etc.)

sebres: Oh yeah, forget to mention: to exclude wrong failures, I've of course taken a look for similar error in mysql (currently only this 2 messages):
https://github.com/mysql-inception/inception/search?utf8=%E2%9C%93&q=%22Access+denied+for+user%22&type=Code

yarikoptic: We indeed have talked before on this, and in that case I agreed since number of possible suffixes was growing large and hard to manage.  It was not the case here.
My overall philosophy could be described here as 'better be safe than sorry'.  Fixing up regexes across all releases of Debian/Ubuntu when we have discovered possibility of DDoS via injection was more costly than staying "safe".  So I would really prefer to be "safe" even if it would cost a tiny bit more upfront by needing to adjust regexes whenever new suffix gets discovered.

And my main fear is that without matching a full line we might fall victim again in case e.g. if underlying service manages to log user-provided input at the beginning of the log string in some log message which we simply didn't observe.  To that service it would be a no bigger -- it is just a log, thus who cares...  For us it would be a major pain! We simply can't assure that this will not happen, we can only assume that there always will be some prefix, and that is what you heavily rely upon by stripping down trailing anchoring.
Situation could be even more gruesome in case of centralized logs (syslog, etc) since then we still might then react to logs from other services and no service identifier gets logged (which I believe we allow to be optional since again out of our control).
So, overall -- I would still prefer to specify failregexes as stringently as possible as  a low-cost preventative measure to avoid us possible major pain and suffering in the future.

Here, the "crime" is even two-fold: you have stripped it during fixing something else, without even a changelog note of that.  Ideally fixes should only concern with fixing issue raised by a bug report and have no side-effects, and not "fix" something which is not broken.  FWIW "Updated ..." without description of what for is imho is not informative

sebres: revert?

PR #1345, opened 2016-03-07
datedetector: epoch time expression fix 

sebres:
- epoch time expression bug fixed: wrong recognition of epoch time (now accepts only whole number {10,11} - anchored ^...\b or by special case within [], audit())
- test cases extended (positive/negative case)

### DIFF
--- a/fail2ban/tests/utils.py
+++ b/fail2ban/tests/utils.py
@@ -208,6 +208,18 @@ def gatherTests(regexps=None, no_network=False):
 	return tests
 
 
+# forwards compatibility of unittest.TestCase for some early python versions
+if not hasattr(unittest.TestCase, 'assertIn'):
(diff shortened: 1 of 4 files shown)


fail2ban/tests/utils.py
```diff
@@ -208,6 +208,13 @@ def addTest(self, suite):
 	return tests
 
 
+# forwards compatibility of unittest.TestCase for some early python versions
+if not hasattr(unittest.TestCase, 'assertIn'):
+	def __assertIn(self, a, b):
+		return self.assertTrue(a in b)
```
yarikoptic: I would appreciate if you add a `, msg="%r was not found in %r" % (a, b)"` here so if we use this adapter, we would see what actually caused the failure

sebres: Agree

PR #1350, opened 2016-03-08
Fixed cyrus imap filter as the resolved hostname is optional

majorcs:


config/filter.d/cyrus-imap.conf
```diff
@@ -13,7 +13,7 @@ before = common.conf
 
 _daemon = (?:cyrus/)?(?:imap(d|s)?|pop3(d|s)?)
 
-failregex = ^%(__prefix_line)sbadlogin: \S+ ?\[<HOST>\] \S+ .*?\[?SASL\(-13\): (authentication failure|user not found): .*\]?$
+failregex = ^%(__prefix_line)sbadlogin: (\S+)? ?\[<HOST>\] \S+ .*?\[?SASL\(-13\): (authentication failure|user not found): .*\]?$
```
sebres: `(\S+)? ?` -\> `(?:\S+ )?`
- correct
- non capturing group

Additionally (because we touch this regexp) I tend to remove `.*\]?$` at end of regexp (greedy catch-all anchored at end makes no sense, also if that contains optional bracket)...

So my regexp would be:
```
failregex = ^%(__prefix_line)sbadlogin: (?:\S+ )?\[<HOST>\] \S+ .*?\[?SASL\(-13\): (authentication failure|user not found):
```
@yarikoptic objections?

yarikoptic: 1. alternatively just `\S* ?` (but I like sebres's more explicit one I guess... I don't care if group is capturing or not and would value more "readability" of the regex, or should we care about capturing one?
2. Please do not remove any trailing anchoring unless it is indeed greedy catch-all.  Here ideally it should get that lookback to demand trailing `\]`  if leading one was present

sebres: 1. `(?:...)` - no capturing
2. Definitelly no, because as I already wrote, `\]?` is optional, so expression `.*\]?$` matchs always, no matter what it is
```python
re.search(r"""(.*\]?$)""", r"""test""").groups()
('test',)
```

yarikoptic: 1.  I know that this one is non-capturing, but does it give us any advantage over a more concise capturing `(...)`?
2. I know that it is optional ATM, but it imho should be made non optional if leading `\[` is present

sebres: 1) In regexp (no matter which re-engine used) is **always** better not to use capturing groups, if you not need to get it. At least it's faster by searching of match and retrieving of really expected groups and fewer resource "intensively".
```python
## non-capture (host only)...
>>> timeit.timeit('m = r.search("test host"); m.group("host")', 
... setup='import re; r = re.compile(r"""^(?:\S*\s)?(?P<host>host)$""")', number=1000000)
0.77633031683956233
## capture (host + no-named one)...
>>> timeit.timeit('m = r.search("test host"); m.group("host")',
... setup='import re; r = re.compile(r"""^(\S*\s)?(?P<host>host)$""")', number=1000000)
0.84158292268296009
```
And it is just a simple test...

2) I was in the mode, where a new expression would be qualitative better, without changing it completely (leading `\[` or not, etc.)

yarikoptic: thanks @sebres -- so non capturing one please @majorcs,  trailing anchor here is superficial/non-effective anyways so could indeed be gone until someone comes with a better one (become having non optional `]` if leading one is present)

sebres:
Fixed in gh-1708. Thus closing.

Issue #1354, opened 2016-03-09
mysql log path for fedora (Fedora 23 SE)


GitzJoey:
### Environment:
Fedora 23 Server Edition
Fail2Ban installed via OS/distribution mechanisms

### The issue:
I know it override-able via path-overrides.local but it would be better since the paths file already specific to each os, The correct log path for mysql, for fedora distributions is /var/log/mysqld.log

so changes required in file paths-fedora.conf

sebres:
The question is - it also applies to most another versions of Fedora?

GitzJoey:
been using fedora since 17, the actual log always in /var/log/mysqld.log, 
i think its also depend on the mysql version, i'm using mysql since version 5.6 (now 5.7.11) and the log in fedora dist is always /var/log/mysqld.log

require other to verify.
thanks for the fast attention, appreciate it :)

opoplawski:
Another twist is that the default DB in Fedora is actually mariadb and that logs to /var/log/mariadb/mariadb.log.

weberhofer:
Isn't mysql/mariadb using the systemd-backend on fedora? This is globally set in paths-fedora.conf and should work with the latest version 0.9.4.

opoplawski:
No, mysql/mariadb do not log the failed login attempts to the journal, only to the log file.

sebres:
fixed in #1356

PR #1356, opened 2016-03-09
Fedora use mariadb by default, fix log path

opoplawski:
Fixes issue #1354

### DIFF
--- a/config/paths-fedora.conf
+++ b/config/paths-fedora.conf
@@ -34,7 +34,7 @@ apache_access_log = /var/log/httpd/*access_log
 
 exim_main_log = /var/log/exim/main.log
 
-mysql_log = /var/lib/mysql/mysqld.log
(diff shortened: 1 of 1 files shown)


config/paths-fedora.conf
```diff
@@ -34,7 +34,7 @@ apache_access_log = /var/log/httpd/*access_log
 
 exim_main_log = /var/log/exim/main.log
 
-mysql_log = /var/lib/mysql/mysqld.log
+mysql_log = /var/log/mariadb/mariadb.log
```
sebres: what's about both? I've not tried it, but think it should be possible in includes also, since we support multiline there... I mean, something like:
```
mysql_log = /var/log/mysqld.log
            /var/log/mariadb/mariadb.log
```

opoplawski: Ah, I'd totally forgotten about that.  That would be the best solution indeed.

PR #1359, opened 2016-03-10
BF: Include iptables-common (not removed iptables-blocktype) within firewallcmd- actions

### LINKED ISSUE #1254: firewallcmd-multiport.conf and firewallcmd-allports.conf have iptables-blocktype.conf dependency but that file is not installed by EPEL v7 fail2ban RPMs
On the latest stable EPEL v7 RPM after installing fail2ban (v0.9.3-1), using firewallcmd-multiport.conf and firewallcmd-allports.conf fails because they have a 
before = iptables-blocktype.conf

statement.  However that file is missing from the RPM's.  That file should either be added to the RPMs or the dependency changed to iptables-common.conf which already defines the blocktype.

I am currently using this file the get around it for now.
https://github.com/breauxaj/puppet-fail2ban/blob/master/files/original/rhel/action.d/iptables-blocktype.conf

yarikoptic:

- [x] **LIST ISSUES** this PR resolves Closes #1254

### DIFF
--- a/config/action.d/firewallcmd-allports.conf
+++ b/config/action.d/firewallcmd-allports.conf
@@ -6,7 +6,7 @@
 
 [INCLUDES]
 
-before = iptables-blocktype.conf
+before = iptables-common.conf
 
 [Definition]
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
should be safe, thus will merge tomorrow unless e.g. @sebres points to some problem! ;)

sebres:
speak no firewallcmd, but seems to be ok )))

PR #1362, opened 2016-03-11
define journalmatch setting for pure-ftps

weberhofer:


yarikoptic:
Thanks! Just to make sure - it does with for you if you use systemd backend for this jail?

weberhofer:
Yes, it's working. Currently I have hundreds of blocked IPs on one of my servers using this setting.

yarikoptic:
OK, then either you as changelog entry or I will do that later when get home and merge

weberhofer:
@yarikoptic please do the merge, I can't.

PR #1365, opened 2016-03-16
SmartOS PAM Authentication failed (not failURE)

jblachly:
On SmartOS (and likely other Illumos platforms -- OmniOS, OpenIndiana, Delphix, etc.) sshd + PAM enters authentication failure entries in the log file of the form:

`Authentication failed for USER from HOST`

However, the current regex in `sshd.conf` matches fail**ure** not fail**ed**. I added an additional match to support SmartOS and other platforms.

This is a one-line change; I did not run any automated testing.

### DIFF
--- a/config/filter.d/sshd.conf
+++ b/config/filter.d/sshd.conf
@@ -18,7 +18,7 @@ before = common.conf
 
 _daemon = sshd
 
-failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failed|failure|error) for .* from <HOST>( via \S+)?\s*$
+failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|error|failed) for .* from <HOST>( via \S+)?\s*$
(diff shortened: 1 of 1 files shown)


jblachly:
Looks like a bug in coveralls.io -- a -0.0% change causes a failure.

config/filter.d/sshd.conf
```diff
@@ -18,7 +18,7 @@ before = common.conf
 
 _daemon = sshd
 
-failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|error) for .* from <HOST>( via \S+)?\s*$
+failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failed|failure|error) for .* from <HOST>( via \S+)?\s*$
```
sebres: I tend to merge it, but would prefer another order here (because all other OS have `failURE`)...
So please change to 
```
...(?:failure|error|failed)...
```

Thx

jblachly:
Done

PR #1367, opened 2016-03-17
Two new firewalld actions

TorontoMedia:
Added two new firewalld actions for a cleaner iptables layout and also fixed firewallcmd allports and multiport.

### DIFF
--- a/config/action.d/firewallcmd-allports.conf
+++ b/config/action.d/firewallcmd-allports.conf
@@ -1,6 +1,10 @@
 # Author: Donald Yandt 
 # Because of the --remove-rules in stop this action requires firewalld-0.3.8+
 
+[INCLUDES]
+
+before = iptables-common.conf
+
 [Definition]
 
 actionstart = firewall-cmd --direct --add-chain ipv4 filter f2b-<name>
@@ -15,9 +19,9 @@ actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -j f2b-<n
 
 actioncheck = firewall-cmd --direct --get-chains ipv4 filter | sed -e 's, ,\n,g' | grep -q '^f2b-<name>$'
 
-actionban = firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j REJECT
+actionban = firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j <blocktype>
 
-actionunban = firewall-cmd --direct --remove-rule ipv4 filter f2b-<name> 0 -s <ip> -j REJECT
+actionunban = firewall-cmd --direct --remove-rule ipv4 filter f2b-<name> 0 -s <ip> -j <blocktype>
 
 [Init]
 
--- a/config/action.d/firewallcmd-allports.conf
+++ b/config/action.d/firewallcmd-allports.conf
@@ -1,12 +1,5 @@
-# Fail2Ban configuration file
-#
 # Author: Donald Yandt 
 # Because of the --remove-rules in stop this action requires firewalld-0.3.8+
-
-
-[INCLUDES]
-
-before = iptables-blocktype.conf
 
 [Definition]
 
@@ -18,28 +11,25 @@ actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -j f2b-<n
              firewall-cmd --direct --remove-rules ipv4 filter f2b-<name>
              firewall-cmd --direct --remove-chain ipv4 filter f2b-<name>
 
-
 # Example actioncheck: firewall-cmd --direct --get-chains ipv4 filter | sed -e 's, ,\n,g' | grep -q '^f2b-recidive$'
 
 actioncheck = firewall-cmd --direct --get-chains ipv4 filter | sed -e 's, ,\n,g' | grep -q '^f2b-<name>$'
 
-actionban = firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j <blocktype>
+actionban = firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j REJECT
 
-actionunban = firewall-cmd --direct --remove-rule ipv4 filter f2b-<name> 0 -s <ip> -j <blocktype>
+actionunban = firewall-cmd --direct --remove-rule ipv4 filter f2b-<name> 0 -s <ip> -j REJECT
 
 [Init]
 
-# Default name of the chain
-#
 name = default
 
+# Default name of the chain
 chain = INPUT_direct
 
 # DEV NOTES:
 #
 # Author: Donald Yandt 
 # Uses "FirewallD" instead of the "iptables daemon".
-#
 #
 # Output:
 
@@ -50,4 +40,3 @@ chain = INPUT_direct
 # success
 # $ sudo firewall-cmd --direct --add-rule ipv4 filter INPUT_direct 0 -j f2b-recidive
 # success
-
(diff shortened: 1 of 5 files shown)


config/action.d/firewallcmd-allports.conf
```diff
@@ -18,21 +11,19 @@ actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -j f2b-<n
              firewall-cmd --direct --remove-rules ipv4 filter f2b-<name>
              firewall-cmd --direct --remove-chain ipv4 filter f2b-<name>
 
-
 # Example actioncheck: firewall-cmd --direct --get-chains ipv4 filter | sed -e 's, ,\n,g' | grep -q '^f2b-recidive$'
 
 actioncheck = firewall-cmd --direct --get-chains ipv4 filter | sed -e 's, ,\n,g' | grep -q '^f2b-<name>$'
 
-actionban = firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j <blocktype>
+actionban = firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j REJECT
```
sebres: > and also fixed firewallcmd allports and multiport.

??? What was fixed here?

Block type can also be specified in Init block as default, and can be rewritten in jail or in includes (if expected):
```
[Init]
blocktype = REJECT
```

BTW, I do not understand why following test does not fail by this action (https://github.com/fail2ban/fail2ban/blob/5d0d96a5cbe7b8b45173f440568386ecd7d1924b/fail2ban/tests/clientreadertestcase.py#L652)

TorontoMedia: Sorry that was a mistake. I wasn't synced with [eed1a0d686ba8bb0aae7ec1546a2d4aa2415ba74]((https://github.com/fail2ban/fail2ban/pull/1359/commits/eed1a0d686ba8bb0aae7ec1546a2d4aa2415ba74)) and was testing dependencies and copied the wrong file since I have three versions of fail2ban in my environment(gcm, xmpp, i18n).

config/action.d/firewallcmd-allports.conf
```diff
@@ -29,10 +25,9 @@ actionunban = firewall-cmd --direct --remove-rule ipv4 filter f2b-<name> 0 -s <i
 
 [Init]
 
-# Default name of the chain
-#
 name = default
```
yarikoptic: It is a default name for the fail2ban chain ;-)

config/action.d/firewallcmd-multiport.conf
```diff
@@ -3,17 +3,18 @@
...
 
-before = iptables-blocktype.conf
+before = iptables-common.conf
 
 [Definition]
 
 actionstart = firewall-cmd --direct --add-chain ipv4 filter f2b-<name>
               firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 1000 -j RETURN
-              firewall-cmd --direct --add-rule ipv4 filter <chain> 0 -m state --state NEW -p <protocol> -m multiport --dports <port> -j f2b-<name>
+              firewall-cmd --direct --add-rule ipv4 filter <chain> 0 -m conntrack --ctstate NEW -p <protocol> -m multiport --dports <port> -j f2b-<name>
```
yarikoptic: Would this new form be compatible with older versions of the firewall-cmd?

config/action.d/firewallcmd-rich-logging.conf
```diff
@@ -0,0 +1,41 @@
...
+
+# log levels are "emerg", "alert", "crit", "error", "warning", "notice", "info" or "debug"
+level = info
+
+# log rate per minute
+rate = 1
+
+zone = public
+
+service = ssh
```
yarikoptic: Is service entry a mere label or it points to the ports?

config/action.d/firewallcmd-rich-logging.conf
```diff
@@ -0,0 +1,41 @@
+# Fail2Ban configuration file
```
yarikoptic: Might be worth adding short description what is so special about this rich one

config/action.d/firewallcmd-allports.conf
TorontoMedia: I'll fix that. I don't know why I got the input chain mixed up with the f2b chains.

config/action.d/firewallcmd-multiport.conf
TorontoMedia: I could be wrong but this is what I've gathered on nf_conntrack support:

- in modules.py function get_firewall_modules() it just reads from "/proc/modules" and loads dependencies if they exist so no compatibility checks.

- further research into all the talk about firewalld getting conntrack support suggests that they are talking about nf_conntrack userspace tools and not the kernel conntrack module.

- all the services have nf_conntrack module names. Example "<module name="nf_conntrack_amanda"/>"

with that stated and module "state"  being depreciated I suggest we move to nf_conntrack.

TorontoMedia:
Sorry about all the commits my grammar is terrible today. 

If someone could write something that executes the command for every port if a comma delimiter is used between ports that would be awesome. Since these actions are limited to a single port or a range of ports.

sebres:
> If someone could write something that executes the command for every port if a comma delimiter is used between ports that would be awesome. 

Something like this (should work in default shell):
```bash
ports="80,8080,88,8088" ; for p in $(echo $ports | tr "," " "); do echo execute command for port "$p"; done
```
or this (`bash -c '...'`):
```bash
ports="80,8080,88,8088" ; for p in ${ports//,/ }; do echo execute command for port "$p"; done
```

TorontoMedia:
I'm sorry but could you please give me an example that works in the config file since I can't for the life of me get it working in the config. I can only echo the commands in terminal but nothing is executed... My bash knowledge is poor.

sebres:
ConfigParser seems to have a bug with spaces in front of semicolon (I think it's a bug, because imho the comments should begin with new line...)
```python
>>> s = '[def]\ntest1=1;2\ntest2=3 ;-) 4\n'; print(s); p = ConfigParser.ConfigParser(); p.readfp(StringIO.StringIO(s)); p.items('def')
[def]
test1=1;2
test2=3 ;-) 4

[('test1', '1;2'), ('test2', '3')]
```
No way to escape, we can prospective extend our config reader to replace something like `<SEMCOL>` with `;`. But atm it can be used still without leading space...

So either remove leading space for each `;` or make a new sh-file (as executable) and put it in config... Below is an example for functioning config entry:
```python
>>> s = '[def]\nactionban = ports="<port>"; for p in $(echo $ports | tr "," " "); do ban <HOST> --port "$p"; done\n'; print(s); p = ConfigParser.ConfigParser(); p.readfp(StringIO.StringIO(s)); p.items('def')
```
```
[def]
actionban = ports="<port>"; for p in $(echo $ports | tr "," " "); do ban <HOST> --port "$p"; done

[('actionban', 'ports="<port>"; for p in $(echo $ports | tr "," " "); do ban <HOST> --port "$p"; done')]
```

TorontoMedia:
I was able to get it working and it's awesome. Thank you. This is now ready to be committed if you want to.

TorontoMedia:
Sorry I'm done editing them now.

sebres:
:+1:

TorontoMedia:
I have a question regarding "state" should we not be banning established connections too. Since blocking new connections works in a stateless connection but what about an established connection that has passed the queue and authentication and is now generating fail logs?

sebres:
Too many irrelevant commits, so I've rebased it onto current master...
(at the same time put away indifferent changes like styles, comments, white spaces etc.)

PR #1368, opened 2016-03-19
[WIP] add ability to use memory db while running and save to file on …

tjikkun:
Early pull-request to check if this is wanted functionality, and if I am on the right track.
Code is not tested yet
What it does:
If you put dbfile = :memory:,/path/to/backup/dbfile.sqlite3 in your config, this will on startup read the db in /path/to/backup/dbfile.sqlite3 and copy it to memory. Then it will use the db in memory, and finally on shutdown it will copy the memory db to /path/to/backup/dbfile.sqlite3 again.

I wrote this because I noticed that on a busy server fail2ban is doing quite a few iops, when switching to memory this is gone but that means a longer startup time

Comments welcome!

### DIFF
--- a/fail2ban/server/database.py
+++ b/fail2ban/server/database.py
@@ -34,6 +34,14 @@ from .mytime import MyTime
 from .ticket import FailTicket
 from ..helpers import getLogger
 
+try:
+	import sqlitebck
+except ImportError:
+	# Dont print error here, as sqlitebck may not even be used
+	sqlite_backup = False
+else:
+	sqlite_backup = True
+	
 # Gets the instance of the logger.
 logSys = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


fail2ban/server/database.py
```diff
@@ -34,6 +34,14 @@
 from .ticket import FailTicket
 from ..helpers import getLogger
 
+try:
+	import sqlitebck
```
yarikoptic: in general we have managed to avoid depending on anything but stock python and only few core libraries... this one seems to be quite concise, but not even present in debian and seems might have issues on osx (https://github.com/husio/python-sqlite3-backup/issues/5)... so to jump that way it would be nice to have concrete numbers of how beneficial this endeavor would be

tjikkun: Ok understood. I am a Fedorian myself and at the moment it is not packaged there either, but I will do that if this patch gets accepted.
As it is written it adds a new feature, but only if sqlitebck is available. Another possibility would be to copy the database another way if sqlitebck is not available (https://docs.python.org/release/2.6/library/sqlite3.html#sqlite3.Connection.iterdump)
As to the benefit: When we started using fail2ban we saw the number of writes/sec to disk increase from an average of 100 to an average of 600. This dropped again to 100 when we switched to :memory: (but this causes minutes of 100% CPU on startup)
Anyway, thanks for responding!

tjikkun: Hmm thinking about it these numbers seem a bit high, I will re-check that soon, probably tomorrow. But in any case they were high enough to immediately draw our attention.

tjikkun:
Ok I ran fail2ban with dbfile = /var/lib/fail2ban/fail2ban.sqlite3 for a little while on 2 of our servers. The results on both of them are similar, so I will only post about one of them here:
The server was running with dbfile = :memory:
At 09:36:43 I restarted with dbfile = /var/lib/fail2ban/fail2ban.sqlite3
Till about 10:47 fail2ban was doing 100% CPU (To be expected since /var/lib/fail2ban/fail2ban.sqlite3 was empty and fail2ban has a lot of catching up to do)
Starting from 10:47 our OS disks jumped from about 10-15 writes/sec to about 2k writes/sec
screenshot from 2016-03-29 11-34-12
These are from zabbix vfs.dev.write[sdc,operations] and vfs.dev.write[sdd,operations]
At 11:27 fail2ban was restarted again with dbfile = :memory:

tjikkun:
@yarikoptic Do you have any input on this? Shall I pursue this further or is there no interest?

sebres:
I think, that I've made some speedups of the fail2ban, also in direction sqlite writing (so #1346 together with #716 on same servers produce signifier much fewer IO last).

So I'm not against this PR at all (I think this is a nice idea), just to describe the problem from other side.

sebres:
@tjikkun can you repeat please your test with #1436, because I will merge this feature in our 0.9 and 0.10 branches, and would be not bad to know, how beneficial would your FR comparable with #1436

tjikkun:
Closing this because #1436 is a much better idea, getting the same benefits.

sebres:
Thx for testing, I don't think it has really the same performance, but good to know it is at least comparable.

PR #1373, opened 2016-03-24
Show the most recent 100 lines instead of the first 1000

leto:
  I recently had an IP get banned that had many thousands of entries in today's log file. Showing the first 1000 lines gives no context to why the IP is banned if they had 10,000 valid requests today and then "sped up" and got banned. Showing the most recent X lines is probably what most people want.


yarikoptic:
I am not sure on this one... I guess "it all depends".  So far showing first ones was useful (IIRC I have introduced it awhile back) to see indeed how far back in time these attempts were going on.  I could argue that the last ones might be less informative.  I guess we could all make it an option so you could easily override in your settings within .local or smth like that... but still not sure

sebres:
Implemented in 0.10 within #1622 for all similar actions. Thus close.

Issue #39, opened 2012-03-05
IPv6 Support


raccettura:
fail2ban doesn't appear to have any IPv6 support.  With lots of networks (and thus hosts) trying to turn it on for World IPv6 Launch on June 6[1], this is a blocker.

http://www.worldipv6launch.org/

mgrant0:
Is there any ipv6 support in 0.9.1?  I see it mentioned above but I don't see it in the release notes for 0.9.1.

yarikoptic:
there is no ipv6 support in stock fail2ban releases yet.  You can find mentioned above some tentative solutions you could try, thus this issue is not closed yet

monkeyhybrid:
I have to agree with @brianjmurrell in that the whole talk of subnet blocking is getting in the way of the more straightforward, and painfully absent, functionality for just working with IPv6 addresses. I think Fail2ban is great, but it does seem rather ridiculous that in this day and age, IPv6 is still not officially supported.

Three years on, I don't see any consensus on what a resolution to this issue should look like. Would it not be best to at least implement support for IPv6 addresses, and then think about subnet blocking as a separate issue?

calestyo:
It's simply totally useless to "support" v6, without subnet blocking. Or what would your use case be?

johnpatcher:
I want to backup samir3 on this. Don't let perfection be the opponent of usefulness on this one. Adding support for IPv6 (without support for subnets) is perfectly fine and there are some (limited) use cases for it. Support for subnets can then be added and also implemented for IPv4 while you're at it.

Blocking whole subnets for IPv6 will get tricky anway, because simply blocking a /64 subnet might be based on the wrong assumption, so I don't see in which way these two issues are related.

monkeyhybrid:
@calestyo - It's far from useless to support IPv6 addresses without subnet blocking functionality. Whilst IPv6 makes it trivial for attackers to work from a large pool of IPs, in practice, I'm not seeing that happen any more than I already see with IPv4.

calestyo:
@monkeyhybrid You say it... it's trivial,... and I bet with you that 1 day (if not one hour) after nodes would actually start to employ blocking techniques at the v6 level (e.g. via fail2ban) attackers would start to adapt.

While the last commenting people said it would be useful, no one really could explain for what (other than if we assume good faith by attackers not to cycle through their ranges).

So right now I can only see disadvantages.

E.g. giving the user a wrong feeling for beeing more secure.

[...]

PR #1374, opened 2016-03-24
IPv6 support

koeppea:
Hi,

I'm loving fail2ban but really suffered lacking IPv6 support which motivated me working on this pull request.

Please forgive me that I couldn't keep to one of your principals (Keep the PR small) as this change is huge by nature.

Therefore I've kept the commits detailed to help understanding the changes. When everything is fine, I can squash them down to a *ENH* oneline commit.

Next steps would be, finalizing nftables action including v6 support and maybe adding some more IPv6 related tests (I have get bit more used to the test framework).


sebres:
Thx for the PR!
I will take a look after Easter...

config/action.d/iptables-allports.conf
```diff
@@ -42,6 +53,7 @@ actioncheck = <iptables> -n -L <chain> | grep -q 'f2b-<name>[ \t]'
 # Values:  CMD
 #
 actionban = <iptables> -I f2b-<name> 1 -s <ip> -j <blocktype>
+actionban6 = <ip6tables> -I f2b-<name>6 1 -s <ip> -j <blocktype6>
```
yarikoptic: as previous discussions pointed out -- it might be not effective to ban solely originating IPv6 ip alone but requires some reasonable (but what is reasonable?) netmask... what is your take on that?

config/jail.conf
```diff
@@ -47,7 +47,7 @@ before = paths-debian.conf
 # "ignoreip" can be an IP address, a CIDR mask or a DNS host. Fail2ban will not
 # ban a host which matches an address in this list. Several addresses can be
 # defined using space (and/or comma) separator.
-ignoreip = 127.0.0.1/8
+ignoreip = 127.0.0.1/8 ::1
```
yarikoptic: how in given approach you enable/disable ipv6 support altogether?

fail2ban/server/actions.py
```diff
@@ -178,7 +179,7 @@ def setBanTime(self, value):
 	def getBanTime(self):
 		return self.__banManager.getBanTime()
 
-	def removeBannedIP(self, ip):
+	def removeBannedIP(self, ipstr):
```
yarikoptic: if renamed -- docstring should be adjusted accordingly

fail2ban/server/filter.py
```diff
@@ -354,34 +375,22 @@ def getIgnoreIP(self):
 	#
 	# Check if the given IP address matches an IP address/DNS or a CIDR
 	# mask in the ignore list.
-	# @param ip IP address
+	# @param ip IP address object
 	# @return True if IP address is in ignore list
 
 	def inIgnoreIPList(self, ip, log_ignore=False):
```
yarikoptic: looking at all the tests needing wrapping of `ip:str -> IpAddr(ip)`, may be worth supporting here both str and IpAddr specification with a quick check or even making a "copy constructor" form `IpAddr(ipaddr)` or `asanyipaddr` (numpy rotten soul speaking as for analog to `asanyarray`) helper constructor which would simply assure having an IpAddr object out of string or passing along original IpAddr object if that one given?  That would remove majority of necessary tune ups needed for tests

fail2ban/server/filter.py
```diff
@@ -920,19 +913,234 @@ def textToIp(text, useDns):
...
+class IPAddr:
+	""" provide functions to handle IPv4 and IPv6 addresses 
+	"""
+
+	IP_CRE = re.compile("^(?:\d{1,3}\.){3}\d{1,3}$")
+	IP6_CRE = re.compile("^[0-9a-fA-F]{4}[0-9a-fA-F:]+:[0-9a-fA-F]{1,4}|::1$")
+
+	# object attributes
+	addr = 0
+	family = 0
```
yarikoptic: "explicit better than implicit"  -- `socket.AF_UNSPEC`?

fail2ban/server/filter.py
```diff
@@ -920,19 +913,234 @@ def textToIp(text, useDns):
...
+			# (e.g. DNS resolution)
+			self.raw = ipstring
+
+	def __repr__(self):
+		return self.ntoa()
+
+	def __str__(self):
+		return self.ntoa()
+
+	def __eq__(self, other):
```
yarikoptic: I guess if here that `asanyipaddr` used, then comparisons could be carried out between ipaddr and string representations which could be a handy feature making code more lightweight...?

yarikoptic:
BTW, no need to squash anything IMHO.  Keep PR small principle is primarily to state that it should not include unrelated to proposed fix/enhancement changes.  # of commits doesn't matter (at least to me), so keep it growing. One of the first priorities is it to make tests pass ;)  @sebres and I (and hopefully others) would look into details as time goes and provide more feedback -- hopefully we will settle it down for merge sooner or later

config/action.d/iptables-allports.conf
koeppea: Honestly, I don't see the point in blocking anything else than the effective IP address.
It's the same as with IPv4 where also just hosts are blocked; not networks. This behavior I only know from the army where a group is being punished for failures of individuals.
Even though in IPv6 a host can have multiple IPv6 address, especially with Privacy Extensions enabled, it's very unlikely that it will change it's outbound IPv6 address faster than the ban will expire.
So again, I can't really join this opinion.

config/jail.conf
koeppea: To add *::1* reflects just the default value for IPv4 and gives a hint, that also IPv6 addresses or networks can be added to this configuration variable.
I don't know why *fail2ban*'s IPv6 support should be disabled. If it runs on a host with IPv4 connectivity only, then only the IPv4 specific parts of the code are being executed. The presence of the only depended functions **inet_pton()** and **inet_ntop()** are bound to the *python* version. This could be the only decision point IMHO.
Generally my personal opinion on this is, that the time is over to consider IPv6 being optional in network related software.

fail2ban/server/actions.py
koeppea: Thanks, will be fixed...

fail2ban/server/filter.py
koeppea: Yes as with this PR, an IP address is being treated as a IPAddr object, the tests need to honor this change/fact. Unfortunately I don't know/understand what you mean with `asanyipaddr` or `asanyarray`. Do you have any reference to what you mean by that? If anything can be simplified, why not?

fail2ban/server/filter.py
koeppea: I'll check on that ....

fail2ban/server/filter.py
koeppea: Again, here I don't understand what you exactly mean. I'd be glad if you could explain a bit more.

koeppea:
> One of the first priorities is it to make tests pass

for sure, I'm also working on it. It seems that travis is doing more tests than when I execute the test scipts on my machines.

fail2ban/server/actions.py
koeppea: Fixed with commit  	0e65c96

fail2ban/server/filter.py
koeppea: Done with 6a46227

koeppea:
I don't have experience with this "Coverall". What is necessary to fix the coverage decrease?

yarikoptic:
quick one re coverage:  I would recommend to click on "Review entire Coverage Diff " in codecov-io box (which should get updated after each commit, thus not posting a link) and go through the diff paying attention to new lines which are in red -- that means they aren't tested.  Ideally every line should get tested, but some times it might be tricky ;)

fail2ban/server/filter.py
yarikoptic: pretty much smth like
```python

def asanyipaddr(ip):
  return ip if isinstance(ip, IpAddr) else IpAddr(ip)
```

```python
def __eq__(self, other):
   other = asanyipaddr(other)
   ...
```

we could even add more "semantic" to variables at few places, to have eg "ip" -- either str or IpAddr, "ipaddr" -- as you already have IpAddr, "ipstr" -- str representation

fail2ban/server/filter.py
```diff
@@ -920,19 +913,234 @@ def textToIp(text, useDns):
...
+		if not isinstance(other, IPAddr): return False
+		if not self.valid and not other.valid: return self.raw == other.raw
+		if not self.valid or not other.valid: return False
+		if not self.addr == other.addr: return False
```
yarikoptic: btw I would prefer in general use of "!=" to "not x == y".  Here in particular I would prefer 
```python
return (
   self.valid and other.valid and 
   self.addr == other.addr and 
   self.family == other.family and 
   self.plen == self.plen
)
```

fail2ban/server/filter.py
yarikoptic: to go fancy you might even want to come up with a decorator which would take care about converting the first positional argument (or kwarg ip), so code could then look like
```
@iparg
def __eq__(self, other)
  ...
```
thus making it a bit more explicit in reading

TorontoMedia:
I'm going to stay out of this discussion since I don't know enough to have a opinion but I have a concern were ipv6 is mapped to ipv4 using the prefix of `::ffff:0:0/96` and that in rfc 6052(4.2) they recommend mapping to /96 due to translation problems when sending, so if they recommend client side routers to stop prefixing with ipv4 mapping then we should probably take that into consideration...

fail2ban/server/filter.py
koeppea: Thanks, this makes definetely sense. Let me work this into the PR.

fail2ban/server/filter.py
koeppea: Let me think this through. Especially the approach with the decorator could take out complications out of this.

koeppea:
I built in the recommendations and optimized the changes in the tests to the necessary parts only. In order to keep the history clean, I rebased the branch.
Now I need to get Conginuous happy....

config/action.d/iptables-allports.conf
```diff
@@ -21,6 +23,10 @@ actionstart = <iptables> -N f2b-<name>
               <iptables> -A f2b-<name> -j <returntype>
               <iptables> -I <chain> -p <protocol> -j f2b-<name>
 
+actionstart6 = <ip6tables> -N f2b-<name>6
+               <ip6tables> -A f2b-<name>6 -j <returntype>
+               <ip6tables> -I <chain> -p <protocol> -j f2b-<name>6
```
yarikoptic: I see no problem of having the same name (`f2b-<name>`) in both iptables and ip6tables -- those seems to be maintained independently... so what is the purpose of the 6 suffix?

fail2ban/client/beautifier.py
```diff
@@ -78,13 +78,13 @@ def beautify(self, response):
...
 							prefix2 = prefix1 + ("`-" if m == len(res1[1]) - 1 else "|-")
-							val = " ".join(res2[1]) if isinstance(res2[1], list) else res2[1]
+							val = " ".join(map(str,res2[1])) if isinstance(res2[1], list) else res2[1]
 							msg.append("%s %s:\t%s" % (prefix2, res2[0], val))
 				else:
 					msg = ["Status"]
 					for n, res1 in enumerate(response):
 						prefix1 = "`-" if n == len(response) - 1 else "|-"
-						val = " ".join(res1[1]) if isinstance(res1[1], list) else res1[1]
+						val = " ".join(map(str,res1[1])) if isinstance(res1[1], list) else res1[1]
```
yarikoptic: very minor... sorry for whining... please add a space after `,` in map call

fail2ban/tests/filtertestcase.py
```diff
@@ -251,21 +257,21 @@ def testIgnoreIPNOK(self):
 
 	def testIgnoreIPCIDR(self):
 		self.filter.addIgnoreIP('192.168.1.0/25')
-		self.assertTrue(self.filter.inIgnoreIPList('192.168.1.0'))
-		self.assertTrue(self.filter.inIgnoreIPList('192.168.1.1'))
-		self.assertTrue(self.filter.inIgnoreIPList('192.168.1.127'))
-		self.assertFalse(self.filter.inIgnoreIPList('192.168.1.128'))
-		self.assertFalse(self.filter.inIgnoreIPList('192.168.1.255'))
-		self.assertFalse(self.filter.inIgnoreIPList('192.168.0.255'))
+		self.assertTrue(self.filter.inIgnoreIPList(IPAddr('192.168.1.0')))
```
yarikoptic: so the same automagic conversion could have been done in `inIgnoreIPList` (which would be at least the 4th place of the `x = x if isinstance(x, ...) else X(x)` hopefully warranting a helper function ;) e.g. `asip` or alike ;)).  That would avoid all the IPAddr'ing here, making diff smaller, less duplication, etc

config/action.d/iptables-allports.conf
koeppea: I think just to be more obvious.
Here and there you have to care the naming, e.g. when using *ipset*. 
I don't have a strong opinion on this. If you like, I can omit the `6' at the end where it isn't necessary to be distinguished.

yarikoptic: it is just that without custom suffix you can note that both of the invocations are identical, and only command is different. so we are arriving to some elderly discussion on may be possibly somehow allowing for a single specification ... see e.g. https://github.com/fail2ban/fail2ban/commit/28b44d2655abdf3768b30751a581def2f5f4b5a5 (referenced from https://github.com/fail2ban/fail2ban/issues/39#issuecomment-4324711) .  Just wonder if you have thoughts on that?

koeppea: Understood. I will omit it where ´possible.
However I agree that it's desireable having a general way of handling this.
But for such a cross-plattform software like fail2ban, it's heavily dependent on the plattform specifics being used to enforce the blocking. E.g.

1. Linux iptables needs dedicated start/stop/ban and unban action commands
2. Linux iptables with ipset currently as well but set of type inet may really become protocol independent so that only the start/stop commands would be come protocol dependent.
3. For iptables, not only the binary name differs but also protocol dependent things like the blocktype
4. hostsdeny is completely protocol independent (no dedicated ban/unban commands)
5. route as well
6. PF on *BSD is also completely protocol independent and works like iptables with ipset
7. ...

So your approach is very challenging and I'm not sure if we can reach this right away. I personally would concentrate on introducing v6 support at all.

What happens at the moment when a real v6 address is captured from the logs: the ban/unban commands just return with a unsuccessful return code.

So any thing we do in this matter is a improvement and may not be perfect and leave room for further improvements. For the KISS principal, I see such a enhancement as such a future improvement. Another approach to achieve this could also be a script that offloads this logic from fail2ban instead of calling the actual commands directly.

koeppea:
I think I've now catched all the explicit IPAddr conversions by the decorator from @yarikoptic so we hopefully reduced the impact of the pull request.
Lets see what Travis and Coverage, ... says to it....

If it's fine so far, the next step would be to get coverage reduction back to 0%

fail2ban/ipaddr.py
```diff
@@ -0,0 +1,294 @@
...
+
+from functools import wraps
+
+##
+# Helper functions / decorator
+# Thanks to Yaroslav Halchenko (yarikoptic)
+#
+def asip(ip):
+	"""A little helper to guarantee ip being an IPAddr instance"""
+	return ip if isinstance(ip, IPAddr) or ip == None else IPAddr(ip)
```
yarikoptic: Have I done that? It is a good practice in python to 'is None' and not '== None' since None must be a singleton construct

fail2ban/ipaddr.py
```diff
@@ -0,0 +1,294 @@
...
+
+			return "%s.%s" % (reversed_ip, suffix)
+			
+		else:
+			return ""
+
+	def isIPv4(self):
+		""" return true if the IP object is of address family AF_INET
+		"""
+		return True if self.family == socket.AF_INET else False
```
yarikoptic: Just return the result of comparison

yarikoptic:
@sebres what do you think? I might give it a pass to make some pieces more pythonic flavor but overall...?

fail2ban/ipaddr.py
koeppea: Thanks for the advise. Will take care of it....

fail2ban/ipaddr.py
koeppea: you're right. That's better.

fail2ban/ipaddr.py
koeppea: BTW. That was me as a result of tests adaption, where "getBansMerged" is been called w/o any arguments, but the decorator has expected at least one madatory.

koeppea:
This is strange.
The test to fetch the ASN for 10.0.0.0 fails because a reply comes back actually:
```
$ dig TXT 0.0.0.10.origin.asn.cymru.com. @8.8.8.8

; <<>> DiG 9.9.5-9+deb8u6-Debian <<>> TXT 0.0.0.10.origin.asn.cymru.com. @8.8.8.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42582
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;0.0.0.10.origin.asn.cymru.com.	IN	TXT

;; ANSWER SECTION:
0.0.0.10.origin.asn.cymru.com. 14399 IN	TXT	"4565 | 10.0.0.0/8 |  | other |"

;; Query time: 137 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Sat Apr 23 20:20:16 CEST 2016
;; MSG SIZE  rcvd: 101
```
I think I'll get in touch with cymru.com to check.

yarikoptic:
Fwiw Failing test issue was fixed by #1440

Aeyoun:
I’ve tested the branch on a production server with systemd journal. It seems to work very well. The only thing I had to adjust was the `<HOST>` substitution in `failregex.py`. @koeppea, this needs to be updated as well to work with all the default filters.

This will work for domains and IPv4, and add support for IPv6 with the least amount of modifications to the original regex:

    regex = regex.replace("<HOST>", "(?:::f{4,6}:)?(?P<host>[\w\-.:_]*\w)")

(I just added a colon to the charter group at the end.) This extra could have unintended consequences but I can’t find a concrete example. The original regex is kind of horrible to begin with so I’m not sure if I’m making things worse or better. I’d say this should be sorted out before integrating this branch as every default filter relies on the `<HOST>` substitution mechanism.

PS: I’ll contribute a separate merge request with updated firewalld actions after this request is merged. :wink:

yarikoptic:
On Thu, 28 Apr 2016, Daniel Aleksandersen wrote:

> I’ve tested the branch on a production server with systemd journal. It seems to work very well. The only thing I had to adjust was the `<HOST>` substitution in `failregex.py`. @koeppea, this needs to be updated as well to work with all the default filters.

> This will work for domains and IPv4, and add support for IPv6 with the least amount of modifications to the original regex:

>     regex = regex.replace("<HOST>", "(?:::f{4,6}:)?(?P<host>[\w\-.:_]*\w)")

> (I just added a colon to the charter group at the end.) This extra could have unintended consequences but I can’t find a concrete example. The original regex is kind of horrible to begin with so I’m not sure if I’m making things worse or better. I’d say this should be sorted out before integrating this branch as every default filter relies on the `<HOST>` substitution mechanism.

THANK YOU for testing and the feedback!  We will look into HOST issue.

> PS: I’ll contribute a separate merge request with updated firewalld actions after this request is merged. :wink: 

no need to wait -- just send a PR against koeppea:ipv6-support so things
do not get lost/forgotten

Cheers!

sebres:
So if we speak about HOST regex, I had already something better:
```diff
-regex = regex.replace("<HOST>", "(?:::f{4,6}:)?(?P<host>[\w\-.^_]*\w)")
+regex = regex.replace("<HOST>", r"""(?:::f{4,6}:)?(?P<host>(?:\d+\.){3}\d+|[0-9a-fA-F:]+|[\w\-.^_]*\w)""")
```

More interesting would be another one, but it grasps several logic of the failregex...

The difference and the idea I followed, will be understandable here:

```python
test=("1.2.3.4", "192.0.2.255", "::1", "1::", "64:ff9b::1ab:2cd:3ef:467", "github.com")
[re.search(
  r"""(?:::f{4,6}:)?(?P<host>(?:\d+\.){3}\d+|[0-9a-fA-F:]+|[\w\-.^_]*\w)""", 
  s
).groupdict() for s in test]
[
  {'host': '1.2.3.4'}, 
  {'host': '192.0.2.255'}, 
  {'host': '::1'}, 
  {'host': '1::'}, 
  {'host': '64:ff9b::1ab:2cd:3ef:467'}, 
  {'host': 'github.com'}
]
```

```python
test=("1.2.3.4", "192.0.2.255", "::1", "1::", "64:ff9b::1ab:2cd:3ef:467", "github.com")
[re.search(
  r"""(?:::f{4,6}:)?(?:(?P<IPv4>(?:\d+\.){3}\d+)|(?P<IPv6>[0-9a-fA-F:]+)|(?P<HOST>[\w\-.^_]*\w))""",
  s
).groupdict() for s in test]
[
  {'HOST': None, 'IPv4': '1.2.3.4', 'IPv6': None}, 
  {'HOST': None, 'IPv4': '192.0.2.255', 'IPv6': None}, 
  {'HOST': None, 'IPv4': None, 'IPv6': '::1'}, 
  {'HOST': None, 'IPv4': None, 'IPv6': '1::'}, 
  {'HOST': None, 'IPv4': None, 'IPv6': '64:ff9b::1ab:2cd:3ef:467'}, 
  {'HOST': 'github.com', 'IPv4': None, 'IPv6': None}
]
```
Pro would be a differentiation of Host, IPv4 and IPv6 direct after finding of the hits...

sebres:
Additionally because of standard spelling of IPv6 (with port) inside bracket, I think we should add this regex also.

Example:
```
#IPv4:
127.0.0.1:55555
#IPv6:
[a:b:c:d::1]:55555
```
to bypass the necessarily changing of many lines in filter definitions...

```python
test=(
  "http://1.2.3.4:55555", "http://192.0.2.255:55555", 
  "http://[::1]:55555", "http://[1::]:55555", "http://[64:ff9b::1ab:2cd:3ef:467]:55555", "http://[a:b:c:d::1]:55555", 
  "http://github.com:55555"
)
regex = "^http://<HOST>:\d+$"
regex = regex.replace("<HOST>", 
  r"""(?:::f{4,6}:)?(?:(?P<IPv4>(?:\d+\.){3}\d+)|\[?(?P<IPv6>[0-9a-fA-F:]+)\]?|(?P<HOST>[\w\-.^_]*\w))"""
)
regex = re.compile(regex)
[regex.search(s).groupdict() for s in test]
```

```
[
  {'HOST': None, 'IPv4': '1.2.3.4', 'IPv6': None}, 
  {'HOST': None, 'IPv4': '192.0.2.255', 'IPv6': None}, 
  {'HOST': None, 'IPv4': None, 'IPv6': '::1'}, 
  {'HOST': None, 'IPv4': None, 'IPv6': '1::'}, 
  {'HOST': None, 'IPv4': None, 'IPv6': '64:ff9b::1ab:2cd:3ef:467'}, 
  {'HOST': None, 'IPv4': None, 'IPv6': 'a:b:c:d::1'}, 
  {'HOST': 'github.com', 'IPv4': None, 'IPv6': None}
]
```

ATM no idea how I can made this with single group `host`...

sebres:
Ah yeah, of course the prefix `::ffff:` for the IPv4 mapped addresses can be moved to the IPv4 "group".

But I really don't know if for example the mapped address `::ffff:1.2.3.4` should be banned and in reality only IPv4 `1.2.3.4` will be prohibited via iptables etc., it will be then really blocked as mapped IPv6 (I mean in all actions we use)?

BTW. Someone know why `::ffff:` will be to `(?:::f{4,6}:)?` and not to `(?:::f{4}:)?`? I mean, the mapped has exactly 4 and can't have 5 or 6...

Aeyoun:
@yarikoptic, I’ll keep the firewalld stuff separate because I don’t want to slow this merge request. After sorting out the `<HOST>` replacement replacement, I recommend merging this (assuming it meet the project’s quality criteria). This branch has already lived for over a month, and @koeppea and everyone else would probably like to see it merged sooner rather than later.

yarikoptic:
On Thu, 28 Apr 2016, Daniel Aleksandersen wrote:

>    @yarikoptic, Ia**ll keep the firewalld stuff separate because I dona**t
>    want to slow this merge request. After sorting out the <HOST> replacement
>    replacement, I recommend merging this (assuming it meet the projecta**s
>    quality criteria). This branch has already lived for over a month, and

living in a PR is one thing... living in a wild on the servers is another ;)

but yeah -- now sooner than later it would initiate 0.9 branch  to
continue the stable line

and then master would be aiming for the shiny 0.10.0 release with
IPv6 support and other major enhancements (e.g. #1346).

-- 
Yaroslav O. Halchenko

sebres:
I'm for the new another "master" branch (and for #1346 I'll not say "no":)) despite that I will merge this here very very long.

yarikoptic:
anyways this branch needs a bit more work (further "Pythonization", <HOST>, ...) before you should bother with the merge ;)

sebres:
I know, I made already a short review (no time to make the results public) and have a small branch above, that "fixes" a lot (the branch grows still constantly :))).
Must push to github, will try to do that next week (mon, tue)

yarikoptic:
coolio, iirc there were also some TODOs you wanted TODO before merging it ;)

as for HOST, while at it, shouldn't we `\d+` to `\d{1,3}` and similarly somewhat more restrictive for IPv6: `[0-9a-fA-F:]+` -> `(?:[0-9a-fA-F]{,4}:){,7}[0-9a-fA-F]{,4}` or alike?  There is actually a bunch of sample regexes online, e.g. `(?<![:.\w])(?:[A-F0-9]{1,4}:){7}[A-F0-9]{1,4}(?![:.\w])` for looking up within the text (src: https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9781449327453/ch08s17.html) which pretty much matches mine just without boundary checks which shouldn't hurt I guess but needs checking at the boundaries of the line I guess

sebres:
Well we want not to exaggerate :)
BTW I'm right away agree with `\d+` to `\d{1,3}`, but `(?:[0-9a-fA-F]{,4}:){,7}[0-9a-fA-F]{,4}` would be wrong, imho...
With my expression i want just to differentiate between types of host, and do it more precise as it is...
But I want to think a little bit about...

sebres:
As test case with correct (longer) expression:
```python
test=(
  "http://1.2.3.4:55555", "http://192.0.2.255:55555", "http://::ffff:5.6.7.8:55555",
  "http://[::1]:55555", "http://[1::]:55555", "http://[64:ff9b::1ab:2cd:3ef:467]:55555", "http://[a:b:c:d::1]:55555", 
  "http://github.com:55555", 
  "http://:test:80 http://test::80 http://abcd::80 http://:1234:80 http://1.1234.1.5:80"
)
regex = "http://<HOST>:\d+"
regex = regex.replace("<HOST>", 
  r"""(?:(?:::f{4,6}:)?(?P<IPv4>(?:\d{1,3}\.){3}\d{1,3})|\[?(?P<IPv6>(?:[0-9a-fA-F]{1,4}::?|::){1,7}(?:[0-9a-fA-F]{1,4}|(?<=:):))\]?|(?P<HOST>[\w\-.^_]*\w))"""
)
regex = re.compile(regex)
for s in test:
  print("\"%s\"\n  %s" % (s, {k:v for k,v in regex.search(s).groupdict().items() if v is not None}))
```

```python
"http://1.2.3.4:55555"
  {'IPv4': '1.2.3.4'}
"http://192.0.2.255:55555"
  {'IPv4': '192.0.2.255'}
"http://::ffff:5.6.7.8:55555"
  {'IPv4': '5.6.7.8'}
"http://[::1]:55555"
  {'IPv6': '::1'}
"http://[1::]:55555"
  {'IPv6': '1::'}
"http://[64:ff9b::1ab:2cd:3ef:467]:55555"
  {'IPv6': '64:ff9b::1ab:2cd:3ef:467'}
"http://[a:b:c:d::1]:55555"
  {'IPv6': 'a:b:c:d::1'}
"http://github.com:55555"
  {'HOST': 'github.com'}
"http://:test:80 http://test::80 http://abcd::80 http://:1234:80 http://1.1234.1.5:80"
  {'HOST': '1.1234.1.5'}
```
The last one contains intentionally all wrong, except the last legal HOST (not IP) entry.

koeppea:
Thanks for the testings and feedback.
Of course the regex may be updated as well. Especially for the special notation `[a:b:1:c::d]:8080`.

> Pro would be a differentiation of Host, IPv4 and IPv6 direct after finding of the hits...

However there is no need to enforce the distinguishment using regex as it becomes pretty much  complex and hence easily prone to error. 
In my opinion, the regex should match the string where the host is expected in the log file while being as simple as possible. The challenge is to separate the optional port from the IP address. The IPAddr class takes then care of the distinguishment between v4,v6 or being a hostname. This can potentially as well handled in the `__init__` method of the class. This way the regex just need to match whatever is expected as the host without caring of the exact details.

koeppea:
> Fwiw Failing test issue was fixed by #1400

Thanks  @yarikoptic 
Seems that team-cymru didn't react on my email I've sent them one week ago. But your solution is smart way around.

sebres:
I want to continue my review and have found that my branch is a "little bit" outdated.
Will nevertheless try to make it ready...
Short feedback: My greatest trouble with this PR was the overcomplication in configs and unnecessarily duplication of all action-parameters... 

So as an appetizer hier is an example for 2 configs, that (hopefully) describe, how I will rewrite the config readers for IPv6...

https://github.com/sebres/fail2ban/compare/ipv6-support...sebres:ipv6

- new section [Init6] that holds a set of parameters for IPv6;
- all config files except commons are reverted back to the master (see for example `config/action.d/iptables.conf`);
- the code inside jail, action, etc. is smaller (no duplication) and that's why easy to read;
- I'll create new test cases to check the parameters for start/stop, ban/unban are the same;

yarikoptic:
>    Short feedback: My greatest trouble with this PR was the overcomplication
>    in configs and unnecessarily duplication of all action-parameters...
 
that was exactly  my concern as well.  But I thought to do it one step 
at a time, i.e. 

step 1 -- finalize it in current shape of configs .  What was left is to RF ipaddr to use subclassing instead of all of `if ip.family`.. I believe we have discussed it with @koeppea in quick private emails so there is no public record atm.  But in large it should touch only ipaddr.py so shouldn't conflict much with e.g. step 2

step 2 -- bring back question of providing support for flexible  "family-dependent" specification.   Previous old ideas could be seen at https://github.com/fail2ban/fail2ban/issues/39#issuecomment-118543134 .  Your proposed specs/approach @sebres sounds feasible but to avoid duplication between Init and Init6 may be there should be 3: e.g.  `Init` (common for both), `Init:ipv=4` and `Init:ipv=6`?  why such confounded section name -- may be eventually we could expand support for Init sections based on values for other tags?

sebres:
I would not separate step1 and step2 because of possible "release" and then possible backwards incompatibility between configs, etc.
> e.g. Init (common for both) ...

Yes it was also my plan and so implemented, but I found currently no usage for Init4 (resp. the changes made for a IPv6 so are much better visible)

> based on values for other tags

The idea is good, and I can do that without giant overhead (exactly it would be a simple rule changing all around action reader, if that already contains my other customization);

koeppea:
I'll try to implement your suggestions; I find both make sense.
But I'm currently concentrating getting to a state where all checks pass. Then moving further on.

sebres:
I'll do it :) The PR was self-assigned to me... (I'm possibly the only one who is currently able to fully disassemble in the head our config readers functionality, believe me it not so easy ;)

BTW - JFYI: I've found a bug in current implementation of ban/unban of IPv6 (I think somewhere close by actioncheck, etc.) as I wrote my new test cases - see an excerpt of executed commands in my test cases (start, ban-unban ipv4, ipv6, stop) below.
This is exactly the kind of possible errors I mean with "overcomplication and unnecessarily duplication of params"...
```
 2016-05-02 21:21:45,946 140286590965568 exec-cmd: `# === start ===`
 2016-05-02 21:21:45,947 140286590965568 exec-cmd: `iptables -w -N f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,948 140286590965568 exec-cmd: `iptables -w -A f2b-j-w-iptables-mp -j RETURN`
 2016-05-02 21:21:45,948 140286590965568 exec-cmd: `iptables -w -I INPUT -p tcp -m multiport --dports http,https -j f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,948 140286590965568 exec-cmd: `ip6tables -w -N f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,948 140286590965568 exec-cmd: `ip6tables -w -A f2b-j-w-iptables-mp -j RETURN`
 2016-05-02 21:21:45,949 140286590965568 exec-cmd: `ip6tables -w -I INPUT -p tcp -m multiport --dports http,https -j f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,949 140286590965568 exec-cmd: `# === ban ipv4 ===`
 2016-05-02 21:21:45,949 140286590965568 exec-cmd: `iptables -w -n -L INPUT | grep -q 'f2b-j-w-iptables-mp[ \t]'`
 2016-05-02 21:21:45,950 140286590965568 exec-cmd: `iptables -w -I f2b-j-w-iptables-mp 1 -s 192.0.2.1 -j REJECT --reject-with icmp-port-unreachable`
 2016-05-02 21:21:45,950 140286590965568 exec-cmd: `# === unban ipv4 ===`
 2016-05-02 21:21:45,951 140286590965568 exec-cmd: `iptables -w -n -L INPUT | grep -q 'f2b-j-w-iptables-mp[ \t]'`
 2016-05-02 21:21:45,951 140286590965568 exec-cmd: `iptables -w -D f2b-j-w-iptables-mp -s 192.0.2.1 -j REJECT --reject-with icmp-port-unreachable`
 2016-05-02 21:21:45,951 140286590965568 exec-cmd: `# === ban ipv6 ===`
 2016-05-02 21:21:45,952 140286590965568 exec-cmd: `iptables -w -n -L INPUT | grep -q 'f2b-j-w-iptables-mp[ \t]'`
 2016-05-02 21:21:45,953 140286590965568 exec-cmd: `ip6tables -w -I f2b-j-w-iptables-mp 1 -s 2001:db8:: -j REJECT --reject-with icmp6-port-unreachable`
 2016-05-02 21:21:45,953 140286590965568 exec-cmd: `# === unban ipv6 ===`
 2016-05-02 21:21:45,953 140286590965568 exec-cmd: `iptables -w -n -L INPUT | grep -q 'f2b-j-w-iptables-mp[ \t]'`
 2016-05-02 21:21:45,954 140286590965568 exec-cmd: `ip6tables -w -D f2b-j-w-iptables-mp -s 2001:db8:: -j REJECT --reject-with icmp6-port-unreachable`
 2016-05-02 21:21:45,954 140286590965568 exec-cmd: `# === stop ===`
 2016-05-02 21:21:45,955 140286590965568 exec-cmd: `iptables -w -D INPUT -p tcp -m multiport --dports http,https -j f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,955 140286590965568 exec-cmd: `iptables -w -F f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,955 140286590965568 exec-cmd: `iptables -w -X f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,955 140286590965568 exec-cmd: `ip6tables -w -D INPUT -p tcp -m multiport --dports http,https -j f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,956 140286590965568 exec-cmd: `ip6tables -w -F f2b-j-w-iptables-mp`
 2016-05-02 21:21:45,956 140286590965568 exec-cmd: `ip6tables -w -X f2b-j-w-iptables-mp`
```

I believe very strongly the lines between `# === ban ipv6 ===` and  `# === stop ===` should contain only `ip6tables` (currently contains `iptables`)...

But as I already said - I'll do it (and repair also)... This was only a clue for some people to the subject "production ready"... ;)))

sebres:
So I will preliminary commit into fail2ban:ipv6-support (just commited a bug fix and test cases):
  https://github.com/fail2ban/fail2ban/compare/ipv6-support

For comparison to the PR see also:
  https://github.com/koeppea/fail2ban/compare/ipv6-support...fail2ban:ipv6-support

@yarikoptic, @fail2ban/developers everyone having permissions to push in upstream-origin welcome to review resp. develop with me, but to hold the branch simple, please don't merge - always rebase your commits above if someone other has pushed in the meantime;

everyone without permissions - please review, propose PR resp. changeset to me or @yarikoptic

Short clue: I will rewrite the configs, so please still avoid the changes in config-files

All the discussions about current development should be made here (in this PR)

sebres:
Current TODO for me:
- [x] BF and test coverage for ipv6 actions: wrong actioncheck (IPv4) for IPv6 used;
- [ ] BF: rollback the addIgnoreIP logic with wrapping to IP, because it can contains really HOSTs, and in case of something like dyndns, the IPs can change;
- [ ] remove unnecessarily code aggravation with explicit converting with ntoa, str to IPAddr, etc. Make it implicit (resp. in the function self, not in argument);
- [x] configs rollback (almost exclusively over includes);

yarikoptic:
there must be a PR with proposed changes so tests are excercised etc.  
Just to make sure -- your branch incorporates all changes within this PR, so we could replace this one with the new from your branch?  then indeed subsequent PRs could be sent against it.

NB not sure what strict "rebase" policy should be necessary besides only if you want to rebase your branch against anything else at some point

sebres:
So I'm preliminarily ready - preliminarily, because I've developed it relative my featured branch (#1346) for several reasons:
- because the changes are very extensive, there should always pass all the test cases (and it is extrem fast here in the comparison to current master)
- I can exactly execute targeted one test case, in the comparison to current master, where always the whole class will be tested;
- it is so easier to merge (for me);

If someone want to test it - here you are (sebres:ipv6-support-rl1346-perf716):
https://github.com/sebres/fail2ban/compare/f2b-perfom-prepare-716...sebres:ipv6-support-rl1346-perf716

To test it fast, just use:
```bash
# fast
?PYTHONPATH=.? bin/fail2ban-testcases -f
# fast / no network
?PYTHONPATH=.? bin/fail2ban-testcases -fn
```

I will rebase this relative current master tomorrow (unfortunately it will be not so easy), and will make then a new PR as replacement for this one.

@yarikoptic I could like to refrain if we accept #1346 :)) and I will implement then hereafter my todo-list from there (I promise :)))

yarikoptic:
What if, instead of making master a 'bleeding edge' with #1346 and IPv6 we establish '0.10' branch which will aim for 0.10.0 release with your original and IPv6 support changes?
And then we keep a 'staging' PR which would aim for "merge" of "0.10" into master (well, should be just a fast forward if you agree for no rebasing for that piece;)), although meanwhile we could as well keep releasing both 0.9.x series with stable fixes and 0.10.x with new and exciting improvements, eventually finally making it the master and continuing bug fixes for 0.9.x may be just for a bit in its dedicated 0.9 branch which we establish later.  This way, if someone introduces (via PR) a fix against master, we would be in better chance to have it in both stable (currently master) and development 0.10.

To that end: pushed 0.10 branch with your current state of #1346, so you can close it.  Opened a "staging" PR: https://github.com/fail2ban/fail2ban/pull/1410 replacing 1346 for a merge from 0.10 branch into master.  So now we just need IPv6 feature branch to be PRed against 0.10 branch, not the master

Aeyoun:
Isn’t it more common to make release branches and keep master bleeding edge? Like, exactly the reverse of what you’re suggesting @yarikoptic. You can then still fix issues on the stale but stable release branch whilst developing on master until at some point deciding that master is “good enough” and branch that off to become the new stable branch. Entirely opinion based, of course; but this is how I’ve seen it done pretty much everywhere.

koeppea:
I can do the rebase against 0.10. Branch.

sebres:
@koeppea the rebase of what? 
if you mean my current branch, it is already over 0.10.
or you mean instead of 0.10?

yarikoptic:
@Aeyoun workflows vary widely... in our case we have many small fixes and enhancements coming about from a large number of contributors who are otherwise not active in the project.  Merging them into 'master' being a development/bleeding edge and then picking them manually one by one into 'stable' would be imho much more pain than merging them all into master (given they are all easy to verify them not breaking existing stuff) and then once in a while merging master into 'bleeding edge', until the point when 'bleeding edge' can simply become the stablish master.  I also saw many projects with per-release branches, where it was later really difficult to figure out relation of the master to any of those releases -- can't `git describe` reasonably etc.  Quite often I also saw some feature/bug fixes branches being missed to be picked up into the stable branch solely because, once again, it is yet additional step to carry out.

koeppea:
> @koeppea the rebase of what?

@sebres I mean, that this PR is rebased over master. Not 0.10. I meant that I can rebase this PR over 0.10 instead.

Honestly, at the moment I lost a bit of sight what's actually going to be merged.
Haven't had yet the time to look at your branch.

koeppea:
Had a quick glance. Will focus on testing your branch now.

yarikoptic:
Just to make sure -- 0.10 now carries IPv6 implementation and all additional changes should go through PRs submitted against 0.10 branch.  It seems to incorporate all the changes from this PR, so I will close this one.

koeppea:
ACK

PR #1376, opened 2016-03-29
Added ZoneMinder filter

j-marz:
Added filter to ban failed HTTP and HTTPS logins for ZoneMinder.
Tested on ZoneMinder v1.29.0

### DIFF
--- a/config/filter.d/zoneminder.conf
+++ b/config/filter.d/zoneminder.conf
@@ -3,7 +3,7 @@
 
 [Definition]
 
-# pattern: [client 10.1.1.1:38022] WAR [Login denied for user "test"], referer: https://zoneminderurl/
+# pattern: [Wed Apr 27 23:12:07.736196 2016] [:error] [pid 2460] [client 10.1.1.1:47296] WAR [Login denied for user "test"], referer: https://zoneminderurl/index.php
 #
 #
 # Option:  failregex
@@ -13,7 +13,7 @@
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
 
 
-failregex = ^.*? [[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user \S*], referer: \S*$
+failregex = ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]
 
 ignoreregex =
 
--- a/config/filter.d/zoneminder.conf
+++ b/config/filter.d/zoneminder.conf
@@ -1,6 +1,5 @@
 # Fail2Ban filter for Zoneminder login failures
 
-
 [Definition]
 
 # pattern: [Wed Apr 27 23:12:07.736196 2016] [:error] [pid 2460] [client 10.1.1.1:47296] WAR [Login denied for user "test"], referer: https://zoneminderurl/index.php
@@ -11,7 +10,6 @@
 #          host must be matched by a group named "host". The tag "<HOST>" can
 #          be used for standard IP/hostname matching and is only an alias for
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
-
 
 failregex = ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]
 
--- a/config/filter.d/zoneminder.conf
+++ b/config/filter.d/zoneminder.conf
@@ -12,6 +12,7 @@
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
 
 failregex = ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]
+	    ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\], referer: \S*$
 
 ignoreregex =
 
--- a/config/filter.d/zoneminder.conf
+++ b/config/filter.d/zoneminder.conf
@@ -3,7 +3,7 @@
 
 [Definition]
 
-# patern :      [client 10.1.1.1:38022] WAR [Login denied for user "test"], referer: https://zoneminderurl/
+# pattern: [client 10.1.1.1:38022] WAR [Login denied for user "test"], referer: https://zoneminderurl/
 #
 #
 # Option:  failregex
@@ -13,7 +13,7 @@
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
 
 
-failregex = [[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user
+failregex = ^[[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user \S*], referer: \S*$
 
 ignoreregex =
 
--- a/config/filter.d/zoneminder.conf
+++ b/config/filter.d/zoneminder.conf
@@ -1,5 +1,5 @@
 # Fail2Ban filter for Zoneminder login failures
-#
+
 
 [Definition]
 
--- a/config/filter.d/zoneminder.conf
+++ b/config/filter.d/zoneminder.conf
@@ -13,7 +13,7 @@
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
 
 
-failregex = ^[[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user \S*], referer: \S*$
+failregex = ^.*? [[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user \S*], referer: \S*$
 
 ignoreregex =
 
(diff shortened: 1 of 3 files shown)


config/jail.conf
```diff
@@ -839,3 +839,14 @@ logencoding = utf-8
...
 # maxretry and findtime.
 logpath  = /var/log/haproxy.log
+
+
+[zoneminder]
+# Zoneminder HTTP/HTTPS web interface auth
+# Logs auth failures to apache2 error log
+enabled = false
+port    = http,https
+filter  = zoneminder
```
yarikoptic: no need for filter entry if name matches section name (which it does)

config/jail.conf
```diff
@@ -839,3 +839,14 @@ logencoding = utf-8
...
+logpath = /var/log/apache*/*error.log
```
yarikoptic: use `%(apache_error_log)s` instead so could be overloaded in a single place per distribution

config/filter.d/zoneminder.conf
```diff
@@ -0,0 +1,23 @@
...
+#
+#
+# Option:  failregex
+# Notes.:  regex to match the password failure messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
+#          be used for standard IP/hostname matching and is only an alias for
+#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
+
+
+failregex = [[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user
```
yarikoptic: needs to be anchored, ideally on both sides (leading ^ and trailing $), e.g. for the end 
`\], referer: \S*$`

config/filter.d/zoneminder.conf
```diff
@@ -0,0 +1,23 @@
+# Fail2Ban filter for Zoneminder login failures
+#
+
+[Definition]
+
+# patern :      [client 10.1.1.1:38022] WAR [Login denied for user "test"], referer: https://zoneminderurl/
```
yarikoptic: minor spelling  pat**t**ern

config/jail.conf
j-marz: Fixed

config/jail.conf
```diff
@@ -839,3 +839,13 @@ logencoding = utf-8
 # See "haproxy-http-auth" filter for a brief cautionary note when setting
 # maxretry and findtime.
 logpath  = /var/log/haproxy.log
+
+
+[zoneminder]
+# Zoneminder HTTP/HTTPS web interface auth
+# Logs auth failures to apache2 error log
+enabled = false
```
yarikoptic: once again -- no need for 'enabled=false' -- just see the other jail definitions

config/jail.conf
```diff
@@ -839,3 +839,13 @@ logencoding = utf-8
...
+port    = http,https
+logpath = %(apache_error_log)s
+maxretry = 3
```
yarikoptic: neither this one is so special to warrant a custom maxretry

config/filter.d/zoneminder.conf
```diff
@@ -0,0 +1,23 @@
...
+#
+#
+# Option:  failregex
+# Notes.:  regex to match the password failure messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
+#          be used for standard IP/hostname matching and is only an alias for
+#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
+
+
+failregex = ^.*? [[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user \S*], referer: \S*$
```
yarikoptic: and paranoid me loves avoiding '.*' greedy or not.  What about just replacing it with smth like
`^\[\S*\] (?:[:error] )?(?:[pid \d*] )?`
(initial beast is probably needed to absorb the beginning of the line with date stripped out)

config/jail.conf
j-marz: I will remove

config/jail.conf
j-marz: I will remove

config/filter.d/zoneminder.conf
j-marz: What is the benefit?

sebres: Because `.*?` is very "evil" at all (anchored or not). Please prevent usage of catch-all's (also if it would be greedy) if not really expected.
The benefit is:
- it will match exactly that (resp. precisely) and not any sub string inside log message.
- It finds matched entry or bypasses non-matched much faster (because the regular expressions with non-greedy catch-all's are slowest, and the bigger the message the slower it does searching);

So I'm with @yarikoptic and I will recommend following one:
```diff
-failregex = ^.*? [[]client <HOST>:\d\d\d\d\d] WAR [[]Login denied for user \S*], referer: \S*$
+failregex = "^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]"
```
I've removed referer from regexp, because it doesn't matter (possible miss if empty?)

j-marz: Thanks for explaining and good point about the referer.
I've tried your suggested regex, but it doesn't work... Here is an example log entry to test against.
`[Wed Apr 27 23:12:07.736196 2016] [:error] [pid 2460] [client 10.1.1.1:47296] WAR [Login denied for user "test"], referer: https://zoneminderurl/index.php`
Issue appears to be with the timestamp `^\[\]`
Do you want to capture the timestamp as well?

sebres: Sorry my error (enclosing in quotes is not good for the config, I've corrected above). Here is a test in bash:
```bash
test='[Wed Apr 27 23:12:07.736196 2016] [:error] [pid 2460] [client 10.1.1.1:47296] WAR [Login denied for user "test"], referer: https://zoneminderurl/index.php'
echo $test

fail2ban-regex "$test" '^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]'

Running tests
=============

Use   failregex line : ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <H...
Use      single line : [Wed Apr 27 23:12:07.736196 2016] [:error] [pid 24...


Results
=======

Failregex: 1 total
|   1) [1] ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]
`-

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [1] (?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?
`-

Lines: 1 lines, 0 ignored, 1 matched, 0 missed [processed in 0.00 sec]
```

sebres: > Do you want to capture the timestamp as well?

Ah yeah, the fail2ban cuts timestamp completely from the message. I don't like it (because confusingly and this operation is slow - as memcpy) but it is historical, so...

j-marz: Regex has been changed as per your recommendations

j-marz:
Why did tests fail for commit f5f0f73?

sebres:
> Why did tests fail for commit

Is not your bug, fixed within \#1404 
(just amend commit the same, and hereafter push -force (with know changes))

j-marz:
All checks passed. Will this be merged?

config/filter.d/zoneminder.conf
```diff
@@ -0,0 +1,21 @@
...
+# pattern: [Wed Apr 27 23:12:07.736196 2016] [:error] [pid 2460] [client 10.1.1.1:47296] WAR [Login denied for user "test"], referer: https://zoneminderurl/index.php
+#
+#
+# Option:  failregex
+# Notes.:  regex to match the password failure messages in the logfile. The
+#          host must be matched by a group named "host". The tag "<HOST>" can
+#          be used for standard IP/hostname matching and is only an alias for
+#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
+
+failregex = ^\[\](?: \[:error\])?(?: \[pid \d+\])? \[client <HOST>:\d+\] WAR \[Login denied for user "[^"]*"\]
```
yarikoptic: I would have added optional entry to match referer and anchor at the end

j-marz: I removed as per Sebres recommendation.
> I've removed referer from regexp, because it doesn't matter (possible miss if empty?)

j-marz: @yarikoptic I have added the optional referer and anchor in commit a2d088eca76ece95e72bfec086adb442d35dd56d

j-marz:
@yarikoptic @sebres , will this ZoneMinder filter be merged?

config/filter.d/zoneminder.conf
sebres: I'm sure @yarikoptic said as "optional entry" something like this:
```regex
(?:, referer: \S*)$
```
You've added it as second regex (that in addition will never match, because the first one always wins)...
Since the part before `<HOST>` is not variable I don't see any reason for this referer part. IMHO the message `Login denied for user` is precise enough no matter what for data follows this at end of line.

I'll merge it without second regex. Thanks.

sebres:
rebased (conflicts resolved, code review, etc.) and merged with acd9e8155b9ac0e40e20d5ca18fb6e2f799d731c

Issue #1372, opened 2016-03-24
Fail2ban 9.4.1 fails with systemd


shanff:
```
Mar 23 21:01:49 wrangler systemd[1]: Starting Fail2Ban Service...
Mar 23 21:01:49 wrangler fail2ban-client[27881]: ERROR  No file(s) found for glob /var/log/auth.log
Mar 23 21:01:49 wrangler fail2ban-client[27881]: ERROR  Failed during configuration: Have not found any log file for sshd jail
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Control process exited, code=exited status=255
Mar 23 21:01:49 wrangler systemd[1]: Failed to start Fail2Ban Service.
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Unit entered failed state.
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Failed with result 'exit-code'.
Mar 23 21:01:49 wrangler sudo[27874]: pam_unix(sudo:session): session closed for user root
Mar 23 21:01:49 wrangler polkitd[1155]: Unregistered Authentication Agent for unix-process:27875:784907 (system bus name :1.77, object path /org/freedesktop/PolicyKit1/AuthenticationAgent, locale en_US.utf8) (disconnected from bus)
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Service hold-off time over, scheduling restart.
Mar 23 21:01:49 wrangler systemd[1]: Stopped Fail2Ban Service.
Mar 23 21:01:49 wrangler systemd[1]: Starting Fail2Ban Service...
Mar 23 21:01:49 wrangler fail2ban-client[27945]: ERROR  No file(s) found for glob /var/log/auth.log
Mar 23 21:01:49 wrangler fail2ban-client[27945]: ERROR  Failed during configuration: Have not found any log file for sshd jail
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Control process exited, code=exited status=255
Mar 23 21:01:49 wrangler systemd[1]: Failed to start Fail2Ban Service.
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Unit entered failed state.
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Failed with result 'exit-code'.
Mar 23 21:01:49 wrangler systemd[1]: fail2ban.service: Service hold-off time over, scheduling restart.
[...]

Issue #1270, opened 2015-12-12
port is not inherited from [DEFAULT] section in jail.local


wodry:
fail2ban 0.9.2

I copied
/etc/fail2ban/jail.conf
to
/etc/fail2ban/jail.local
There is written:
```
[DEFAULT]
port = 0:65535
```
In the [sshd] section is no port defined, so I would guess the default is inherited. But looking at iptables, I see

-A INPUT -p tcp -m multiport --dports 22 -j f2b-sshd

So I added
port = 0:65535

to the [sshd] section.

Now it works:
-A INPUT -p tcp -m multiport --dports 0:65535 -j f2b-sshd

So the default port definition is not inherited. Please fix code or documentation.

yarikoptic:
Copying entire jail.conf into . local is a sin ;-)

port is defined per jail (this overriding the default), this if you want to change it for sshd have it overridden in sshd section, cheers

wodry:
I think the subject is pretty clear.
I think You did not understand the subject.
I do _not_ want to change the port defined in the [DEFAULT] section.
In the [DEFAULT] section port is defined as 0:65535.
But if I do not define port in the [sshd] Jail (so I should _not_ overwrite the default port definition), only port 22 is blocked.

sebres:
The subject is pretty clear.
I think You did not understand the fail2ban works (resp. reads its config files).
Firstly it will always load `x.conf` and substitutes it using `[default]` from `x.conf`.
Hereafter it loads `x.local`, substitutes it exact the same way and only then overwrites entries from `x.conf`.
In [sshd] of `jail.conf` we have `port=ssh`, so no matter which port you will have in `[default]` of your `jail.local`.
Secondly, if you really want to block all ports, you can use `banaction = %(banaction_allports)s` in your local changeset for sshd jail (or/and you can just enable jail `pam-generic`).

wodry:
I think values defined in the [DEFAULT] section of jail.local should overwrite the values in the [DEFAULT] section of jail.conf _before_ the other sections of jail.conf are parsed (and afterwards the other sections of jail.local).

Otherwise, as of my understanding now, values in the [DEFAULT] section of jails.local have only impact on jail sections, that are not defined in jail.conf which is pretty confusing, IMHO.

But that would be another issue because the port is needed in sshd jail.local section anyway because it is already set explicit in jail.conf sshd section.

sebres:
[...]

PR #1378, opened 2016-03-30
bug fix: retracts wrong disregarding of defaults from includes like 'jail.d/*.conf' etc. (#1225 was to blame for this)

sebres:
move explicit jail parameters within jail specifications to section defaults (from [jail] to [default.jail]);

logic of config reader extended: introduced per section defaults functionality (section with named convention [default.name]);

Pros:
  - we can precise define defaults for separate sections (for example without define it for all jails);
  - we can define defaults for sections, without that they involve currently not existing section;
  - resolves unclear defaults/locals handling between multiple includes and standard values (for example by issues like #1270);

Cons:
   - atm unknown, because it should be backwards compatibly

closes gh-1372

### DIFF
--- a/man/jail.conf.5
+++ b/man/jail.conf.5
@@ -158,6 +158,9 @@ This sets the age at which bans should be purged from the database.
 
 .SH "JAIL CONFIGURATION FILE(S) (\fIjail.conf\fB)"
 The following options are applicable to any jail. They appear in a section specifying the jail name or in the \fI[DEFAULT]\fR section which defines default values to be used if not specified in the individual section.
+Another way to define default options per jail is to specify those in the section \fI[default.jail-name]\fR.
+The option is looked up in \fI[jail]\fR section, in jail defaults \fI[default.jail]\fR and in global \fI[DEFAULT]\fR section in exactly that order.
+Setting of some options in \fI[DEFAULT]\fR overwrites per section defaults for this options of the previous includes. This applies to all configuration files that file2ban uses by read of jail parameters.
 .TP
 .B filter
 name of the filter -- filename of the filter in /etc/fail2ban/filter.d/ without the .conf/.local extension.
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -376,8 +376,10 @@ class Fail2banClient:
 			logSys.setLevel(logging.WARNING)
 		elif verbose == 2:
 			logSys.setLevel(logging.INFO)
-		else:
+		elif verbose == 3:
 			logSys.setLevel(logging.DEBUG)
+		else:
+			logSys.setLevel(logging.HEAVYDEBUG)
 		# Add the default logging handler to dump to stderr
 		logout = logging.StreamHandler(sys.stderr)
 		# set a format which is simpler for console use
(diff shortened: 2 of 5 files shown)


sebres:
Because of better verbosity (switchable through `-vvv`) in my branch sebres/f2b-perfom-prepare-716-cs it will be possible to show which config file overwrite resp. set which defaults or direct parameter... (currently not ported here)

sebres:
trebly increasing of verbosity of fail2ban-client (`-vvv` up to heavy debug) ported from sebres/f2b-perfom-prepare-716-cs

man/jail.conf.5
```diff
@@ -158,6 +158,9 @@ This sets the age at which bans should be purged from the database.
 
 .SH "JAIL CONFIGURATION FILE(S) (\fIjail.conf\fB)"
 The following options are applicable to any jail. They appear in a section specifying the jail name or in the \fI[DEFAULT]\fR section which defines default values to be used if not specified in the individual section.
+Another way to define default options per jail is to specify those in the section \fI[default.jail-name]\fR.
+The option is looked up in \fI[jail]\fR section, in jail defaults \fI[default.jail]\fR and in global \fI[DEFAULT]\fR section in exactly that order.
```
yarikoptic: you really like hacking @sebres ;)
this imho even further increases difficulty of comprehension of our configuration files.  At least before it was easy and more or less inline with standard Python's treatment of DEFAULT section specifying default values for sections.  Inclusion of additional files provided remedy to allow to override parameters.   Now we would also get this additional `[default.` construct... blows up jail.conf even further... not sure...
going back to issues which arouse recently which lead to this PR
- per jail port overrides (#1270) - well, a single user missed the point of how values are picked up from DEFAULT.  Given that it is quite a rare report (if not the first), I would say majority of users get the idea (when they do customizations) on how DEFAULT settings work.  So I wouldn't try to fix anything to address it anyhow (besides may be even more documentation? ;))
- backends hardcoded to each section (change in #1225 which lead to #1372) thus making it impossible to change backend in DEFAULT to come in effect for all the sections.  Well, as I have suggested in https://github.com/fail2ban/fail2ban/issues/1372#issuecomment-202951440 , imho the most straightforward approach is just to revert #1225 and ask distro maintainers to define those backends for necessary jails only (since otherwise the backend overload in DEFINE) in per-distribution shipped overload file `jail.d/defaults-distro.conf` .  I already do something like that to enable sshd jail out of the box in Debian packages
```
$> cat jail.d/defaults-debian.conf
[sshd]
enabled = true
```

I am just afraid that adding this default.section feature and by increasing complexity of "defaults handling" we would not only make config file longer and harder to digest and customize but possibly overlook possible gotcha... the simpler we keep it all, more robust it would stay.

sebres:
Well I was even not really sure, so what - I will close it... 
Immortalized here if expected at some point.

I will merge only second commit (about verbosity)...

yarikoptic:
>    I will merge only second commit (about verbosity)...

+1 for verbosity boost

sebres:
already done!

sebres:
Better solution for the same handling provided in #1750

PR #1380, opened 2016-04-01
revised exim regexes to cover entries with rDNS

theDogOfPavlov:
The original regexes I added were skipping hosts with rDNS entries, so I fixed them. 

Also added additional tests (this time anonymised).

changelog still stands so not updated.


theDogOfPavlov:
managed to add two pull requests, closed this one.

PR #1381, opened 2016-04-01
Tightened up exim regexes to catch rDNS entries

theDogOfPavlov:
Tightened up exim regexes as some hosts with rDNS entries weren't being picked up.

Added additional tests to cover.

Changelog still stands


yarikoptic:
Looks good to me. thanks. hopefully I wouldn't forget to add some changelog entry later on ;)

theDogOfPavlov:
no longer required...

PR #1390, opened 2016-04-12
ensure /var/run/fail2ban is created in systemd service file

khumarahn:
Hi, this pull request adds one line to the systemd service file, making sure that the folder `/var/run/fail2ban` is created.

I ran into this problem in gentoo, when I installed fail2ban, but it did not start because `/run/fail2ban` did not exist. Gentoo installs bundled tmpfiles.d file, but it takes effect only on reboot, or on manual invocation of `systemd-tmpfiles`. Making sure that the directory is created before the service is run by means of systemd command seems clean and simple.

### DIFF
--- a/files/fail2ban.service
+++ b/files/fail2ban.service
@@ -6,6 +6,7 @@ PartOf=iptables.service firewalld.service
 
 [Service]
 Type=forking
+ExecStartPre=/bin/mkdir -p /var/run/fail2ban
 ExecStart=/usr/bin/fail2ban-client -x start
 ExecStop=/usr/bin/fail2ban-client stop
 ExecReload=/usr/bin/fail2ban-client reload


yarikoptic:
yeah... I have read now about tmpfile.d being a "proper" way to accomplish the mission and wondered what could possibly be the drawback of your direct call to mkdir, I see none. May be someone of @fail2ban/contributors and @fail2ban/developers has stronger opinion? if not -- I will just merge.

mrc0mmand:
I've been using this solution for a while and the only drawback so far is that it won't clean up the folder after service shutdown. Also, the `-p` parameter suppresses all possible errors during path creation, which is not a good thing either. So, you're right, [tmpfiles.d](http://fedoraproject.org/wiki/Packaging:Tmpfiles.d) is a "proper" way, how to solve this issue. 

I guess creating a file `/usr/lib/tmpfiles.d/fail2ban.conf` with something like:
```
d /var/run/fail2ban 755 root root
d /run/fail2ban 755 root root
``` 
should do the same thing as a `mkdir` in a service file. But it's up to you which solution you prefer more - both should accomplish the same thing.

khumarahn:
tmpfiles.d also does not clean up the folder on service shutdown. Also, `/bin/mkdir -p /var/run/fail2ban` would not do anything if such a folder already exists (say, created by tmpfiles.d).

There is an alternative option for systemd service file:
```
RuntimeDirectory=fail2ban
```
this creates a folder `/run/fail2ban`, and deletes it after the service is stopped. I feel that it is a bit less good than the one with mkdir, because mkdir would not interfere with tmpfiles.d, while `RuntimeDirectory` could delete the directory. Also, all the configs for fail2ban are written for `/var/run`, not for `/run`.

yarikoptic:
well, `/var/run` is for historical compatibility... on any modern (and systemd requires relatively modern distro) `var/run` would be a symlink to `/run` so I actually would prefer the native systemd way to define run directory (i.e. RuntimeDirectory) which should accomplish the mission

sebres:
already a long time a part of [0.10](https://github.com/fail2ban/fail2ban/blob/0.10/files/fail2ban.service#L9), so I don't see why it should be not merged into master also...

PR #1391, opened 2016-04-14
Fix postfix/smtps/smtpd matching.

kAworu:
Hi there,

Fail2Ban v0.9.4 has trouble to ban postfix auth attempts via the old (deprecated) smtps port (465). 

After study it seems that the filters doesn't match when `syslog_name=postfix/smtps` which looks like to be the default since 2011 in postfix (see https://github.com/vdukhovni/postfix/commit/9639819fd08c692e480958669b0e00975a86a04f).

Ok?

Tests in the commit.

Could not find a related issue.

Previous test cases still pass.

Can't do smaller.

Ok.

Done.

### DIFF
--- a/config/filter.d/postfix-sasl.conf
+++ b/config/filter.d/postfix-sasl.conf
@@ -7,7 +7,7 @@ before = common.conf
 
 [Definition]
 
-_daemon = postfix(-\w+)?/((?:submission|smtps)/)?smtpd
+_daemon = postfix(-\w+)?/(?:submission/|smtps/)?smtp[ds]
 
 failregex = ^%(__prefix_line)swarning: [-._\w]+\[<HOST>\]: SASL ((?i)LOGIN|PLAIN|(?:CRAM|DIGEST)-MD5) authentication failed(: [ A-Za-z0-9+/:]*={0,2})?\s*$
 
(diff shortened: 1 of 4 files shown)


config/filter.d/postfix-sasl.conf
```diff
@@ -7,7 +7,7 @@ before = common.conf
 
 [Definition]
 
-_daemon = postfix(-\w+)?/(submission/)?smtp(d|s)
+_daemon = postfix(-\w+)?/((?:submission|smtps)/)?smtpd
```
sebres: Are you sure that daemon prefix like `postfix-.../submission/smtps` resp. `postfix-.../smtps` (both with `s` at end) will never occur anymore?
Because if not, it should be:
```diff
-_daemon = postfix(-\w+)?/(submission/)?smtp(d|s)
+_daemon = postfix(-\w+)?/(?:submission/|smtps/)?smtp[ds]
```
Either way I'll prefer this one, to be really backwards compatible...

kAworu: Hi @sebres 

To answer your question: no I am not sure that it will never occur anymore since the daemon name is configurable by the postfix administrator. I can only say that is it ok (and actually a fix) for the default postfix config (which might even get patched by the vendor / distribution / OS).

I'll update the PR with your suggestion :+1:

sebres:
Well full backwards compatible, so merged direct now

kAworu:
@serbes thanks !

PR #1394, opened 2016-04-24
WiP: Replace <HOST> with default host regex

da2x:
Match the regex used in fail2ban/server/failregex.py

### DIFF
--- a/fail2ban/client/fail2banregex.py
+++ b/fail2ban/client/fail2banregex.py
@@ -58,7 +58,7 @@ from ..helpers import FormatterWithTraceBack, getLogger
 logSys = getLogger("fail2ban")
 
 def debuggexURL(sample, regex):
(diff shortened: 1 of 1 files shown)


yarikoptic:
could you please elaborate why? or what for?

Aeyoun:
* Sharing the same regex replacement as used elsewhere in Fail2Ban
* Adds IPv6 support

This regex replacer was probably forgotten when IPv6 support was added to `fail2ban/server/failregex.py`.

Aeyoun:
So, don’t merge this. Leaving it open as it should use the same regex as is decided on in #1374.

yarikoptic:
retitled with WiP prefix (for work in progress == do not merge)

sebres:
Fixed in #1715, because of better solution by general tag replacements (we've more tags as `<HOST>` now)

PR #1395, opened 2016-04-24
Use HTTPS in Debuggex URLs

Aeyoun:


yarikoptic:
why? i.e. is it no longer working with initial http:// link?

Aeyoun:
`http://www.debuggex.com` always redirects to `https://www.debuggex.com`. Pointing users directly to https will save them an extra round trip to he server and speed up their experience. Plus, of course, https is more secure and private than http so the whole web is moving towards https.

buanzo:
Are we checking server certificates when doing HTTPS requests?

PR #1403, opened 2016-04-30
New logtarget: systemd-journal

da2x:
Removed closing of all files as `python-systemd` uses persistent file descriptors that there is no way to identify nor reopen. Logging could be initialized after forking, but that would mean no logging prior to forking as the file descriptors aren’t reopened once they’ve been closed. Alternatively still, `os.closerange(0, maxfd)` could be set to start closing at `4` rather than `0` which is kind of magical and not guaranteed to work in all situations on all systems. I didn’t see any good reasons for why all file descriptors are brutally murdered in the first place. Looks to me like a case of premature optimization that even has caused problems in the past.

Didn’t add tests as I’m not familiar with the test framework and assumed the project probably doesn’t want to add `systemd` to the list of requirements for running tests anyways.

### DIFF
--- a/fail2ban/server/server.py
+++ b/fail2ban/server/server.py
@@ -389,7 +389,10 @@ class Server:
 			self.__loggingLock.acquire()
 			# set a format which is simpler for console use
 			formatter = logging.Formatter("%(asctime)s %(name)-24s[%(process)d]: %(levelname)-7s %(message)s")
-			if target == "SYSLOG":
+			if target == "SYSTEMD-JOURNAL":
+				from systemd.journal import JournalHandler
+				hdlr = JournalHandler(SYSLOG_IDENTIFIER='fail2ban')
+			elif target == "SYSLOG":
 				# Syslog daemons already add date to the message.
 				formatter = logging.Formatter("%(name)s[%(process)d]: %(levelname)s %(message)s")
 				facility = logging.handlers.SysLogHandler.LOG_DAEMON
@@ -438,9 +441,12 @@ class Server:
 					if (2,6,3) <= sys.version_info < (3,) or \
 							(3,2) <= sys.version_info:
 						raise
-			# tell the handler to use this format
-			hdlr.setFormatter(formatter)
+
+			if target != "SYSTEMD-JOURNAL":
+				# tell the handler to use this format
+				hdlr.setFormatter(formatter)
 			logger.addHandler(hdlr)
+
 			# Does not display this message at startup.
 			if not self.__logTarget is None:
 				logSys.info(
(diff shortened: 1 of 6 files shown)


Aeyoun:
Remaining failing test resolved in issue #1404.

fail2ban/server/server.py
```diff
@@ -576,28 +582,6 @@ def __createDaemon(self): # pragma: no cover
...
-			for fd in range(0, maxfd):
-				try:
-					if not os.path.sameopenfile(urandom_fd, fd):
-						os.close(fd)
-				except OSError:   # ERROR (ignore)
-					pass
-			os.close(urandom_fd)
-		else:
-			os.closerange(0, maxfd)
-
```
yarikoptic: how this removal is warranted?

man/fail2ban-client.1
```diff
@@ -87,8 +87,8 @@ gets the logging level
 .TP
 \fBset logtarget <TARGET>\fR
 sets logging target to <TARGET>.
-Can be STDOUT, STDERR, SYSLOG or a
-file
+Can be STDOUT, STDERR, SYSLOG,
+SYSTEMD-JOURNAL or a file
```
yarikoptic:
```shell
  $> head -n1 man/fail2ban-client.1
.\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.47.2.
```
no harm though

fail2ban/server/server.py
Aeyoun: See the PR description.

rumpelsepp:
Whats the advantage over logging to stdout when started via systems?

Aeyoun:
> Whats the advantage over logging to stdout when started via systemd?

Formatting. You get just the message in the log and the journal handles the rest of the info relatively. Meaning you can filter and sort messages by priority and time using journalctl.

sebres:
In our major (0.10th) branch we can already long time log into systemd without usage of python JournalHandler, etc.  See for example in #1618

Prerequisites:
- you've fail2ban 0.10
- fail2ban runs as systemd.service
- fail2ban-server started direct (in foreground, service type is `simple`, not `forking`), see fail2ban.service from #1618

So let the logtarget = STDOUT and fail2ban logging goes in systemd-journal.

Therefore I tend to close this PR, but ... 
@yarikoptic your conclusion?

sebres:
> @yarikoptic your conclusion?

ping?!

polyzen:
This can be closed.

sebres:
> This can be closed.

I don't think so, as for better formatting and because there are issues like #2482.
I'll try to resolve conflicts and merge it later.

Thaodan:
> > This can be closed.
> 
> I don't think so, as for better formatting and because there are issues like #2482. I'll try to resolve conflicts and merge it later.

Are you still planing to do so?

sebres:
Rebased in d6896eb26d81355b3e0eac0da7153f44b2a98c44.
I didn't remove part responsible for closing file descriptors (by forking), since it is unused if fail2ban started from `fail2ban-server` in foreground without daemonization (and forking) at all, see https://github.com/fail2ban/fail2ban/blob/d6896eb26d81355b3e0eac0da7153f44b2a98c44/files/fail2ban.service.in#L11
Tested, it seems to work as expected - thus pushed to master branch now.

PR #1413, opened 2016-05-12
0.10-cs: client-server rewritten as modules / start in foreground / test coverage for cs

### LINKED ISSUE #1121: Move all bin/fail2ban- functionality under fail2ban/cmdline/ module
And possibly rely on setuptools entrypoints mechanism to establish command line "scripts".

See #1107 for initial discussion

@sebres You don't carry substantial changes within bin/fail2ban- scripts do you?

### LINKED ISSUE #1139: Support foreground with fail2ban-server
With the era of new-style init (such as systemd) or daemon managers (such as supervisord), it seems safer to write program that doesn't try to be a daemon itself, since there are a lot of corner cases and pitfalls. Fail2ban seems to do it nicely, however, I see at least two reasons to support a foreground mode:
  * If some platform has some specific initialization pattern, it could be handled by the init/daemon manager.
  * Log management is far easier when the init/daemon manager handles it, usually by fetching the stdout of the launched program. I'm not sure that setting the logtarget to STDOUT with the current daemonizing behavior will be successful.

Is there any good reason not to support a foreground behavior? Or is it a simple historical thing?

sebres:
closes #1121
closes  #1139 
Several enhancements, bug fixed and code optimization:

* starting of the server (and client/server communication behavior during start and daemonize) completely rewritten:

  - client/server functionality moved away from bin and using now the common interface (introduced in fail2bancmdline);
  - start in foreground fixed;
  - server can act as client corresponding command line;
  - command "restart" added: in opposite to "reload" in reality restarts the server (new process);
  - several client/server bugs during starting process fixed.

* possibility to increase verbosity up to heavy debug (command line parameter `-vvv`)
* increase readability and details level by increased verbosity
* reader bug fix: prevent to silent "load" of not existing jail;
* coverage of test cases increased;
* several bug fixed: fork in client-server test cases prohibited, all worker threads daemonized (to prevent hanging on exit).


bin/fail2ban-testcases
```diff
@@ -119,10 +119,14 @@ else:
...
-	stdout.setFormatter(Formatter(' %(asctime)-15s %(thread)s' + fmt))
-else: # pragma: no cover
-	# just prefix with the space
-	stdout.setFormatter(Formatter(fmt))
+	if verbosity > 3:
+		fmt = ' | %(module)15.15s-%(levelno)-2d: %(funcName)-20.20s |' + fmt
+	if verbosity > 2:
+	  fmt = ' +%(relativeCreated)5d %(thread)X %(levelname)-5.5s' + fmt
+	else:
+		fmt = ' %(asctime)-15s %(thread)X %(levelname)-5.5s' + fmt
```
yarikoptic: Seems mixed indentation with spaces and tabs

fail2ban/server/server.py
```diff
@@ -402,8 +434,14 @@ def getLogLevel(self):
...
 	
 	def setLogTarget(self, target):
-		try:
-			self.__loggingLock.acquire()
+		with self.__loggingLock:
+			# don't set new handlers if already the same
+			# or if "INHERITED" (foreground worker of the test cases, to prevent stop logging):
+			if self.__logTarget == target:
+				return True
+			if target == "INHERITED":
```
yarikoptic: I don't think docs were adjusted to reflect having this new setting I am yet to comprehend

yarikoptic:
Why didn't you listen to a wise man? https://github.com/fail2ban/fail2ban/issues/1139#issuecomment-125960786
It would be much easier to review if we first just merged the move files away PR, without any functionality change. Now impossible to see what was moved and what was moved and then changed

fail2ban/tests/utils.py
```diff
@@ -386,6 +412,9 @@ def assertNotLogged(self, *s, **kwargs):
 	def pruneLog(self):
 		self._log.truncate(0)
 
+	def pruneLog(self):
+		self._log.truncate(0)
+
```
yarikoptic: Duplicate

fail2ban/tests/fail2banclienttestcase.py
```diff
@@ -0,0 +1,593 @@
...
+				self.assertTrue(phase.get('end', None))
+				self.assertLogged("Shutdown successful", "Exiting Fail2ban")
+		finally:
+			_kill_srv(tmp)
+			if th:
+				th.join()
+
+	@withtmpdir
+	def testServerFailStart(self, tmp):
+		try:
```
yarikoptic: This massive try finally begs for having it done in eg `@kill_srv_after` or alike

yarikoptic:
So please do first the move, send it as a PR, we merge it after tests pass, and hopefully this diff becomes manageable

fail2ban/server/server.py
sebres: INHERITED is just for the test cases (it tests now starting of the server, and server should log in the same test environment, so possible to capture and analyse it hereafter)

sebres:
> Why didn't you listen to a wise man? #1139 (comment)

Because: 
- it can't be test covered there;
- in single commit I've moved bin/fail2ban-client -\> fail2banclient and bin/fail2ban-server -\> fail2banserver
- I brought together CS in command line (double code -> single code)
- the wise man can try to do it self, if he says it is lighter :))) With almost 30 years of developing behind back am I already wise enough to determine, that it is not so easy :)))

> Now impossible to see what was moved and what was moved and then changed

Wrong, it is possible - just compare:
- old bin/fail2ban-client with new fail2banclient / fail2bancmdline
- old bin/fail2ban-server with new fail2banserver / fail2bancmdline

My git does it automatically (it can find the code lines, moved from other files)

> So please do first the move, send it as a PR, we merge it after tests pass, and hopefully this diff becomes manageable

Move what and where?

bin/fail2ban-testcases
sebres: :+1:

fail2ban/tests/utils.py
sebres: :+1:

fail2ban/tests/fail2banclienttestcase.py
sebres: ??
I can not quite follow what you mean here

yarikoptic:
On Thu, 12 May 2016, Serg G. Brester wrote:

>      Why didn't you listen to a wise man? #1139 (comment)

>    Because:

>      * it can't be test covered there;

I am not arguing against moving altogether, and understand that it
couldn't test covered there

>      * in single commit I've moved bin/fail2ban-client -> fail2banclient and
>        bin/fail2ban-server -> fail2banserver

good!

>      * I brought together CS in command line (double code -> single code)

good -- that could be a good point [lets call it POINTX] to send a
PR at without piling up additional fixes/etc.  Just code move +
refactoring to reduce duplication

>      * the wise man can try to do it self, if he says it is lighter :))) With
>        almost 30 years of developing behind back am I already wise enough to
>        determine, that it is not so easy :)))

not sure what was the difficulty to have a PR for POINTX

>      Now impossible to see what was moved and what was moved and then changed

>    Wrong, it is possible - just compare:

>      * old bin/fail2ban-client with new fail2banclient / fail2bancmdline
>      * old bin/fail2ban-server with new fail2banserver / fail2bancmdline

I meant while reviewing PR on github.   I often do reviews from the
phone so can't easily checkout /compare files otherwise

>    My git does it automatically (it can find the code lines, moved from other
>    files)

>      So please do first the move, send it as a PR, we merge it after tests
>      pass, and hopefully this diff becomes manageable

>    Move what and where?

could there be a PR for POINTX mentioned above?

-- 
Yaroslav O. Halchenko

sebres:
I don't know all the test cases of this commit pass, I'll make a commit 556ddaabd7a0b1f1ac94d014680fa83d980f770f as separated PR.

I can remember a "bug" for 'flushed' vs 'rolled over' in testFlushLogs, was fixed later from me...

fail2ban/tests/fail2banclienttestcase.py
yarikoptic: ok -- https://github.com/sebres/fail2ban/pull/5   where I called it `@with_kill_srv`.

yarikoptic:
btw -- tests fail for me (only) strangely since seems to (rightfully?) need a sample log file which is not in git... confused (don't have time atm to investigate) how it passed on travis:
```
$> bin/fail2ban-testcases testCheckStockJailActions
Fail2ban 0.9.4.dev0 test suite. Python 2.7.11+ (default, Apr 17 2016, 14:00:29) [GCC 5.3.1 20160409]. Please wait...
F
FAIL: testCheckStockJailActions (fail2ban.tests.servertestcase.ServerConfigReaderTests)
  File "./fail2ban/tests/servertestcase.py", line 1046, in testCheckStockJailActions
    self.fail("Command %r has failed. Received %r" % (cmd, e))
AssertionError: Command ['set', 'pass2allow-ftp', 'addlogpath', './fail2ban/tests/files/logs/pass2allow-ftp', 'head'] has failed. Received IOError(2, 'No such file or directory')

Ran 1 test in 0.397s

FAILED (failures=1)

$> ls -l './fail2ban/tests/files/logs/pass2allow-ftp' 
ls: cannot access './fail2ban/tests/files/logs/pass2allow-ftp': No such file or directory
```

sebres:
You have not clean copy - perhabs one custom config `pass2allow-ftp` local (uncommited) and for it no logs. 
testCheckStockJailActions loads all stock filters / action and transmit it to the TestServer...

I can rewrite this test case, that will ignore all configs, that don't have corresponding test log file...

Or just ignore this one...
```
bin/fail2ban-testcases -i testCheckStockJailActions
```

sebres:
Have rabased without commit, that set log level to debug in `.travis.yml`.
In the meantime our logging with this log level is too powerful (many new tests in branch 0.10), so the travis had failed with:
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated
```

sebres:
@yarikoptic I want to merge this next week, just to complete a still one building lot, so I can concentrate to work on #1437, #1460, etc, without to consider compatibility to it, etc. I think we have 2-3 PR also that conflict with this rewritten cs-functionality, so it may be resolved and rebased later...
What do you think?

fail2ban/client/fail2banclient.py
```diff
@@ -0,0 +1,462 @@
...
+			self._alive = False
+			for s, sh in _prev_signals.iteritems():
+				signal.signal(s, sh)
+
+
+##
+# Wonderful visual :)
+#
+
+class _VisualWait:
```
yarikoptic: Some docs would be really appreciated, otherwise I am just guessing what is the intended use for these two

fail2ban/protocol.py
```diff
@@ -26,6 +26,9 @@
 
 import textwrap
 
+def output(s):
+	print(s)
```
yarikoptic: So what is the purpose of having it a function? To ease mocking in the tests or ...? Please describe in the docstring

fail2ban/server/server.py
```diff
@@ -371,16 +407,15 @@ def statusJail(self, name, flavor="basic"):
...
-			self.__loggingLock.release()
+		value = value.upper()
+		with self.__loggingLock:
+			if self.__logLevel == value:
+				return
+			try:
+				getLogger("fail2ban").setLevel(getattr(logging, value))
+				self.__logLevel = value
+			except AttributeError:
+				raise ValueError("Invalid log level")
```
yarikoptic: As the actual invalid value to the exception msg

fail2ban/tests/fail2banclienttestcase.py
```diff
@@ -0,0 +1,598 @@
...
+fail2banserver.output = \
+protocol.output = _test_output
+
+def _test_exit(code=0):
+	logSys.debug("Exit with code %s", code)
+	if code == 0:
+		raise ExitException()
+	else:
+		raise FailExitException()   
+fail2bancmdline.exit = \
```
yarikoptic: Since we don't like folks who later contribute all the styling fixes, let's from the beginning try to stay close to pep8 and sensible spacing? At least a single empty line here... generally advised to have 2 lines at module level and single one within classes (between methods) etc

yarikoptic: note that mocking it this way would then 1. be permanent as soon as this module gets imported, 2. would be dependent on order of tests imports.  That is why in general better to mock specifically in setup/teardown methods.  In general we might eventually start using mock module (standard since 3.4 iirc) -- very powerful and nice

fail2ban/tests/fail2banclienttestcase.py
```diff
@@ -0,0 +1,598 @@
...
+		#logSys.debug('--- interact command: %r', INTERACT[0])
+		return INTERACT.pop(0)
+	else:
+		return "exit" 
+fail2banclient.input_command = _test_input_command
+
+# prevents change logging params, log capturing, etc:
+fail2bancmdline.PRODUCTION = \
+fail2banclient.PRODUCTION = \
+fail2banserver.PRODUCTION = False
```
yarikoptic: also with such an approach, we will miss testing those lines conditioned on PRODUCTION.  Marking them as " # pragma: no cover" is just hiding the problem -- imho a really inferior way to "mock things out".

fail2ban/client/fail2banclient.py
```diff
@@ -0,0 +1,462 @@
...
+			args = self._args
+
+			# Interactive mode
+			if self._conf.get("interactive", False):
+				# no readline in test:
+				if PRODUCTION: # pragma: no cover
+					try:
+						import readline
+					except ImportError:
+						raise ServerExecutionException("Readline not available")
```
yarikoptic: what is the side-effect of merely importing readline if running the tests (ie when not PRODUCTION)?

fail2ban/client/fail2banclient.py
```diff
@@ -0,0 +1,462 @@
...
+				try:
+					ret = True
+					if len(args) > 0:
+						ret = self.__processCommand(args)
+					if ret:
+						if PRODUCTION: # pragma: no cover
+							readline.parse_and_bind("tab: complete")
```
yarikoptic: if we were to use mock, something like
```
with mock.patch('readline.parse_and_bind') as parse_and_bind:
    .... tested functionality...
```
should be sufficient and we could test that actuaally parse_and_bind was called with such an argument  (smth like `parse_and_bind.assert_called_with('tab: complete')` IIRC)

fail2ban/client/fail2banclient.py
```diff
@@ -0,0 +1,462 @@
...
+							cmd = input_command()
+							if cmd == "exit" or cmd == "quit":
+								# Exit
+								return True
+							if cmd == "help":
+								self.dispUsage()
+							elif not cmd == "":
+								try:
+									self.__processCommand(shlex.split(cmd))
+								except Exception, e: # pragma: no cover
```
yarikoptic: even if we had it -- imho "no cover" is not appropriate here -- this code could and should be tested

fail2ban/client/fail2bancmdline.py
```diff
@@ -0,0 +1,279 @@
...
+
+	def dispVersion(self):
+		output("Fail2Ban v" + version)
+		output("")
+		output("Copyright (c) 2004-2008 Cyril Jaquier, 2008- Fail2Ban Contributors")
+		output("Copyright of modifications held by their respective authors.")
+		output("Licensed under the GNU General Public License v2 (GPL).")
+		output("")
+		output("Written by Cyril Jaquier <cyril.jaquier@fail2ban.org>.")
+		output("Many contributions by Yaroslav O. Halchenko <debian@onerussian.com>.")
```
yarikoptic: within this PR or some time later, we should remove those 2 lines here and in places alike.  Fail2Ban now is far beyond Cyril+mine's work.  Generic copyright statement should be sufficient I guess

fail2ban/client/fail2bancmdline.py
```diff
@@ -0,0 +1,279 @@
...
+	@staticmethod
+	def dumpConfig(cmd):
+		for c in cmd:
+			output(c)
+		return True
+
+	@staticmethod
+	def exit(code=0): # pragma: no cover - can't test
+		logSys.debug("Exit with code %s", code)
+		if os._exit:
```
yarikoptic: is that what should be `hasattr(os, '_exit')`? when (on which OS?) there is no `os._exit`?

fail2ban/client/fail2bancmdline.py
```diff
@@ -0,0 +1,279 @@
...
+		return ret, stream
+
+	@staticmethod
+	def dumpConfig(cmd):
+		for c in cmd:
+			output(c)
+		return True
+
+	@staticmethod
+	def exit(code=0): # pragma: no cover - can't test
```
yarikoptic: nope -- pragma is incorrect -- could be tested via mocking at least for the smoke testing and/or logic

fail2ban/client/fail2banserver.py
```diff
@@ -0,0 +1,226 @@
...
+	# 	Fail2banCmdLine.__init__(self)
+
+	##
+	# Start Fail2Ban server in main thread without fork (direct, it can fork itself in Server if daemon=True).
+	#
+	# Start the Fail2ban server in background/foreground (daemon mode or not).
+
+	@staticmethod
+	def startServerDirect(conf, daemon=True):
+		logSys.debug("-- direct starting of server in %s, deamon: %s", os.getpid(), daemon)
```
yarikoptic: so now we got some debug messages with `--` and `---` prefixes spread around this PR -- are they somehow special?

fail2ban/tests/fail2banclienttestcase.py
```diff
@@ -0,0 +1,598 @@
...
+		"""Call before every test case."""
+		LogCaptureTestCase.setUp(self)
+
+	def tearDown(self):
+		"""Call after every test case."""
+		LogCaptureTestCase.tearDown(self)
+
+	def _wait_for_srv(self, tmp, ready=True, startparams=None):
+		try:
+			sock = tmp+"/f2b.sock"
```
yarikoptic: generally we do still use 'os.path.join' for composing full filenames

fail2ban/tests/fail2banclienttestcase.py
```diff
@@ -0,0 +1,598 @@
...
+		if logSys.level < logging.DEBUG: # if HEAVYDEBUG
+			_out_file(cfg+"/fail2ban.conf")
+			_out_file(cfg+"/jail.conf")
+	# parameters (sock/pid and config, increase verbosity, set log, etc.):
+	return ("-c", cfg, "-s", tmp+"/f2b.sock", "-p", tmp+"/f2b.pid",
+					"-vv", "--logtarget", logtarget, "--loglevel", "DEBUG", "--syslogsocket", "auto",
+					"--timeout", str(fail2bancmdline.MAX_WAITTIME),
+	)
+
+def _kill_srv(pidfile): # pragma: no cover
```
yarikoptic: again, please don't mark code which could be tested with a bit of mocking as untestable.

fail2ban/tests/utils.py
```diff
@@ -50,6 +54,10 @@
 	else:
 		CONFIG_DIR = '/etc/fail2ban'
 
+# In not installed env (setup, test-cases) use fail2ban modules from main directory:
+if 1 or os.environ.get('PYTHONPATH', None) is None:
```
yarikoptic: ie shouldn't it be just `if os.environ.get('PYTHONPATH'):` (not sure if empty PYTHONPATH should have some magical treatment) ?

sebres: No, in the test case we should always add the current root path as PYTHONPATH variable, because the f2b-modules are not installed and in forked processes it is wrong otherwise. I've just forgot to remove `if 1 or ...` line.

About others of your comments: thanks for review, I'll take a look tomorrow...

fail2ban/tests/fail2banclienttestcase.py
sebres: > 1. be permanent as soon as this module gets imported, 2. would be dependent on order of tests imports.

This one is interested only if clientserver will be tested (so in this test module only);

> Since we don't like folks who later contribute all the styling fixes, let's from the beginning try to stay close to pep8 and sensible spacing?

I prefer to leave it together, because the function above belongs to this wrapper, and I don't want, that later it will be separated from someone (if someone inserts something between).

Because yes, we could rewrite it later using mock module.

Currently, the job of this branch is:
- to fix starting fail2ban in foreground;
- to cover fail2ban client-server with test cases (even with minimal overhead);

fail2ban/tests/fail2banclienttestcase.py
sebres: Currently used for prevent usage of readline module during test cases of client interactive mode (minimal invasive). But the interactive mode self will be tested...

And as the comment says, it prevents change of logging parameters (because that all takes place in test cases and we should capture logging output in LogCaptureTestCase.

Additionally see my comment above (later, minimal overhead).

>  Marking them as " # pragma: no cover" is just hiding the problem

- this branch enormous increase coverage anyway (neither client nor server was covered);
- you can see ignore lines in both cover-tools, so it can be easy found later;

fail2ban/client/fail2banclient.py
sebres: It's there not possible - :boom:

fail2ban/client/fail2banclient.py
sebres: Partially agree, because it is unfortunately not so easy with readline.

fail2ban/client/fail2banclient.py
sebres: This code could be unfortunately not so easy tested, because `__processCommand` will fail in rare case (for example only if client-server communication breaks in the meantime, but should be still active before). As already said it is not so easy, otherwise I had already covered it

fail2ban/client/fail2bancmdline.py
sebres: This is legacy code of @lostcontrol in 704a0b834f8a55240e2db1e4fc15bb3d434c2489...
Maybe any standard snippet for previously python versions...
I'll leave it (but add your suggestion with hasattr)

fail2ban/client/fail2bancmdline.py
sebres: Currently the global exit handler is overwritten in test cases (to test forked processes exit code), so it can't (we need a return code). 

Indeed, we could rewrite it using mock module, but the overhead is unnecessarily (several times larger) for coverage of this 2 lines.

fail2ban/client/fail2banserver.py
sebres: Just for easy distinguish the debug communication between parent/child (cs) and all around by heavydebug in the test cases...

fail2ban/tests/fail2banclienttestcase.py
sebres: This is only test case helper, used in helper decorator with_kill_srv only for safe removing server if some test case failed (remove process, free socket, etc.). 
So imho at all not expected to be covered - even test case helper only.

sebres:
I'm done, ready to merge...

Still again: that is a first shot (minimal invasive surgery), better mocking can be made later, because the test cases are almost wholehearted (I mean expected behavior of CS).

fail2ban/client/fail2banclient.py
yarikoptic: `__processCommand` could be mocked to just throw and exception on some argument... and then it would be very easy to test ;)
once again -- pretty much everything could be tested this days, `no cover` should be used really sparingly... may be almost never, since otherwise we just blind ourselves to uncovered code

fail2ban/client/fail2banserver.py
yarikoptic: then don't import logSys from elsewhere, define it here as `logSys = getLogger("fail2ban.client.server")`  and you get your dedicated (but also consistent with the rest) marker for such log lines.  custom prefixes just clutter it, without clear semantic attached etc.

fail2ban/tests/fail2banclienttestcase.py
yarikoptic: so, should we later investigate failures within the helpers, happen they come to help us whenever main thing goes kaboom?  nope... so they should (theoretically) be tested as well.  Thus -- `no cover` is not warranted.

yarikoptic:
sent you a PR https://github.com/sebres/fail2ban/pull/6

fail2ban/client/fail2banserver.py
sebres: nope, because the same code will be executed in server also (e.g. starting server in foreground).
Are we arguing about each additionally character now?

fail2ban/tests/fail2banclienttestcase.py
sebres: If the helper goes kaboom, you'll see it in the result of the test case (with the call stack). No matter `no cover` or not.
The helper self does not redound to "productive" codebase and in success test case returns immediatelly, that's why `no cover` imho justified

fail2ban/client/fail2banserver.py
yarikoptic: It is not about adding a character but about consistency, and readability... if everyone through the history of development of fail2ban was adding a new character, log would have become a Christmas tree. We need consistency in his cider is written, docs, logs looks like etc so future folks have a better chance at reading and using it.
The name I came up with was based on the name of the module here... anyways, was fixed already, but then I saw you returning it... heh

sebres: Nope, you've "fixed" similar log in "fail2banclienttestcase.py", to here ...
What's about standard and Christmas tree: as already said, I've consistently used sometimes the prefix "--" in some debug messages, so that it is immediately in the eye. 
So I think it's consistently enough :)

yarikoptic: On Tue, 12 Jul 2016, Serg G. Brester wrote:
>  > +
>  > +     @staticmethod
>  > +     def startServerDirect(conf, daemon=True):
>  > +             logSys.debug("-- direct starting of server in %s, deamon: %s", os.getpid(), daemon)

>    Nope, you've "fixed" similar log in "fail2banclienttestcase.py", to here
>    ...
>    What's about standard and Christmas tree: as already said, I've
>    consistently used sometimes the prefix "--" in some debug messages, so
>    that it is immediately in the eye.
>    So I think it's consistently enough :)

and then I would start using ++ consistently, and joe will start using
>< consistently etc...

myself I found those "hints" useful right when I need them, but then,
whenever I need to look at the log for some other issue, they stick like
a sore into the eye, since I need to tune my visual attention to other
lines.  That is why uniformity of the entire log, not just per author
consistency, is important in the long run.

-- 
Yaroslav O. Halchenko

fail2ban/client/fail2banclient.py
yarikoptic: hm... I have removed all PRODUCTION readline conditioning in this file and tests still passed no problem (Python 2.7.11+ , Debian system)... what is the catch? ;)

fail2ban/client/fail2banserver.py
sebres: > and then I would start using ++ consistently, and joe will start using
Then we'll beat them on the fingers ;)

Good, good, I'll replace it with space (You've already accepted such "indentation" as I made our newest cached ConfigReaders)...

yarikoptic:
I still don't like having tests specific logic within code itself (all the PRODUCTION thing) and persistent mocking of stuff within test module, but I guess we will leave it for the future

PR #1414, opened 2016-05-12
Preliminary PR for 0.10-cs (gh-1413)

sebres:
Desired in-between PR for 0.10-cs (gh-1413), that will be rebased hereafter above if this PR will be merged.

This is in-between PR, so client/server does not work at all (because no more in bin).

I can remember a "fail" for 'flushed' vs 'rolled over' in testFlushLogs, was fixed later from me...
So just ignore this if occurred.


sebres:
Close this preliminary in-between PR now, because #1413 is in pre-merge state

PR #1415, opened 2016-05-12
0.10 fixes

sebres:
small fixes for 0.10:
* prevent to fail stock configs test case, if any jail custom config does not have own test log-file (perhaps not clean copy), see https://github.com/fail2ban/fail2ban/pull/1413#issuecomment-218764683

* allow using of IPv6 address style mask (analog to the IPv4), for example: `2606:28ff::/ffff:ff80::` -> `2606:2880::/25`
* fast calculating of maskplen using map table MAP_ADDR2MASKPLEN, with pre-calculated addr->maskplen values;
test cases extended;

### DIFF
--- a/fail2ban/server/ipdns.py
+++ b/fail2ban/server/ipdns.py
@@ -171,7 +171,7 @@ class IPAddr(object):
 		# IP address without CIDR mask
 		if len(s) > 2:
 			raise ValueError("invalid ipstr %r, too many plen representation" % (ipstr,))
-		if "." in s[1]: # 255.255.255.0 style mask
+		if "." in s[1] or ":" in s[1]: # 255.255.255.0 resp. ffff:: style mask
 			s[1] = IPAddr.masktoplen(s[1])
 		s[1] = long(s[1])
 		return s
@@ -235,7 +235,7 @@ class IPAddr(object):
 		return self.ntoa
 
 	def __reduce__(self):
-		"""IPAddr pickle-handler, that simple wrap IPAddr to the str
+		"""IPAddr pickle-handler, that simply wraps IPAddr to the str
 
 		Returns a string as instance to be pickled, because fail2ban-client can't
 		unserialize IPAddr objects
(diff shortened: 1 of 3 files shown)


fail2ban/server/ipdns.py
```diff
@@ -235,7 +235,7 @@ def __str__(self):
 		return self.ntoa
 
 	def __reduce__(self):
-		"""IPAddr pickle-handler, that simple wrap IPAddr to the str
+		"""IPAddr pickle-handler, that simply wrap IPAddr to the str
```
yarikoptic: I hate to say but it is "wraps "
Feel free to ignore

sebres: fixed, "german-russian" and so on :))

fail2ban/tests/servertestcase.py
```diff
@@ -1021,7 +1021,11 @@ def testCheckStockJailActions(self):
 						cmd[2] = 'polling'
 					# change log path to test log of jail (to prevent "Permission denied" on /var/logs/ for test-user):
 					elif len(cmd) > 3 and cmd[0] == 'set' and cmd[2] == 'addlogpath':
-						cmd[3] = os.path.join(TEST_FILES_DIR, 'logs', cmd[1])
+						fn = os.path.join(TEST_FILES_DIR, 'logs', cmd[1])
+						# fallback to testcase01 if jail has not an own test log-file (currently should be no matter):
+						if not os.path.exists(fn): # pragma: no cover
+							fn = os.path.join(TEST_FILES_DIR, 'testcase01.log')
+						cmd[3] = fn
```
yarikoptic: I think that just assumptions in the test were somewhat "wrong" --  We do not have sample log files per jail -- we have them for each filter and there is a test (fail2ban.tests.samplestestcase.FilterSamplesRegex) to verify that we do.  So if you are to replace log file with something appropriate, it should ideally match the filter name.  For majority of jails, jail name matches filter name, so I guess it kinda worked.  Without this "fix" I ran into the same issue on a clean clone as well (so nothing adhockish about my setup), that is why I was puzzled why failure didn't trigger on travis.
But in the big realm of things for this test it doesn't care which file we feed here to the server, as long as it exists.  So, could as well all be testcase01.log, and a little comment that it shouldn't really matter if server doesn't get confused too much before getting killed ;)
So, I will merge and then will push a little DOC/PEP8ing enhancement PR

PR #1419, opened 2016-05-13
fixes gh-1417 Grave tags substitution bug. [part. cherry-picked from 0.10]

### LINKED ISSUE #1417: wrong recursion check in some constellation within substituteRecursiveTags
Just to reminder for me...

Invalid recursion check in substituteRecursiveTags.

For example action `bsd-ipfw` produced ValueError('properties contain self referencing definitions and cannot be resolved...')

Almost fixed in 0.10 (building test cases)

chery-pick to 0.9 (master)

sebres:
Grave tags substitution bug fixed [part. cherry-picked from 0.10]: 
Wrong recursion recognition by some constellation within substituteRecursiveTags. Affected actions like `bsd-ipfw`, etc.

closes #1417

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,9 @@ ver. 0.9.5 (2016/XX/XXX) - wanna-be-released
      - failregex of previous monit version merged as single expression.
    * filter.d/postfix.conf, filter.d/postfix-sasl.conf
      - extended failregex daemon part, matching also `postfix/smtps/smtpd` now (gh-1391)
+   * fixed a grave bug within tags substitutions because of incorrect detection of recursion 
+     in case of multiple inline substitutions of the same tag (affected actions: `bsd-ipfw`, etc). 
+     Now tracks the actual list of the already substituted tags (per tag instead of single list)
 
 - New Features:
    * New Actions:
--- a/fail2ban/server/action.py
+++ b/fail2ban/server/action.py
@@ -400,12 +400,16 @@ class CommandAction(ActionBase):
 				value = str(tags[tag])
 				# search and replace all tags within value, that can be interpolated using other tags:
 				m = t.search(value)
-				done = []
+				done = {}
+				last_found = tag
 				#logSys.log(5, 'TAG: %s, value: %s' % (tag, value))
 				while m:
 					found_tag = m.group(1)
 					#logSys.log(5, 'found: %s' % found_tag)
-					if found_tag == tag or found_tag in done:
+					curdone = done.get(last_found)
+					if curdone is None:
+						done[last_found] = curdone = []
+					if found_tag == tag or found_tag in curdone:
 						# recursive definitions are bad
 						#logSys.log(5, 'recursion fail tag: %s value: %s' % (tag, value) )
 						return False
@@ -417,7 +421,8 @@ class CommandAction(ActionBase):
 						continue
 					value = value.replace('<%s>' % found_tag , tags[found_tag])
 					#logSys.log(5, 'value now: %s' % value)
-					done.append(found_tag)
+					curdone.append(found_tag)
+					last_found = found_tag
 					m = t.search(value, m.start())
 				#logSys.log(5, 'TAG: %s, newvalue: %s' % (tag, value))
 				# was substituted?
--- a/fail2ban/tests/actiontestcase.py
+++ b/fail2ban/tests/actiontestcase.py
@@ -29,6 +29,7 @@ import time
 import tempfile
 
 from ..server.action import CommandAction, CallingMap
+from ..server.actions import OrderedDict
 
 from .utils import LogCaptureTestCase
 from .utils import pid_exists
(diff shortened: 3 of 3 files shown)


sebres:
BTW. Because we habe 2 active resp. major branches, I will prefer fast-forwarding push-merge (from git) if possible, instead of merge via gihub.
That makes the both branches better "readable", provides better overview, which of the commits are already in, etc.

yarikoptic:
```
BTW. Because we habe 2 active resp. major branches, I will prefer fast-forwarding push-merge (from git) if possible, instead of merge via gihub.
That makes the both branches better "readable", provides better overview, which of the commits are already in, etc.
```
I was of the same opinion for a while, but got to like having explicit merges, as e.g. github generates and you can do locally (merge --no-ff) -- that is actually when you can easily overview which development was done for that feature/fix (which might be multiple commits).  If all lined up within the same master branch -- there is no easy way to figure that out

sebres:
I said "prefer", of course what you mean is correct, but I think for one or two simple commits would be really better the fast-forward.

ChangeLog
```diff
@@ -15,6 +15,8 @@ ver. 0.9.5 (2016/XX/XXX) - wanna-be-released
      - failregex of previous monit version merged as single expression.
    * filter.d/postfix.conf, filter.d/postfix-sasl.conf
      - extended failregex daemon part, matching also `postfix/smtps/smtpd` now (gh-1391)
+   * grave tags substitution bug fixed: wrong recursion recognition by some constellation 
+     within substituteRecursiveTags (gh-1417). Affected actions like `bsd-ipfw`, etc.
```
yarikoptic: Such changelog describes the symptom which was fixed, which is indeed also useful (for users), but not the change which was done (useful for precious us).  Also not sure what "by some constellation" means.  What about
```
* fixed a bug within tags substitutions because of incorrect detection of recursion in case of multiple substitutions
  (affected actions: `bsd-ipfw`). Now also tracks the actual content of the already substituted tags
```

yarikoptic:
once again -- with FF you don't see the boundary of the change.  So please let's stay consistent with previous practice? ;-)  and also... since I am planing to have master merged into 0.10 periodically.  For such bug fixes best to just fix in master (only) and get it merged into 0.10 to avoid duplication of effort and to avoid future possible collisions

sebres:
> For such bug fixes best to just fix in master (only) and get it merged into 0.10 to avoid duplication of effort and to avoid future possible collisions

I can't do that sometimes in 0.9, here it was commited in 0.10 because:
- better error handling and increased verbosity in 0.10 - I had the problem at all to find this error in 0.9, because no call stack, just an error:
```
Failed to execute ban jail 'proftpd' action 'bsd-ipfw' info ... 
# with loglevel debug:
TypeError: 'bool' object is not iterable
```
compare with 0.10 (in addition to **full call stack** if verbosity mode):
```
RuntimeError: Error starting action Jail('proftpd')/bsd-ipfw: 
ValueError('properties contain self referencing definitions and cannot be resolved, fail tag: actionstart value: ipfw show | fgrep -q \'table(1)\' || ( ipfw show | awk \'BEGIN { b = 1 } { if ($1 <= b) { b = $1 + 1 } else { e = b } } END { if (e) exit e <br> else exit b }\'; num=$?; ipfw -q add $num unreach port ip from table\\(1\\) to me ; echo $num > "/var/run/fail2ban/ipfw-started-table_<table>" )',)
```
- the conflicts, because the substituteRecursiveTags totally rewritten in 0.10 (you know "conditional" support because of IPv6)
- one of the test case completely missing (resp. at all not available in 0.9).
Just compare both solutions.

BTW. If I have fixed it exactly in this way, weil I thought so it better, not only to make you angry :))

ChangeLog
sebres: Done (but little bit modified)

sebres:
For general understanding: 

This fix is concerned not only action `bsd ipfw`. Theoretical any action can be affected, because the order of elements by iterating over `__dict__` is undefined, so changing or adding of single parameter in jail, action or filter, can completely tie the banning within this jail.

That is, what I called as "some constellation"

sebres:
The story was continued: in certain circumstances it will again occurred, that the cyclic recursion will be wrong recognized. 
I've fixed it and added new test cases to cover it in my 0.10 branch and then cherry picked it in this branch (and merged to master), see 7a428d332739cba2562c25f4b48dbe0d626f12a2

Normally it works also by `MAX_TAG_REPLACE_COUNT = 1`, but to be sure I had set it to 10, for maximal 10 repetition for replacement of the same tag (I think there would be possible in very complex composite interpolations)

sebres:
If someone need a diff for manually update of this bug, look at fail2ban/server/action.py
in https://github.com/fail2ban/fail2ban/compare/bdc2d07946d2b10ccf4634db6117e798ab079555...a80043ce808cfb96ae5908c056ce8ccb969d9bc0

yarikoptic:
Could wet please stay with the establish with the established workflow of changes going through PRs????

yarikoptic:
Could we please stay with established workflow of changes going through PR testing and review and giving other folks at least a day to react????

sebres:
It is the same PR!? And I've fixed it in the same branch, to hold the tree PR-consistently...
What the "review and giving other folks to react" - sorry, I had just not thought about that. I'm rather almost alone here the most time :)))

But they can still review and react... :)) just without new PR for the same thing.

yarikoptic:
This PR was merged more than a week ago. As such it is done. New changes should be reviewed independently. You can keep pushing to the same branch and submitting new PRs from it.
You aren't alone - "big brother" is watching (as you might see ) ;-)

PR #1421, opened 2016-05-13
filter.d/common.conf: fixes unexpected extra regex-space in generic `__prefix_line` (gh-1405)

### LINKED ISSUE #1405: Extra regex space in __prefix_line
### Environment:

- Fail2Ban: 0.8.13-1
- OS: Debian   8.4
- firehol  1.297-1

### The issue:

Firehol delivers lines to syslog that cannot be matched by regex's in a filter.

#### Steps to reproduce

Filter, ./fh.local:
```
[INCLUDES]
before = /etc/fail2ban/filter.d/common.conf

[Definition]
failregex = ^%(__prefix_line)sIN-world:IN=eth0 .* SRC=<HOST> .*$
```
run:
```
$ fail2ban-regex \
  'Apr 19 09:02:51 thismachine kernel: [  970.699396] IN-world:IN=eth0 OUT= MAC= SRC=123.123.1.2 DST=192.168.1.31 LEN=248 TOS=0x00 PREC=0x00 TTL=64 ID=17598 DF PROTO=UDP SPT=138 DPT=138 LEN=228' \
  ./fh.local

Running tests
=============

Use   failregex file : ./fh.local
Use      single line : Apr 19 09:02:51 thismachine kernel: [  970.699396] IN-worl...


Results
=======

Failregex: 0 total

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [1] MONTH Day Hour:Minute:Second
`-

Lines: 1 lines, 0 ignored, 0 matched, 1 missed
```

Firehol's log strings would match if they had a second space in front of
'IN-world'.

Example:
```
$ fail2ban-regex \
  'Apr 19 09:02:51 thismachine kernel: [  970.699396]  IN-world:IN=eth0 OUT= MAC= SRC=123.123.1.2 DST=192.168.1.31 LEN=248 TOS=0x00 PREC=0x00 TTL=64 ID=17598 DF PROTO=UDP SPT=138 DPT=138 LEN=228' \
  ./fh.local

Running tests
=============

Use   failregex file : ./fh.local
Use      single line : Apr 19 09:02:51 thismachine kernel: [  970.699396]  IN-wor...


Results
=======

Failregex: 1 total
|   1) [1] ^\s*(<[^.]+\.[^.]+>)?\s*(?:\S+ )?(?:kernel: \[ *\d+\.\d+\] )?(?:@vserver_\S+ )?(?:(?:\[\d+\])?:\s+[\[\(]?\S*(?:\(\S+\))?[\]\)]?:?|[\[\(]?\S*(?:\(\S+\))?[\]\)]?:?(?:\[\d+\])?:?)?\s(?:\[ID \d+ \S+\])?\s*IN-world:IN=eth0 .* SRC=<HOST> .*$
`-

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [1] MONTH Day Hour:Minute:Second
`-

Lines: 1 lines, 0 ignored, 1 matched, 0 missed
```

#### Expected behavior

Log strings of firehol can be matched out of the box.

#### Observed behavior

Log strings of firehol cannot be matched.

My findings:
I guess this issue does affect recent versions of fail2ban.

Line 54 of config/filter.d/common.conf (current master):
```
__prefix_line = \s*%(__bsd_syslog_verbose)s?\s*(?:%(__hostname)s )?(?:%(__kernel_prefix)s )?(?:@vserver_\S+ )?%(__daemon_combs_re)s?\s%(__daemon_extra_re)s?\s*
```

introduces an extra space at column 133-134 '\s'. The space is in between '__daemon_combs_re' and '__daemon_extra_re'. Both strings do not match for firehol log strings. But the space remains. 

This extra space has been introduced on 29 Apr 2013 by commit d2a953756802bd7cf63f5f5f792371f52f5cba8c.

I suggest commit bc2e89c026732eb39f19404ae5c91345c80bbfbf. Does this commit break existing filters?

#### Relevant lines from monitored log files in question:

```
Apr 19 09:02:51 thismachine kernel: [  970.699396] IN-world:IN=eth0 OUT= MAC= SRC=123.123.1.2 DST=192.168.12.12 LEN=248 TOS=0x00 PREC=0x00 TTL=64 ID=17598 DF PROTO=UDP SPT=138 DPT=138 LEN=228
```

sebres:
misleadingly committed in d2a953756802bd7cf63f5f5f792371f52f5cba8c;

all optional spaces normalized in generic include `common.conf` + test cases are extended (using new example pseudo-filter and test log `zzz-generic-example`)

closes  #1405

### DIFF
--- a/config/filter.d/common.conf
+++ b/config/filter.d/common.conf
@@ -26,11 +26,11 @@ __daemon_re = [\[\(]?%(_daemon)s(?:\(\S+\))?[\]\)]?:?
 
 # extra daemon info
 # EXAMPLE: [ID 800047 auth.info]
-__daemon_extra_re = (?:\[ID \d+ \S+\])
+__daemon_extra_re = \[ID \d+ \S+\]
 
 # Combinations of daemon name and PID
 # EXAMPLES: sshd[31607], pop(pam_unix)[4920]
-__daemon_combs_re = (?:%(__pid_re)s?:\s+%(__daemon_re)s|%(__daemon_re)s%(__pid_re)s?:?)
+__daemon_combs_re = %(__pid_re)s?:\s+%(__daemon_re)s|%(__daemon_re)s%(__pid_re)s?:?
 
 # Some messages have a kernel prefix with a timestamp
 # EXAMPLES: kernel: [769570.846956]
@@ -44,14 +44,14 @@ __md5hex = (?:[\da-f]{2}:){15}[\da-f]{2}
 
 # bsdverbose is where syslogd is started with -v or -vv and results in <4.3> or
 # <auth.info> appearing before the host as per testcases/files/logs/bsd/*.
-__bsd_syslog_verbose = (<[^.]+\.[^.]+>)
+__bsd_syslog_verbose = <[^.]+\.[^.]+>
 
 # Common line prefixes (beginnings) which could be used in filters
 #
 #      [bsdverbose]? [hostname] [vserver tag] daemon_id spaces
 #
 # This can be optional (for instance if we match named native log files)
-__prefix_line = \s*%(__bsd_syslog_verbose)s?\s*(?:%(__hostname)s )?(?:%(__kernel_prefix)s )?(?:@vserver_\S+ )?%(__daemon_combs_re)s?\s%(__daemon_extra_re)s?\s*
+__prefix_line = \s*(?:(?:%(__bsd_syslog_verbose)s)\s*)?(?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?(?:(?:@vserver_\S+)\s*)?(?:(?:%(__daemon_combs_re)s)\s*)?(?:(?:%(__daemon_extra_re)s)\s*)?
 
 # PAM authentication mechanism check for failures, e.g.: pam_unix, pam_sss,
 # pam_ldap
--- a/config/filter.d/common.conf
+++ b/config/filter.d/common.conf
@@ -30,7 +30,7 @@ __daemon_extra_re = \[ID \d+ \S+\]
 
 # Combinations of daemon name and PID
 # EXAMPLES: sshd[31607], pop(pam_unix)[4920]
-__daemon_combs_re = %(__pid_re)s?:\s+%(__daemon_re)s|%(__daemon_re)s%(__pid_re)s?:?
+__daemon_combs_re = (?:%(__pid_re)s?:\s+%(__daemon_re)s|%(__daemon_re)s%(__pid_re)s?:?)
 
 # Some messages have a kernel prefix with a timestamp
 # EXAMPLES: kernel: [769570.846956]
@@ -46,14 +46,16 @@ __md5hex = (?:[\da-f]{2}:){15}[\da-f]{2}
 # <auth.info> appearing before the host as per testcases/files/logs/bsd/*.
 __bsd_syslog_verbose = <[^.]+\.[^.]+>
 
-__date_ambit = \[\]
+__vserver = @vserver_\S+
+
+__date_ambit = (?:\[\])
 
 # Common line prefixes (beginnings) which could be used in filters
 #
 #      [bsdverbose]? [hostname] [vserver tag] daemon_id spaces
 #
 # This can be optional (for instance if we match named native log files)
-__prefix_line = (?:%(__date_ambit)s)?\s*(?:(?:%(__bsd_syslog_verbose)s)\s*)?(?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?(?:(?:@vserver_\S+)\s*)?(?:(?:%(__daemon_combs_re)s)\s*)?(?:(?:%(__daemon_extra_re)s)\s*)?
+__prefix_line = %(__date_ambit)s?\s*(?:%(__bsd_syslog_verbose)s\s+)?(?:%(__hostname)s\s+)?(?:%(__kernel_prefix)s\s+)?(?:%(__vserver)s\s+)?(?:%(__daemon_combs_re)s\s+)?(?:%(__daemon_extra_re)s\s+)?
 
 # PAM authentication mechanism check for failures, e.g.: pam_unix, pam_sss,
 # pam_ldap
--- a/config/filter.d/common.conf
+++ b/config/filter.d/common.conf
@@ -46,12 +46,14 @@ __md5hex = (?:[\da-f]{2}:){15}[\da-f]{2}
 # <auth.info> appearing before the host as per testcases/files/logs/bsd/*.
 __bsd_syslog_verbose = <[^.]+\.[^.]+>
 
+__date_ambit = \[\]
+
 # Common line prefixes (beginnings) which could be used in filters
 #
 #      [bsdverbose]? [hostname] [vserver tag] daemon_id spaces
 #
 # This can be optional (for instance if we match named native log files)
-__prefix_line = \s*(?:(?:%(__bsd_syslog_verbose)s)\s*)?(?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?(?:(?:@vserver_\S+)\s*)?(?:(?:%(__daemon_combs_re)s)\s*)?(?:(?:%(__daemon_extra_re)s)\s*)?
+__prefix_line = (?:%(__date_ambit)s)?\s*(?:(?:%(__bsd_syslog_verbose)s)\s*)?(?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?(?:(?:@vserver_\S+)\s*)?(?:(?:%(__daemon_combs_re)s)\s*)?(?:(?:%(__daemon_extra_re)s)\s*)?
 
 # PAM authentication mechanism check for failures, e.g.: pam_unix, pam_sss,
 # pam_ldap
--- a/config/filter.d/nsd.conf
+++ b/config/filter.d/nsd.conf
@@ -22,7 +22,7 @@ _daemon = nsd
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
 # Values:  TEXT
 
-failregex =  ^\[\]%(__prefix_line)sinfo: ratelimit block .* query <HOST> TYPE255$
-              ^\[\]%(__prefix_line)sinfo: .* <HOST> refused, no acl matches\.$
+failregex =  ^%(__prefix_line)sinfo: ratelimit block .* query <HOST> TYPE255$
+             ^%(__prefix_line)sinfo: .* <HOST> refused, no acl matches\.$
 
 ignoreregex =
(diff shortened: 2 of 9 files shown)


sebres:
BTW. I was thinking about extending of `__prefix_line` with optional prefix `(?:\[])?` to automatically match if date pattern included in brackets (and it will be cut out from the log line, before match).
Again and again occured the issues with this question. 
Just to resolve this problem for all times. I think, it would also fit in this PR, because already touch a prefix line resp. common.conf.
@yarikoptic, @fail2ban/contributors Any thoughts?

config/filter.d/zzz-generic-example.conf
```diff
@@ -0,0 +1,17 @@
+# Fail2Ban generic example resp. test filter
```
yarikoptic: so it would be shipped with fail2ban installations... but if solely for testing, why not to keep it in testing?

yarikoptic:
I don't see harm in having such a prefix, indeed we had few cases like that already I believe

config/filter.d/zzz-generic-example.conf
sebres: Because we should build then a biggish test case (that doing the same as our testSampleRegexs-Builder).
Or we say now easy: the filters begins with zzz-* are example's or just for test cases.

I'll check later how I can extract the testSampleRegexs-Builder functionality to separate test case with some common method... Then we can move it to the tests-directory. Currently impossible, because of include also.

And now we have 2 major branches... Hmm

sebres:
generic `__prefix_line` extended with optional brackets for the date ambit, added as new parameter `__date_ambit`

config/filter.d/zzz-generic-example.conf
sebres: zzz-* moved to tests folder, will push it now

sebres:
Factory testSampleRegexs extended here for sample filters from folder tests (the name of filter should start with zzz-*)

sebres:
After merging this to master (0.9) I'll merge the master to 0.10 (to resolve current conflicts within pair commits in current master and 0.10)

sebres:
Codecov says sometimes bullshit also: coverage decreased, but if you'll see the report, it was really increased with +7 lines... 
Well, so are the AIs today :)

config/filter.d/common.conf
```diff
@@ -44,14 +44,16 @@ __md5hex = (?:[\da-f]{2}:){15}[\da-f]{2}
...
+
+__date_ambit = \[\]
 
 # Common line prefixes (beginnings) which could be used in filters
 #
 #      [bsdverbose]? [hostname] [vserver tag] daemon_id spaces
 #
 # This can be optional (for instance if we match named native log files)
-__prefix_line = \s*%(__bsd_syslog_verbose)s?\s*(?:%(__hostname)s )?(?:%(__kernel_prefix)s )?(?:@vserver_\S+ )?%(__daemon_combs_re)s?\s%(__daemon_extra_re)s?\s*
+__prefix_line = (?:%(__date_ambit)s)?\s*(?:(?:%(__bsd_syslog_verbose)s)\s*)?(?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?(?:(?:@vserver_\S+)\s*)?(?:(?:%(__daemon_combs_re)s)\s*)?(?:(?:%(__daemon_extra_re)s)\s*)?
```
yarikoptic: I will repeat my whining to keep previously hardcoded spaces,  within optional groups, and not to make them all `\s*`

config/filter.d/nsd.conf
```diff
@@ -22,7 +22,7 @@ _daemon = nsd
 #          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
 # Values:  TEXT
 
-failregex =  ^\[\]%(__prefix_line)sinfo: ratelimit block .* query <HOST> TYPE255$
-              ^\[\]%(__prefix_line)sinfo: .* <HOST> refused, no acl matches\.$
+failregex =  ^%(__prefix_line)sinfo: ratelimit block .* query <HOST> TYPE255$
+             ^%(__prefix_line)sinfo: .* <HOST> refused, no acl matches\.$
```
yarikoptic: Unaltered to this fix... But it looks worth investigating if any log line monitored by this filter could contain user provided content at the end if the line... If can - injection can be possible due to completely absent anchoring at the front

sebres: I would have to completely remove `^%(__prefix_line)sinfo: .*`, because it makes at all no sense (independent whether greedy or not). Imho, only anchored end after `.*` is important.

But I know some owners, that will immediately beef about this :)))

config/filter.d/common.conf
sebres: Then I'll repeat my https://github.com/fail2ban/fail2ban/issues/1405#issuecomment-219214953
This is a minimalist invasive fix for #1405 (yes, optional brackets also, but even - optional).
I want not implement suggested mandatory spaces here. 
If you want do that, make a separate PR - because I want not to take blame on me...

yarikoptic: On Wed, 18 May 2016, Serg G. Brester wrote:
>  > -__prefix_line = \s*%(__bsd_syslog_verbose)s?\s*(?:%(__hostname)s )?(?:%(__kernel_prefix)s )?(?:@vserver_\S+ )?%(__daemon_combs_re)s?\s%(__daemon_extra_re)s?\s*
>  > +__prefix_line = (?:%(__date_ambit)s)?\s*(?:(?:%(__bsd_syslog_verbose)s)\s*)?(?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?(?:(?:@vserver_\S+)\s*)?(?:(?:%(__daemon_combs_re)s)\s*)?(?:(?:%(__daemon_extra_re)s)\s*)?

>    Then I'll repeat my #1405 (comment)
>    This is a minimalist invasive fix for #1405 (yes, optional brackets also,
>    but even - optional).
>    I want not implement suggested mandatory spaces here.
>    If you want do that, make a separate PR - because I want not to take blame
>    on me...

but here I can blame you removing the spaces already in other spots! ;)
so the fix cannot be called minimalist by that measure alone --
minimalist fix would have fixed only that portion of the regex.
Here you are doing all other stuff, so I can hardly call it
minimalist by any measure... ok -- let me then just take pride (I guess
that is anti-blame) by submitting you a PR to adjust your PR -- just ack
if that would be ok.

sebres: No, it is really minimalist, because as well as before and hereafter all the groups all separated with \\s*, so it is totally backwards compatible. Also if someone will rewrite two or more groups vars in .local, it will still work. But definitely not if we will use here a mandatory space...
But you are the boss... and no matter for me, because all my expressions are anyway custom.
The question would be to do it in this PR, that fixes a concretely bug...

yarikoptic: >  because as well as before and hereafter all the groups all separated with \s*

Are we looking at the same diff? ;) :
`(?:%(__hostname)s )?(?:%(__kernel_prefix)s )?` -> `?:(?:%(__hostname)s)\s*)?(?:(?:%(__kernel_prefix)s)\s*)?`
and others

Moreover, I would prefer to keep groupping (`(?:...)`) within the "variables" which then would make final regexp easier -- now it is difficult to match all the parantheses. Not to say that again, this goes away from "minimal change" -- what if someone has created a custom filter which relied on groups being present in those variables?

sebres: well, about that was already mandatories - no problem (i made is just consistently equal), what I mean was the ALL spaces...
And the grouping outside is necesarry, because if someone will use pipe as or operand, that applied to elms in the group, not for whole expression right from pipe...

yarikoptic: >    And the grouping outside is necesarry, because if someone will use pipe as
>    or operand, that applied to elms in the group, not for whole expression
>    right from pipe...

sorry, I don't get it ...  could you demonstrate on an example for an
old me on what you mean?

sebres: `x=(?:<a>\s)?`
if someone defines now:
`a=b|c`
the space will belong to `c` only, `b` has no space.
Otherwise it will be by:
`x=(?:(?:<a>)\s)?`
And the regex compiler should optimize the second non-matching group anyway, if `a` is a single token

yarikoptic: So far your example only confirms what I have recommended -- to use groups within variable definitions. E.g. in your case if you had defined `a=(?:b|c)` you would be just fine in your `x` definition, wouldn't you?
It is because this way all possible logical operands applied within variable definition have a clear defined boundary, and an evil someone can then use it without realizing that he needed to surround it by an additional group just to guarantee correct application of logical operrands within it!

But indeed, if we are  to please both of our paranoias we would ultimately need both -- groups while defining variables and then, to prevent silly user's mistake of not using groups within his varaibles, while using those variables along side with additional characters (i.e. space in this case ;-) ).  But it would render many regexps hard to comprehend.  In our particular case here though we rely on variable definitions which aren't generally modified or passed from jail specs somehow but simply used for string substitutions pretty much right here.  Thus IMHO we could safely stick to previously used groupping within variables definitions, which will assure that any logical operand within them, would have it bound just fine and trailng space added will be applied outside of that boundary without  needing double-protection.
Makes sense, or am I still missing? ;)

sebres: > So far your example only confirms what I have recommended 

No It's not, because what I mean, was the `a` can be redefined from any (end-users, \*nix maintainers, etc.)

If `x` is `x=(?:(?:<a>)\s)?`, that is not so error prone for the future, because no matter whether `a` is self-grouped or not.
Just to be understandable, I meant as `x` the `__prefix_line`, as `a` for example the `__hostname`...

yarikoptic: well, to be fully clear, by `<a>` (not just `a` name of the regex group) you meant `%(__hostname)s` (not just `_hostname`), correct?

Then again -- **any** of those variables can be redefined! 
So, the question is then either we should wrap around **every** string interpolation variable we ever use?
That would make many if not all of the failregex definitions really hard to comprehend, especially after all the string interpolations are substituted. Or, as I suggest and as we did in the past, we just adhere to the principle that if there is any logical operand within the variable definition -- it should be groupped at that point of definition to guarantee correct application of the logical operrand?

sebres: > Then again -- any of those variables can be redefined!
So, the question is then either we should wrap around every string...

Well, that would be a too large number :) 
But why not to make it, where we can make it with mini effort? 
Resp. it is already done, and you want extra ungroup it.

yarikoptic: because "That would make many if not all of the failregex definitions really hard to comprehend, especially after all the string interpolations are substituted" and above should still be "fixed" to bring back non-optional spaces ;)

sebres:
@yarikoptic please take a look for compromise proposal (new parameter `__pl_sep`) in the common.conf... As default value exact one mandatory space will be used as separator. But now there can be at once reconfigured for all groups, by changing of its value.

BTW. I think the common space (\s) is more better as space-character (\x20) because no matter how it is really separated (for example a tab character would be good also, because as already said the spaces did not wear any connotation at all).

yarikoptic:
sorry - I don't see a point to define a new variable there making it again harder to read.  Do you see it being redefined/overriden by someone some time?  majority of those small variables were defined to make regexp more readable, and this one goes against that.
I am ok with `\s` or even `\s+`

sebres:
Amended now without `__pl_sep`. As replacement used `\s+` as more flexible as an exact one space.
I want merge it, if travis not griped

PR #1410, opened 2016-05-10
WiP: do not merge!  0.10 -> master  merge whenever it is ready to be named a 'stable'-ish one

yarikoptic:
This PR is staging to make sure that we can progress from current state of master into the world of 0.10.  To be "merged" after a few of 0.10.x experimental releases

Issue #1293, opened 2016-01-07
action hostsdeny throws ERROR in logfile but does the job


koeppea:
Hi

When I try to use the default command on a Debian 8  ( *however I had to explicitely install the package ed* ) , I always get an error when the *unban action* in the **hostsdeny.conf** is being invoked.
However, the line disappears in the end.
The following output is produced using a *bantime* of 60 and running `tail -f /var/log/fail2ban.log` in the background.
```
2016-01-07 23:21:08,176 fail2ban.actions[26268]: WARNING [ssh] Ban 192.168.1.1

root@localhost /etc/fail2ban # tail /etc/hosts.deny 
#
# The PARANOID wildcard matches any host whose name does not match its
# address.
#
# You may wish to enable this to ensure any programs that don't
# validate looked up hostnames still leave understandable logs. In past
# versions of Debian this has been the default.
# ALL: PARANOID

ALL: 192.168.1.1
root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # 2016-01-07 23:22:08,198 fail2ban.actions[26268]: WARNING [ssh] Unban 192.168.1.1
2016-01-07 23:22:08,230 fail2ban.actions.action[26268]: ERROR   echo -e "/^ALL: 192.168.1.1$/\nd\nw\nq" | ed /etc/hosts.deny returned 100

root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # tail /etc/hosts.deny 
# for further information.
#
# The PARANOID wildcard matches any host whose name does not match its
# address.
#
# You may wish to enable this to ensure any programs that don't
# validate looked up hostnames still leave understandable logs. In past
# versions of Debian this has been the default.
# ALL: PARANOID

root@localhost /etc/fail2ban # 

```

Why shows the logfile 
**ERROR   echo -e "/^ALL: 192.168.1.1$/\nd\nw\nq" | ed /etc/hosts.deny returned 100** ?

When I try to reproduce this by explicitely running these commands on the console, the return code is 0:
```

root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # echo "ALL: 192.168.1.1" >> /etc/hosts.deny 
root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # 
root@localhost /etc/fail2ban # tail /etc/hosts.deny 
#
[...]

PR #1422, opened 2016-05-13
0.10 fix multi set: 

sebres:
Fixes:
*  csocket multi-set fix: prevent to convert `list`, `dict`, `set` during transfer (send), this offers a sending of 'multi-set' arrays 
(forgotten by cherry-picking from my multi-set branch)
* a little bit more details (in debug logging) if multi-set used
* database: always explicit convert `ip` to `str`, because may be an IPAddr, that will be unsupported type by bind parameter (as long as we've found any default wrapper handler for sqlite3)

### DIFF
--- a/fail2ban/client/csocket.py
+++ b/fail2ban/client/csocket.py
@@ -44,7 +44,9 @@ class CSocket:
 	
 	def send(self, msg):
 		# Convert every list member to string
-		obj = dumps([str(m) for m in msg], HIGHEST_PROTOCOL)
+		obj = dumps(map(
+			lambda m: str(m) if not isinstance(m, (list, dict, set)) else m, msg),
+		  HIGHEST_PROTOCOL)
 		self.__csock.send(obj + CSPROTO.END)
 		return self.receive(self.__csock)
 
--- a/fail2ban/server/database.py
+++ b/fail2ban/server/database.py
@@ -455,7 +455,7 @@ class Fail2BanDb(object):
 			queryArgs.append(MyTime.time() - bantime)
 		if ip is not None:
 			query += " AND ip=?"
-			queryArgs.append(ip)
+			queryArgs.append(str(ip))
 		query += " ORDER BY ip, timeofban desc"
 
 		return cur.execute(query, queryArgs)
--- a/fail2ban/server/server.py
+++ b/fail2ban/server/server.py
@@ -266,8 +266,10 @@ class Server:
 		flt = self.__jails[name].filter
 		if multiple:
 			for value in value:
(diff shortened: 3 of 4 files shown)


yarikoptic:
re " csocket multi-set fix:" -- what was the issue this one is fixing?

sebres:
> re " csocket multi-set fix:" -- what was the issue this one is fixing?

Second piece from https://github.com/fail2ban/fail2ban/pull/1410#issuecomment-219131023
```
ERROR  NOK: ("Unable to compile regular expression '['",)
ERROR  NOK: ('string index out of range',)
```

If you'll try to start any jail having more as one failregex without this fix - you'll get an error, because it has always convert command array to the string (by `multi-set` command the failregex's is a list).

yarikoptic:
sorry... still not grasping since nearly all of our filters have multi-line failregexes... how was it working before?

sebres:
I've cherry-picket `multi-set` functionality from my another branch, as I've implement conditional sections (and forget the csocket).
And tested only with test jails using single `failregex`, that uses `set` instead of `multi-set` :)

yarikoptic:
aha... and our unittests which load filters they don't bother sending them through the csocket where the problem got revealed... could we also get a unittest for the communication please then so it fails without this fix? ;)

sebres:
sure, but let us make it later, it could be one number greater (our artificial tests send not really somewhat now from filters).
In the time, I would merge this one, therefore the people can continue test 0.10.

yarikoptic:
Go ahead... Cant do from the phone since coverage decreased

koeppea:
In this branch, the manual banip works and the IP actually arrives in the **/etc/hosts.deny** file.
Trying to manually unban the IP hits an error I've had before (#1293):
```
2016-05-14 07:58:58,144 fail2ban.utils          [6877]: ERROR    -- stderr: '/bin/sh: 4: ed: not found'
```
Using the standard action *iptables-multiport* , manually ban und unban works fine for both v4 and v6.
Is this a resurrected issue limited to the hostsdeny action config file?

koeppea:
> Is this a resurrected issue limited to the hostsdeny action config file?

Sorry, it's again my fault. I forgot that I've exchanged my laptop in the meantime and ed is not included in Debian's minimal package selection.
So having **ed** now installed, hostsdeny works of course as it should be.

koeppea:
route action is also working as expected.
Continuing testing ipset....

koeppea:
Testing action **iptables-ipset-proto6** was also successful.
Testing action pf later today.

PR #1424, opened 2016-05-14
ENH: Branch 0.10 - updated firewallcmd actions

TorontoMedia:
- Created firewallcmd-common.
- Updated all firewallcmd actions to handle ipv6.

### DIFF
--- a/config/action.d/firewallcmd-multiport.conf
+++ b/config/action.d/firewallcmd-multiport.conf
@@ -5,59 +5,28 @@
 
 [INCLUDES]
 
-before = iptables-common.conf
+before = firewallcmd-common.conf
 
 [Definition]
 
-actionstart = firewall-cmd --direct --add-chain ipv4 filter f2b-<name>
-              firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 1000 -j RETURN
-              firewall-cmd --direct --add-rule ipv4 filter <chain> 0 -m conntrack --ctstate NEW -p <protocol> -m multiport --dports <port> -j f2b-<name>
+actionstart = firewall-cmd --direct --add-chain <family> filter f2b-<name>
+              firewall-cmd --direct --add-rule <family> filter f2b-<name> 1000 -j RETURN
+              firewall-cmd --direct --add-rule <family> filter <chain> 0 -m conntrack --ctstate NEW -p <protocol> -m multiport --dports <port> -j f2b-<name>
 
-actionstop = firewall-cmd --direct --remove-rule ipv4 filter <chain> 0 -m conntrack --ctstate NEW -p <protocol> -m multiport --dports <port> -j f2b-<name>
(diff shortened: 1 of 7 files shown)


config/action.d/firewallcmd-multiport.conf
```diff
@@ -5,59 +5,28 @@
...
-chain = INPUT_direct
-
 # Could also use port numbers separated by a comma. 
 port = 1:65535
 
-
-# Option:  protocol
-# Values:  [ tcp | udp | icmp | all ]
-
-protocol = tcp
```
yarikoptic: I believe this one got lost...?

TorontoMedia: I can add protocol and port in common but I wrote it I felt that it wasn't needed due to being defined in jail.conf. Should I add them to common?

sebres:
@TorontoMedia looks good and as long as I can estimate, all are firewallcmd conform.
Thx for contribution to IPv6! @fail2ban/contributors Still someone? :))

So will merge now (let the people test it :)))

sebres:
@TorontoMedia would be really good if you could take a look in the test case `testCheckStockCommandActions` at least for the firewallcmd-action's.
I wrote a new test cases for these.
see: https://github.com/fail2ban/fail2ban/blob/0.10/fail2ban/tests/servertestcase.py#L1394
For list of command's will be started, use:
```
bin/fail2ban-testcases -l heavydebug testCheckStockCommandActions
```

Additionaly I've modified `action.d/firewallcmd-ipset.conf`: different name of the match set's for IPv4/IPv6, using conditional `<ipmset>`, analog to the iptables-ipset action.
I thought, the 2 ipsets with the same name are not allowed

config/action.d/firewallcmd-multiport.conf
yarikoptic: All of these fields are to provide default values, so we better have then set explicitly without relying on being defined in jail.conf... And that is how it is in other action files, so we better stay consistent. Since @sebres merged already, please push a new commit and open a new PR. Thanks!

sebres: Why? You can still override all that (independent in which common-include or directly in action these are take place).
So if you write something followed in jail, you will overwrite a protocol with `udp`:
```
banaction=firewallcmd-multiport[name=%(__name__)s, protocol="udp", chain="INPUT" ...]
```
The same concerns to firewallcmd-multiport.local etc.

Or I understand something not? What you mean with "stay consistent"?

sebres: Ah, did you mean, we've no default now?

yarikoptic: Yes. When moved it was not copied into -common in this PR

TorontoMedia:
I wrote the rest of the firewallcmd action testcases and added ipmset to firewallcmd-common. I'm going to start a new pull-request since this one is closed.

Issue #990, opened 2015-03-16
Use #!/usr/bin/python -Es to avoid local changes


opoplawski:
See https://bugzilla.redhat.com/show_bug.cgi?id=1202151

fail2ban probably should not attempt to load local python modules.  Thoughts?

yarikoptic:
On Mon, 16 Mar 2015, Orion Poplawski wrote:

>    See https://bugzilla.redhat.com/show_bug.cgi?id=1202151

>    fail2ban probably should not attempt to load local python modules.
>    Thoughts?

so the recommendation is to  use "#!/usr/bin/python -Es" as the shebang
line to prevent side-effects of importing from PYTHONPATH etc

certainly sounds like a good strategy for the versions deployed
system-wide from the distribution packages.  But I am not sure if we
should do it in the mainline codebase -- would complicate testing etc.
So IMHO we should just provide this recommendation and adopt it in
Debian packages

-- 
Yaroslav O. Halchenko, Ph.D.
http://neuro.debian.net http://www.pymvpa.org http://www.fail2ban.org
Research Scientist,            Psychological and Brain Sciences Dept.
Dartmouth College, 419 Moore Hall, Hinman Box 6207, Hanover, NH 03755
Phone: +1 (603) 646-9834                       Fax: +1 (603) 646-1419
WWW:   http://www.linkedin.com/in/yarik

opoplawski:
Yeah, does seem like a packaging option.  Is there a way to get this done at build/install time other than using sed?

tomspur:
Could you maybe try `python setup.py build --executable="%{__python} -Es"`?

yarikoptic:
On Wed, 18 Mar 2015, Thomas Spura wrote:

>    Could you maybe try python setup.py build --executable="%{__python} -Es"?

sweet ! thank you Thomas for the hint... pity (and strange) that
install  command doesn't support --executable 

-- 
Yaroslav O. Halchenko, Ph.D.
http://neuro.debian.net http://www.pymvpa.org http://www.fail2ban.org
Research Scientist,            Psychological and Brain Sciences Dept.
Dartmouth College, 419 Moore Hall, Hinman Box 6207, Hanover, NH 03755
Phone: +1 (603) 646-9834                       Fax: +1 (603) 646-1419
WWW:   http://www.linkedin.com/in/yarik

opoplawski:
[...]

PR #1426, opened 2016-05-15
no hardcoded python interpreter path

koeppea:
Under FreeBSD and also on other systems, pyhton is not located as it's hardcoded in the shebang of the start scripts.
Therefore this PR aims for makeing the shebang more generic by leveraging the helper binary **env**.

### DIFF
--- a/fail2ban/client/fail2banregex.py
+++ b/fail2ban/client/fail2banregex.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: t -*-
 # vi: set ft=python sts=4 ts=4 sw=4 noet :
 #
--- a/fail2ban/client/fail2banregex.py
+++ b/fail2ban/client/fail2banregex.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: t -*-
 # vi: set ft=python sts=4 ts=4 sw=4 noet :
 #
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: t -*-
 # vi: set ft=python sts=4 ts=4 sw=4 noet :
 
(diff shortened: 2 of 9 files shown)


yarikoptic:
Usually I was against using env'ed python in shebangs... And we even had discussions on this before. But recently I saw even more radical move to use `!python `, eg https://github.com/nipy/nibabel/blob/master/bin/nib-dicomfs to rely on change of shebang by setuputils upon installation. Similarly distribution specific helpers would adjust even this enved shebang... So may be it does make sense...

yarikoptic:
Interestingly http://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my mentions that some exotic systems don't carry /use/bin/env . @fail2ban/maintainers - anything fresh against using env'ed shebang?

koeppea:
Just experienced it when starting testing on FreeBSD, hence thought I'm not the only one. Of course, package maintainers may invisibly correct such platform specific things.
I can as well live with always changing the shebang when testing on other platforms. It's up to you.

chtheis:
On FreeBSD the port system changes the shebang line to where the 
(default) python version is installed during the stage phase.

yarikoptic:
Related: #990 (use of `python -Es`) -- which would then require direct `/usr/bin/pyhon` use.  Even if decided to be so, would be done only within unified distributions and not good for a portable (upstream) installation.
Since ATM we do not require any flags to the interpreter, distributions would replace shebang with the most appropriate at packaging/stage phase, and we would like to aim for the most portable installation, I would second to accept this PR.  If no objections - I (or someone else) will merge within a day or so

koeppea:
Finally of course. Through the maintainer.
But of one wants to test or use the latest version from github directly, then the shebang needs to be changed manually. 
As I said, a matter of opinion.

fail2ban/client/fail2banregex.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
```
yarikoptic: @sebres this one is not intended to be executed directly right? Then we don't need shebang here and should remove executable but if set

koeppea:
the executable bit is set at the moment. let me know then I can correct this quickly.

fail2ban/client/fail2banregex.py
sebres: right

yarikoptic:
per https://github.com/fail2ban/fail2ban/pull/1426/files/975608dfb629c637622f0d638335637c1c688970#r63496051 indeed not needed there, please correct/update PR

koeppea:
Why the heck has this commit decreased the coverage?

yarikoptic:
;-)

yarikoptic:
for some reason coveralls do not ignore comment/shebang line, so the ratio of uncovered lines of code increased because total number of lines decreased ;)  merging now

PR #1425, opened 2016-05-14
ENH/TST: Branch: 0.10 - firewallcmd action tests

TorontoMedia:
- added ipmset to firewallcmd-common.
- added 3 firewallcmd action tests[ firewallcmd-new,  firewallcmd-rich-rules,  firewallcmd-rich-logging ] in 

>  testCheckStockCommandActions

PR #1429, opened 2016-05-17
split PF config files between all- and multi port

koeppea:
This is a consequence for matter of fact, that "any" is a invalid specification for "port" in the pfctl rule.
So this PR just splits up the PF action configuration files based on the model of the *iptables-allports.conf* and *iptables-multiport.conf* files.


sebres:
Here https://github.com/fail2ban/fail2ban/pull/1425#issuecomment-219284809 we've discussed how we can reduce number of actions... And you add almost the same action as pf. :)))

koeppea:
When trying to use `banaction = pf[actionoptions=<multiport>]`, <name> in the command isn't replaced, rather than appended somewher in between:
```
2016-05-17 16:32:02,538 fail2ban.utils          [1314]: ERROR   echo "table <f2b-<name>> persist counters" | pfctl -f-
echo "block proto tcp from <f2b-<name>> to any port ssh];[name=sshd" | pfctl -f- -- returned 1
```

sebres:
Hmm, strange...
I think it could be a different order in dict supplied to substituteRecursiveTags, that causes that name gets lost (by \<\> around), what may be actual an error (because has not correct direct replacement)

sebres:
BTW you've added a commit 802d1c5a0f36dd487687221a5c7913a5b20e05b6 here, that, imho, belong to another PR

sebres:
@koeppea support for multiple option groups is ready in #1431
If you rebase or squash your branch `0.10-freebsd-fix-pf` over this commit, it should then running without errors (don't forget to skip 802d1c5a0f36dd487687221a5c7913a5b20e05b6 that would be wrong here)

koeppea:
> (don't forget to skip 802d1c5 that would be wrong here)

oh yes you're right. This comes from manually patching fixing the shebang issue. Commit removed.
I rebased it over #1431 and will test it on FreeBSD now....

koeppea:
Perfect. Works perfectly fine.
Ban triggered via v4, DNS reverse, and forward lookup, and all IPs blocked at once.
Let me check the travis issue...

sebres:
Already, I've made a PR for your branch...

sebres:
One reads not his mail :smile:

koeppea:
> One reads not his mail

Your're right. Had to struggle a little fight with FreeBSD before I could dig into it.
Hopefully the tests go well so this PR should be fine then.

koeppea:
I think I will never understand coveralls. A file is now claimed which hasn't been touched by this PR....

sebres:
Well it could be: some except in previous commit by any test case has covered 2 lines more:))
Corresponding this one 2 lines fewer...

config/action.d/pf.conf
```diff
@@ -16,7 +16,7 @@
 # we don't enable PF automatically; to enable run pfctl -e 
 # or add `pf_enable="YES"` to /etc/rc.conf (tested on FreeBSD)
 actionstart = echo "table <<tablename>-<name>> persist counters" | pfctl -f- 
-              echo "block proto <protocol> from <<tablename>-<name>> to any port <port>" | pfctl -f-
+              echo "block proto <protocol> from <<tablename>-<name>> to any<actiontype>" | pfctl -f-
```
yarikoptic: Shouldn't there be a space before `<actiontype>`?

yarikoptic: Ah, I see now :-)
Please just add to the doc of actiontype that it is appended as is without any space separation (is there any negative effect from a trailing space BTW? Just wondering... Since we could also simply just strip lines on the right I guess before execution...

koeppea: I think it doesn't harm and would improve readablity.
But I think the intention comes through the tests.

sebres: > Since we could also simply just strip lines on the right I guess before execution...

Impossible here, because after the space we've piping... But the test cases are exact :)

yarikoptic: >  > -              echo "block proto <protocol> from <<tablename>-<name>> to any port <port>" | pfctl -f-
>  > +              echo "block proto <protocol> from <<tablename>-<name>> to any<actiontype>" | pfctl -f-
>      Since we could also simply just strip lines on the right I guess before
>      execution...
>    Impossible here, because after the space we've piping... But the test
>    cases are exact :)

d'oh -- right!   but then here it is ok to just leave space after any
and have "any <actiontype>" without requiring any specific leading
"<sp>" in actiontype, which could easily be omitted by users in their
"customizations" etc.  I am just trying to get it more obvious -- you
already saw how ppl can shoot them in the foot with even simpler more
obvious cases ;)

sebres: Of course we can do that, but I would not so "accurate" with possible evil customizations in actions, BUT not in filters:
- if action defined wrong, it does not work (and shits with errors in f2b.log);
- if filter defined wrong, it will never find some failures (and can easily be overlooked, because no errors directly)

yarikoptic: Yeap, correct, but here we are talking about an action. If relevant,  I am not arguing against utility of `<sp>`. I am taking about this surviving case, where demand on having it explicitly not warranted since we can have that space hardcoded above

sebres: Good, persuaded... I'll correct this directly in PR-branch and then push which merge directly to 0.10.

yarikoptic: >    Good, persuaded... I'll correct this directly in PR-branch and then push
>    which merge directly to 0.10.
ok

sebres: done with 4cdca8c258befd6848cef61d12bbcba224372072

PR #1431, opened 2016-05-17
jailreader: support multiple option groups, syntax `action = act[p1=.…

sebres:
jailreader: support multiple option groups, syntax `action = act[p1=...][p2=...]` + test case for it
(see https://github.com/fail2ban/fail2ban/pull/1425#issuecomment-219285524, gh-1429)


yarikoptic:
First I thought that may be requiring/supporting escaping withing passed values would be better, ... But I was wrong... So indeed such support of multiple groups would scale better... Thus +1 from me

PR #1433, opened 2016-05-19
BF: maintain previous default beh for pf -- default ban type is multiport 

yarikoptic:
I felt that it is not time/place to change default behavior of pf banning a single port, as the majority (if not all) actions do by default.  So I have complemented to merged #1429, and reverted back to original behaviour -- ban a single port, BUT changed which port to ban (ssh, not telnet). Adjusted doc also to note that port specification is desired if multiport is used

- [ ] **AVOID** making unnecessary stylistic changes in unrelated code (did remove one trailing space thing right before my changes)

### DIFF
--- a/config/action.d/pf.conf
+++ b/config/action.d/pf.conf
@@ -72,10 +72,6 @@ tablename = f2b
 protocol = tcp
 
 
-# Option:  port
-# Notes.:  specifies port to monitor
-# Values:  [ NUM | STRING ]  Default:
-port = ssh
 
 # Option: actiontype
 # Notes.: defines additions to the blocking rule
@@ -90,5 +86,5 @@ allports = any
 
 # Option: multiport
 # Notes.: addition to block access only to specific ports
-# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>, port="%(port)s"]"
+# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>]"
 multiport = any port <port>
--- a/config/action.d/pf.conf
+++ b/config/action.d/pf.conf
@@ -75,14 +75,13 @@ protocol = tcp
 # Option:  port
 # Notes.:  specifies port to monitor
 # Values:  [ NUM | STRING ]  Default:
-#
-#port = telnet
+port = ssh
 
 # Option: actiontype
 # Notes.: defines additions to the blocking rule
 # Values: leave empty to block all attempts from the host
-# Default: Value of the allports
-actiontype = any
+# Default: Value of the multiport
+actiontype = <multiport>
 
 # Option: allports
 # Notes.: default addition to block all ports
@@ -91,5 +90,5 @@ allports = any
 
 # Option: multiport
 # Notes.: addition to block access only to specific ports
-# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>]"
+# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>, port="%(port)s"]"
 multiport = any port <port>
--- a/fail2ban/tests/servertestcase.py
+++ b/fail2ban/tests/servertestcase.py
@@ -1370,14 +1370,14 @@ class ServerConfigReaderTests(LogCaptureTestCase):
 					),
 					'ip6-unban': (
 						r"`echo -2001:db8:: > /proc/net/xt_recent/f2b-j-w-iptables-xtre6`",
-					),					
+					),
 				}),
-				# pf allports --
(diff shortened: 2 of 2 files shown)


config/action.d/pf.conf
```diff
@@ -75,14 +75,13 @@ protocol = tcp
 # Option:  port
 # Notes.:  specifies port to monitor
 # Values:  [ NUM | STRING ]  Default:
-#
-#port = telnet
+port = ssh
```
sebres: I don't like such defauts values:
I think better it fails, as silently bans the IP for wrong port :)))
If used multiport, the port SHOULD be defined in jail, imho.

config/action.d/pf.conf
```diff
@@ -91,5 +90,5 @@ allports = any
 
 # Option: multiport
 # Notes.: addition to block access only to specific ports
-# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>]"
+# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>, port="%(port)s"]"
```
sebres: You are wrong here: the `port` will be specified through `action_` or `action`, but no longer with `banaction`.

config/action.d/pf.conf
sebres: Ahh, I see the port defined here will be not used at all, because uses always the port defined in jail.conf (or .local)

config/action.d/pf.conf
yarikoptic: >  > -# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>]"
>  > +# Usage.: use in jail config: "banaction = pf[actiontype=<multiport>, port="%(port)s"]"
>    You are wrong here: the port will be specified through action_ or action,
>    but no longer with banaction.

yes -- you are right!  didn't spot that it was for banaction and not for
action.  I will revert this change, thanks!

config/action.d/pf.conf
yarikoptic: >  > -#port = telnet
>  > +port = ssh
>    Ahh, I see the port defined here will be not used at all, because uses
>    always the port defined in jail.conf (or .local)

yes!  This is just a "sensible default".  And in general agree, as with
you other comment, ideally we fail than operate on some "random" port.
But if we are to make such decision to change behavior, we better apply
it across all actions, not just one at a time.  That is why this change
was to stay consistent with other actions/current "ideology".

sebres:
Please take a look in my branch: remotes/sebres/yarikoptic/bf-0.10-pf-prevbeh
I've rebased your commit as continuation to the #1429 branch and added my commit as possible solution...

yarikoptic:
>    Please take a look in my branch:
>    remotes/sebres/yarikoptic/bf-0.10-pf-prevbeh
>    I've rebased your commit as continuation to the #1429 branch and added my
>    commit as possible solution...

I will look in the evening... need to carry out main work ATM

koeppea:
> (ssh, not telnet)

this was just to double check the replacement from the jail. **ssh** for sure makes more sense in most cases.

> Ahh, I see the port defined here will be not used at all, because uses always the port defined in jail.conf (or .local)

That's actually what I tried to tell in 4cdca8c.

yarikoptic:
@sebres regarding the rebased/"fixed" branch -- in general I could only repeat  https://github.com/fail2ban/fail2ban/pull/1433#discussion_r63902594 (i.e. I would have preferred to have it done uniformly across all actions which assume the port).  But I don't care much enough atm about this one to insist. I will repush your branch into this PR now.

config/action.d/pf.conf
sebres: I think, the port (as well as protocol) defined in action self, should't be used at all, because there are defined in definition section (that normally expected `%(port)s` substitution) and currently both parameters will be substituted using tag syntax `<port>`, that are external options (or options from init section).

koeppea: But wouldn't this apply then as well for *iptables-common.conf*?

yarikoptic: >  > -#port = telnet
>  > +port = ssh
>    But wouldn't this apply then as well for *iptables-common.conf*?
That is exactly what I was saying that many of our actions do carry
defaults for every variable they potentially use, even if we expect them
to be (re)defined from jail action specs.  That is why I thought it
would be nicer to make such (remove default definitions for those which
we expect to be redefined explicitly from the jails) change in 1 step.

yarikoptic:
I guess it just awaits your blessing/merge @sebres

Issue #1434, opened 2016-05-19
Gentoo init script fails with "start-stop-daemon: start does not exist"


vwegert:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.9.4
- OS, including release name/version: Gentoo (Portage 2.2.28)
- [X] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)

### The issue:

I've upgraded my system to net-analyzer/fail2ban-0.9.4 (among other packages) as well as performing a kernel update, and after the reboot, fail2ban refuses to start. The error message:

```
# /etc/init.d/fail2ban start
* Starting fail2ban ...
* start-stop-daemon: start does not exist
* Failed to start fail2ban                                                                                                                                                                                 * ERROR: fail2ban failed to start
```

#### Any additional information

Running the init script in debug mode yields

```
# /etc/init.d/fail2ban -d start
+ _conf_d=/etc/init.d/../conf.d
+ _c=fail2ban
+ '[' -n fail2ban -a fail2ban '!=' fail2ban ']'
+ unset _c
+ source -e /etc/init.d/../conf.d/fail2ban.default
+ '[' -e = -e ']'
+ shift
+ '[' -e /etc/init.d/../conf.d/fail2ban.default ']'
+ return 1
+ source -e /etc/init.d/../conf.d/fail2ban
+ '[' -e = -e ']'
+ shift
+ '[' -e /etc/init.d/../conf.d/fail2ban ']'
+ . /etc/init.d/../conf.d/fail2ban
++ FAIL2BAN_OPTIONS=-x
+ unset _conf_d
+ source -e /etc/rc.conf
+ '[' -e = -e ']'
+ shift
+ '[' -e /etc/rc.conf ']'
+ . /etc/rc.conf
++ rc_shell=/sbin/sulogin
++ unicode=YES
++ rc_tty_number=12
+ '[' -d /etc/rc.conf.d ']'
+ source /lib64/rc/sh/s6.sh
+ '[' /lib64/rc/sh/s6.sh = -e ']'
+ . /lib64/rc/sh/s6.sh
++ '[' -z '' ']'
++ s6_service_path=/var/svc.d/fail2ban
+ source /lib64/rc/sh/start-stop-daemon.sh
+ '[' /lib64/rc/sh/start-stop-daemon.sh = -e ']'
[...]

PR #1322, opened 2016-02-15
gentoo-initd: Use start-stop-daemon in order to handle crashes better

jsuter1:
Currently, if fail2ban is killed (or crashes), its status will be
reported by '/etc/init.d/fail2ban status' as 'running' even though it
is not. Attempting to restart the service also fails, because Gentoo
unsuccessfully tries to stop the service.

By using start-stop-daemon and providing a pidfile, Gentoo will
instead report the status as 'crashed' and allow the service to be
restarted as normal.

PR #1435, opened 2016-05-19
fixed --pidfile bug, introduced in gh-1322

sebres:
gentoo-initd fixed --pidfile bug: `--pidfile` is option of start-stop-daemon, not argument of fail2ban (see gh-1434)
closes gh-1434


sebres:
I was not yet ready, because `--exec` expected command line (I think exact one parameter), but no command with parameters. Amended a fix to branch and pushed replacement to master.

yarikoptic:
Good. Sorry that I have rushed it... assigned that it was tested... we need a test farm of dockers for such tests :-/

PR #1436, opened 2016-05-20
extremely speedup of all database operations:

sebres:
- (synchronous = OFF) write data through OS without syncing
- (journal_mode = MEMORY) use memory for the transaction logging
- (temp_store = MEMORY) temporary tables and indices are kept in memory

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,11 @@ ver. 0.9.5 (2016/XX/XXX) - wanna-be-released
    * New Actions:
      - action.d/firewallcmd-rich-rules and action.d/firewallcmd-rich-logging (gh-1367)
 - Enhancements:
+   * Extreme speedup of all sqlite database operations (gh-1436),
+     by using of following sqlite options:
+     - (synchronous = OFF) write data through OS without syncing
+     - (journal_mode = MEMORY) use memory for the transaction logging
+     - (temp_store = MEMORY) temporary tables and indices are kept in memory
    * journald journalmatch for pure-ftpd (gh-1362)
    * Add additional regex filter for dovecot ldap authentication failures (gh-1370)
    * added additional regex filters for exim (gh-1371)
--- a/fail2ban/server/database.py
+++ b/fail2ban/server/database.py
@@ -181,12 +181,24 @@ class Fail2BanDb(object):
 				filename, e.args[0])
 			raise
 
+		# differentiate pypy: switch journal mode later (save it during the upgrade), 
+		# to prevent errors like "database table is locked":
+		try:
+			import __pypy__
+			pypy = True
+		except ImportError:
+			pypy = False
+
 		cur = self._db.cursor()
 		cur.execute("PRAGMA foreign_keys = ON")
 		# speedup: write data through OS without syncing (no wait):
 		cur.execute("PRAGMA synchronous = OFF")
 		# speedup: transaction log in memory, alternate using OFF (disable, rollback will be impossible):
-		cur.execute("PRAGMA journal_mode = MEMORY")
+		if not pypy:
+			cur.execute("PRAGMA journal_mode = MEMORY")
+		# speedup: temporary tables and indices are kept in memory:
+		cur.execute("PRAGMA temp_store = MEMORY")
+
 		try:
 			cur.execute("SELECT version FROM fail2banDb LIMIT 1")
 		except sqlite3.OperationalError:
@@ -205,6 +217,9 @@ class Fail2BanDb(object):
 						Fail2BanDb.__version__, version, newversion)
 					raise RuntimeError('Failed to fully update')
 		finally:
+			# pypy: set journal mode after possible upgrade db:
+			if pypy:
+				cur.execute("PRAGMA journal_mode = MEMORY")
 			cur.close()
 
 	@property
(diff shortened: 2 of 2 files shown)


sebres:
Just currently realized, that I somehow forgot a long time implemented enhancement to speedup the database in fail2ban. It is already more as year by me (and my customers) in production, so at all production-ready. The feature should be in 0.10 but somehow this branch was not merged (sorry :)))

Have rebased it now in master, thus users of 0.9 benefit from also.

It can be also a good alternate to FR #1368 

You can see here how quickly became the test cases (now really comparable with pure in memory db)
```
# without speedup-fix:
bin/fail2ban-testcases database
Fail2ban 0.10.0a1 test suite. Python 2.7.6 (default, Jun 22 2015, 17:58:13) [GCC 4.8.2]. Please wait...
..................
Ran 18 tests in 10.129s

# with speedup-fix:
bin/fail2ban-testcases database
Fail2ban 0.10.0a1 test suite. Python 2.7.6 (default, Jun 22 2015, 17:58:13) [GCC 4.8.2]. Please wait...
..................
Ran 18 tests in 0.071s
```

The same is not really visible in 0.9 test cases, because they are too slow there (too many sleep)

sebres:
The risks (really minimal) are known to me, and the most unlikely case of loss of the database (by crash of the server) can we really neglect (there are to create the backup of database, if it is really important).

sebres:
This travis and both coverages make one really crazy.

BTW. I suspect, at some point, we need to rewrite the badips test cases to something artifical (like our smtp-test-server) or something several, because in many cases it is a reason for all the timing problems

yarikoptic:
Ideally I want also to have a buildbot running tests across platforms and reporting as well ;-) see eg how it looks in datalad project

Badips, we need indeed to mock it;-)

On May 20, 2016 7:43:19 AM EDT, "Serg G. Brester" <notifications@github.com> wrote:
>This travis and both coverages make one really crazy.
>
>BTW. I suspect, at some point, we need to rewrite the badips test cases
>to something artifical (like our smtp-test-server) or something
>several, because in many cases it is a reason for all the timing
>problems
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/fail2ban/fail2ban/pull/1436#issuecomment-220584569

-- 
Sent from a phone which beats iPhone.

ChangeLog
```diff
@@ -30,6 +30,7 @@ ver. 0.9.5 (2016/XX/XXX) - wanna-be-released
    * New Actions:
      - action.d/firewallcmd-rich-rules and action.d/firewallcmd-rich-logging (gh-1367)
 - Enhancements:
+   * extremely speedup of all sqlite database operations (gh-1436)
```
yarikoptic: Extreme ... by using temp_store memory

Also while at it might be worth unifying leading casing

Shouldn't it be an option in fail2ban.conf so if it doesn't work for them they could quickly disable it?

sebres: > Extreme ... by using temp_store memory

There are 3 pragmas in this FR, and temp_store currenly increase speed of upgrade only
A reference to PR is in the same line (gh-1436) where described all this, otherwise there is `git blame`, that can show all the changes;

> Shouldn't it be an option in fail2ban.conf so if it doesn't work for them they could quickly disable it?

:+1: 
I'll do that in 0.10 together with any other options, that I promised (and then cherry pick back this options to 0.9)

fail2ban/server/database.py
```diff
@@ -201,6 +216,11 @@ def __init__(self, filename, purgeAge=24*60*60):
 						": updated from '%i' to '%i'",
 						Fail2BanDb.__version__, version, newversion)
 					raise RuntimeError('Failed to fully update')
+
+			# pypy: set journal mode after possible upgrade db:
+			if pypy:
+				cur.execute("PRAGMA journal_mode = MEMORY")
+
```
tjikkun: Won't this mean that if pypy is true and db does not exist yet that "PRAGMA journal_mode = MEMORY" will not be done?
I think you need less indentation here

ChangeLog
yarikoptic: >    There are 3 pragmas in this FR, and temp_store currenly increase speed of
>    upgrade only
>    A reference to PR is in the same line (gh-1436) where described all this,
>    otherwise there is git blame, that can show all the changes;
I (as a user) would hate to be demanded to figure out what git, where
the PR, etc for every change whenever upgrading.  So all that doesn't
warrant opaque changelog msgs -- there should be some kind of a summary
on what was done, not only what was achieved since there might be
multiple things which could lead to speed up.  In this case it is the
memory journaling/caching, and thus should be at least briefly mentioned

yarikoptic: >      Shouldn't it be an option in fail2ban.conf so if it doesn't work for
>      them they could quickly disable it?
>    I'll do that in 0.10 together with any other options, that I promised (and
>    then cherry pick back this options to 0.9)
IMHO this should be atomic with this change/PR.  awaiting on it in 0.10
and then cherry-picking sounds like more work for no benefit.

sebres: > sounds like more work for no benefit

No, because I should then just remove pure 0.10-options, that are not involved here.

fail2ban/server/database.py
sebres: Thx, you are right (it was done only by upgrade, what is very rare), moved in finally scope now

ChangeLog
sebres: Oh yeah, about of the ChangeLog, extended now

sebres:
Updated https://github.com/fail2ban/fail2ban/pull/1436#issuecomment-220571011 with 0.9 test examples - 8.5s vs. 0.06s

Previously I had tried it with a wrong filter, that has pinched the cases, that were too slow in 0.9 (made too many sleep, that was in 0.10 no matter, because few sleep at all).

fail2ban/server/database.py
```diff
@@ -202,6 +217,9 @@ def __init__(self, filename, purgeAge=24*60*60):
 						Fail2BanDb.__version__, version, newversion)
 					raise RuntimeError('Failed to fully update')
 		finally:
+			# pypy: set journal mode after possible upgrade db:
+			if pypy:
+				cur.execute("PRAGMA journal_mode = MEMORY")
```
yarikoptic: just expressing my excitement that we do still have pypy3 in travis and thus this line is test covered!
BTW -- are you guys using codecov extension for firefox?  whenever it works -- it is wonderful.  I can see what is covered or not right here in the PR on github : http://www.onerussian.com/tmp/gkrellShoot_05-20-16_094317.png   and here all of the thick greens are for the new code which is covered.

yarikoptic:
I will wait for the config option before releasing any of the branches! 😉

PR #1443, opened 2016-05-21
correct MANIFEST file

koeppea:
This PR just corrects the path of **ipdns.py** in the MANIFEST file.


yarikoptic:
Bloody coveralls... need to figure out how to exclude those non python files, can't merge on the phone now

PR #1447, opened 2016-05-24
ENH/TST: Branch: 0.10 - firewallcmd and iptables actions plus servertestcase

TorontoMedia:
I don't how I closed my other request #1425 but here it is again but improved:
- combined firewallcmd actions and added dport action to firewallcmd.conf
- combined iptable actions
- added action tests. I left out a lot of them due to code size. I recommend creating a actiontest xml and load them from there instead.


TorontoMedia:
I'm having some issues with git. Some how a couple files had git code inserted into them and I thought I had fixed them as it works on my system but I guess not,  so I'll fix it tomorrow.

sebres:
Looks good, but:
- `splitwords` resp. current master merged (or cherry picked) somehow partially, because the test cases miss `splitwords`. This merge (or cherry pick) is not really expected in this PR, so I suggest, please rebase this branch without this commits
- the test cases show, that some defaults are unlike previuos values, for example by some tests the default protocol was changed from `all` to `tcp` (may be right, but...)

sebres:
Ah, and you has unresolved conflicts in source code.

TorontoMedia:
Sorry, I'm having some issues between origin, upstream and my local/remote branches. I'm going to rebase it in a couple hours and look over what happened in the tests regarding protocol.

Should I add the tests for logging on? It'll be a couple hundred lines.

sebres:
I've rebased your branch over new 0.10 (correct merged with master without conflicts), but many test cases fails there, because of usage of old actions like iptables-multiport, iptables-allports, etc.

I've repaired it, so it works now, but...

https://github.com/fail2ban/fail2ban/compare/0.10...sebres:TorontoMedia/combined

But I don't like it for several reasons.
I'll write my feedback later, because very busy currently

TorontoMedia:
Thanks for repairing it...I also fixed my local git repo and tracking so hopefully I don't have any more issues. 

I'm going to hold off working on it until further feedback.

sebres:
Here is a short feedback, why I don't like it (I meant not your "put-together-realisation" of iptables/firewallcmd, that I found alright, but the complete current solution at all):
- **Complexity** The complexity and partially inconsistency of the substitution:
  - the interpolation of `%(param)s` takes place only in respective configreader's (and directly includes) and will be then transferred to the server. It is impossible to specify or override this parameters via syntax `a = b[param=d]`. Only the passing of direct parameters is possible: `a = b[c=%(param)s]`
  - the interpolation of `<param>` takes place during execution of the action, but 99% of all parameters will be never modified. Good, it is cached now, but it's not all about performance.
  - all the parameters (whether used or not) will be transferred to the server also. If you look at current stream transferred to the server for one jail, you will see a big lot of that.
- **Relations** All the parameters are perfect structured in filter, actions and it includes in the client side, but will be transferred as unordered heap (I mean the stream) without any structure. Where 99% of all parameters will be repeated many times, direct and indirect (as part of something). But it may be transferred once without any repetition (I mean the stream should get relations).
- **Conditions** The current conditional realization makes impossible to do many things directly, so it will be workaround-ed like in your realization of parameter `logging=on|off` for iptables. So arise confusing constructs like `<logging-<logging>-actionban>`, `<logging-<logging>-actionunban>` etc.

My conclussion would be to freeze this PR (and further development of actions) as long as we don't reorganize our config readers and parameters structures in the client- and server-side processors.

sebres:
> I also fixed my local git repo

Can you push your current rebased branch `combined` to your origin, just to be up-to-date here in this PR.
```
git.exe push --force-with-lease origin combined
```

TorontoMedia:
I'm sorry I deleted and restore 'combined' by accident and forced an update and now I can't reopen the pr. I can leave this closed for now and pull again later in the future if warranted.

sebres:
If the branch was deleted, the PR will be auto-closed by github. No way to reopen it then...

That's why I wrote the command line :)) 

You can namely rebase, reset, merge or whatever you will do with the remote origin branch, so the github actualized the PR automatically, but never delete this remote branch...

TorontoMedia:
I used your command minus .exe(ubuntu) but I have origin as fail2ban's git and upstream as my git but  I was getting ref and fast-forward errors and I accidentally deleted my combined branch...I'm still working on my git commands :)

TorontoMedia:
When I get a chance I can split them up to firewallcmd.conf and firewallcmd-logging.conf, would that work?

sebres:
> When I get a chance I can split them up ...

No, please, I'm not against your approach by putting its all together... Just read my feedback again. 
My problem here are the readers, client-server communication, streams, etc.

TorontoMedia:
No more actiontype?

sebres:
However, but the `actiontype` is almost mandatory thing. 
I've said about all the optional things.

TorontoMedia:
I don't see why splitting them into two files(.conf and -logging.conf) and moving some definitions out of ---common back to the action config would be much of a problem then?

sebres:
Because there would be nice:
- to have exact one action file for one type of processing (it begins only again with 2 files, tomorrow it will be 10 again);
- to have better conditional handling
- to bring a little bit structure in the client/server streams
- (maybe) better to distinguish between `%(param)s` and `<param>`, to handle allmost static, not modified parameters as `%(param)s`, and dynamical like conditional resp. run-time set from the server by executing of action as `<param>`;

Issue #67, opened 2012-07-29
RF to allow arbitrary set of keys (not just HOST) to track the failures


yarikoptic:
From the mailing list:
http://sourceforge.net/mailarchive/forum.php?thread_name=50143F62.6050904%40whyscream.net&forum_name=fail2ban-users

```
> On 28-07-12 16:18, Gerry Magill wrote:
> > VoIP has to go through a SIP aware firewall in My scenario to resolve
> > NAT-traversal issues with the SIP protocol. So my SIP Registrar sees all
> > connections as coming from my SBCs IP. Each new distinct user appears to
> > me as traffic from a new port. This port binding never changes until my
> > SBC is rebooted.

> > I can identify SIP SPIT and SPAM using snort but blocking the whole IP
> > would kill my SBC.

> > So I am looking to black list IP port bindings I identify as being
> > malicious and release them after 3 days in jail.
```

So this use case would require to track not only by IP but by (IP, port) pair.  We could implement this based on the named groups in the failregex (ATM only ```P<host>``` is utilized) and pass all those to the actions complementing the variables defined in Init and passed into the jail definition

chetan-prime:
I have a particular requirement with my email server to detect compromised users regardless of IP. I have a regex ready to detect when a user is sending emails and a script ready to disable that user temporarily.
All I need is a way to run the user disable script with fail2ban passing the email address captured using the regex.
Ideally this requires changes to fail2ban to allow grouping matches by a regex variable,or to allow setting <HOST> to a named regex match, so fail2ban can detect if the same user is sending emails too fast (irrespective of IP)
#1110

tguild:
Another person who needs more data from matched log line than just IPv4 address.  Hope this gets released soon and surprised this has been open for nearly 4 years.

sebres:
@chetan-prime please test #1454

sebres:
Issue can be reopened later:
- if someone needs multiple groups (the "host" as ban-identifier should be accumulated from multiple "tags")
- there is no unique ban identifier (user, pair ip:port, mac-address) available or there should not be unique: for example "banning" process does not really ban, but should just execute some action after X failures (for example just sends a mail to admin);

chetan-prime:
Thank you @sebres , this looks exactly like what I wanted. I'm testing and will report back in case of issues

sebres:
[...]

PR #1454, opened 2016-05-30
[gh-67] no-host ban

sebres:
Partially closes gh-67:
- there is any unique ban identifier available (ban-id)
- this ban-id (user, ip:port pair, something else) stand together and can be retrieved with exact one group

Example for test jail to ban users, config `jail.local`:
```bash
[test]
# don't use dns, because host group is not hostname (and not resolvable ip):
usedns = raw
ignoreip =
ignorecommand =
# if used some filter:
#filter = no-host-filter[HOST=(?P<host>\S+)]
#
# if used failregex:
filter =
failregex = ^\s*(?:\S+\s+)?(?:[^:]+:auth\[\d+\]:\s+)?pam_unix(?:\(\S+\))?:?\s+authentication failure; login=(?P<host>\S+)
# action that bans users
banaction = test-ban-user[name=%(__name__)s]
#
logpath  = %(syslog_authpriv)s
enabled = true
```
Action config file `action.d/test-ban-user.conf`:
```bash
[Definition]
actionstart = 
actionstop =
actioncheck =
actionban = echo 'ban f2b-<name> --user <ip>'
actionunban = echo 'unban f2b-<name> --user <ip>'
```

To test, the user "xxx" will be banned, just execute:
```bash
logger -t 'test:auth' -i -p auth.info "pam_unix(test:auth): authentication failure; login=xxx"
logger -t 'test:auth' -i -p auth.info "pam_unix(test:auth): authentication failure; login=xxx"
logger -t 'test:auth' -i -p auth.info "pam_unix(test:auth): authentication failure; login=xxx"
```

To test regular expression, use new option `--raw` or `-r`, to prevent dns resolving errors:
```bash
fail2ban-regex --raw /var/log/auth.log '^\s*(?:\S+\s+)?(?:[^:]+:auth\[\d+\]:\s+)?pam_unix(?:\(\S+\))?:?\s+authentication failure; login=(?P<host>\S+)'
```

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -94,6 +94,7 @@ backend = auto
 #        but it will be logged as a warning.
 # no:    if a hostname is encountered, will not be used for banning,
 #        but it will be logged as info.
+# raw:   use raw value (no hostname), allow use it for no-host filters/actions (example user)
 usedns = warn
 
 # "logencoding" specifies the encoding of the log files handled by the jail
--- a/fail2ban/client/fail2banregex.py
+++ b/fail2ban/client/fail2banregex.py
@@ -126,6 +126,8 @@ Report bugs to https://github.com/fail2ban/fail2ban/issues
 			   help="set custom pattern used to match date/times"),
 		Option("-e", "--encoding",
 			   help="File encoding. Default: system locale"),
+		Option("-r", "--raw", action='store_true',
+			   help="Raw hosts, don't resolve dns"),
 		Option("-L", "--maxlines", type=int, default=0,
 			   help="maxlines for multi-line regex"),
 		Option("-m", "--journalmatch",
@@ -239,6 +241,7 @@ class Fail2banRegex(object):
 			self.encoding = opts.encoding
 		else:
 			self.encoding = locale.getpreferredencoding()
+		self.raw = True if opts.raw else False
 
 	def decode_line(self, line):
 		return FileContainer.decode_line('<LOG>', self.encoding, line)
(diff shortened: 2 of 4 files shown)


sebres:
Because `usedns = no` should ignore no ip failures, new value `raw` introduced here for option `usedns`

yarikoptic:
I think this is a nice quick and dirty workaround indeed.

sebres:
I've another one for that, that is also "quick" but fewer "dirty", but I want not "publish" it as long as we've not merged 0.10 (and 0.10-cs).

yarikoptic:
haven't merged into where -- into master?    then -- "until 0.10 becomes stable"?

sebres:
No, I meant until 0.10 will not be released

sebres:
@yarikoptic I know, I know - PR etc. - not to annoy you, but because this PR makes no sense without this amend-fix b853474, because `usedns` goes with fallback again to `no`, so the fail2ban will ignore such failures (because are not valid)

the0ne:
This is **fantastic**! Finally i can use fail2ban for all the server tasks i want that i could not use it for because those logs don't have IP numbers or host names in it. Now fail2ban is a real **swiss-army-knife**!

As the server runs on a debian variant i have ported back the changes to debian/0.8.11-1
I tried creating a pull request to that version but i somehow could not get it to work. So the required changes based on Ubuntu 14.04.5 LTS can be found here:
https://github.com/the0ne/fail2ban/tree/debian---0.8.11-1---gh-67-ban-wo-host

sebres:
glad it works!
BTW, we don't plan to support 0.8 (at least such major changes), our favour branch is 0.10, and 0.9 will be supported still 0.10 grews up from children slacks

PR #1455, opened 2016-06-01
Update Kerio Connect filter

herrbischoff:
Fixed regex for some log entries that did not get recognized and some additional error formats are added.

### DIFF
--- a/config/filter.d/kerio.conf
+++ b/config/filter.d/kerio.conf
@@ -21,6 +21,6 @@ datepattern = ^\[%%d/%%b/%%Y %%H:%%M:%%S\]
 # DEV NOTES:
 # 
 # Author: A.P. Lawrence
-# Updated by M. Bischoff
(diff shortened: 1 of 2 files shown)


sebres:
The test cases are must (and you'll see, travis will fail).
So for each new failregex, you should add at least one log entry into https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/kerio that will match each expression.

herrbischoff:
Thanks for pointing that out. I'll get on that today as soon as I find the time.

sebres:
Still interrested to complete? Or should I close this PR (I've no kerio to test resp. you've not provided any log-excerpt)

herrbischoff:
Please excuse the long silence. I'll try to get to this in the coming days.

herrbischoff:
@sebres: I have added the missing test, all checks pass now. Please review.

sebres:
I've seen it. 
Looks good, I'll merge it in 24h if no objections follow

PR #1456, opened 2016-06-01
Asterisk pjsip

GMLudo:
Hi,

This PR adds the new format of logs with PJSIP and Asterisk 13+.

Thanks for the merge.

Have a nice day.

--
- [X] **CONSIDER adding a unit test** if your PR resolves an issue
- [X] **LIST ISSUES** this PR resolves
- [X] **MAKE SURE** this PR doesn't break existing tests 
- [X] **KEEP PR small** so it could be easily reviewed.
- [X] **AVOID** making unnecessary stylistic changes in unrelated code
- [X] **ACCOMPANY** each new `failregex` for filter `X` with sample log lines 
      within `fail2ban/tests/files/logs/X` file

PR #1458, opened 2016-06-09
Improve PJSIP log support for asterisk 13+ with different callID

nturcksin:
Hi,

This PR is a child of  Asterisk pjsip #1456 . We push the new filter on prod server and we see that sometimes the CallID can be different so we modify the asterisk's filter to match all callID.

Thanks for the merge.

Have a Nice day !

### DIFF
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -27,7 +27,7 @@ failregex = ^%(__prefix_line)s%(log_prefix)s Registration from '[^']*' failed fo
             ^%(__prefix_line)s%(log_prefix)s hacking attempt detected '<HOST>'$
             ^%(__prefix_line)s%(log_prefix)s SecurityEvent="(FailedACL|InvalidAccountID|ChallengeResponseFailed|InvalidPassword)",EventTV="([\d-]+|%(iso8601)s)",Severity="[\w]+",Service="[\w]+",EventVersion="\d+",AccountID="(\d*|<unknown>)",SessionID=".+",LocalAddress="IPV[46]/(UDP|TCP|WS)/[\da-fA-F:.]+/\d+",RemoteAddress="IPV[46]/(UDP|TCP|WS)/<HOST>/\d+"(,Challenge="[\w/]+")?(,ReceivedChallenge="\w+")?(,Response="\w+",ExpectedResponse="\w*")?(,ReceivedHash="[\da-f]+")?(,ACLName="\w+")?$
             ^%(__prefix_line)s%(log_prefix)s "Rejecting unknown SIP connection from <HOST>"$
-            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' \(callid: \w*\) - No matching endpoint found$
+            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' \(callid: .*\) - No matching endpoint found$
 
 ignoreregex =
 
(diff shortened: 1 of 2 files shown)


config/filter.d/asterisk.conf
```diff
@@ -27,7 +27,7 @@ failregex = ^%(__prefix_line)s%(log_prefix)s Registration from '[^']*' failed fo
             ^%(__prefix_line)s%(log_prefix)s hacking attempt detected '<HOST>'$
             ^%(__prefix_line)s%(log_prefix)s SecurityEvent="(FailedACL|InvalidAccountID|ChallengeResponseFailed|InvalidPassword)",EventTV="([\d-]+|%(iso8601)s)",Severity="[\w]+",Service="[\w]+",EventVersion="\d+",AccountID="(\d*|<unknown>)",SessionID=".+",LocalAddress="IPV[46]/(UDP|TCP|WS)/[\da-fA-F:.]+/\d+",RemoteAddress="IPV[46]/(UDP|TCP|WS)/<HOST>/\d+"(,Challenge="[\w/]+")?(,ReceivedChallenge="\w+")?(,Response="\w+",ExpectedResponse="\w*")?(,ReceivedHash="[\da-f]+")?(,ACLName="\w+")?$
             ^%(__prefix_line)s%(log_prefix)s "Rejecting unknown SIP connection from <HOST>"$
-            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' \(callid: \w*\) - No matching endpoint found$
+            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' \(callid: .*\) - No matching endpoint found$
```
sebres: Please don't use catch-all if it possible, defying anchor - in your case I doubt the callid can contain ")", so you can replace:
```diff
-            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' \(callid: .*\) - No matching endpoint found$
+            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' \(callid: [^\)]*\) - No matching endpoint found$
```
BTW, I self prefer non-anchored expressions, that has no catch-alls, because imho `.*\) - No matching endpoint found$` only have a meaning if the conclusion "No matching endpoint found" is important. Otherwise not anchored one would be good enough:
```
^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?' 
```

nturcksin:
I change the filter with @sebres observation.

Thanks you

sebres:
I looked at source code in https://github.com/asterisk/asterisk/blob/9e4efe9e1644938f256d2135003e8b9ccbf24104/res/res_pjsip/pjsip_distributor.c#L512 and have 2 findings:
- in this version `log_failed_request` will be executed within different cases with different messages:
  - No matching endpoint found
  - Not match Endpoint ACL
  - Not match Endpoint Contact ACL
  - Failed to authenticate
  - Error to authenticate

The question would be - is the message really not important?
- in this version `log_failed_request` will write little bit changed message, so the regexp should get optional part between "Request" and "from", that corresponding source code contains a method string:
```diff
-            ^%(__prefix_line)s%(log_prefix)s Request from '[^']*' failed for '<HOST>(?::\d+)?'
+            ^%(__prefix_line)s%(log_prefix)s Request (?:'[^']*')? from '[^']*' failed for '<HOST>(?::\d+)?'
```

GMLudo:
> The question would be - is the message really not important?

To my understanding, all errors could indicate an hack attempt, the difference are only the hack step process: For example, if the hacker has found a right account but not right password.

If you want to be sure, we can ask the question on asterisk-dev mailing-list.

>  the regexp should get optional part between "Request" and "from"

Good catch, thanks for the suggestion.

nturcksin:
updated with the optional part between "Request" and "from" .

nturcksin:
Hi,

Could you merge the PR ? :) 

Thanks a lot,

Nicolas

yarikoptic:
Fwiw paranoid me still dislikes relaxing regexp anchoring. If there is a limited number of possible messages I would prefer then all be listed, and then anchored at the end as it was before!

sebres:
@yarikoptic а я знал, я знал!

IMHO, it is alright so, but if the boss says...

nturcksin:
done with @yarikoptic observation and @sebres blessing :)

Added all log possibility from asterisk and log exemple

nturcksin:
it's possible to merge or i still have to do something ? sorry for the spam but we need it for our work

sebres:
I found it no so good:
- `[^']*` after callid is wrong, it should be `[^\)]*`
- almost the same regex can be grouped to one...

Will merge it soon with my changes above;

nturcksin:
Thanks :)

PR #1459, opened 2016-06-09
0.10 fid host ip tags

sebres:
Here is a continuation of #1454, resp. extended version of it for 0.10 branch:
Pros:
- possibility to differentiate between failure ID and failure IP on tag levels;
- possibility to define a start/end tag for each group (failure id, ip4, ip6, host, etc.);
- possibility to define different `failregex` for IPv4, IPv6 and for both (included dns host match);
- more precise regular expressions for IPv4 and IPv6 (and host) - separated now;

New tags:
- `<HOST>` - used new separated groups for `ip4`, `ip6`, `dns`;
- `<IP4>` (+ alias `<F-IP4/>`) - matches IPv4 only;
- `<IP6>` (+ alias `<F-IP6/>`) - matches IPv6 only;
- `<DNS>` (+ alias `<F-DNS/>`) - matches host name (usedns should be enabled);
- `<F-ID/>` - Failure ID, typically not IP-address (for example user name);
- `<F-PORT/>` - port, will be wrapped into the group "port";

New start/end tags to use it together around any free configurable own regexp:
- `<F-IP4>...</F-IP4>` - regexp between tags matches IPv4 only;
- `<F-IP6>...</F-IP6>` - regexp between tags matches IPv6 only;
- `<F-DNS>...</F-DNS>` - regexp between tags matches host name (usedns should be enabled);
- `<F-ID>...</F-ID>` - regexp between tags corresponds to failure ID (for example user name);
- `<F-PORT>...</F-PORT>` - regexp between tags corresponds to port, will be wrapped into the group "port";

Planned (TODO):
- additionally to the "solution" from #1454 a new substitution tags `<fid>`, `<ip4>`, `<ip6>`, `<port>`, `<dns>` should be used in addition to tag `<ip>` in action definitions.

### DIFF
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -533,21 +533,17 @@ class Filter(JailThread):
 						# failure-id:
 						fid = fail.get('fid')
 						# ip-address or host:
-						host = fail.get('ip4')
+						host = fail.get('ip4') or fail.get('ip6')
 						if host is not None:
 							raw = True
 						else:
-							host = fail.get('ip6')
-							if host is not None:
-								raw = True
-							else:
-								host = fail.get('dns')
-								if host is None:
-									# if no failure-id also (obscure case, wrong regex), throw error inside getFailID:
-									if fid is None:
-										fid = failRegex.getFailID()
-									host = fid
-									cidr = IPAddr.CIDR_RAW
+							host = fail.get('dns')
+							if host is None:
+								# if no failure-id also (obscure case, wrong regex), throw error inside getFailID:
+								if fid is None:
+									fid = failRegex.getFailID()
+								host = fid
+								cidr = IPAddr.CIDR_RAW
 						# if raw - add single ip or failure-id,
 						# otherwise expand host to multiple ips using dns (or ignore it if not valid):
 						if raw:
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -418,6 +418,9 @@ class Filter(JailThread):
 			ip = element[1]
 			unixTime = element[2]
 			lines = element[3]
+			fail = {}
+			if len(element) > 4:
+				fail = element[4]
 			logSys.debug("Processing line with time:%s and ip:%s", 
 					unixTime, ip)
 			if unixTime < MyTime.time() - self.getFindTime():
@@ -429,7 +432,7 @@ class Filter(JailThread):
 			logSys.info(
 				"[%s] Found %s - %s", self.jail.name, ip, datetime.datetime.fromtimestamp(unixTime).strftime("%Y-%m-%d %H:%M:%S")
 			)
-			tick = FailTicket(ip, unixTime, lines)
+			tick = FailTicket(ip, unixTime, lines, data=fail)
 			self.failManager.addFailure(tick)
 
 	##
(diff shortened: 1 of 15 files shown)


sebres:
Sometimes I hate github, coveralls and co... 
For example this PR should contain **exactly one** commit. And I see 10 currently... WTF

yarikoptic:
I think we just need to merge master into 0.10 first

sebres:
It was merged

yarikoptic:
you are correct... it indeed looks like a github glitch

fail2ban/server/filter.py
```diff
@@ -518,19 +526,45 @@ def findFailure(self, tupleLine, date=None, returnRawHost=False,
 						 % ("\n".join(failRegex.getMatchedLines()), timeText))
 				else:
 					self.__lineBuffer = failRegex.getUnmatchedTupleLines()
+					# retrieve failure-id, host, etc from failure match:
+					raw = returnRawHost
```
yarikoptic: a bit confused here with initializing 'raw' with returnRawHost... so if that one is True (return raw), the at line 553, een if no host address was specified (so you get dns), you will then at 554 call IPAddr  on that fid.  Sounds incorrect.  Note that lines 547 - 550  weren't covered by test (codecov extension rules) that is why may be you didn't trigger this behaviour

fail2ban/server/filter.py
```diff
@@ -518,19 +526,45 @@ def findFailure(self, tupleLine, date=None, returnRawHost=False,
...
 					try:
-						host = failRegex.getHost()
-						if returnRawHost:
-							failList.append([failRegexIndex, IPAddr(host), date,
-								 failRegex.getMatchedLines()])
+						fail = failRegex.getGroups()
+						# failure-id:
+						fid = fail.get('fid')
+						# ip-address or host:
+						host = fail.get('ip4')
```
yarikoptic: so is all this spagetti may be identical to
```python
host = fail.get('ip4') or fail.get('ip6') or fail.get('dns')
if host is None:
   # if no failure-id
   ...
```
?

yarikoptic: ok - may be first doing separate first or on ips to figure if 'raw'

fail2ban/server/filter.py
sebres: raw is exactly that: it should return failure-id, without checking it by resolve dns etc.
It is just:
- backwards compatibility to the current master, see the workaround in #1454 (using not new `fid` but `host`);
- raw handling in fail2ban-regex command line;
- possibility to add a host as failure (to put in raw in some special action, that can ban host names);

fail2ban/server/filter.py
sebres: Exactly: now, the pure IPs would be converted to IPAddr directly and only hosts (dns), if not raw, will be expanded to list of IPs via textToIp or ignored corresonding value of usedns;

BTW: I don't like such sugar, that "breaks" thoughts logic, as if it were boolean...

sebres: But pro rata you are right here, following code looks better in this case, so I will add a commit for it:
```diff
-			host = fail.get('ip4')
+			host = fail.get('ip4') or fail.get('ip6')
 			if host is not None:
 				raw = True
 			else:
-				host = fail.get('ip6')
-				if host is not None:
-					raw = True
-				else:
```

sebres:
I think, I've found a pragmatically solution for sporadically test failures in badips, just using increased timeout (30 seconds) inside our test cases - see e39126f6301fdf86348f43283b802dc28ad2a76f

This commit may be cherry-picked into master...

yarikoptic:
What about @szepeviktor comment? Shouldn't that suffix (not sure why it was part of this PR anyways) be optional?

yarikoptic:
Doh right this was github glitch and iirc did make it optional, so I will check actual state after merge, since who knows what other ramification of this gh behaviour is

PR #1460, opened 2016-06-10
0.10 ban-time-incr

### LINKED ISSUE #1267: Database will be not purged at all (currently no db.purge call, except test cases)
If we restart the computer where fail2ban is running then after the computer boots and fail2ban starts the database is not purged (old entries are still there and banned). So dbpurge value is ignored (in my case it is set in fail2ban.local).
It seems database purging must be performed when fail2ban starts (before starting to ban anything) to avoid this behavior.

sebres:
Currently included:
- ban-time-incr #716 (previously)
- cs-fix #1413 (already merged into 0.10)

Closes #1267


sebres:
Grrrr...

I suggest to exclude bad-ip test cases (BadIPsActionTest), as long as the stability of this "doubtful" service will not be good enough or we don't make a mokeup-dummy-server as replacement for it.

Another question would be why this one (foreign service) should be at all a part of fail2ban test cases.

coder4life:
I am curious if this plans to be released soon. Really could take advantage of increasing ban times.

mydexterid:
So.. this feature was planned for 0.10 but got postponed to 0.11?

Can it be released in 0.10 instead? Or can it be released at all soon?

sebres:
> Can it be released in 0.10 instead? 

No, it is impossible due to certain policies of several distributions (updates only in minor releases).
> Or can it be released at all soon?

Yes, it will be released soon.

mydexterid:
I see. Is there an issue about the 0.11 release, so that I can subscribe to it to be notified when released (closed)?

Although I don't think a new minor version will be included in current distribs (like Ubuntu Bionic)

sebres:
> Is there an issue about the 0.11 release, so that I can subscribe to it to be notified when released (closed)?

You don't need it. Just click on Watch button above (and select at least "Releases only", so you'll be notified if we'll create a new release here).

> Although I don't think a new minor version will be included in current distribs (like Ubuntu Bionic)

This is the matter of maintainers. But you'd then surely find several repositories where you can obtain it from.

asymmetric:
Are there any plans to release this anytime soon?

hostup:
Any planned release date?

asymmetric:
@sebres has this been released as part of 0.10.5? I can't see this PR number listed in the [Changelog](https://github.com/fail2ban/fail2ban/blob/0.10.5/ChangeLog).

sebres:
Nope, it is part of 0.11, which release is planned for tomorrow...

sebres:
Release [0.11.1](https://github.com/fail2ban/fail2ban/releases/tag/0.11.1) is out!

PR #1478, opened 2016-07-08
adding openldap slapd filter

gips0n:
adding openldap slapd filter with multi-line search functionality

### DIFF
--- a/config/filter.d/slapd.conf
+++ b/config/filter.d/slapd.conf
@@ -13,7 +13,7 @@ before = common.conf
 
 _daemon = slapd
 
-failregex = ^(?P<__prefix>%(__prefix_line)s).* conn=(?P<pid>\d+) fd=\d+ ACCEPT from IP=<HOST>\:\d+ .+$<SKIPLINES>(?P=__prefix).+ conn=(?P=pid) .* RESULT .* err=49 .*$
+failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s
 
 ignoreregex =
 
--- a/config/filter.d/slapd.conf
+++ b/config/filter.d/slapd.conf
@@ -3,9 +3,17 @@
 # Detecting invalid credentials: error code 49
 # http://www.openldap.org/doc/admin24/appendix-ldap-result-codes.html#invalidCredentials (49)
 
+[INCLUDES]
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
 [Definition]
 
-failregex = ^.* conn=(?P<pid>\d+) fd=\d+ ACCEPT from IP=<HOST>\:\d+ .+$<SKIPLINES>^.+ conn=(?P=pid) .* RESULT .* err=49 .*$
+_daemon = slapd
+
+failregex = ^(?P<__prefix>%(__prefix_line)s).* conn=(?P<pid>\d+) fd=\d+ ACCEPT from IP=<HOST>\:\d+ .+$<SKIPLINES>(?P=__prefix).+ conn=(?P=pid) .* RESULT .* err=49 .*$
 
 ignoreregex =
 
--- a/config/filter.d/slapd.conf
+++ b/config/filter.d/slapd.conf
@@ -13,7 +13,7 @@ before = common.conf
 
 _daemon = slapd
 
-failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s
+failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s$
 
 ignoreregex =
 
--- a/config/filter.d/slapd.conf
+++ b/config/filter.d/slapd.conf
@@ -1,0 +1,17 @@
+# slapd (Stand-alone LDAP Daemon) openldap daemon filter
+#
+# Detecting invalid credentials: error code 49
+# http://www.openldap.org/doc/admin24/appendix-ldap-result-codes.html#invalidCredentials (49)
+
+[Definition]
+
+failregex = ^.* conn=(?P<pid>\d+) fd=\d+ ACCEPT from IP=<HOST>\:\d+ .+$<SKIPLINES>^.+ conn=(?P=pid) .* RESULT .* err=49 .*$
+
+ignoreregex =
+
+[Init]
+
+# "maxlines" is number of log lines to buffer for multi-line regex searches
+maxlines = 20
+
+# Author: Andrii Melnyk
--- a/config/filter.d/slapd.conf
+++ b/config/filter.d/slapd.conf
@@ -13,7 +13,7 @@ before = common.conf
 
 _daemon = slapd
 
-failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s$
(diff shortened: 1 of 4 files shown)


config/filter.d/slapd.conf
```diff
@@ -0,0 +1,25 @@
...
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition]
+
+_daemon = slapd
+
+failregex = ^(?P<__prefix>%(__prefix_line)s).* conn=(?P<pid>\d+) fd=\d+ ACCEPT from IP=<HOST>\:\d+ .+$<SKIPLINES>(?P=__prefix).+ conn=(?P=pid) .* RESULT .* err=49 .*$
```
sebres: - too many greedy catch-alls (bad performance, high load, apart from that dangerously, because fault-prone);
- catch-all `.*` directly after prefix line is not required;
- anchored regex `.*$` - such expressions make no sence at all;
- `pid` means connection id here;

So I'll suggest:
```diff
- failregex = ^(?P<__prefix>%(__prefix_line)s).* conn=(?P<pid>\d+) fd=\d+ ACCEPT from IP=<HOST>\:\d+ .+$<SKIPLINES>(?P=__prefix).+ conn=(?P=pid) .* RESULT .* err=49 .*$
+ failregex = failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s
```
Expression `(?:\s(?!err)\S+=\S*)*` means - any count of "tags" in form of `tag=?anything?`, but only if tag not equal "err", so will match ` a=1 b=2 c= ...` or nothing before `err=...`.

fail2ban/tests/files/logs/slapd
```diff
@@ -0,0 +1,18 @@
...
+# failJSON: { "match": false }
+Jul  8 01:47:19 ldap-server slapd[1183]: conn=1022 op=0 EXT oid=1.3.6.1.4.1.6.1
+# failJSON: { "match": false }
+Jul  8 01:47:19 ldap-server slapd[1183]: conn=1022 op=0 STARTTLS
+# failJSON: { "match": false }
+Jul  8 01:47:19 ldap-server slapd[1183]: conn=1022 op=0 RESULT oid= err=1 text=TLS already started
+# failJSON: { "match": false }
+Jul  8 01:47:20 ldap-server slapd[1183]: conn=1022 op=1 BIND dn="uid=gipson,ou=people,dc=example,dc=com" method=128
+# failJSON: { "time": "2005-07-08T01:47:20", "match": true , "host": "8.8.8.8", "desc": "Multiline match for invalid credentials" }
+Jul  8 01:47:20 ldap-server slapd[1183]: conn=1022 op=1 RESULT tag=97 err=49 text=
```
sebres: Are you sure, that this error (49) means unauthorized?
Because I see above another error (err=1) with same conn-id (1022) - "TLS already started".

Can you browse you log files to find a failure without `err=1` or if possile take a look at documentation (resp. source code) of slapd...

gips0n: Hi, @sebres 
Thanks for review!

* updated failregex according to your suggestion
* you can find result codes [here](http://www.openldap.org/doc/admin24/appendix-ldap-result-codes.html), and it seems, that error code 49 is for [invalid credentials](http://www.openldap.org/doc/admin24/appendix-ldap-result-codes.html#invalidCredentials%20(49))

sebres:
FWIW (because in hidden comment of me) just to facilitate future reviews:

Expression `(?:\s(?!err)\S+=\S*)*` means - any count of "tags" in form of tag=?anything?, but only if tag not equal "err", so will match a=1 b=2 c= ... or nothing before err=....

config/filter.d/slapd.conf
```diff
@@ -0,0 +1,25 @@
...
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition]
+
+_daemon = slapd
+
+failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s
```
yarikoptic: trailing anchor? ;)

gips0n: I'm not sure, that it'll be useful here
@sebres, what do you think?

sebres: he is the boss ;)
PS I'm basically agree, it's not really "useful" here, but ... he pursued another strategy...
See for example a small discussion here: https://github.com/fail2ban/fail2ban/pull/1479#issuecomment-231331273

sebres:
That will fail in the test cases: You did not understand - the line should fully match the test expression.

What @yarikoptic means was not only add a trailing anchor, but to extend the expression between current end and this anchor :)

But please not extend it with `.*$` (so not `...\s.*$`), then this "expression" resp. anchor makes not sense at all

gips0n:
@sebres yes, I missed log structure.

According to current log records regex must be something like this:
> failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49 text=$

But, `text` is empty because this error code (49) don't have description in slapd code (some [error codes described](https://github.com/fail2ban/fail2ban/pull/1478/files#diff-38b90fee2e4d134a04a9368f1054a0efR10)), but in next releases it can be added, so it better to use next variant:
> failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49 text=[\w\s]*$

Is it ok?

config/filter.d/slapd.conf
```diff
@@ -0,0 +1,25 @@
...
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition]
+
+_daemon = slapd
+
+failregex = ^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>[^\n]+<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49\s$
```
yarikoptic: could you please also make `[^\n]+` more explicit, since it is analogous to absent anchoring in the end of the line.  E.g. ` \(IP=\S+\)\s*`?

yarikoptic:
well, to my taste `[\w\s]*` is a bit too broad, adding a token to @sebres 's arguments for not requiring the trailing anchoring ;)   but at least it doesn't allow any punctuation symbols, so very unlikely to absorb some at least minimal punctuation upstream would add if for some reason starts including user-provided data straight into the beginning of the log line....  so ok -- I can swallow this pill here ;)

config/filter.d/slapd.conf
gips0n:
```
^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>:\d{1,5} \(IP=[\d\w\.:]+\)\s*<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49 text=[\w\s]*$
```
@yarikoptic what about this variant?

gips0n: or we can try even something like this
```
^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>:\d{1,5} \(IP=[\d\w\.:]+:\d{1,5}\)\s*<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49 text=[\w\s]*$
```

yarikoptic: We already support ipv6 in 0.10 branch... so I wouldn't bother being that specific ;-)

yarikoptic: Previous variant is probably better in that regard

gips0n:
```
^(?P<__prefix>%(__prefix_line)s)conn=(?P<_conn_>\d+) fd=\d+ ACCEPT from IP=<HOST>:\d{1,5} \(IP=\S+\)\s*<SKIPLINES>(?P=__prefix)conn=(?P=_conn_) op=\d+ RESULT(?:\s(?!err)\S+=\S*)* err=49 text=[\w\s]*$
```
maybe this one?

yarikoptic: Looks ok, but I am having difficulty visually diffing with Ali prev just commit, push please ;-)

PR #1479, opened 2016-07-08
sshd-filter: recognized "Failed publickey for" now (gh-1477)…

### LINKED ISSUE #1477: Included filter.d/sshd.conf does not identify failed public keys
### Environment:

_Fill out and check (`[x]`) the boxes which apply. If your Fail2Ban version is outdated, 
and you can't verify that the issue persists in the recent release, better seek support 
from the distribution you obtained Fail2Ban from_
- Fail2Ban version (including any possible distribution suffixes):
- OS, including release name/version: 
- [ ] You have not applied any additional foreign patches to the codebase
### The issue:

Fail2Ban does not trigger a ban when connection attempts are made using an invalid public key.  On this box, sshd setup running on port 15001 and only allows keyfiles on Arch Linux using Fail2Ban v0.9.4.
#### Steps to reproduce
1. Enable the fail2ban systemd service using the below `/etc/fail2ban/jail.local`
2. Connect from an external IP using an unauthorized key to ssh.
#### Expected behavior

After 3 attempts, F2B should lock out the offending IP.
#### Observed behavior

No ban is triggered.
#### Any customizations done to /etc/fail2ban/ configuration

The only modification I have made is to create /etc/fail2ban/jail.local

```
[DEFAULT]
ignoreip = 192.168.1.1/24
bantime  = 1200
maxretry = 3
backend  = auto

[sshd]
enabled = true
filter  = sshd
backend = systemd
```
#### Relevant parts of /var/log/fail2ban.log file:

_preferably obtained while running fail2ban with `loglevel = 4`_

```
2016-07-05 18:22:29,760 fail2ban.server         [4636]: INFO    Changed logging target to /var/log/fail2ban.log for Fail2ban v0.9.4
2016-07-05 18:22:29,764 fail2ban.database       [4636]: INFO    Connected to fail2ban persistent database '/var/lib/fail2ban/fail2ban.sqlite3'
2016-07-05 18:22:29,987 fail2ban.database       [4636]: WARNING New database created. Version '2'
2016-07-05 18:22:29,994 fail2ban.jail           [4636]: INFO    Creating new jail 'sshd'
2016-07-05 18:22:30,127 fail2ban.jail           [4636]: INFO    Jail 'sshd' uses systemd
2016-07-05 18:22:30,412 fail2ban.jail           [4636]: INFO    Initiated 'systemd' backend
2016-07-05 18:22:30,440 fail2ban.filter         [4636]: INFO    Set findtime = 600
2016-07-05 18:22:30,442 fail2ban.filter         [4636]: INFO    Set maxRetry = 3
2016-07-05 18:22:30,444 fail2ban.actions        [4636]: INFO    Set banTime = 1200
2016-07-05 18:22:30,451 fail2ban.filter         [4636]: INFO    Set maxlines = 10
2016-07-05 18:22:31,231 fail2ban.filtersystemd  [4636]: INFO    Added journal match for: '_SYSTEMD_UNIT=sshd.service + _COMM=sshd'
2016-07-05 18:22:31,318 fail2ban.jail           [4636]: INFO    Jail 'sshd' started
```

Below is the output from `journalctl -u sshd` while I am pounding the box with an unauthorized key:

```
Jul 05 18:22:41 mercury sshd[4659]: Failed publickey for graysky from 76.58.20.166 port 37944 ssh2: RSA SHA256:v3dpapGleDaUKf$4V1vKyR9ZyUgjaJAmoCTcb2PLljI
Jul 05 18:22:41 mercury sshd[4659]: Connection closed by 76.58.20.166 port 37944 [preauth]
Jul 05 18:22:41 mercury sshd[4661]: Connection from 76.58.20.166 port 37946 on 192.168.1.110 port 15001
Jul 05 18:22:42 mercury sshd[4661]: Failed publickey for graysky from 76.58.20.166 port 37946 ssh2: RSA SHA256:v3dpapGleDaUKf$4V1vKyR9ZyUgjaJAmoCTcb2PLljI
Jul 05 18:22:42 mercury sshd[4661]: Connection closed by 76.58.20.166 port 37946 [preauth]
Jul 05 18:22:42 mercury sshd[4663]: Connection from 76.58.20.166 port 37948 on 192.168.1.110 port 15001
Jul 05 18:22:42 mercury sshd[4663]: Failed publickey for graysky from 76.58.20.166 port 37948 ssh2: RSA SHA256:v3dpapGleDaUKf$4V1vKyR9ZyUgjaJAmoCTcb2PLljI
Jul 05 18:22:42 mercury sshd[4663]: Connection closed by 76.58.20.166 port 37948 [preauth]
Jul 05 18:22:42 mercury sshd[4665]: Connection from 76.58.20.166 port 37950 on 192.168.1.110 port 15001
Jul 05 18:22:43 mercury sshd[4665]: Failed publickey for graysky from 76.58.20.166 port 37950 ssh2: RSA SHA256:v3dpapGleDaUKf$4V1vKyR9ZyUgjaJAmoCTcb2PLljI
Jul 05 18:22:43 mercury sshd[4665]: Connection closed by 76.58.20.166 port 37950 [preauth]
Jul 05 18:22:43 mercury sshd[4667]: Connection from 76.58.20.166 port 37952 on 192.168.1.110 port 15001
Jul 05 18:22:43 mercury sshd[4667]: Failed publickey for graysky from 76.58.20.166 port 37952 ssh2: RSA SHA256:v3dpapGleDaUKf$4V1vKyR9ZyUgjaJAmoCTcb2PLljI
Jul 05 18:22:43 mercury sshd[4667]: Connection closed by 76.58.20.166 port 37952 [preauth]
Jul 05 18:22:43 mercury sshd[4669]: Connection from 76.58.20.166 port 37954 on 192.168.1.110 port 15001
Jul 05 18:22:44 mercury sshd[4669]: Failed publickey for graysky from 76.58.20.166 port 37954 ssh2: RSA SHA256:v3dpapGleDaUKf$4V1vKyR9ZyUgjaJAmoCTcb2PLljI
Jul 05 18:22:44 mercury sshd[4669]: Connection closed by 76.58.20.166 port 37954 [preauth]
```

sebres:
- improved regexp (not anchored now to recognize all of `Failed method for ... from <HOST>`

Closes #1477

### DIFF
--- a/config/filter.d/sshd.conf
+++ b/config/filter.d/sshd.conf
@@ -20,7 +20,7 @@ _daemon = sshd
 
 failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|error|failed) for .* from <HOST>( via \S+)?\s*$
             ^%(__prefix_line)s(?:error: PAM: )?User not known to the underlying authentication module for .* from <HOST>\s*$
-            ^%(__prefix_line)sFailed \S+ for .*? from <HOST>(?: port \d*)?(?: ssh\d*)?(: (ruser .*|(\S+ ID \S+ \(serial \d+\) CA )?\S+ %(__md5hex)s(, client user ".*", client host ".*")?))?\s*$
+            ^%(__prefix_line)sFailed \S+ for (?P<cond_inv>invalid user )?(?P<user>(?P<cond_user>\S+)|(?(cond_inv)(?:(?! from ).)*?|[^:]+)) from <HOST>(?: port \d+)?(?: ssh\d*)?(?(cond_user):|(?:(?:(?! from ).)*)$)
             ^%(__prefix_line)sROOT LOGIN REFUSED.* FROM <HOST>\s*$
             ^%(__prefix_line)s[iI](?:llegal|nvalid) user .* from <HOST>\s*$
             ^%(__prefix_line)sUser .+ from <HOST> not allowed because not listed in AllowUsers\s*$
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,6 +32,9 @@ releases.
     - Extended failregex and test cases to handle ASSP V1 and V2 (gh-1494)
 * `filter.d/postfix-sasl.conf`
     - Allow for having no trailing space after 'failed:' (gh-1497)
+* `filter.d/vsftpd.conf`
+    - Optional reason part in message after FAIL LOGIN (gh-1543)
+
 
 ### New Features
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -54,6 +54,11 @@ releases.
 * New forward compatibility method assertRaisesRegexp (normally python >= 2.7).
   Methods assertIn, assertNotIn, assertRaisesRegexp, assertLogged, assertNotLogged
   are test covered now
+* Jail configuration extended with new syntax to pass options to the backend (see gh-1408),
+  examples:
+    - `backend = systemd[journalpath=/run/log/journal/machine-1]`
+    - `backend = systemd[journalfiles="/run/log/journal/machine-1/system.journal, /run/log/journal/machine-1/user.journal"]`
+    - `backend = systemd[journalflags=2]`
 
 
 ver. 0.9.5 (2016/07/15) - old-not-obsolete
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,7 @@ releases.
 * Fixed test case "testSetupInstallRoot" for not default python version (also
   using direct call, out of virtualenv);
 * Fixed ambiguous wrong recognized date pattern resp. its optional parts (see gh-1512);
+* FIPS compliant, use sha1 instead of md5 if it not allowed (see gh-1540)
 * `filter.d/asterisk.conf`
     - Fixed to match different asterisk log prefix (source file: method:)
 * `filter.d/ignorecommands/apache-fakegooglebot`
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,7 @@ releases.
   using direct call, out of virtualenv);
 * Fixed ambiguous wrong recognized date pattern resp. its optional parts (see gh-1512);
 * FIPS compliant, use sha1 instead of md5 if it not allowed (see gh-1540)
+* Monit config: scripting is not supported in path (gh-1556)
 * `filter.d/asterisk.conf`
     - Fixed to match different asterisk log prefix (source file: method:)
 * `filter.d/ignorecommands/apache-fakegooglebot`
--- a/ChangeLog
+++ b/ChangeLog
@@ -39,6 +39,8 @@ releases.
 
 
 ### New Features
+* New Actions:
+    - `action.d/npf.conf` for NPF, the latest packet filter for NetBSD
 
 ### Enhancements
 * DateTemplate regexp extended with the word-end boundary, additionally to 
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,8 @@ releases.
     - Allow for having no trailing space after 'failed:' (gh-1497)
 * `filter.d/vsftpd.conf`
     - Optional reason part in message after FAIL LOGIN (gh-1543)
+* `filter.d/sendmail-reject.conf`
+    - removed mandatory double space (if dns-host available, gh-1579)
 
 
 ### New Features
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,8 @@ releases.
 * Fixed test case "testSetupInstallRoot" for not default python version (also
   using direct call, out of virtualenv);
 * Fixed ambiguous wrong recognized date pattern resp. its optional parts (see gh-1512);
+* `filter.d/asterisk.conf`
+    - Fixed to match different asterisk log prefix (source file: method:)
 * `filter.d/ignorecommands/apache-fakegooglebot`
     - Fixed error within apache-fakegooglebot, that will be called 
       with wrong python version (gh-1506)
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,10 @@ releases.
     - Optional reason part in message after FAIL LOGIN (gh-1543)
 * `filter.d/sendmail-reject.conf`
     - removed mandatory double space (if dns-host available, gh-1579)
+* filter.d/sshd.conf
+    - recognized "Failed publickey for" (gh-1477);
+    - optimized failregex to match all of "Failed any-method for ... from <HOST>" (gh-1479)
+    - eliminated possible complex injections (on user-name resp. auth-info, see gh-1479)
 
 
 ### New Features
--- a/FILTERS
+++ b/FILTERS
@@ -227,7 +227,7 @@ Regular expressions (failregex, ignoreregex) assume that the date/time has been
 removed from the log line (this is just how fail2ban works internally ATM).
 
 If the format is like '<date...> error 1.2.3.4 is evil' then you need to match
-the < at the start so regex should be similar to '^<> <HOST> is evil$' using
+the <> at the start so regex should be similar to '^<> error <HOST> is evil$' using
 <HOST> where the IP/domain name appears in the log line.
 
 The following general rules apply to regular expressions:
--- a/bin/fail2ban-client
+++ b/bin/fail2ban-client
@@ -176,7 +176,7 @@ class Fail2banClient:
 					if showRet:
 						self.__logSocketError()
 					return False
-				except Exception, e:
+				except Exception as e:
 					if showRet:
 						logSys.error(e)
 					return False
@@ -429,7 +429,7 @@ class Fail2banClient:
 						elif not cmd == "":
 							try:
 								self.__processCommand(shlex.split(cmd))
-							except Exception, e:
+							except Exception as e:
 								logSys.error(e)
 			except (EOFError, KeyboardInterrupt):
 				print
@@ -451,7 +451,7 @@ class Fail2banClient:
 			ret = self.__configurator.getOptions(jail)
 			self.__configurator.convertToProtocol()
 			self.__stream = self.__configurator.getConfigStream()
-		except Exception, e:
+		except Exception as e:
 			logSys.error("Failed during configuration: %s" % e)
 			ret = False
 		return ret
--- a/bin/fail2ban-server
+++ b/bin/fail2ban-server
@@ -127,7 +127,7 @@ class Fail2banServer:
 								self.__conf["pidfile"],
 								self.__conf["force"])
 			return True
-		except Exception, e:
+		except Exception as e:
 			logSys.exception(e)
 			self.__server.quit()
 			return False
--- a/config/action.d/npf.conf
+++ b/config/action.d/npf.conf
@@ -1,0 +1,61 @@
+# Fail2Ban configuration file
+#
+# NetBSD npf ban/unban
+#
+# Author: Nils Ratusznik <nils@NetBSD.org>
+# Based on pf.conf action file
+#
+
+[Definition]
+
+# Option:  actionstart
+# Notes.:  command executed once at the start of Fail2Ban.
+# Values:  CMD
+#
+# we don't enable NPF automatically, as it will be enabled elsewhere
+actionstart = 
+
+
+# Option:  actionstop
+# Notes.:  command executed once at the end of Fail2Ban
+# Values:  CMD
+#
+# we don't disable NPF automatically either
+actionstop = 
+
+
+# Option:  actioncheck
+# Notes.:  command executed once before each actionban command
+# Values:  CMD
+#
+actioncheck = 
+
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    <ip>  IP address
+#          <failures>  number of failures
+#          <time>  unix timestamp of the ban time
+# Values:  CMD
+#
+actionban = /sbin/npfctl table <tablename> add <ip>
+
+
+# Option:  actionunban
+# Notes.:  command executed when unbanning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    <ip>  IP address
+#          <failures>  number of failures
+#          <time>  unix timestamp of the ban time
+# Values:  CMD
+#
+# note -r option used to remove matching rule
+actionunban = /sbin/npfctl table <tablename> rem <ip>
+
+[Init]
+# Option:  tablename
+# Notes.:  The pf table name.
+# Values:  [ STRING ]
+#
+tablename = fail2ban
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -16,7 +16,7 @@ __pid_re = (?:\[\d+\])
 iso8601 = \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+[+-]\d{4}
 
 # All Asterisk log messages begin like this:
-log_prefix= (?:NOTICE|SECURITY|WARNING)%(__pid_re)s:?(?:\[C-[\da-f]*\])? [^:]+:\d*( in \w+:)?
+log_prefix= (?:NOTICE|SECURITY|WARNING)%(__pid_re)s:?(?:\[C-[\da-f]*\])? [^:]+:\d*(?:(?: in)? \w+:)?
 
 failregex = ^%(__prefix_line)s%(log_prefix)s Registration from '[^']*' failed for '<HOST>(:\d+)?' - (Wrong password|Username/auth name mismatch|No matching peer found|Not a local domain|Device does not match ACL|Peer is not supposed to register|ACL error \(permit/deny\)|Not a local domain)$
             ^%(__prefix_line)s%(log_prefix)s Call from '[^']*' \(<HOST>:\d+\) to extension '[^']*' rejected because extension not found in context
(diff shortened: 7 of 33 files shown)


sebres:
@yarikoptic A best example for "paranoid you", why I always say, that sometimes we should stop using always-end-anchoring (if we can define the regexp precise enough to exclude injecting), because:
- here we had already a correct regexp for that, but totally non-interesting information at end of log line like "RSA SHA256:sign" prevents to ban, because failregex was end-anchored;
- we can't predict what for information may be added at end, for example after `ssh2:` because it is dependent on method, security key, etc. 
But, imho, totally "non-interesting" as long as this message according to source code means "unauthorized", and at the same time totally "bad" - because according to source code also - the parts at end of log line are very-very dynamically.

For example, I could not reproduce this bug in my own bundle (then I always use my own custom filter without unnecessary end-anchoring);

yarikoptic:
Unless we get a guarantee from upstream (and possibly with assurance through some kind of a test on their end) and all of the maintainers of the service in distributions we kinda support (which we can't really) we are to monitor, that no user-provided input would ever be logged at the beginning of the line, we must not relax end of the line anchoring, especially for such critical services as ssh.
On a really quick look at ssh code base I see that logging with some content being placed into the beginning of the log line, other than `__func__` is not unprecedented:
```shell
$> git grep 'logit("%' | grep -v '", __func'

auth2-hostbased.c:		logit("%s: key type %s not in HostbasedAcceptedKeyTypes",
auth2-pubkey.c:		logit("%s: unsupported public key algorithm: %s",
auth2-pubkey.c:		logit("%s: key type %s not in PubkeyAcceptedKeyTypes",
debian/patches/debian-banner.patch: 		logit("%s line %d: Deprecated option %s",
loginrec.c:			logit("%s: Couldn't seek to tty %d slot in %s",
loginrec.c:			logit("%s: Couldn't seek to tty %d slot in %s",
moduli.c:	logit("%.24s Sieve next %u plus %u-bit", ctime(&time_start),
moduli.c:	logit("%.24s Sieved with %u small primes in %ld seconds",
moduli.c:	logit("%.24s Found %u candidates", ctime(&time_stop), r);
moduli.c:		logit("%.24s processed %lu in %s", ctime(&time_now),
moduli.c:	logit("%.24s processed %lu of %lu (%lu%%) in %s, ETA %s",
moduli.c:	logit("%.24s Found %u safe primes of %u candidates in %ld seconds",
mux.c:		logit("%s: streamlocal and dynamic forwards "
servconf.c:		logit("%s line %d: Deprecated option %s",
servconf.c:		logit("%s line %d: Unsupported option %s",
sftp-client.c:			logit("%s: not a regular file\n", new_src);
sftp-client.c:			logit("%s: lstat failed: %s", filename,
sftp-client.c:			logit("%s: not a regular file\n", filename);
sftp-server.c:		logit("%s%sclose \"%s\" bytes read %llu written %llu",
sftp-server.c:		logit("%s%sclosedir \"%s\"",
ssh-keygen.c:				logit("%s:%ld: ignoring host name "
ssh-keygen.c:		logit("%s:%ld: invalid line", l->path, l->linenum);
ssh-keygen.c:			logit("%s:%ld: invalid line", l->path, l->linenum);
ssh.c:		logit("%s, %s", SSH_RELEASE,
sshconnect.c:				logit("%s host key for IP address "
sshconnect.c:			logit("%s", msg);
sshconnect.c:			logit("%s", msg);
sshconnect.c:			logit("%s", ra);
sshconnect2.c:		logit("%s", info);
sshconnect2.c:		logit("%s", name);
sshconnect2.c:		logit("%s", inst);
```
some of those are pertinent to client, and probably could be ignored, and hopefully the others are safe too.  BUT the point it is that it is not unprecedented within ssh codebase, and good that it is followed by some suffix seems me, so in case of trailing anchoring we would not cause a match (as it it would happen in case of it being absent).

Fixing up for a security/DoS flaw is much more gruesome (especially for maintainers, like yours truly) than trying to catch most frequently used trailings of the lines.  I would prefer miss some hits than allow a single vector of attack.  Now the points

*  `because failregex was end-anchored`, not quite -- more of "because we did not specify additional trailing entry"
* `ssh2: because it is dependent on method, security key, etc`... as long as it is structured and not entirely freeform, we can come up with a regex entry for it.

sebres:
I knew it :)
Not that I totally do not agree, but ...

>  I would prefer miss some hits than allow a single vector of attack.

Well, it cuts both ways - imho, we open hereby another vector of attack (and I don't know what would be worser), then "miss some hits" means "let's the attacker play"...

As already said, myself still prefer not-anchored expressions (if the information at the end of log entry does no matter).

> especially for such critical services as ssh.

It will be a little bit better, if we'll release at some point the ban-time-incr, because we can then radical decrease a initial (resp. default) bantime. So the service can be faster accessible again after possible mistaken failure (if the IP is not totally bad).

yarikoptic:
> Well, it cuts both ways - imho, we open hereby another vector of attack (and I don't know what would be worser), then "miss some hits" means "let's the attacker play"...

Such "view" does come from users from time to time -- "oh, fail2ban failed to ban, thus it is a security issue since my system became  less protected."
But the answer is that fail2ban missing the hit(s), and/or crashing and no longer working at all is not making system less secure, since if you have security vulnerability (besides weak passwords) -- fail2ban doesn't really help to avoid it.  But if fail2ban itself introduces an attack vector (e.g. ability to cause DoS) -- that is the security vulnerability I am talking about we need to avoid.  Unbanning faster is not really a solution for that -- attacker would just keep injecting victim's IP making those bantimes longer and longer.

and again -- it is not about information in the end of entry matter or not, it is about being able to inject any malicious line within the log line and then us falling into the trap.

sebres:
> since if you have security vulnerability (besides weak passwords)

I spoke not about theoretical vulnerabilities and possible completely replacement of the security with fail2ban :) I know (better than the other maybe;), that the fail2ban is no panacea (just as the anti-virus software is not it), but if some use it, they rely to it sometimes more than it should be...

And I'm definitively against "Security through obscurity", but "Security together with something else (inclusive these obscurity)" is imho always better as only "Security".

My individually objection was only: I'll always prefer "**possibly a wrong ban**" (I can then correct the expression) in place of "**possible miss some bans**" (then this attacker-failures will be possibly never or too late seen)... 
You are following a different strategy - to hedge one's bets...

Apart from that, "missed ban" makes sometimes possible to executes some "brute" unimaginable attacks (segfaulting, stack overflows, all that, including unknown vulnerabilities, etc), that would never have been possible otherwise... (believe me, I know what I'm talking about... heh good olden times :))

yarikoptic:
On Wed, 13 Jul 2016, Serg G. Brester wrote:

>    My individually objection was only: I'll always prefer "possibly a wrong
>    ban" (I can then correct the expression) in place of "possible miss some
>    bans" (then this attacker-failures will be possibly never or too late
>    seen)...
>    You are following a different strategy - to hedge one's bets...

"wrong ban" == I can ban arbitrary <HOST> == DoS attack vector == security vulnerability

"correct the expression" == all versions of fail2ban in the wild will
need to get a security update with the corrected expression.  It is lots
of work, spread through lots of people to do it.  That is what  I am
trying to avoid.  And that is what I had to experience before, see e.g.
https://www.google.com/search?q=CVE+fail2ban&ie=utf-8&oe=utf-8
and that is what for I have got even some bounty pounds when
discovered and demonstrated such vulnerability in other systems ;)

>    Apart from that, "missed ban" makes sometimes possible to executes some
>    "brute" unimaginable attacks (segfaulting, stack overflows, all that,
>    including unknown vulnerabilities, etc), that would never have been
>    possible otherwise... (believe me, I know what I'm talking about... heh
>    good olden times :))

example? ;)

sebres:
> "wrong ban" == I can ban arbitrary \<HOST\> == DoS attack vector == security vulnerability 

Nope, nope, nope: you're wrong here (see the begin of the discussion). I wrote "if we can define the regexp precise enough to exclude injecting". And it is almost always the case, if the regexp has leading anchor and contains only precise sub-expressions (no catch-alls, no ambiguous sub-expressions, etc).
So the attacker **can't ban arbitrary host**... he can ban at the most **itself**. 

So we should really compare only between:
- "wrong ban" == I (my \<HOST\>) was misleadingly banned != security vulnerability 
- "missing ban" == attacker can do brute force, search unhindered for vulnerabilities, etc. == 
theoretical security vulnerability == at least fail2ban does not the work, for which it was installed.

Imho, the choice is unambiguous...

> example? ;)

You want really, that I would put some hacking tips how to outwit fail2ban (and-or some security tools) ... in the repository of fail2ban self? :)

config/filter.d/sshd.conf
```diff
@@ -20,7 +20,7 @@ _daemon = sshd
 
 failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|error|failed) for .* from <HOST>( via \S+)?\s*$
             ^%(__prefix_line)s(?:error: PAM: )?User not known to the underlying authentication module for .* from <HOST>\s*$
-            ^%(__prefix_line)sFailed \S+ for .*? from <HOST>(?: port \d*)?(?: ssh\d*)?(: (ruser .*|(\S+ ID \S+ \(serial \d+\) CA )?\S+ %(__md5hex)s(, client user ".*", client host ".*")?))?\s*$
+            ^%(__prefix_line)sFailed \S+ for .*? from <HOST>(?: port \d*)?(?: ssh\d*)?(?:: | (?!from )|$)
```
sebres: BTW we've already injecting test cases for the sshd-filter ... 
And the regexp, that suggested in this PR, is definitely injection-safe. Because otherwise this test case will fail (wrong 10.10.1.1 instead of correct 127.0.0.1).

See for example the test case "Injecting on username ssh ..." 2 lines above my newer test case ...
https://github.com/fail2ban/fail2ban/pull/1479/files#diff-affd3f5d7acb03d5fe1c1a6e773f05a6R118

yarikoptic:
On Thu, 14 Jul 2016, Serg G. Brester wrote:

>      "wrong ban" == I can ban arbitrary <HOST> == DoS attack vector ==
>      security vulnerability

>    Nope, nope, nope: you're wrong here (see the begin of the discussion). I
>    wrote "if we can define the regexp precise enough to exclude injecting".
>    And it is almost always the case, if the regexp has leading anchor and
>    contains only precise sub-expressions (no catch-alls, no ambiguous
>    sub-expressions, etc).
>    So the attacker can't ban arbitrary host... he can ban at the most itself.

if ever upstream/maintainers would log user provided input at the
beginning of the line without any custom prefix, e.g.

log("%s data from the client was malformed", user_provided_data)

then an attacker could possibly inject arbitrary input there, providing
victim's IP to be matched using one of our expressions, and we
would be doomed, unlike if we precisely match anchoring at the end,
since then such a line would not match the trailing portion.

How likely to see such type of logging?  I don't know... but I wouldn't
be surprised if it would happen either straight in upstream, or by some
maintainer in a distribution adding some patch to provide
additional logging.  Sure thing if someone just adds

log("%s", user_provided_data)

we are doomed regardless, but hopefully such evil wouldn't happen

>    So we should really compare only between:

>      * "wrong ban" == I (my <HOST>) was misleadingly banned != security
>        vulnerability
>      * "missing ban" == attacker can do brute force, search unhindered for
>        vulnerabilities, etc. == theoretical security vulnerability == at
>        least fail2ban does not the work, for which it was installed.

>    Imho, the choice is unambiguous...

>      example? ;)

>    You want really, that I would put some hacking tips how to outwit fail2ban
>    (and-or some security tools) ... in the repository of fail2ban self? :)

you can send a private email, I would really be interested to know where
we fall short

-- 
Yaroslav O. Halchenko

sebres:
@yarikoptic How we will decide about this PR? I mean I made the suggested expression safe against possible injection with `(?!from )`, and we've already a injecting test case for the sshd-filter, see https://github.com/fail2ban/fail2ban/pull/1479#discussion-diff-70815578

I had flowed through the source code of openssh, the logging of such messages occurred clearly on authentication failures only... But at the same time it is too complex and IMHO the end-anchoring here does not matter at all:
```c
...
	authmsg = authenticated ? "Accepted" : "Failed";
	authlog("%s %s%s%s for %s%.100s from %.200s port %d ssh2%s%s",
	    authmsg,
	    method,
	    submethod != NULL ? "/" : "", submethod == NULL ? "" : submethod,
	    authctxt->valid ? "" : "invalid user ",
	    authctxt->user,
	    ssh_remote_ipaddr(ssh),
	    ssh_remote_port(ssh),
	    authctxt->info != NULL ? ": " : "",
	    authctxt->info != NULL ? authctxt->info : "");
...
```

Of course, I can follow your arguments and I can not swear that will never changed, but I see really no occasion why the ssh-developers should change in the future this logic so, that this will not concern the authentication.

Of course we can define regexp more precisely, for example with enumerating of all auth-methods, but that are a lot (password, challenge-response, publickey, hostbased, gssapi-with-mic etc) and not statically, because everyone can extend it with new methods...

So if will be accepted, I'll resolve the conflicts and merge...

Or you'll nevertheless insist to have the end-anchoring in this regexp? So please let me know, because I'll like to close #1477

sebres:
I've currently realized that regexp changes resp. injecting test cases implemented in #426 are wrong resp. because of ambiguous interpretation could be really dangerous (at least I've working recipe, that is complex, expected customized ssh-client, but it makes fail2ban to ignore my IP and ban foreign IP).

It comes to rows:
1. `Failed password for user from 127.0.0.1 port 20000 ssh1: ruser from 1.2.3.4`
https://github.com/fail2ban/fail2ban/blob/5d6cead99694efae8710ab12cd68a924a83d7396/fail2ban/tests/files/logs/sshd#L113
2.  `Failed password for invalid user from 10.10.1.1 from 127.0.0.1 port 58946 ssh2`
https://github.com/fail2ban/fail2ban/blob/5d6cead99694efae8710ab12cd68a924a83d7396/fail2ban/tests/files/logs/sshd#L119

Both say, that `127.0.0.1` would be here an IP of the attacker, that would be banned.

Exactly that I mean as ambiguous and it is really wrong - then es either 1st is correct or the 2nd.
The expressions are currently artificial made to match the "correct" IP address of both. But which address is "correct"? Only one is true by remote use name (the name before `from <HOST>` or the name after `: `). So as it is currently constructed - it is manipulable

Therefore this is currently a "security" issue for me and I classify this as grave...
We should resolve this "conflict" or just forbid the recognizing that as failure in both cases (i.e. ignore it)

BTW it will be anyway banned, because another rule will match this failure, there sshd additionally produces `pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=<HOST>` that our sshd-filter accepts also as failure (this is also a second thing I want criticize here, because one login attempt should produce one failure, but we will have 2 failures with both expressions).

Bellow is an excerpt from sshd journal if I try password-login with "illegal" user name `ttt from 172.16.0.5 port 63260 ssh2:` :
```
Aug 18 19:47:49 srv sshd[9643]: Invalid user ttt from 172.16.0.5 port 63260 ssh2 from 172.16.0.5
Aug 18 19:47:49 srv sshd[9643]: input_userauth_request: invalid user ttt from 172.16.0.5 port 63260 ssh2 [preauth]
Aug 18 19:47:57 srv sshd[9643]: pam_unix(sshd:auth): check pass; user unknown
Aug 18 19:47:57 srv sshd[9643]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=172.16.0.5
Aug 18 19:47:59 srv sshd[9643]: Failed password for invalid user ttt from 172.16.0.5 port 63260 ssh2 from 172.16.0.5 port 56095 ssh2
```

Therefore I believe that the first case (implemented in #426) is wrong... The correct one is 2th only.
But because I think, that `:` in the user name not allowed (at least with OpenSSH v.7.2  I can't produce any log-entry contains my `:`, that I've generated), I think it would be theoretical possible to make a correct expression for both cases...
Additionally I've found another injecting test-case that I'll add here...

So I continue to analyze and will post here a possible solution... WiP.

sebres:
I belive it is correct now...

Still one "problem", as already said: one login attempt should produce exact one failure, but we will sometimes have 2 failures with both expressions... I don't know how we can solve this to be backwards compatible for some previous versions and at the same time to match all failures (pam and not pam)...

yarikoptic:
Great analysis, thanks... Yet to wrap my head fully around it and either we already should expect or ourselves issue a cve and deal with fixing versions in the wild ;-) as for two failures for one - hardly we can do much about it besides starting to track count per each line in regexp... Meanwhile could be mitigated by recommending a higher setting for allowed number of failures I guess if you find your host banning legit users forgetting their passwords

config/filter.d/sshd.conf
```diff
@@ -20,7 +20,7 @@ _daemon = sshd
 
 failregex = ^%(__prefix_line)s(?:error: PAM: )?[aA]uthentication (?:failure|error|failed) for .* from <HOST>( via \S+)?\s*$
             ^%(__prefix_line)s(?:error: PAM: )?User not known to the underlying authentication module for .* from <HOST>\s*$
-            ^%(__prefix_line)sFailed \S+ for .*? from <HOST>(?: port \d*)?(?: ssh\d*)?(: (ruser .*|(\S+ ID \S+ \(serial \d+\) CA )?\S+ %(__md5hex)s(, client user ".*", client host ".*")?))?\s*$
+            ^%(__prefix_line)sFailed \S+ for (?P<cond_inv>invalid user )?(?P<user>(?P<cond_user>\S+)|(?(cond_inv)(?:(?! from ).)*?|[^:]+)) from (?:::f{4,6}:)?(?P<host>[\w\-.^_]*\w)(?: port \d+)?(?: ssh\d*)?(?(cond_user):|(?:(?:(?! from ).)*)$)
```
sebres: the expression is complex conditionally variable:
- it matches end-anchored if we have simple user name, exactly one `from` (or no `from` before `:` and-or no `from` after `:`)
- or it have no `:` (normally ends with ssh2), this case prefers the last `from`
- otherwise the first `from`

Additionally both conditions make it a bit faster

sebres: Oh, I forgot to replace `(?:::f{4,6}:)?(?P<host>[\w\-.^_]*\w)` again back with `<HOST>` (tested expression in shell)... I amend it now

sebres:
<sub>Because now expression description hidden in the outdated diff, here again:</sub>

The expression is complex conditionally variable:

- it matches end-anchored if we have simple user name, exactly one ` from ` (or no ` from ` before `:` and-or no ` from ` after `:`); the end-anchor is then important, because it should verify this all
- or it have no `:` (normally ends with ssh2), this case prefers the last ` from `
- otherwise prefers the first ` from `

Additionally both conditions make it a bit faster

sebres:
Merged with master.
It has been a while now (and IMHO issue is grave), so I merge it

PR #1483, opened 2016-07-13
Pr 1413: BF for isStarted (should it be just removed?) and centralize reassignment of signal handlers

yarikoptic:

isStarted is not used anywhere and difficult to test for it to return True within the same process ;)


yarikoptic:
d'oh, forgot to adjust target repo for this PR ;)

PR #1485, opened 2016-07-15
BF: do not rely on long relative path to upstairs config - symlink dereferenced copied during install

yarikoptic:
I know that original idea was that the zzz is never shipped/installed.  But e.g. since I base debian packaging straight on the git repo, and it is easier to maintain a thorough MANIFEST, I have decided to just workaround by making that common.conf a symlink and then rely on it being dereferenced during installation.  It is unlikely someone would be installing fail2ban on crummy filesystem not supporting symlinks.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -16,6 +16,8 @@ config/action.d/firewallcmd-allports.conf
 config/action.d/firewallcmd-ipset.conf
 config/action.d/firewallcmd-multiport.conf
 config/action.d/firewallcmd-new.conf
+config/action.d/firewallcmd-rich-logging.conf
+config/action.d/firewallcmd-rich-rules.conf
 config/action.d/hostsdeny.conf
 config/action.d/ipfilter.conf
 config/action.d/ipfw.conf
@@ -124,6 +126,7 @@ config/filter.d/sendmail-auth.conf
 config/filter.d/sendmail-reject.conf
 config/filter.d/sendmail-spam.conf
 config/filter.d/sieve.conf
+config/filter.d/slapd.conf
 config/filter.d/sogo-auth.conf
 config/filter.d/solid-pop3d.conf
 config/filter.d/squid.conf
@@ -205,6 +208,7 @@ fail2ban/tests/config/fail2ban.conf
 fail2ban/tests/config/filter.d/simple.conf
 fail2ban/tests/config/filter.d/test.conf
 fail2ban/tests/config/filter.d/test.local
+fail2ban/tests/config/filter.d/zzz-generic-example.conf
 fail2ban/tests/config/jail.conf
 fail2ban/tests/config/paths-common.conf
 fail2ban/tests/config/paths-debian.conf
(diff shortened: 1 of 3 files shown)


yarikoptic:
ok, tests passed, so I will have it merged into master to facilitate the 0.9.5 release ;)

sebres:
Good idea, merge...

sebres:
I've reverted this in c4dc698d98c751826e40c0abad6f75a41774fcc7, because symlink may be very annoying - if I develop over net share, that does not support symlink (routers, netbox, etc), or using VM in windows host etc, then I cannot use git (merge, checkout, reset, etc. fails resp. conflicts on this file).
It is sometimes very evil (and horrid disturbs).
So please, if not really expected don't use symlinks within git-repo (I'm not against dynamical creation of it somewhere in the test-cases, but here it is 4 times updir `..`, so imho does not really needed.

yarikoptic:
I guess you did run into a real case when this was a stumbling stone, so I guess the necessity is warranted.
But then how should issue be fixed to have tests usable "as installed" where the relative path would be incorrect since fail2ban installed under /usr while configuration under /etc???

yarikoptic:
I only see "copy the common.conf" into the tests, but that is somewhat is against the purpose... unless we just do it in some `setup` fixture automagically and cleanup in `teardown` . @sebres?

sebres:
We can hang it depending to the variable `STOCK` (how it made in clientreadertestcase):
```python
STOCK = os.path.exists(os.path.join('config', 'fail2ban.conf'))
...
if STOCK:
   # do the test
```

yarikoptic:
which then wouldn't allow to troubleshoot possibly broken/incompatible `common.conf` by running the tests on an installed instance of fail2ban... but now I recalled -- we have `CONFIG_DIR` defined in fail2ban/tests/utils.py  which we probably should use here...?

sebres:
you meant something like:
```diff
if not CONFIG_DIR:
# Use heuristic to figure out where configuration files are
	if os.path.exists(os.path.join('config','fail2ban.conf')):
		CONFIG_DIR = 'config'
	else:
		CONFIG_DIR = '/etc/fail2ban'
+ 		os.symlink(CONFIG_DIR, ...)
```

sebres:
> which then wouldn't allow to troubleshoot possibly broken/incompatible common.conf by running the tests on an installed instance of fail2ban...

But only test-cases from test/files (and currently only have name begins with `zzz-` - exact this one affected from c4dc698d98c751826e40c0abad6f75a41774fcc7).

yarikoptic:
I guess no symlinks since that is what you wanted to avoid, and it would be not kosher to add any (or copy anything, as I have erroneously suggested) on an installed system.  Given your comment on `zzz-` may be indeed easiest just to skip that single test.  Alternative -- is to add handling of having `<CONFIG_DIR>` to `SafeConfigParserWithIncludes` or smth like that but sounds like an overkill atm

sebres:
I wanted to avoid in git-repo only, then the target sub-system can normally use links, but it is not implicitly possible using git on host-system, resp. by communication between host / guest (if guest under NAT, I can use git from host only via share/samba).

About the rest, I'll think about how we can do that with minimal overhead.

PR #1486, opened 2016-07-16
new postfix filters

rcwschaller:


config/filter.d/postfix-address.conf
```diff
@@ -0,0 +1,22 @@
...
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition]
+
+_daemon = postfix/(submission/)?smtp(d|s)
+
+failregex = ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 550 5\.1\.1 <\S+>: Recipient address rejected: User unknown in virtual mailbox table;.*$
```
sebres: Regular expression `.*$` makes no sense at all.
I had said - remove it, but the top boss here says - please supply precise anchored expression...
For the discussion about anchors see https://github.com/fail2ban/fail2ban/pull/1479#issuecomment-231331273

fail2ban/tests/files/logs/postfix-address
```diff
@@ -0,0 +1,2 @@
+# failJSON: { "time": "2016-07-15T17:40:29", "match": true , "host": "1.1.1.1" }
```
sebres: All our test cases are running in year 2005, that's why the test cases fail... Please correct both times in failJSON.

rcwschaller:
implemented @sebres suggestions

sebres:
You see, the test cases pass now :)

But, now about the anchors: is not a suggestion of me (I'm definitely against end-anchoring of this regexp here), because it imho very-very "vulnerable", because it is firstly - dynamically and secondly - can contain "unsafe" user input. 
Just imagine:
- attacker uses as parameter `helo` something, this regexp does not covered (for example in your case a string contains space);
- after update of postfix (or changing of some parameters) this line contains additionally information (for example encryption method etc.) or `proto=ESMTP` will called `proto=ESMTPv2`;

What I said, was just - `.*$` makes no sense at all, and there are other opinions on the subject...
But it was definitely not my proposal, and hereby I dissociate myself from ;)

yarikoptic:
Use `proto=\S+` if we want to allow any protocol

yarikoptic:
The question is, Could helo contain space here? What happens if newline character or > is provided within the helo input?

rcwschaller:
Okay sebres, then i just missunder your frist comment. Totally agree that there should be no end-anchor here, for the reasons the both of you explained well above. However, I was trying to follow the existing definitions in postfix and postfix-sasl as close as possible.

Since all above agree, that possibly malicious user inputs should be avoided as failregex, i will widen my proposal to this other two filters.

rcwschaller:
okay, also regrouped the some related filters from postfix into my two new proposals. Reason for proposing separate filters is that mail server admins might want to assign different levels of rate limits and penalties to it:

- postfix: protocol violations and non-existing hostnames, attack or misconfiguration -> the client is definitively evil
- postfix-relay: relaying mails failed, can happen legitimately in some rare use-cases -> the client might or might not be evil depending on your configuration
- postfix-address: recipient address does not exist, happens in every-day life -> the client is likely not evil, except if it happens too often.

yarikoptic:
Now you need to swear that postfix logs do not (in any version used in a wild) and would never include client provided data at the beginning of the line.

sebres:
> Now you need to swear that postfix logs do not (in any version used in a wild) and would never include client provided data at the beginning of the line.

But if they do that, the end-anchoring don't save us, because for example the attacker can send a parameter with a newline character, so can build a "correct" log entry matched end-anchored regexp.

Just in this case not-anchored expression will firstly possibly ban the attacker and then secondly somebody else (if he make a flood).
But in other case (by anchored expression) we will definitively ignore the first line with "incorrect" data, that contains attacker IP. So the attacker can unhindered continue his attack...

rcwschaller:
So guys, no new suggestions for 10 days, cause:
- I am happy with it and
- I don't see specific suggestions how to improve them by anyone else here.

The actual valid range of possible characters for mail-addresses, domain names, ... (aka attacker provided data) is governed by RFCs and if postfix puts something like line-feed or <> unchecked in a log-file - that's a postfix bug; to be fixed there, not a job for fail2ban...

So it is now time to decide if you want to merge it or not. I would vote for merging, since it is definitely (see discussed points above) an improvement over the current version...

sebres:
could you please short describe the close reason

PR #1490, opened 2016-07-25
easy-fix for distributions compatible test cases

### LINKED ISSUE #1353: FAIL: testReadStockJailConfForceEnabled (fail2ban.tests.clientreadertestcase.JailsReaderTest)
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.9.4
- OS, including release name/version: Fedora Rawhide (F25)

### The issue:

~~~~~
FAIL: testReadStockJailConfForceEnabled (fail2ban.tests.clientreadertestcase.JailsReaderTest)
  File "./fail2ban/tests/clientreadertestcase.py", line 636, in testReadStockJailConfForceEnabled
    self.assertTrue(['add', j, 'auto'] in comm_commands)
AssertionError: False is not true
~~~~~
#### Steps to reproduce
~~~~~
Executing(%check): /bin/sh -e /var/tmp/rpm-tmp.LC5buM
+ umask 022
+ cd /builddir/build/BUILD
+ cd fail2ban-0.9.4
+ export LANG=en_US.UTF-8
+ LANG=en_US.UTF-8
+ ./fail2ban-testcases-all-python3 --no-network
Testing using /usr/bin/python3.5
Fail2ban 0.9.4 test suite. Python 3.5.1 (default, Mar  4 2016, 13:43:50) [GCC 6.0.0 20160302 (Red Hat 6.0.0-0.14)]. Please wait...
~~~~~

sebres:
(e.g. fedora default backend is 'systemd'), see gh-1353

cherry pick into 0.9 branch

Closes #1353

### DIFF
--- a/fail2ban/tests/clientreadertestcase.py
+++ b/fail2ban/tests/clientreadertestcase.py
@@ -649,8 +649,9 @@ class JailsReaderTest(LogCaptureTestCase):
 
 			# and we know even some of them by heart
 			for j in ['sshd', 'recidive']:
-				# by default we have 'auto' backend ATM
-				self.assertTrue(['add', j, 'auto'] in comm_commands)
+				# by default we have 'auto' backend ATM, but some distributions can overwrite it, 
+				# (e.g. fedora default is 'systemd') therefore let check it without backend...
+				self.assertTrue([True for i in comm_commands if len(i) == 3 and i[0:2] == ['add', j]])
 				# and warn on useDNS
 				self.assertTrue(['set', j, 'usedns', 'warn'] in comm_commands)
 				self.assertTrue(['start', j] in comm_commands)


sebres:
Although I don't like self, how I've fixed it here, but I see no better solution atm, without huge (unnecessary) effort.

sebres:
I've rebased it, so have better handling now (the same criteria for all distributions).
If no objection, I'll merge it tomorrow

fail2ban/tests/clientreadertestcase.py
```diff
@@ -649,8 +649,9 @@ def testReadStockJailConfForceEnabled(self):
 
 			# and we know even some of them by heart
 			for j in ['sshd', 'recidive']:
-				# by default we have 'auto' backend ATM
-				self.assertTrue(['add', j, 'auto'] in comm_commands)
+				# by default we have 'auto' backend ATM, but some distributions can overwrite it, 
+				# (e.g. fedora default is 'systemd') therefore let check it without backend...
+				self.assertTrue([True for i in comm_commands if len(i) == 3 and i[0:2] == ['add', j]])
```
yarikoptic: I would have checked for `['add', j] in [c[:2] for c in comm_commands` which is imho easier to read

sebres: yep, but another is a little bit faster, because iterate over generator with few entries;

yarikoptic: performance difference is imho negligible but readability is somewhat ... if you want the same -- make 2nd one a generator: `['add', j] in (c[:2] for c in comm_commands)` ;)

sebres: > performance difference is imho negligible

Collection of 2000 negligible non performant differences ea. 1ms, makes the whole thing 2 seconds longer...
Since then I've optimized our test-cases in 0.10, I respect it very strong, in particular by such easy-going things, that everyone can read (I'm about readability)...

>  if you want the same -- make 2nd one a generator

nope, this generator is larger, because contains more entries...

Secondly that will be slower and slower with each extending of jail, regex of jail or filter, etc...

yarikoptic: On Wed, 27 Jul 2016, Serg G. Brester wrote:
>  >                       # and we know even some of them by heart
>  >                       for j in ['sshd', 'recidive']:
>  > -                             # by default we have 'auto' backend ATM
>  > -                             self.assertTrue(['add', j, 'auto'] in comm_commands)
>  > +                             # by default we have 'auto' backend ATM, but some distributions can overwrite it,
>  > +                             # (e.g. fedora default is 'systemd') therefore let check it without backend...
>  > +                             self.assertTrue([True for i in comm_commands if len(i) == 3 and i[0:2] == ['add', j]])

>      performance difference is imho negligible

>    Collection of 2000 negligible non performant differences ea. 1ms, makes
>    the whole thing 2 seconds longer...
>    Since then I've optimized our test-cases in 0.10, I respect it very
>    strong, in particular by such easy-going things, that everyone can read
>    (I'm about readability)...

>      if you want the same -- make 2nd one a generator

>    nope, this generator is larger, because contains more entries...

>    Secondly that will be slower and slower with each extending of jail, regex
>    of jail or filter, etc...

nope 

1. your construct is not a generator at all -- it is a list
comprehension, it will go through **all** the entries

2. mine is a generator, it will stop going through the entries as soon
as it finds a match, it doesn't store anything along the way.

$> cat /tmp/1.py
def visit(i):
   print "visited ", i
   return i


print 3 in (visit(x) for x in range(10))

$> python /tmp/1.py
visited  0
visited  1
visited  2
visited  3
True


but interestingly you are right for the case whenever desired entry is the last
in the list, but then very much off if it is in the beginning ;) 

regardless of those -- it takes miniscule time to run, and I have
spent already more time altogether on comprehending original code and
then arguing to simplify it than probably all the runs of this test in
the next year ;)  so I would still vote for more readable code.

$> cat /tmp/time_check.py
from timeit import timeit

j = "ssh"

global f1

def f1(comm_commands):
    return bool([True for i in comm_commands if len(i) == 3 and i[0:2] == ['add', j]])

def f2(comm_commands):
    return ["add", j] in (x[:2] for x in comm_commands)


comm_commands = [["add", str(x)] for x in range(1000)]
comm_commands_last = comm_commands + [['add', j]]
comm_commands_first = [['add', j]] + comm_commands


kw = dict(
    setup="from __main__ import f1, f2, comm_commands_first, comm_commands_last",
    number=1000
)

print "checks"
print "f1: %s, %s" % (f1(comm_commands_first), f1(comm_commands_last))
print "f2: %s, %s" % (f2(comm_commands_first), f2(comm_commands_last))

if __name__ == '__main__':

    print "Last entry"
    for f in ('f1', 'f2'):
        print "%s: %.7f" % (f, timeit("%s(comm_commands_last)" % (f,), **kw)/kw['number'])

    print "First entry"
    for f in ('f1', 'f2'):
        print "%s: %.7f" % (f, timeit("%s(comm_commands_first)" % (f,), **kw)/kw['number'])


$> python /tmp/time_check.py
checks
f1: False, False
f2: True, True
Last entry
f1: 0.0000471
f2: 0.0001337
First entry
f1: 0.0000476
f2: 0.0000006

yarikoptic: Actually optimal line would be

any(len(i) == 3 and i[0:2] == ['add', j] for i in comm_commands)

! :-)

yarikoptic: ok -- here is the results of the test:
```
$> python /tmp/time_check.py
checks
f1: True, True
f2: True, True
Last entry
f1: 0.0000450
f2: 0.0000766
First entry
f1: 0.0000444
f2: 0.0000007
```
where f2 is `any(len(x) == 3 and ["add", j] == x[:2] for x in comm_commands)` . Most efficient and compromised between our two ;)

sebres: Already long convinced... (well as regards 1000 entries - by test case enabling all jails, permanent growing of jail params, etc. we are long ago about)

Have rewritten it using assertIn (btw many other things, you know me :), new PR follows...

yarikoptic: What about keeping it concise and filing against 0.9?

sebres: If you'll take a look on #1492, you'll don't ask :) some things are better implementable over 0.10

I'll cherry pick and re-merge 0.9 to 0.10 hereafter (to create a clean merge point backwards, to resolve possible conflicts)...
I swear!

PR #1494, opened 2016-07-31
Fix ASSP filter to work with both ASSP V1 and V2

rhardy613:
ASSP V1 development stopped at the end of 2014 and it is now deprecated.
All users were urged to upgrade to ASSP V2 which is still actively
developed. For some reason fail2ban 0.9.5 (and trunk) still have code
which only understands ASSP V1 logs. This means the filter ignores brute
force attacks against ASSP.

### DIFF
--- a/config/filter.d/assp.conf
+++ b/config/filter.d/assp.conf
@@ -1,5 +1,5 @@
-# Fail2Ban filter for Anti-Spam SMTP Proxy Server (ASSP) Version 2.5.1 (or later)
-# 
+# Fail2Ban filter for Anti-Spam SMTP Proxy Server (ASSP)
+#    Filter works in theory for both ASSP V1 and V2. Recommended ASSP is V2.5.1 or later.
 #    Homepage:    http://sourceforge.net/projects/assp/
 #    ProjectSite: http://sourceforge.net/projects/assp/?source=directory
 #
@@ -7,14 +7,23 @@
 
 [Definition] 
 
-failregex = ^\s*(?:[m0-9\-]+\s+)*(?:\[\S+\]\s+)*<HOST> (?:\<\S+@\S+\.\S+\> )*(?:to: \S+@\S+\.\S+ )*relay attempt blocked for(?: \(parsing\))*: \S+$
-            ^\s*(?:[m0-9\-]+\s+)*(?:\[\S+\]\s+)*<HOST> \[SMTP Error\] 535 5\.7\.8 Error: authentication failed:\s+(?:\S+|Connection lost to authentication server|Invalid authentication mechanism|Invalid base64 data in continued response)*$
+__assp_actions = (?:dropping|refusing)
+
+failregex = ^(:? \[SSL-out\])? <HOST> max sender authentication errors \(\d{,3}\) exceeded -- %(__assp_actions)s connection - after reply: \d{3} \d{1}\.\d{1}.\d{1} Error: authentication failed: \w+;$
+			^(?: \[SSL-out\])? <HOST> SSL negotiation with client failed: SSL accept attempt failed with unknown error.*:unknown protocol;$
+			^ Blocking <HOST> - too much AUTH errors \(\d{,3}\);$
+			^\s*(?:[m0-9\-]+\s+)*(?:\[\S+\]\s+)*<HOST> (?:\<\S+@\S+\.\S+\> )*(?:to: \S+@\S+\.\S+ )*relay attempt blocked for(?: \(parsing\))*: \S+$
+			^\s*(?:[m0-9\-]+\s+)*(?:\[\S+\]\s+)*<HOST> \[SMTP Error\] 535 5\.7\.8 Error: authentication failed:\s+(?:\S+|Connection lost to authentication server|Invalid authentication mechanism|Invalid base64 data in continued response)*$
 
 ignoreregex = 
 
 # DEV Notes:
+# V1 Examples matches:
+#   Apr-27-13 02:33:09 Blocking 217.194.197.97 - too much AUTH errors (41);
+#   Dec-29-12 17:10:31 [SSL-out] 200.247.87.82 SSL negotiation with client failed: SSL accept attempt failed with unknown errorerror:140760FC:SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol;
+#   Dec-30-12 04:01:47 [SSL-out] 81.82.232.66 max sender authentication errors (5) exceeded 
 #
-# Examples matches:
+# V2 Examples matches:
 #   Jul-29-16 16:49:52 m1-25391-06124 [Worker_1] [TLS-out] [RelayAttempt] 0.0.0.0 <user@example.com> to: user@example.org relay attempt blocked for: someone@example.org
 #   Jul-30-16 16:59:42 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed: UGFzc3dvcmQ6
 #   Jul-30-16 00:15:36 m1-52131-09651 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed: UGFzc3dvcmQ6
@@ -28,4 +37,4 @@ ignoreregex =
 
 #
 # Author: Enrico Labedzki (enrico.labedzki@deiwos.de)
-# Updated: Robert Hardy (rhardy@webcon.ca)
+# V2 Filters: Robert Hardy (rhardy@webcon.ca)
--- a/config/filter.d/assp.conf
+++ b/config/filter.d/assp.conf
@@ -7,18 +7,25 @@
 
 [Definition] 
 
-failregex = <HOST> \<\S+@\S+\.\S+\> to: \S+@\S+\.\S+ relay attempt blocked for: \S+$
-            <HOST> \[SMTP Error\] 535 5\.7\.8 Error: authentication failed.*$
+failregex = ^\s*(?:[m0-9\-]+\s+)*(?:\[\S+\]\s+)*<HOST> (?:\<\S+@\S+\.\S+\> )*(?:to: \S+@\S+\.\S+ )*relay attempt blocked for(?: \(parsing\))*: \S+$
+            ^\s*(?:[m0-9\-]+\s+)*(?:\[\S+\]\s+)*<HOST> \[SMTP Error\] 535 5\.7\.8 Error: authentication failed:\s+(?:\S+|Connection lost to authentication server|Invalid authentication mechanism|Invalid base64 data in continued response)*$
 
 ignoreregex = 
 
 # DEV Notes:
 #
-# Examples:
-#               Jul-29-16 16:49:52 m1-25391-06124 [Worker_1] [TLS-out] [RelayAttempt] 0.0.0.0 <user@example.com> to: user@example.org relay attempt blocked for: someone@example.org
-#               Jul-30-16 16:59:42 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed: UGFzc3dvcmQ6
-#               Jul-30-16 00:15:36 m1-52131-09651 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed: UGFzc3dvcmQ6
-#               Jul-31-16 06:45:59 [Worker_1] [TLS-in] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed:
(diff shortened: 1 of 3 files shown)


config/filter.d/assp.conf
```diff
@@ -1,24 +1,24 @@
...
 
 [Definition] 
 
-__assp_actions = (?:dropping|refusing)
-
-failregex = ^(:? \[SSL-out\])? <HOST> max sender authentication errors \(\d{,3}\) exceeded -- %(__assp_actions)s connection - after reply: \d{3} \d{1}\.\d{1}.\d{1} Error: authentication failed: \w+;$
-			^(?: \[SSL-out\])? <HOST> SSL negotiation with client failed: SSL accept attempt failed with unknown error.*:unknown protocol;$
-			^ Blocking <HOST> - too much AUTH errors \(\d{,3}\);$
+failregex = <HOST> \<\S+@\S+\.\S+\> to: \S+@\S+\.\S+ relay attempt blocked for: \S+$
+            <HOST> \[SMTP Error\] 535 5\.7\.8 Error: authentication failed.*$
```
yarikoptic: Shouldn't we keep both old and new patterns?

yarikoptic: Should be anchored upfront... Can we anchor in the back? `.*$` is of no effect anchoring

rhardy613: It isn't my call but I will give my two cents: I would say no. ASSP is a spam firewall and it is useless if not kept current. As a security tool fail2ban should only support safe supported versions and clearly list that as a requirement. That is why I explicitedly listed version 2.5.1 or later in the config file. Any supported version of ASSP will not generate those old patterns. I also suspect the old regex actually was wrong in that it would fail to match lines where the flags were different. My regex is deliberately not anchored. The log lines have a whole list of flags which may or may not appear such as [SSL-in] [TLS-in] [TLS-out] [RelayAttempt] etc. I also believe the order can change. There were also other patterns which varied such as m1-52131-09651 at the start of the line. My attempts at anchoring the patterns caused many missed attacks during my testing. The reason for the .* is other content sometimes follows. I could just remove .*$ if you want. Some examples:
Feb-10-16 15:06:02 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed: UGFzc3dvcmQ6
Feb-10-16 15:06:12 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed: Connection lost to authentication server
Feb-10-16 15:06:12 [Worker_1] [TLS-out] 0.0.0.0 [SMTP Error] 535 5.7.8 Error: authentication failed:

rhardy613:
I've asked the ASSP authors whether they can give me an anchored pattern. Finding all the possible tags in a 70000+ line perl program was problematic, especially with the use of [ in the tags.

rhardy613:
I asked and here is what one of the authors said:
[http://assp.sourceforge.net/forum/viewtopic.php?f=8&t=2887&sid=bbafe0bd0cdbbf87c570819339302c21](url)
"This is not possible, because every part of a log line in assp can be switched off or on in the configuration. Even the content of some parts is configurable."
Given this is the patch ok as is?

yarikoptic:
Thanks for asking, but:

>    Given this is the patch ok as is?

unlikely since I still see no reason to not to provide more exhaustive 
anchored regexp

we would need to have it anchored, in particular at the beginning to start with.

"every part of a log line in assp can be switched off or on"

regular expression could include those parts, and be marked as optional.
e.g. we could probably easily allow a number of leading [TAGS] with smth like
`^(?:\[\S+\]\s+)*<HOST> ....`.  Give us more examples of possible lines, and then we could adjust it even further if necessary

rhardy613:
That stretched my Regex... I'm sure the patterns in the 2nd commit could be optimized further but they are anchored on both sides and they didn't miss anything I thought should have been caught in my last 6 months of logs. Please let me know if the commit is good enough for a merge now.

yarikoptic:
awesome -- I like these ones better!
My only concern is my previous whining -- can we be sure noone who uses fail2ban is not still using older assp v1? so here is my quick look:

```
$> whohas --strict assp                          
FreeBSD     assp                                   1.9.9.14158,1                      mail                      http://www.freebsd.org/cgi/ports.cgi?stype=all&query=assp
MacPorts    assp                                   1.1.0                                                        https://trac.macports.org/browser/trunk/dports/mail/assp/Portfile
Tried fetching "http://dl.fedoraproject.org/pub/fedora/linux/releases/21/Everything/i386/os/Packages/a/" five times. Giving up.
Tried fetching "http://dl.fedoraproject.org/pub/fedora/linux/releases/20/Everything/i386/os/Packages/a/" five times. Giving up.
Gentoo      assp                                   1.9.8.13030                                                  http://packages.gentoo.org/package/mail-filter/assp 
Gentoo      assp                                   1.9.4.9                                                      http://packages.gentoo.org/package/mail-filter/assp 
Gentoo      assp                                   1.9.4.8                                                      http://packages.gentoo.org/package/mail-filter/assp 
Gentoo      assp                                   1.9.4.7                                                      http://packages.gentoo.org/package/mail-filter/assp 
Gentoo      assp                                   1.8.5.9                                                      http://packages.gentoo.org/package/mail-filter/assp 
```
so I actually do not see v2 being used in any of the distributions yet (at least as far as whohas managed to fetch info about)... so ... I would really prefer to keep older regexes in as well, at least for some foreseable future so we don't break ppls setups

rhardy613:
To quote the website "V1 development has been stopped at the end of 2014 (1.10.1 build 16060)."
It is safe to say those packages have been abandoned years ago. I think the reason why there were abandoned is the dependancies footprint for ASSP is large and it grew with when V2 was released so no one is packaging it. The concept of pristine sources which I consider a requirement of properly packaged software, does not seem popular in the ASSP community. It is the only production software I use without proper packaging, even though I have packaged all the missing dependancies for Ubuntu in my personal PPA on Launchpad. My current ASSP system depends on 52 Perl modules and I had to package around 16 of them myself. Some day I may try to package ASSP but it won't be a quick undertaking.

rhardy613:
Hopefully this is ready to merge now.

yarikoptic:
Perfect
@sebres, how does it look to you?

yarikoptic:
Oh, merge recent master please or rebase and push . Conflicts now. I guess in changelog

sebres:
looks good, I had made `[\w\-]+\s+` instead of `[m0-9\-]+\s+`, is IMHO precisely enough

sebres:
Merged, but... by the way...

I've a little branch with extended interpolation (for both main-streams, but it will be especially needed for 0.10, to allow me and @TorontoMedia to continue later our frozen work with new 0.10-th firewallcmd-* actions), that I'll soon push partially as a PR.

So the suggested ASSP config can and will be rewritten as first example of using of the extended interpolation...
As foretaste the config example:
### assp.conf:
```
[DEFAULT]
extended_interpolation = on

# Note: Specify the ASSP version (e.g. in jail.local, recommended), for compatibility reasons currently both are active.
version = DEFAULT

__assp_actions = (?:dropping|refusing)

failregex = ${v2:failregex}
            ${v1:failregex}

[v1]
failregex = ^(:? \[SSL-out\])? <HOST> max sender authentication errors \(\d{,3}\) exceeded -- %(__assp_actions)s connection - after reply: \d{3} \d{1}\.\d{1}.\d{1} Error: authentication failed: \w+;$
            ^(?: \[SSL-out\])? <HOST> SSL negotiation with client failed: SSL accept attempt failed with unknown error.*:unknown protocol;$
            ^ Blocking <HOST> - too much AUTH errors \(\d{,3}\);$

[v2]
failregex = ^\s*(?:[\w\-]+\s+)*(?:\[\S+\]\s+)*<HOST> (?:\<\S+@\S+\.\S+\> )*(?:to: \S+@\S+\.\S+ )*relay attempt blocked for(?: \(parsing\))?: \S+$
            ^\s*(?:[\w\-]+\s+)*(?:\[\S+\]\s+)*<HOST> \[SMTP Error\] 535 5\.7\.8 Error: authentication failed:\s+(?:\S+|Connection lost to authentication server|Invalid authentication mechanism|Invalid base64 data in continued response)?$

[Definition] 

failregex = ${%(version)s:failregex}
```
### jail.local:
```
[assp]
enabled = True
filter=assp[version=v2]
```

A new parameter `extended_interpolation = on` enables extended interpolation per config file (in defeault section) or per section (if specified there)... 
This one is to be backwards compatible in some action configs, that already used `${xxx}` as shell variables...

@yarikoptic Because we are python-2.x compatible I've extended our SafeConfigParserWithIncludes with extended interpolation, so I can very fast rewrite the standard python extended interpolation syntax `${section:option}` with something analog our known-syntax, e.g. like `%(section/option)s` or even `%(section:option)s`. 
What do you think about? Which syntax would be reliable resp. better fits our "standard"?

sebres:
@yarikoptic ping (have you my question overlooked? or as usual just too busy :)))
On the whole, the question is only about a new syntax: `${section:option}`,  `%(section/option)s` or even `%(section:option)s`...
Additionally I want to resolve thereby some things for the 0.10-th branch, see https://github.com/fail2ban/fail2ban/pull/1447#issuecomment-221610428

Apart from, the benefit of using of `%(/)s` or `%(:)s` (opposite to `${:}`) would be:
- the same syntax (as already used + anyway default python does not allow mixing of basic and extended interpolation)
- clear syntax: e.g. no mixins that can be confused with shell variables, if that are specified and overload some sections from include-configs resp. local-configs...
- waiver of the new parameter "extended_interpolation = on" - because then it can be enabled per default

Cons would be:
- incompatibility to real extended interpolation of python self;
- minor backwards incompatibility if someone already used `:` or `/` in option names and already available substitutions (which I doubt)

IMHO, the pros outweigh the cons, but ... the boss should exercise one's authority :)

yarikoptic:
@sebres sorry -- missed... I guess worth its own issue or PR to discuss.  In this particular case I would have been ok with having separate filters assp_v1 and assp_v2 which would have accomplished the same without requiring any additional handling.  To say the truth I am already partially lost in all available handling and specifications we added on top of "standard" config parsing.  E.g. how suggested changes relate (replace? compliment? could also be done via existing method) to what we have already for handling sections conditionally on some value e.g. `[Init?family=inet6]`.

rhardy613:
@sebres that new versioned ASSP config looks cool. Either separate files or the merged one would be an improvement. I disliked adding v1 regex i.e. running regex which we know should never match. It wastes resources. I think the versioned one would be better from a backward compatibility point of view. If someone upgraded fail2ban and they didn't notice the new asspv2.conf (or whatever we call it) they would be back into a running filter than will never match. To guarantee that didn't happen you would want the fail2ban packaging (or even the runtime) to figure out what version you are running, but that gets messy quickly, especially when ASSP itself frequently isn't packaged.

yarikoptic:
Note that versions switch is not automagic - sunshine would need to tune config anyways. Eventually we might want to add some versions discovery though

sebres:
(Pseudo)conditionally options is not the same `[Init?family=inet6]`. That controls a "condition" to the runtime (corresponding some failure parameters like ip-family), so it manages "dynamic" things like `<option>`. The syntax `%(...)s` controls a "static" parameter to the load- resp. inlcude-time.

The assp.conf should be here an example only. Separating of the configs is not a solution, IMHO it is a dirty workaround, because exactly so arises 10 config files (filters or actions), doing almost the same, but differ with 1 (or 2) parameters... (copy-paste, unsafer and complex support resp. further development, etc.)
Please read https://github.com/fail2ban/fail2ban/pull/1447#issuecomment-221610428.

Whats about assp.conf self: as already said - it is just an example.
But I think the suggested "extended" config is several times better as currently realized in this PR:
- default do the same as now (both V1 and V2 expressions active)
- for the fine tuning, the version would be controllable with single parameter in jail.local, we should not remove something (V1 expressions) in assp.conf or overload in assp.local (apart from that will be overwritten resp. some expressions can be changed by next package upgrade, so very unsafe, IMHO).
- still only one file (separate is really evil);

Something as auto-recognition of some option is theoretically very nice (and can be realized later), but currently enough a fine tuning. I mean, if you touch a config to enable a jail, specify another port (or even logpath or banaction) etc., you can also specify current version of used product...

bellocarico:
I just want to report that to make the default installation of fail2ban on centos7 I could see the IP appearing in the ipset but not apparent banning happening . From the fail2ban log:

`2018-11-20 20:41:27,609 fail2ban.actions        [107003]: NOTICE  [assp] 45.125.66.105 already banned`

A quick investigation of the iptables reported no reference to fail2ban. So I had to modify the jail.conf 

from this

`action    = iptables-ipset-proto4[name=assp, port="25,465,"]`

to this:

`action    = iptables-ipset-proto4[name=assp, port="25,465"]`

and now it works.

PR #1326, opened 2016-02-17
Added npf.conf for the NPF packet filer

ahpnils:
This file adds support for the NPF packet filter, available on NetBSD since version 6.0

PR #1498, opened 2016-08-07
Added npf.conf for the NPF packet filer, episode 2

ahpnils:
This new action files adds support for the NPF packet filter, available on NetBSD since version 6.0.
This pull request makes https://github.com/fail2ban/fail2ban/pull/1326 obsolete, since I updated the changelog file here.


sebres:
Looks good, as I could judge (corresponding to http://netbsd.gw.com/cgi-bin/man-cgi?npfctl+8+NetBSD-7.0).

But I think it should be master branch (although I can switch base branch to master, but it doesn't work without taking all of the 0.10-th changes also, because this 2 commits only should be rebased to master), so I'll rebase it manually tomorrow if no objections

Issue #1497, opened 2016-08-06
postfix-sasl is not matching failed authentication attempts after upgrade to Ubuntu 16.04.1 LTS


maksyms:
### Environment:
- Fail2Ban version 0.9.3, although according to source code, the issue is also in 0.9.5 and 0.10
- Ubuntu 16.04.1 LTS: 
- [x] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)

### The issue:

After an upgrade to Ubuntu 16.04.1 LTS, fail2ban stopped matching certain postfix-sasl entries. 

#### Steps to reproduce
Example (below is malicious activity, the named server is not supposed to be able to authenticate):

Aug  3 14:28:17 ksusha postfix/smtpd[31211]: warning: mail.foldsandwalker.com[98.191.84.74]: SASL Plain authentication failed: 
Aug  3 15:30:49 ksusha postfix/smtpd[17041]: warning: mail.foldsandwalker.com[98.191.84.74]: SASL Plain authentication failed:

The first line is matched by fail2ban-regex, the second isn't. The reason is that the first line has space after "failed:" and the second doesn't.

#### Any additional information
The issue is in failregex in /etc/fail2ban/filter.d/postfix-sasl.conf. The current value for it is:

failregex = ^%(__prefix_line)swarning: [-._\w]+\[<HOST>\]: SASL ((?i)LOGIN|PLAIN|(?:CRAM|DIGEST)-MD5) authentication failed(: [ A-Za-z0-9+/:]*={0,2})?\s*$

The fix is trivial, to delete space between ":" and "[" at the end of regex:

failregex = ^%(__prefix_line)swarning: [-._\w]+\[<HOST>\]: SASL ((?i)LOGIN|PLAIN|(?:CRAM|DIGEST)-MD5) authentication failed(:[ A-Za-z0-9+/:]*={0,2})?\s*$

I hope this helps someone.

yarikoptic:
Great, thanks for the detailed report and analysis! So we just need to make trailing space optional `\s*$`.  Should we do it or you would like to send a PR?

On August 6, 2016 6:27:29 AM EDT, maksyms <notifications@github.com> wrote:
>### Environment:
>- Fail2Ban version 0.9.3, although according to source code, the issue
>is also in 0.9.5 and 0.10
>- Ubuntu 16.04.1 LTS: 
[...]

PR #1501, opened 2016-08-08
Accept no space after "failed:"

maksyms:
- [X] I hope this is the right branch; I fixed it for 0.10, however, the same issue is present in master - so I will submit two pull requests, to master as well
- [X] No unit test required, as similar conditions do not get tested
- [X] Issue #1497 is resolved
- [X] current tests run, no problems
- [X] PR is only removing 1 space
- [X] No stylistic change
- [X] Added test line to postfix-sasl test

fix issue #1497


yarikoptic:
thanks, merged squashed for easy picking and #1502 changes will bring it into the master

PR #1508, opened 2016-08-11
fail2ban-python / version-selector

sebres:
- Introduces new command "fail2ban-python", as automatically created symlink to python executable, where fail2ban currently installed (resp. its modules are located).<br>
Pros: 
  - allows to use the same version, fail2ban currently running, e.g. in external scripts just via replace python with fail2ban-python:
  ```diff
-#!/usr/bin/env python
+#!/usr/bin/env fail2ban-python
```
  - always the same pickle protocol
  - the same (and also guaranteed available) fail2ban modules
  - simplified stand-alone install, resp. stand-alone installation possibility via setup (like gh-1487) is getting closer
- Fixed pythonic filters and test scripts (running via "fail2ban-python" now);
- Fixed test case "testSetupInstallRoot" for not default python (also using direct call, out of virtualenv);

Closes gh-1506 (and few others issues, we had)

If we assume this, I'll make back-porting and reintegrating of this into 0.9-th branch...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,11 @@ class install_scripts_f2b(install_scripts):
 
 
 # Update fail2ban-python env to current python version (where f2b-modules located/installed)
-updatePyExec(os.path.join(os.path.dirname(__file__), 'bin'))
+rootdir = os.path.realpath(os.path.dirname(
+	# __file__ seems to be overwritten sometimes on some python versions (e.g. bug of 2.6 by running under cProfile, etc.):
+	sys.argv[0] if os.path.basename(sys.argv[0]) == 'setup.py' else __file__
+))
+updatePyExec(os.path.join(rootdir, 'bin'))
 
 if setuptools and "test" in sys.argv:
 	import logging
--- a/setup.py
+++ b/setup.py
@@ -38,11 +38,45 @@ except ImportError:
 	# python 2.x
 	from distutils.command.build_py import build_py
 	from distutils.command.build_scripts import build_scripts
+# all versions
+from distutils.command.install_scripts import install_scripts
+
 import os
 from os.path import isfile, join, isdir, realpath
 import sys
 import warnings
 from glob import glob
+
+
+def updatePyExec(bindir, executable=None):
+	"""Update fail2ban-python link to current python version (where f2b-modules located/installed)
+	"""
+	bindir = os.path.realpath(bindir)
+	if executable is None:
+		executable = sys.executable
+	pypath = os.path.join(bindir, 'fail2ban-python')
+	# if not exists or point to another version - update link:
+	isfile = os.path.isfile(pypath)
+	if not isfile or os.path.realpath(pypath) != os.path.realpath(executable):
+		if isfile:
+			os.unlink(pypath)
+		os.symlink(executable, pypath)
+
+
+# Wrapper to install python binding (to current python version):
+class install_scripts_f2b(install_scripts):
+
+	def get_outputs(self):
+		outputs = install_scripts.get_outputs(self)
+		fn = None
+		for fn in outputs:
+			if os.path.basename(fn) == 'fail2ban-server':
+				break
+		bindir = os.path.dirname(fn)
+		print('creating fail2ban-python binding -> %s' % (bindir,))
+		updatePyExec(bindir)
+		return outputs
+
 
 if setuptools and "test" in sys.argv:
 	import logging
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by sebres** [comment]
Hmm, the run in virtualenv is doubtlessly "different" ;) fixing

File: setup.py [hunk 1]
```diff
@@ -38,12 +38,46 @@
...
+	bindir = os.path.realpath(bindir)
+	if executable is None:
+		executable = sys.executable
+	pypath = os.path.join(bindir, 'fail2ban-python')
+	# if not exists or point to another version - update link:
+	isfile = os.path.isfile(pypath)
+	if not isfile or os.path.realpath(pypath) != os.path.realpath(executable):
+		if isfile:
+			os.unlink(pypath)
+		os.symlink(executable, pypath)
```
yarikoptic: Duplicate code? Could be placed into a module under fail2ban/ and then imported explicitly from setup. Py. This works avoid duplication

yarikoptic:
In general  +1 
Might be better to make it against 0.9 from the beginning so we just merge it into 0.10 .

yarikoptic:
Yet to try etc

yarikoptic:
While at it shouldn't it just take care about standalone deployments? Ie shouldn't be just a symlink but a wrapper which would adjust env to point to configuration files etc?

setup.py
sebres: already done (moved into fail2ban/setup.py)

sebres:
As first enough a symlik... 
Later or even in standalone mode we can replace it with "wrapper", that currently will look as something like that:
```bash
#!/usr/bin/env sh
PYTHONPATH=.. python "$@"
```
It will be easy-fix...

BTW. I prefer currently a symlink (to symlink), because theoretically it holds fail2ban python-version independent, I mean "upgrade compatible" (e.g. 3.3-\>3.4, 3.5-\>3.6, etc.)...

sebres:
BTW does anyone have idea, why the single test case `testSetupInstallRoot` runs in python2.x fewer as 1 second, but within python3.x almost 15 seconds and python 3.5 - 18 seconds?!

To check it, run: `fail2ban-testcases SetupInstallRoot`

Strange is also a profiler-output in python 3.x:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    185/1    0.015    0.000   15.104   15.104 {built-in method exec}
        1    0.000    0.000   15.104   15.104 fail2ban-testcases:5(<module>)
        1    0.000    0.000   14.845   14.845 runner.py:144(run)
        1    0.000    0.000   14.844   14.844 suite.py:83(__call__)
        1    0.000    0.000   14.844   14.844 suite.py:102(run)
        1    0.000    0.000   14.844   14.844 case.py:624(__call__)
        1    0.000    0.000   14.844   14.844 case.py:545(run)
        1    0.000    0.000   14.812   14.812 misctestcase.py:107(testSetupInstallRoot)
        2   14.781    7.390   14.781    7.390 {built-in method system}
   197/28    0.001    0.000    0.237    0.008 <frozen importlib._bootstrap>:2234(_find_and_load)
   197/28    0.001    0.000    0.236    0.008 <frozen importlib._bootstrap>:2207(_find_and_load_unlocked)
   227/30    0.000    0.000    0.232    0.008 <frozen importlib._bootstrap>:313(_call_with_frames_removed)
   188/30    0.001    0.000    0.232    0.008 <frozen importlib._bootstrap>:1186(_load_unlocked)
   163/30    0.000    0.000    0.230    0.008 <frozen importlib._bootstrap>:1122(_exec)
   163/30    0.001    0.000    0.230    0.008 <frozen importlib._bootstrap>:1465(exec_module)
```

I suspect it is `os.system("%s %s install --root=%s >/dev/null" ...` and `os.system("%s %s clean --all >/dev/null 2>&1"` ...  but why?

Here is output in python 2.7:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.001    0.001    0.918    0.918 fail2ban-testcases:5(<module>)
        1    0.000    0.000    0.743    0.743 runner.py:140(run)
        1    0.000    0.000    0.742    0.742 suite.py:69(__call__)
        1    0.000    0.000    0.742    0.742 suite.py:88(run)
        1    0.000    0.000    0.742    0.742 case.py:394(__call__)
        1    0.000    0.000    0.742    0.742 case.py:297(run)
        1    0.000    0.000    0.721    0.721 misctestcase.py:107(testSetupInstallRoot)
        2    0.694    0.347    0.694    0.347 {posix.system}
        1    0.005    0.005    0.139    0.139 utils.py:167(gatherTests)
      194    0.001    0.000    0.044    0.000 re.py:226(_compile)
       62    0.000    0.000    0.043    0.001 sre_compile.py:493(compile)
       71    0.000    0.000    0.042    0.001 re.py:188(compile)
```

sebres:
meantime I know what it is: setup makes 2to3 converting also (but it needless in the test-cases, because should be already done before starting of the test cases (fail2ban-2to3), so I made new setup option "--disable-2to3", that disables it by running from `testSetupInstallRoot`...

Result:
```
$ bin/fail2ban-testcases Setup
Fail2ban 0.10.0a2 test suite. Python 3.4.3 (default, Oct 14 2015, 20:28:29) [GCC 4.8.4]. Please wait...
.
Ran 1 test in 0.664s
```

New easy-fix PR comes after merging of this one... (but only for 0.10, because too different, but also no matter in 0.9, which test-cases are anyway too slow)

sebres:
Sometimes I'll really disable this both coverage tools... WTF?! - because it was definitely increased if we take a look inside both...

So I'll merge if no contradictions.

sebres:
done and back-ported to 0.9 (tree reintegrated with back-merge to 0.10)

PR #1510, opened 2016-08-12
 0.10/optimize setup test case

sebres:
* better handling of setup in test-cases (prevent execution of 2to3 twice, full logging in heavydebug, etc.)
  see https://github.com/fail2ban/fail2ban/pull/1508#issuecomment-239410329
* optimized logging using LogCaptureTestCase (new faster handler _MemHandler) - decreases logging overhead by implementing of lazy converting of log record to string, that prevents of unnecessary converts (important for methods that do many log entries, because LogCaptureTestCase set loglevel to DEBUG);
* backwards compatibility for python 2.6 (SkipTest implemented), so options --fast and --no-network may be used now within python2.6 also;


sebres:
Additionally this PR introduces new setup wrapper `install_command_f2b`, that may be used later to supply some fail2ban own options for installation or build process (e. g. `--standalone` etc.) ...

sebres:
The test cases are successful now, but strange error by executing of `pip install .` in all versions:
```
error: option --single-version-externally-managed not recognized
```
I've found some discussions about, e. g. http://stackoverflow.com/questions/14296531/what-does-error-option-single-version-externally-managed-not-recognized-ind

But no savvy why it occurred now in this PR... 

@yarikoptic do you have an idea?

sebres:
Error with `--single-version-externally-managed` seems to be resolved: I had an idea, that the setup uses install command from `distutils` package instead of another one from `setuptools`.
So 2e530fcf72c9e9bae321b37b625728bf9f88f71b seems to fix it.

fail2ban/tests/utils.py
```diff
@@ -368,6 +448,38 @@ def assertNotIn(self, a, b, msg=None):
 
 class LogCaptureTestCase(unittest.TestCase):
 
+	class _MemHandler(logging.StreamHandler):
+		def __init__(self):
+			self._recs = list()
+			self._strm = StringIO()
+			logging.StreamHandler.__init__(self, self._strm)
+			#
```
yarikoptic: sorry to be a stylistic pain but why to have a `#` here?  we have them nowhere else (and I hope will not). so why?

fail2ban/tests/utils.py
```diff
@@ -368,6 +448,38 @@ def assertNotIn(self, a, b, msg=None):
 
 class LogCaptureTestCase(unittest.TestCase):
 
+	class _MemHandler(logging.StreamHandler):
```
yarikoptic: Please provide the docstring describing its purpose and basic operation (not to delegate logging to StreamHandler at all, format lazily on demand in getvalue, etc)

fail2ban/tests/utils.py
sebres: I could find at least 10 such code pices, but I suspect it will be a large discussion, so I'll replace it with space.

sebres:
I had sometimes sporadicall errors like "not-logged" in 0.10 on very fast computers - now I know why, it was missing lock if multi-threaded (logging in another thread as test case running), see 2b5c5103f463e1fa6ab646a28d813af63389a38f for details...

fail2ban/setup.py
```diff
@@ -32,7 +32,7 @@ def updatePyExec(bindir, executable=None):
 		executable = sys.executable
 	pypath = os.path.join(bindir, 'fail2ban-python')
 	# if not exists or point to another version - update link:
-	isfile = os.path.isfile(pypath)
+	isfile = os.path.isfile(os.path.realpath(pypath))
```
yarikoptic: shouldn't it be just `not exists(pypath)` then if you would dereference it here?

yarikoptic: otherwise I guess it is ok ;)

sebres: "isfile" is stricter as "exists", should it point (however) to a directory...
And dereferencing makes sure that the target exists (because link may point to nonexistent target)

Issue #1507, opened 2016-08-10
Date parse failure if hostname begins with digits


Sut3kh:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.9.3
- OS, including release name/version: CentOS Linux release 7.2.1511 (Core)
- [X] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase
- [X] Some customizations were done to the configuration (provide details below is so)

### The issue:

Using a log format like `Aug 8 11:25:50 hostname ...` causes the date parsing to fail if the hostname begins with digits.

#### Steps to reproduce

- Create logfile in the format of `Aug 4 18:19:37 14430f2329b8 ...`
- Add logfile to a jail
- Run fail2ban

#### Expected behavior

In a perfect world:

- fail2ban-regex should confirm the template as `(?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?`
- it should note that there is no space after '1443' and parse the date as `Aug 4 18:19:37 2016`

#### Observed behavior

The date matches the template `(?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?` but assumes the first 4 digits of the hostname are the year, flooding fail2ban.log with errors every tick.

#### Any additional information

I know this seems like a very rare/strange use case and it probably is in a way but i came across it by running multiple wordpress sites each in their own docker container (hence the random hostnames) and each with the [WP fail2ban](https://en-gb.wordpress.org/plugins/wp-fail2ban/) plugin installed which outputs the date without a year, thereby randomly causing this situation now and then depending on the GUID generated by docker.

It was simple to workaround by updating the plugin to use `$date->format('M j H:i:s Y')` (excluding the year doesn't seem like a great idea anyway) but i figured i should at least report it for others to find, even if its not worth fixing.

#### Any customizations done to /etc/fail2ban/ configuration

n/a

#### Relevant parts of /var/log/fail2ban.log file:

```
2016-08-10 15:45:05,540 fail2ban.filter         [19785]: ERROR   findFailure failed to parse timeText: Aug 8 11:25:50 1443
```

#### Relevant lines from monitored log files in question:

```
Aug 8 11:25:50 14430f2329b8 Authentication attempt for unknown user test_p43jBhTw from 1.2.3.4
```

sebres:
> I know this seems like a very rare/strange use case

[...]

PR #1512, opened 2016-08-15
Fix ambiguous wrong recognized date pattern resp. its optional parts

sebres:
amend to (long time ago fixed left word-boundary) b6bb2f88c1dbb111647269590d80d95f72c81c3e: datepattern right word-boundary - prevents confusions if end of date-pattern (e.g. optional year part) misleadingly match not date values (see gh-1507)

test cases extended to check ambiguous "unbound" patterns in log lines (match/miss resp. positive/negative cases)

### DIFF
--- a/fail2ban/server/datetemplate.py
+++ b/fail2ban/server/datetemplate.py
@@ -64,7 +64,7 @@ class DateTemplate(object):
 	def getRegex(self):
 		return self._regex
 
-	def setRegex(self, regex, wordBegin=True):
+	def setRegex(self, regex, wordBegin=True, wordEnd=True):
 		"""Sets regex to use for searching for date in log line.
 
 		Parameters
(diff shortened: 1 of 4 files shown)


sebres:
This does not fix directly #1507, but will prevent in the future such wrong recognition also without specifying of exact date-pattern. 

In addition, such boundaries (as well as left or right) make the search of date much faster, in particular after merging in our faster branch 0.10...

fail2ban/server/datetemplate.py
```diff
@@ -64,7 +64,7 @@ def name(self, name):
 	def getRegex(self):
 		return self._regex
 
-	def setRegex(self, regex, wordBegin=True):
+	def setRegex(self, regex, wordBegin=True, wordEnd=True):
```
yarikoptic: There is even a docstring with description of wordbegin... Would be nice to note the wordend there I guess

yarikoptic:
Since tests pass, I guess it doesn't ruin any date matching and I am cool to have it merged. But we need a change log entry stating that dates now are checked to end at the words boundary.

sebres:
Done, if tests pass, I'll merge it into both branches (0.9 and 0.10)

PR #1516, opened 2016-08-18
 0.10/let know version

sebres:
- let know the version (and python), also in travis (no output if verbosity is 0)
- small code review: simplify fail2ban-testcases - move some code pieces inclusive option parser from fail2ban-testcases to tests/utils.py (+ coverage)

### DIFF
--- a/bin/fail2ban-testcases
+++ b/bin/fail2ban-testcases
@@ -37,10 +37,8 @@ if os.path.exists("fail2ban/__init__.py"):
 	sys.path.insert(0, ".")
 from fail2ban.version import version
 
-from fail2ban.tests.utils import initProcess, gatherTests
+from fail2ban.tests.utils import getOptParser, initProcess, gatherTests
 from fail2ban.setup import updatePyExec
-
-from optparse import OptionParser, Option
 
 # Update fail2ban-python env to current python version (where f2b-modules located/installed)
 bindir = os.path.dirname(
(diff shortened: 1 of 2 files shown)


fail2ban/tests/utils.py
```diff
@@ -123,7 +167,12 @@ def initProcess(opts):
 	#
 	stdout.setFormatter(Formatter(fmt))
 	logSys.addHandler(stdout)
-	#
+
+	# Let know the version
+	if opts.verbosity != 0: # pragma: no cover
```
yarikoptic: seriously -- why this code could/shouldn't be tested?  it could and should imho, thus 'no cover' is not warranted

sebres: copy-paste (it was already so =) I'll rebase the last commit without that...

bin/fail2ban-testcases
```diff
@@ -103,13 +56,10 @@ opts = initProcess(opts)
 verbosity = opts.verbosity
 
 #
-# Let know the version
+# Gather tests (and filter corresponding options)
 #
-if opts.log_level != logging.CRITICAL: # pragma: no cover
```
sebres: from here ...

fail2ban/tests/utils.py
sebres: done

sebres:
So the travis plays up sometimes also - take a look over current build, python 3.3, - it runs with something else as fail2ban (LLVM, ElasticSearch, etc. =)))

yarikoptic:
travis failure (3.3 only) is irrelevant to this changes and others remained green, so merging.

Issue #1519, opened 2016-08-19
Fail2ban not banning Wrong passwords attempt with asterisk


wonderer007:
I have configured fail2ban with asterisk using [tutorial](http://www.mikeslab.net/?p=353) but its banning IP's with wrongs passwords attempt. It seems like regex is not working, please find my regex and asterisk log below

regex in _asterisk.conf_
```
failregex = NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - Wrong password
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - No matching peer found
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - No matching peer found
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - Username/auth name mismatch
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - Device does not match ACL
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - Peer is not supposed to register
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - ACL error (permit/deny)
            NOTICE.* .*: Registration from '.*' failed for '<HOST>:.*' - Device does not match ACL
            NOTICE.* .*: Registration from '\".*\".*' failed for '<HOST>:.*' - No matching peer found
            NOTICE.* .*: Registration from '\".*\".*' failed for '<HOST>:.*' - Wrong password
            NOTICE.* <HOST> failed to authenticate as '.*'$
            NOTICE.* .*: No registration for peer '.*' \(from <HOST>\)
            NOTICE.* .*: Host <HOST> failed MD5 authentication for '.*' (.*)
            NOTICE.* .*: Failed to authenticate user .*@<HOST>.*
            NOTICE.* .*: <HOST> failed to authenticate as '.*'
            NOTICE.* .*: <HOST> tried  to authenticate with nonexistent user '.*'
            VERBOSE.*SIP/<HOST>-.*Received incoming SIP connection from unknown peer
```
asterisk log

> [2016-08-19 11:11:26] NOTICE[12931]: chan_sip.c:28468 handle_request_register: Registration from '<sip:bob@37.188.117.192>' failed for '39.59.118.146:42406' - Wrong password

sebres:
Your first regex matches it:
```
[...]

PR #1522, opened 2016-08-22
filter.d/asterisk.conf: another part ` chan_sip.c:28468 handle_request_register:` in log prefix

sebres:
Fixed match of different asterisk log prefix (source file:line method:), e.g. ` chan_sip.c:28468 handle_request_register:`
Part ` in` seems to be optional in some versions.

Example log line from #1519:
```
[2016-08-19 11:11:26] NOTICE[12931]: chan_sip.c:28468 handle_request_register: Registration from 'sip:bob@192.0.2.1' failed for '192.0.2.1:42406' - Wrong password
```


sebres:
I've found that this is only a matter of part ` in`, that seems to missing in some versions.
So rebased now with better expression...

PR #1523, opened 2016-08-24
 0.10/systemd journal path option (see gh-1408)

### LINKED ISSUE #1408: Feature: monitor multiple systemd journals (machinectl, systemd-nspawn)
I have a separate `sshd` on a separate port in a `systemd-nspawn` container. Host `sshd` is nicely protected, but container `sshd` gets hammered and `fail2ban` running on the host is not able to do anything about it, because it doesn't see the container's journal. Seeing the container's journal is trivial on the commandline `journalctl -M <machinename>`, so I'd assume it may not be terribly difficult to get `fail2ban` to monitor more than one journal like this. Perhaps nobody has had the need to think about this before - searching issues for `machinectl` returns empty.

Thoughts?

sebres:
- jail configuration extended with new syntax to pass options to the backend (see gh-1408)
examples:
```
  backend = systemd[journalpath=/run/log/journal/machine-1]
  backend = systemd[journalfiles="/run/log/journal/machine-1/system.journal, /run/log/journal/machine-1/user.journal"]
  backend = systemd[journalflags=2]
```
- optimized FilterSystemd method `run`: better wait in idle (no busy-loop), better poll handling, the ban will executed anywhere (at least at 100th log-entry), also if we have never ending logging in this jail (e.g. extremely logging or too many failures)
- systemd test cases extended

Closes #1408


sebres:
several code pieces of this PR moved to master base (0.9) and merged hereafter within 0.10, 
additionally I've found wrong (resp. missing) idle handling in several filters, so fixed and covered here (currently 0.10 only).

Issue #656, opened 2014-03-18
ufw action doesn't call ufw insert


schaal:
Using this jail on fail2ban 0.9.0:

```
[sshd]
enabled = true
action = ufw[application=SSH]
```

And using the default ufw action with this actionban:

```
actionban = [ -n "<application>" ] && app="app <application>" ; ufw insert <insertpos> <blocktype> from <ip> to <destination> $app
```

I get this output when an IP gets banned:

```
2014-03-18 17:53:32,413 fail2ban.server.actions[23588]: NOTICE  [sshd] Ban 192.168.178.11
2014-03-18 17:53:32,419 fail2ban.server.action[23588]: DEBUG   
2014-03-18 17:53:32,421 fail2ban.server.action[23588]: DEBUG   Nothing to do
2014-03-18 17:53:32,425 fail2ban.server.action[23588]: DEBUG   [ -n "SSH" ] && app="app SSH"
2014-03-18 17:53:32,561 fail2ban.server.action[23588]: DEBUG   [ -n "SSH" ] && app="app SSH" -- stdout: ''
2014-03-18 17:53:32,564 fail2ban.server.action[23588]: DEBUG   [ -n "SSH" ] && app="app SSH" -- stderr: ''
2014-03-18 17:53:32,566 fail2ban.server.action[23588]: DEBUG   [ -n "SSH" ] && app="app SSH" -- returned successfully
```

So it seems `ufw insert` never gets called.

It works when the `;` before `ufw insert` is replaced by `&&`:

```
actionban = [ -n "<application>" ] && app="app <application>" && ufw insert <insertpos> <blocktype> from <ip> to <destination> $app
```

PR #1532, opened 2016-08-31
fixed a bug in action ufw when using app with space

magapp:
I was trying to use UFW app "Apache Full", which is installed default in Ubuntu. I've added this in the jail:

ufw[application=Apache Full]

Due to the space in the application argument "Apache Full", the action failed. It does not matter if you escape quotes or space. I even tried double escaping. The only way I could get it to work was to evaluate $app before bash is executing it.

Perhaps there is another way of getting this to work, but in that case it would be great with an example in the action.

### DIFF
--- a/config/action.d/ufw.conf
+++ b/config/action.d/ufw.conf
@@ -13,11 +13,11 @@ actionstop =
 
 actioncheck = 
 
-actionban = [ -n "<application>" ] && app="app <application>"
-            ufw insert <insertpos> <blocktype> from <ip> to <destination> $app
+actionban = [ -n "<application>" ] && app="app '<application>'"
+            eval ufw insert <insertpos> <blocktype> from <ip> to <destination> $app
 
-actionunban = [ -n "<application>" ] && app="app <application>"
-              ufw delete <blocktype> from <ip> to <destination> $app
+actionunban = [ -n "<application>" ] && app="app '<application>'"
+              eval ufw delete <blocktype> from <ip> to <destination> $app
 
 [Init]
 # Option: insertpos
@@ -40,3 +40,4 @@ application =
 # 
 # Author: Guilhem Lettron
 # Enhancements: Daniel Black
+# Enhancements: Magnus Appelquist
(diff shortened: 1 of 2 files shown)


sebres:
Usage of `eval` is evil, at least if you've foreign-input and/or variable arguments...
So you can get dirty data resp. even vulnerability.
No idea currently using pure posix shell (so without some bashism's like arrays, parameters expansion, subst with escape, etc.).
But `eval` is also not a proper way IMHO.

config/action.d/ufw.conf
```diff
@@ -13,11 +13,11 @@ actionstop =
 
 actioncheck = 
 
-actionban = [ -n "<application>" ] && app="app <application>"
-            ufw insert <insertpos> <blocktype> from <ip> to <destination> $app
+actionban = [ -n "<application>" ] && app="app '<application>'"
+            eval ufw insert <insertpos> <blocktype> from <ip> to <destination> $app
```
sebres: please no eval here...

sebres:
From @IaroslavR (#1909):

due to this and https://github.com/fail2ban/fail2ban/issues/656#issuecomment-38027046 actions should be edited for normal operation. I can't understand the meaning of the current template. Whats wrong with simplest solution like this?
```init
actionunban = ufw delete <blocktype> from <ip> to <destination>
actionban = ufw insert <insertpos> <blocktype> from <ip> to <destination>
```

sebres:
> Whats wrong with simplest solution like this?

Nothing is wrong, just our action `uwf` can currently restrict the application too (resp. use application profile, to specify port, protocol etc). But currently it is affected by the issue "if appication name contains space", like "Apache Full"... 
This PR should fix it (but I don't like the `eval` there)...

sebres:
I don't closed it, this looks like a github BUG: several PRs closed - I pushed something in 0.10 (and branch 0.10 disappeared for some reason, together with all PRs for this branch)...

sebres:
implemented in #3018

PR #1344, opened 2016-03-04
WiP BF: decode explicitly the string in systemd after concatenation

yarikoptic:
Before submitting your PR, please review the following checklist:

- [ ] **CONSIDER adding a unit test** if your PR resolves an issue
- [x] **LIST ISSUES** this PR resolves:  Closes #1341
- [ ] **MAKE SURE** this PR doesn't break existing tests 
- [x] **KEEP PR small** so it could be easily reviewed.
- [x] **AVOID** making unnecessary stylistic changes in unrelated code
- [ ] **ACCOMPANY** each new `failregex` for filter `X` with sample log lines 
      within `fail2ban/tests/files/logs/X` file

Not sure yet if this is any correct thing to do -- just want to check

Issue #1341, opened 2016-02-29
sshd jail does not find valid matches in the journal


goeranu:
### Environment:
- Fail2Ban version (including any possible distribution suffixes): 0.9.3-1.fc23
- OS, including release name/version: Fedora 23
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

### The issue:
For some reason, the _sshd_ jail isn't finding the matches it ought to find.  I have configured fail2ban to use the systemd backend by default, and it seems to work for example for the _dovecot_ jail, but not for _sshd_.  I've tried to match the strings from `journalctl` commands, and the expressions do match.  But nothing is banned, and no matches show up in `fail2ban.log`.

#### Steps to reproduce
Enable the _sshd_ jail and wait for attacks.

#### Expected behavior
Matches listed in `/var/log/fail2ban.log`, and after a few matches from the same host, that host being banned.

#### Observed behavior
No obvious action from fail2ban.

### Configuration, dump and another helpful excerpts
This is the status of the jail:
```
mimmi$ sudo fail2ban-client status sshd
[sudo] password for root: 
Status for the jail: sshd
|- Filter
|  |- Currently failed: 0
|  |- Total failed:     0
|  `- Journal matches:  _SYSTEMD_UNIT=sshd.service + _COMM=sshd
`- Actions
   |- Currently banned: 0
   |- Total banned:     0
   `- Banned IP list:
```
#### Any customizations done to /etc/fail2ban/ configuration
I've done these relevant customizations in `jail.local`
```
[DEFAULT]
bantime = 86400
backend = systemd
…
[sshd]
enabled = true
…
```
(Several other jails are also enabled, excluded here.)
#### Relevant parts of /var/log/fail2ban.log file:
The only thing that shows up are the standard messages at startup, like these:
```
2016-02-29 19:00:05,535 fail2ban.jail           [24435]: INFO    Creating new jail 'sshd'
2016-02-29 19:00:05,558 fail2ban.jail           [24435]: INFO    Jail 'sshd' uses systemd
2016-02-29 19:00:05,581 fail2ban.jail           [24435]: INFO    Initiated 'systemd' backend
2016-02-29 19:00:06,902 fail2ban.filter         [24435]: INFO    Set maxRetry = 5
[...]

PR #1542, opened 2016-09-06
fix systemd convert gh 1341

sebres:
systemd-filter: fixed tricky bug, like `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3...: ordinal not in range(128)`

* [BF] fixed journal systemd ascii/utf-8 default converting (closes gh-1341, closes gh-1344)
* [BF+ENH] allow to set default or preferred encoding for other filters as file-filter (e. g. to decode bytes from journal)
* [performance] ignore obsolete log-lines moved before searching of matched failregex
(`if time < now - findtime`)
* extended test cases to cover such bugs...
* [+0.10] optimized test cases after merge (using 0.10-features, see e3a75b4)

To compare changes between 0.10 / 0.9:
- 0.10: https://github.com/fail2ban/fail2ban/compare/0.10...sebres:_0.10/fix-systemd-convert-gh-1341 (this PR, merged from 0.9 + small code review)
- 0.9: https://github.com/fail2ban/fail2ban/compare/master...sebres:_0.9/fix-systemd-convert-gh-1341
(fast-forward)

### DIFF
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -86,6 +86,8 @@ class Filter(JailThread):
 		self.__lastDate = None
 		## External command
 		self.__ignoreCommand = False
+		## Default or preferred encoding (to decode bytes from file or journal):
+		self.__encoding = locale.getpreferredencoding()
 		## Error counter
 		self.__errors = 0
 		## Ticks counter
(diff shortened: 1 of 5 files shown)


sebres:
I did understand now, why coveralls have sometimes strange results (decreasing)...
According to https://coveralls.io/builds/7763452, 2 covered lines was removed (so if it calculate it percentage - it is really decreased)... 
So the codecov seems to be better in the calculating of comparison to previous version... (but we should nevertheless keep both, because e. g. coveralls has good feature to differentiate python-version-related code)

sebres:
looks good for me (and tested from @goeranu) - so merged now in both major branches

PR #1555, opened 2016-09-22
Monit config: scripting is not supported in path into master

szepeviktor:
By the author:

> Yes, scripting is not supported in path.

https://bitbucket.org/tildeslash/monit/issues/372/webadmin-shows-only-the-first-part-of#comment-27946048


szepeviktor:
Big misttake!

Going the console way....

PR #1556, opened 2016-09-22
Monit config: scripting is not supported in path 

szepeviktor:
> Yes, scripting is not supported in path.

https://bitbucket.org/tildeslash/monit/issues/372/webadmin-shows-only-the-first-part-of#comment-27946048

### DIFF
--- a/files/monit/fail2ban
+++ b/files/monit/fail2ban
@@ -1,7 +1,7 @@
 check process fail2ban with pidfile /var/run/fail2ban/fail2ban.pid
     group services
     start program = "/etc/init.d/fail2ban force-start"
-    stop  program = "/etc/init.d/fail2ban stop || :"
+    stop  program = "/etc/init.d/fail2ban stop"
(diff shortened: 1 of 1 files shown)


szepeviktor:
These things make me feel noone uses Fail2ban + Monit.

sebres:
Indeed, but I can merge it if you mean that is so correct :)

szepeviktor:
I had **a lot** of X cases because of script was in the path.

sebres:
> X cases

:question:

szepeviktor:
See https://en.wikipedia.org/wiki/The_X-Files

szepeviktor:
:ship:

sebres:
The next time, please don't forget a changelog entry, like 8e3e333d54e4cef7332fdc4a8eea25bc1c2b30a5

szepeviktor:
Sorry. I was very tired.

PR #1557, opened 2016-09-22
 0.10/reload-and-more

sebres:
- Reload functionality rewritten (now totally without restart, unbanning/rebanning, etc., resp. optional if needed)
Closes #1534 
- Several commands extended and new commands introduced: 
  - `restart [--unban] [--if-exists] <JAIL>` - restarts the jail \<JAIL\> (alias for `reload --restart ... <JAIL>`)
  - `reload [--restart] [--unban] [--all]` - reloads the configuration without restarting of the server, the option `--restart` activates completely restarting of affected jails, thereby can unban IP addresses (if option `--unban` specified)
  - `reload [--restart] [--unban] [--if-exists] <JAIL>` - reloads the jail \<JAIL\>, or restarts it (if option `--restart` specified), at the same time unbans all IP addresses banned in this jail, if option `--unban` specified
  - `unban --all` - unbans all IP addresses (in all jails and database)
  - `unban <IP> ... <IP>` - unbans \<IP\> (in all jails and database), Closes #1388 
- Because of major branch further development, lot of things in all the filters was rewritten (e. g. because of reload-possibility).
Definitelly fixed high-load of pyinotify-backend, see https://github.com/fail2ban/fail2ban/issues/885#issuecomment-248964591
- Now pyinotify and systemd backends are the fastest and least cpu- resp. system-load greedy.

Additionally:
* Switch down log level for some annoying messages to tracedebug or heavydebug (to 7 or even 5);
* Added verification of specified log-level before transmitting to the server (now fails in client if wrong level);
* Numeric log-level allowed now in server (resp. fail2ban.conf);
* Increase default wait operation (sleep time, threshold interval) - avowedly greater inertance, but fewer system load by many jails resp. log files; leave smaller inertance inside test-cases
* Waiting in threads with `wait_for` extended with verifying of active flag. 
* Implemented better error handling in some multi-threaded routines; shutdown of jails rewritten (faster and safer, does not breaks shutdown process if some error occurred);
* Possibility for overwriting some configuration options (read with config-readers) with command line option, e. g.:
```bash
## start server with DEBUG log-level (ignore level read from fail2ban.conf):
fail2ban-client --loglevel DEBUG start
## or
fail2ban-server -c /cfg/path --loglevel DEBUG start
## keep server log-level by reload (without restart it)
fail2ban-client --loglevel DEBUG reload
## switch log-level back to INFO:
fail2ban-client set loglevel INFO
```
* Normalized usage of preferred encoding (and decode any to string);
[...]

PR #1562, opened 2016-09-29
0.10/fix stability and speed optimization

sebres:
* increases performance of executeCmd (actions), thereby introduced new shorter interval for fast operations (leaves unchanged default wait operation intervals (sleep time, threshold interval) - for the same inertance, to save same system (load by many jails resp. log files);
extends wait_for with callable timeout (test case fixed);

* speedup server start/stop (waiting for communicate, etc);
extend server socket with timeouts, extend ping with timeout parameter;

* standardize and normalize logging and verbosity formats, logging level etc between command lines (server, client, test-cases);
test cases could pass (so increase) verbosity to the client (and furthermore client to the server also), usable for debug purposes resp. simplifying read of the log-file;
custom and precise numeric log-levels can be given in test cases now;

* optimized and extended test-cases

amend for #1557


sebres:
[:angry: mode on]
In the meantime I really hate badip, cymru and other similar "very stable" services, that continuously failed in our test cases again and again, so this shit should be removed from the stock fail2ban and/or test cases, until the authors of its bindings do not write good muck-ups for it. 
I'm seriously now! Because imho my patience was strained enough...
[/:angry: mode off]

sebres:
I'm going crazy, barely "fixed" badips-tests (check availability), now cymru makes problems!

sebres:
cymru availability check direct amended - 276759b6c27673be08f84d0997cb107a4f9a0e4c

PR #1563, opened 2016-09-30
make the ipv6 host regex greedy

niklasf:
Previously the regex was lazily matching ``2606:2800:220:1:248:1893:25c8:1946``
as ``2606:2800:220:1:248:1893:25c8:1``.

Expecting unrelated test failures for [everything but Python 2.6](https://github.com/fail2ban/fail2ban/blob/a406c6eb3a20621e43e84cb644ff2c19873d629f/fail2ban/tests/action_d/test_badips.py#L27) because badips.com seems to be down.

### DIFF
--- a/fail2ban/server/failregex.py
+++ b/fail2ban/server/failregex.py
@@ -77,7 +77,7 @@ class Regex:
 		regex = regex.replace("<F-IP4/>", r); # closed
 		r_host.append(r)
 		# separated ipv6:
-		r = r"""(?P<ip6>(?:[0-9a-fA-F]{1,4}::?|::){1,7}(?:[0-9a-fA-F]{1,4}?|(?<=:):))"""
+		r = r"""(?P<ip6>(?:[0-9a-fA-F]{1,4}::?|::){1,7}(?:[0-9a-fA-F]{1,4}|(?<=:):))"""
 		regex = regex.replace("<IP6>", r); # self closed
 		regex = regex.replace("<F-IP6/>", r); # closed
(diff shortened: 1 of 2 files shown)


sebres:
Thx, 
I don't know anymore why this non-greedy operand was earlier important (but I know it was so once).
But the test-cases seem to pass with this fix (and it sounds plausible), so I'll think about this a little bit, and merge it if I don't remember me (I could swear, there was something, so possible a end-of-word anchor may be better...)

sebres:
We've already correct it here - https://github.com/fail2ban/fail2ban/blob/0.10/fail2ban/server/ipdns.py#L128
So seems to be really a bug...

niklasf:
Looks like the `?` appeared first in d3442742716e8253ca592831ae61b51305571fb1.

sebres:
> Looks like the ? appeared first in d344274

No, it is already there. 

Anyway, it concerns to non-anchored failregex, where nothing is after `<HOST>` (e. g. exact your case with nginx-limit-req)

I've already merged it locally, with two small amend from me - removed code duplication (common IP-expressions combined in ipdns.IPAddr), and fix for nginx-limit-req (comma added at end after `<HOST>`). 
I wait for the test-cases, hereafter will merge it.

sebres:
here you can see the amend for the PR - 276759b6c276...ee1727eccaf5

niklasf:
Thanks for dealing with this so fast :+1:

sebres:
You are welcome! :)

PR #1571, opened 2016-10-07
Drop unneeded shebangs from test scripts

opoplawski:
Resolves this rpmlint error:

```
fail2ban-server.noarch: E: non-executable-script /usr/lib/python3.5/site-packages/fail2ban/tests/files/ignorecommand.py 644 /usr/bin/python3.5 
fail2ban-server.noarch: E: non-executable-script /usr/lib/python3.5/site-packages/fail2ban/tests/files/config/apache-auth/digest.py 644 /usr/bin/python3.5 
```

these scripts are executed with "fail2ban-python <script>" directly so no need to shebang.  Also, they are not executable anyway (644).

### DIFF
--- a/fail2ban/tests/files/config/apache-auth/digest.py
+++ b/fail2ban/tests/files/config/apache-auth/digest.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env fail2ban-python
 import requests
 
 try:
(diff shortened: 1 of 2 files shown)


yarikoptic:
Those seemed intended for execution iirc so having shebang might come useful

opoplawski:
If they are intended to be executed directly they should have the execute bit set.

Issue #1533, opened 2016-08-31
sshd-filter: optional port part


sebres:
see comments below

``` diff
- ^%(__prefix_line)s[iI](?:llegal|nvalid) user .* from <HOST>\s*$
+ ^%(__prefix_line)s[iI](?:llegal|nvalid) user .*? from <HOST>(?:\s+port\s+\d+)?\s*$
```

Note the extensions from #1209 (common sshd, sshd-ddos, sshd-aggressive)
It may be possible to unite it all, using options for filter...

sebres:
<sub>moved from another not-related issue)</sub>
> @NoTeefy 

Fail2Ban stopped recognizing login attempts when I updated it to the 9.5.0 version.
My jails look like this:
```
[sshd]

enabled  = true
port    = ssh
filter   = sshd
action   = iptables[name=SSH, port=ssh, protocol=tcp]
           blocklist_de[email="xxx@xxx.xx", apikey="xxxx", service=%(filter)s]
logpath = %(sshd_log)s
backend = systemd
maxretry = 3
findtime = 86400

[sshd-ddos]
# This jail corresponds to the standard configuration in Fail2ban.
# The mail-whois action send a notification e-mail with a whois request
# in the body.
enabled  = true
port    = ssh
filter   = sshd-ddos
action   = iptables[name=SSH, port=ssh, protocol=tcp]
           blocklist_de[email="xxx@xxx.xx", apikey="xxxx", service=%(filter)s]
logpath = %(sshd_log)s
backend = systemd
maxretry = 3
findtime = 86400
```

And from the logging side I get logs like this:
```
2016-08-30 20:40:07,715 fail2ban.filtersystemd  [4203]: DEBUG   Read systemd journal entry: '2016-08-30T20:34:21.063236 hostname sshd[6467]: Disconnected from xx.xxx.xxx.xxx port 46349 [preauth]'
2016-08-30 20:40:07,715 fail2ban.filtersystemd  [4203]: DEBUG   Read systemd journal entry: '2016-08-30T20:34:21.085503 hostname sshd[6469]: Connection from xx.xxx.xxx.xxx port 53353 on xxx.xxx.xxx.xxx port 22'
2016-08-30 20:40:07,715 fail2ban.filtersystemd  [4203]: DEBUG   Read systemd journal entry: '2016-08-30T20:34:23.120815 hostname sshd[6469]: Invalid user operator from xx.xxx.xxx.xxx port 53353'
[...]

PR #1580, opened 2016-10-15
filter.d/sshd.conf: Match 'Invalid user' with 'port \d*'

aclindsa:
On a recent Arch Linux system (OpenSSH_7.3p1), I see lines like:
`Invalid user root from 1.1.1.1 port 37216`

The existing filter.d/sshd.conf failregex doesn't match these because of the ' port 37216' at the end of the line.


sebres:
Closes #1533, but... 

It would be a PR for both branches. So, master branch is more desirable target for it...

yarikoptic:
Needs a sample log line as well I guess among tests/files/logs - easy to add. Indeed better to go against master

aclindsa:
I added a line to tests/files/logs/sshd and re-submitted this as https://github.com/fail2ban/fail2ban/pull/1581 against master. Closing this PR in favor of that one.

PR #1606, opened 2016-11-10
Add Mongodb-auth filter and jail

alex1702:
Added Mongodb-auth filter and jail

Issue #1586, opened 2016-10-21
failregex with Multiline dont work


alex1702:
- Fail2Ban version (including any possible distribution suffixes):
  Fail2Ban 0.9.3-1
- OS, including release name/version: 
  Ubuntu Server 16.04.1 64Bit 
- [X] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)
### The issue:

_Summary here_
I have the Logs of MongoDB with auth activated. I have make a testfile.
#### Steps to reproduce

Filter file :

```
[Init]
maxlines = 3

[Definition]

failregex = ^.* <HOST>\n.*cde\n.*failed$
ignoreregex =
```

Log file:

```
[CEST Oct 20 12:37:14] blub
[CEST Oct 20 12:37:17] bla
[CEST Oct 20 12:39:14] abc 10.44.0.1
[CEST Oct 20 12:39:14] cde
[CEST Oct 20 12:39:14] failed
[CEST Oct 20 12:39:17] bla
[CEST Oct 20 12:40:14] abc 10.44.0.1 
[CEST Oct 20 12:40:14] cde
[CEST Oct 20 12:40:14] failed
```

And the result:

```
root@:~/tmp# fail2ban-regex ./test.log /root/tmp/testfilter.conf 

Running tests
=============

Use   failregex file : /root/tmp/testfilter.conf
Use         maxlines : 3
Use         log file : ./test.log
Use         encoding : UTF-8


Results
=======

Failregex: 1 total
|-  #) [# of hits] regular expression
|   1) [1] ^.* <HOST>\n.*cde\n.*failed$
`-

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [9] (?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?
`-

Lines: 9 lines, 0 ignored, 3 matched, 6 missed [processed in 0.00 sec] 
|- Missed line(s):
|  [CEST Oct 20 12:37:14] blub
|  [CEST Oct 20 12:37:17] bla
|  [CEST Oct 20 12:39:17] bla
|  [CEST Oct 20 12:40:14] abc 10.44.0.1
|  [CEST Oct 20 12:40:14] cde
|  [CEST Oct 20 12:40:14] failed
`-
```

why is this not working?

on the site https://regex101.com/ it works. see screenshot below:
[...]

PR #1607, opened 2016-11-10
Add Mongodb-auth filter and jail

alex1702:
Add Mongodb-auth filter and jail and test.


alex1702:
the test dont work. how must i correct that?

sebres:
For a little discussion about see #1586 or #1606

Issue #1593, opened 2016-10-31
A missing log file prevents fail2ban from starting on Ubuntu


AlexisWilke:
### Environment:

_Fill out and check (`[x]`) the boxes which apply. If your Fail2Ban version is outdated, 
and you can't verify that the issue persists in the recent release, better seek support 
from the distribution you obtained Fail2Ban from_

- Fail2Ban version (including any possible distribution suffixes): 0.9.3-1
- OS, including release name/version: Ubuntu 16.04.1 LTS
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)

### The issue:

fail2ban does not start.

#### Steps to reproduce

* install/setup postfix
* install fail2ban
* enable postfix jails
* try to restart fail2ban
* error says it cannot open `/var/log/mail.warn`

#### Expected behavior

* server starts and warns once about the missing log file

#### Observed behavior

Server refuses to start.

#### Any additional information

Many services do not create all of their log files when getting installed or started. Some systems create their log file only once they actually output the first set of logs, which could take a long time.

### Configuration, dump and another helpful excerpts

I add one file under /etc/fail2ban/jail.d/ with the following to enable postfix (postfix is installed prior to adding this to fail2ban):

    # Enable the various Postfix rules and add custom parameters
    #

    [postfix]
    enabled  = true
    action   = snap-firewall-action[scheme=all,period=year]

    [postfix-rbl]
    enabled  = true
    action   = snap-firewall-action[scheme=all,period=year]

    [postfix-sasl]
    enabled  = true
    action   = snap-firewall-action[scheme=all,period=year]

    # vim: syntax=dosini

#### Any customizations done to /etc/fail2ban/ configuration

No customizations.

#### Relevant parts of /var/log/fail2ban.log file:

This is the journal for fail2ban. I replaced the hostname with just `host`.

```
Oct 31 16:51:59 host systemd[1]: Reloading Fail2Ban Service.
Oct 31 16:51:59 host fail2ban-client[9476]: ERROR  No file(s) found for glob /var/log/mail.warn
Oct 31 16:51:59 host fail2ban-client[9476]: ERROR  Failed during configuration: Have not found any log file for postfix jail
Oct 31 16:51:59 host systemd[1]: fail2ban.service: Control process exited, code=exited status=255
[...]

Issue #865, opened 2014-11-26
Fail2Ban 0.9.0 on CentOS 7 error starting service


zzserg:
Hi, All!
Please help me solve strange problem...
Environment: CentOS Linux 7 (Core) x86_64, Fail2Ban v0.9.0 from repo EPEL 7, OwnCloud 7.0.3 from OwnCloud CentOS 7 repo.
I succesfull setup all environment and manually start fail2ban via command:
[code]
fail2ban-client start
2014-11-26 13:14:05,884 fail2ban.server.server[26908]: INFO    Starting Fail2ban v0.9.0
2014-11-26 13:14:05,885 fail2ban.server.server[26908]: INFO    Starting in daemon mode
[/code]
All working fine!

Than i suceccfully enabling fail2ban.service via:
[code]
systemctl enable fail2ban.service
ln -s '/usr/lib/systemd/system/fail2ban.service' '/etc/systemd/system/multi-user.target.wants/fail2ban.service'
[/code]

But when i try starting fail2ban.service i got error:
[code]
systemctl start fail2ban.service
Job for fail2ban.service failed. See 'systemctl status fail2ban.service' and 'journalctl -xn' for details.[/code]
Result
[code]
systemctl status fail2ban.service
fail2ban.service - Fail2ban Service
   Loaded: loaded (/usr/lib/systemd/system/fail2ban.service; enabled)
   Active: activating (start) since Ср 2014-11-26 13:16:25 MSK; 2s ago
 Main PID: 26844 (code=exited, status=255);         : 27043 (fail2ban-client)
   CGroup: /system.slice/fail2ban.service
           └─27043 /usr/bin/python /usr/bin/fail2ban-client -x start
[/code]

[code]
journalctl -xn
ноя 26 13:29:21 cloud.acconcept.ru systemd[1]: Starting Fail2ban Service...
-- Subject: Unit fail2ban.service has begun with start-up
-- Defined-By: systemd
## -- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel

-- Unit fail2ban.service has begun starting up.
ноя 26 13:29:23 cloud.acconcept.ru fail2ban-client[27566]: ERROR  No file(s) foun
d for glob /var/www/html/owncloud/data/owncloud.log
ноя 26 13:29:23 cloud.acconcept.ru fail2ban-client[27566]: ERROR  Failed during c
onfiguration: Have not found any log file for owncloud jail
[...]

PR #1619, opened 2016-11-22
 0.10/skip wrong confugured jails

sebres:
- really skips invalid jails (because of theirs wrong configuration) - server starts nevertheless, as long as one jail was successful configured;
- message about wrong jail configuration logged in client log (stdout, systemd journal etc.) and in server log as error
- introduced new option `-t` or `--test` to test configuration resp. start server only if configuration is clean (fails by wrong configured jails if option `-t` specified);
- test cases extended: cover skipping invalid jail

### DIFF
--- a/fail2ban/client/configurator.py
+++ b/fail2ban/client/configurator.py
@@ -72,9 +72,9 @@ class Configurator:
 	def getEarlyOptions(self):
 		return self.__fail2ban.getEarlyOptions()
 
-	def getOptions(self, jail=None, updateMainOpt=None):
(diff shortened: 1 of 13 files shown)


sebres:
This fix makes possible to start server despite possible error in configuration of single jail.
IMHO, completelly prevent starting of the server in case of something wrong within single jail is not really good practice at all.

So by such issues as #1593, #865 and several others, broken jail does not prevent fail2ban from starting...

sebres:
Changes without re-indent modifications - https://github.com/fail2ban/fail2ban/pull/1619/files?w=1
(To see the changes with whitespace ignored, add `?w=1` in comparison-URL)

PR #1622, opened 2016-11-25
 0.10/configreader and more

### LINKED ISSUE #727: action_mwl and multi logs breaks start up
Hi,

the patch from issue #231 never made it to a release, did it? AFAIK 0.8.13 still comes with the unpatched file.

jail.local

```
[apache]
enabled = true
port    = http,https
filter  = apache-auth
logpath = /var/log/apache2/error.log
          /var/customers/logs/*error.log
maxretry = 3
action  = %(action_mwl)s
```

Leads to. 

```
[....] Restarting authentication failure monitor: fail2banERROR  Error in action definition sendmail-whois-lines[name=apache, dest="<address>", logpath=<path>
ERROR  Errors in jail 'apache'. Skipping...
```

Using action action_mw it works. 
I tried to patch the file manually in my installation but it does not seem to work also. Is there something more I have to do next to changing the line and compile the file?

### LINKED ISSUE #976: Multiple Logpaths prevent starting when action_mwl is used
When using the setting:

```
    action = %(action_mwl)s
```

any Jail rule defined with multiple logpaths (multiple not wildcards):

```
logpath = /some/path/file.log
           /a/completely/different/path/file.log
```

prevents Fail2ban starting (sample output from /usr/bin/fail2ban-client start)

```
ERROR  Error in action definition sendmail-whois-lines[name=test-jail, dest="root@localhost", logpath=/some/path/file.log
ERROR  Errors in jail 'test-jail'. Skipping...
```

This only occurs using the l option for the action to add the log lines to the email.

Affects versions:

```
0.8.6-3wheezy3 (Debian Wheezy)
0.8.13-1 (Debian Jessie)
0.9.1-1 (Debian Sid)
```

not tested others but seems like it affects all versions.

### LINKED ISSUE #1110: Expose filter regex group captures in actions
I think that if regular expressions are searched and we have already parsed all the data and captured groups, why aren't those available in the action commands.  It would be really useful so that one can do very complex actions and analytics.

sebres:
This prepares for future cherry-pick from my own branch to resolve #1110

* filter/action (and its includes): substitution `%(param)s` may be used now (instead of `<param>`) for init-values specified in jail-configs via `action[param1="...", param2=...]`;
* substitution `<param>` should be used for dynamic interpolation only (todo: review configurations to replace all static parameters with `%(param)s`, if done - `filter::getCombined` can be removed);
* allow newline in extra init-parameters of action/filter (or interpolation of it), e. g. `action[..., logpath="%(logpath)s"]`
* fixed several actions, that could not work with jails using multiple logpath; additionally repaired execution in default shell (bad substitution by `${x//...}` executing in `/bin/sh`);
* added helper "action.d/helpers-common.conf", and `_grep_logs` part-command for actions needed grep logs from multiple log-files
* test cases: executing of some complex actions covered

BTW:
Closes #976
Closes #727

I don't feel like to fix it in 0.9th version at all and don't known whether the changes in action-configs are compatible with 0.9th branch... (at least changes of 65abc639cc7838beaf1f587eae74e98afc4a1473 can be needed to proper usage, but it is certainly different in 0.9).
If someone wants test it, welcome!

### DIFF
--- a/config/action.d/complain.conf
+++ b/config/action.d/complain.conf
@@ -28,6 +28,10 @@
 #
 
 
+[INCLUDES]
+
+before = helpers-common.conf
+
 [Definition]
 
 # Option:  actionstart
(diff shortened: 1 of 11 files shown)


sebres:
see several todo's in #1625

Issue #1626, opened 2016-11-27
modsecurity - remove double-space


phaleas:
### Environment:

- Fail2Ban version 0.9.3-1 all
- Ubuntu 16.04LTS, xenial:
- modsecurity 2.9.0-1
 
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)

### The issue:

Double space in failregex prevents matches:

```
failregex = ^%(_apache_error_client)s ModSecurity:  (\[.*?\] )*Access denied with code [45]\d\d.*$
```

should be:

```
failregex = ^%(_apache_error_client)s ModSecurity: (\[.*?\] )*Access denied with code [45]\d\d.*$
```

In case Git editor removes double-space: It should be ModSecurity:<1 space>(\

PR #1627, opened 2016-11-28
Fix gh-1626: one space after ModSecurity

sebres:
`filter.d/apache-modsecurity.conf`
  - fixed for newer version (one space, closes gh-1626)
  - reviewed and optimized:
    - non-greedy catch-all replaced for safer match
    - unneeded catch-all anchoring removed
    - non-capturing groups


sebres:
easy fix, so merging

PR #1630, opened 2016-12-02
Ensure filtersystemd closes the journal when stopped

AnchorCat:
As of fail2ban 0.9.5, `fail2ban-client reload` or `fail2ban-client stop jail` does not properly close the `journal.Reader` object in `filtersystemd`, relying instead on Python's garbage collection to do that... eventually. Unfortunately, Python may not get around to GCing the object for a long time, and in the meantime all journal files will continue to be held open by the object, even long after they've been rotated and unlinked off the system. In some of my tests, I had journal files held open for several days after they'd been unlinked. The problem is exacerbated since new files are opened each time fail2ban is reloaded.

This patch series introduces a new `cleanup` hook for filters to perform various cleanups when a jail is deleted (which is not necessarily when the jail is stopped, since it looks like it's possible to add and then stop a jail, without it ever being started). `filtersystemd` is updated to use this hook, as are `filtergamin` and `filterpyinotify` which require similar cleanups.

### DIFF
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -290,6 +290,13 @@ class Filter(JailThread):
 		raise Exception("run() is abstract")
 
 	##
+	# Clean up resources allocated by the filter, if necessary.
+	#
+
+	def cleanup(self):
+		pass
+
+	##
 	# Set external command, for ignoredips
 	#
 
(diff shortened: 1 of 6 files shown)


AnchorCat:
Gah, I'll rebase this against the current 0.9 branch.

PR #1637, opened 2016-12-09
DO NOT MERGE FOR NOW:  0.9.6 Release

yarikoptic:

Debian stretch is freezing up so we better kick out a release to get recent stable into upcoming debian stable.  I am already running builds and will perform some real-life  testing before uploading. But this PR to verify that nothing indeed got broken as travis concerned etc.  I might need to push few more tune ups


yarikoptic:
ok -- let's see how it behaves in the wild now ;)

PR #1640, opened 2016-12-15
systemd service: do not set RestartPreventExitStatus=0

marcosfrm:
Minor cleanup.

### DIFF
--- a/files/fail2ban.service
+++ b/files/fail2ban.service
@@ -14,7 +14,7 @@ ExecStop=/usr/bin/fail2ban-client stop
 ExecReload=/usr/bin/fail2ban-client reload
 PIDFile=/var/run/fail2ban/fail2ban.pid
 Restart=on-failure
-RestartPreventExitStatus=0 255
+RestartPreventExitStatus=255
 
 [Install]
 WantedBy=multi-user.target


sebres:
Nope, I've made it explicitly. See #1618

Reason:
- stop the service via fail2ban command line `fail2ban-server stop` or `fail2ban-client stop` or still worse `fail2ban-client reload ...`, that internally decides to restart itself (and supply special additionally arguments), should **definitely prevent triggering of the systemd to restart fail2ban**.
- the systemd should try to restart fail2ban only after possible crash or something like miss-kill (where the exit code is neither 0 nor 255).

Thus I'll close the PR as long as no further arguments for this.

But I think, if you've a case, where it may be not safe enough (fail2ban exits with 0), the manipulating of return code in this case inside fail2ban-core would be more desirable instead of.

sebres:
Please not that this systemd-file is only for 0.10-version (and you made a PR for it).

Fail2ban 0.9th have totally different command-line handling, very different client-server facilities and several other things (e. g. it can not be started in "simple" mode, because of forking).

marcosfrm:
My patch does not change current behaivour. `RestartPreventExitStatus=0` is redundant with `Restart=on-failure`. See `systemd.service(5)`. systemd will not restart it when the exit code is 0. At the end, this PR is cosmetic.

sebres:
You may have right, but:
- on one of my debians it does not ignore 0 exit code;
- what's about if someone will tomorrow change `Restart=on-failure`?
- I'll appreciate yet to have it explicit (one byte more, but for this more understandable :-)

marcosfrm:
- I did not test it on Debian's 215 version (Jessie). It should work at least from CentOS' 219. Fair enough, maybe it is buggy on older versions.
- With my patch, if people change to Restart=always exit code 0 will be ignored and the daemon will be always restarted. If folks want this and change it locally, I supose they know what are doing.

I will test it on Debian Jessie.

Issue #1642, opened 2016-12-19
Default message patterns to protect web apps


benrubson:
Hello,

Would be nice if we could find some default message patterns which could be used by web applications to trigger Fail2Ban, without having to deploy new rules.
These patterns would have to be integrated to existing Apache / Nginx / Lighttpd filters (the auth ones).

I propose the following patterns, which are not far from the already existing ones in Apache & Nginx :
```
user "bob" not found: My Protected Web Application
user "bob" password mismatch: My Protected Web Application
client used invalid token: My Protected Web Application
```

Edit : see [this message](#issuecomment-269220121) below for an update.

Thank you 👍 

Ben

benrubson:
Solved by #1643 👍

szepeviktor:
How would your app tell e.g. Apache to log these messages?

benrubson:
As an example, if it's a PHP app, you can use [error_log()](http://php.net/manual/en/function.error-log.php).

szepeviktor:
That looks like this
```apache
[Fri Dec 23 14:58:26.955220 2016] [proxy_fcgi:error] [pid 8494:tid 139854318851840] [client 54.240.157.87:18275] AH01071: Got error 'PHP message: 404 error: URL not found: https://www.chickenbanana.com/assets/css/lightbox/close.png\n'
```

szepeviktor:
`URL not found: https://www.chickenbanana.com/assets/css/lightbox/close.png\n`
is the output of error_log()

szepeviktor:
I don't see `Got error 'PHP message:` anywhere in your PR.

benrubson:
I did not mention `PHP` because the app could also use Perl etc...
These are intended to be generic messages every app (written in PHP, in Perl... in every langage) could adopt / use as their default auth failure messages.

In PHP, if you do :
`error_log("user \"$username\" not found: MyApp");`
You'll find the corresponding entry in the Apache error log file :
`[Mon Dec 19 19:00:31 2016] [error] [client 1.2.3.4] user "bob" not found: MyApp`

It will then be caught by the generic new rules of this PR 👍

szepeviktor:
> You'll find the corresponding entry in the Apache error log file :

Could you tell me what SAPI are you using? I think it is mod_php.

Please think about production installs where MPM event + PHP-FPM is in use an has this part prepended: `Got error 'PHP message:`

szepeviktor:
I do not know whether Python and Rudy prepend something...

[...]

PR #1643, opened 2016-12-19
Default message patterns to protect web apps

benrubson:
Hello,

This PR solves #1642.

Thank you 👍 

Ben

### DIFF
--- a/config/filter.d/apache-auth.conf
+++ b/config/filter.d/apache-auth.conf
@@ -11,11 +11,11 @@ before = apache-common.conf
 
 
 failregex = ^%(_apache_error_client)s (AH(01797|01630): )?client denied by server configuration: (uri )?\S*(, referer: \S+)?\s*$
-            ^%(_apache_error_client)s (AH01617: )?user .*? authentication failure for "\S*": Password Mismatch(, referer: \S+)?$
-            ^%(_apache_error_client)s (AH01618: )?user .*? not found
+            ^%(_apache_error_client)s (AH01617: )?user \S* authentication failure for "\S*": Password Mismatch(, referer: \S+)?$
+            ^%(_apache_error_client)s (AH01618: )?user \S* not found
             ^%(_apache_error_client)s (AH01614: )?client used (wrong authentication scheme|invalid token)
             ^%(_apache_error_client)s (AH\d+: )?Authorization of user \S+ to access \S* failed, reason: .*$
-            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user .*?:? password mismatch
+            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user \S*:? password mismatch
             ^%(_apache_error_client)s (AH0179[01]: |Digest: )user `.*?' in realm `.+' (not found|denied by provider): \S*(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01631: )?user .*?: authorization failure for "\S*":(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01775: )?(Digest: )?invalid nonce .* received - length is not \S+(, referer: \S+)?\s*$
--- a/config/filter.d/apache-auth.conf
+++ b/config/filter.d/apache-auth.conf
@@ -12,10 +12,10 @@ before = apache-common.conf
 
 failregex = ^%(_apache_error_client)s (AH(01797|01630): )?client denied by server configuration: (uri )?\S*(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01617: )?user .*? authentication failure for "\S*": Password Mismatch(, referer: \S+)?$
-            ^%(_apache_error_client)s (AH01618: )?user .*? not found(: )?.*(, referer: \S+)?\s*$
-            ^%(_apache_error_client)s (AH01614: )?client used (wrong authentication scheme|invalid token): .*(, referer: \S+)?\s*$
+            ^%(_apache_error_client)s (AH01618: )?user \S+ not found
+            ^%(_apache_error_client)s (AH01614: )?client used (wrong authentication scheme|invalid token)
             ^%(_apache_error_client)s (AH\d+: )?Authorization of user \S+ to access \S* failed, reason: .*$
-            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user .*?:? password mismatch: .*(, referer: \S+)?\s*$
+            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user \S+:? password mismatch
             ^%(_apache_error_client)s (AH0179[01]: |Digest: )user `.*?' in realm `.+' (not found|denied by provider): \S*(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01631: )?user .*?: authorization failure for "\S*":(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01775: )?(Digest: )?invalid nonce .* received - length is not \S+(, referer: \S+)?\s*$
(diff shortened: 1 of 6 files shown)


config/filter.d/apache-auth.conf
```diff
@@ -12,10 +12,10 @@ before = apache-common.conf
 
 failregex = ^%(_apache_error_client)s (AH(01797|01630): )?client denied by server configuration: (uri )?\S*(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01617: )?user .*? authentication failure for "\S*": Password Mismatch(, referer: \S+)?$
-            ^%(_apache_error_client)s (AH01618: )?user .*? not found(: )?\S*(, referer: \S+)?\s*$
-            ^%(_apache_error_client)s (AH01614: )?client used wrong authentication scheme: \S*(, referer: \S+)?\s*$
+            ^%(_apache_error_client)s (AH01618: )?user .*? not found(: )?.*(, referer: \S+)?\s*$
```
sebres: A greedy match (`.*`) is very evil here and below, because it could be manipulated with referer, so allows injecting on referer or other data before...

config/filter.d/apache-auth.conf
```diff
@@ -12,10 +12,10 @@ before = apache-common.conf
...
+            ^%(_apache_error_client)s (AH01614: )?client used (wrong authentication scheme|invalid token): .*(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH\d+: )?Authorization of user \S+ to access \S* failed, reason: .*$
-            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user .*?: password mismatch: \S*(, referer: \S+)?\s*$
+            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user .*?:? password mismatch: .*(, referer: \S+)?\s*$
```
sebres: same as above

config/filter.d/apache-auth.conf
benrubson: So we should only allow `\S*` here ?
We won't then be able to use strings with spaces (for example `My Protected Web Application`).

sebres: Did you have a log excerpt for such failure?

benrubson: No, as explained in #1642, this is intended for web applications which want to be protected using default messages.
For example :
`user "bob" not found: My Web App`
This message would come from the web application `My Web App`.
We could also have :
`user "bob" not found: WebPortal2 just protected you`
Or anything else matching the rule.

benrubson: Web applications would use for example log_error() in PHP to generate such messages.
For example :
`log_error('user "bob" not found: My Web App');`

benrubson: We have the case here for example :
https://github.com/fkooman/php-remote-storage/issues/54

benrubson: > A greedy match (...) could be manipulated with referer, so allows injecting on referer or other data before...

As we have `referer: \S+` in the rule, I'm not sure injection is possible, as referer won't be able to use spaces. Don't you think so ?

benrubson:
Would be nice if it could also be backported to 0.9.
I had a look, filter files are identical.
Many thanks 👍

config/filter.d/apache-auth.conf
benrubson: Also this one :
```
[Mon Dec 19 12:07:30 2016] [error] [client 1.2.3.4] client denied by server configuration: /var/www/space dir
```
What we can see here is that due to `\S`, many rules from `apache-auth.conf` will not work as expected.

benrubson: As the referer is not what we are trying to catch, perhaps we could remove all the referer parts at the end of the rules, finishing them with a simple `.*`.
What we want to catch and which is important is at the beginning of each rule, so it would work in any case and would be easier :)

benrubson: Understood, thank you Serg for your clarification !

One last question before modifying the rules, do you think we should use `\S*` to also catch empty user given to the web application ?

sebres: > do you think we should use `\S*` to also catch empty user ...

I don't think so, because that depends. For example if by the handshake, before the client send authentication data, the web-server (or application) will produce 401 (and failure in log in sense of fail2ban), it is also counted. So the client should make `maxretry` of such "failures" - it goes banned.

This question is thus not really answerable as long as too few info... But rather - no, because makes no sence.

benrubson:
Should be OK now @sebres :)
Thank you for your help 👍

config/filter.d/apache-auth.conf
```diff
@@ -12,10 +12,10 @@ before = apache-common.conf
 
 failregex = ^%(_apache_error_client)s (AH(01797|01630): )?client denied by server configuration: (uri )?\S*(, referer: \S+)?\s*$
             ^%(_apache_error_client)s (AH01617: )?user .*? authentication failure for "\S*": Password Mismatch(, referer: \S+)?$
-            ^%(_apache_error_client)s (AH01618: )?user .*? not found(: )?\S*(, referer: \S+)?\s*$
-            ^%(_apache_error_client)s (AH01614: )?client used wrong authentication scheme: \S*(, referer: \S+)?\s*$
+            ^%(_apache_error_client)s (AH01618: )?user .*? not found
```
sebres: Nope, it is vulnerable because of `.*?` ...

config/filter.d/apache-auth.conf
```diff
@@ -12,10 +12,10 @@ before = apache-common.conf
...
+            ^%(_apache_error_client)s (AH01614: )?client used (wrong authentication scheme|invalid token)
             ^%(_apache_error_client)s (AH\d+: )?Authorization of user \S+ to access \S* failed, reason: .*$
-            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user .*?: password mismatch: \S*(, referer: \S+)?\s*$
+            ^%(_apache_error_client)s (AH0179[24]: )?(Digest: )?user .*?:? password mismatch
```
sebres: same here

benrubson:
Yes sorry @sebres for my mistake...
Corrected 👍

benrubson:
Is everything OK for you @sebres ?
Many thanks !

sebres:
Much too much to do, but...

config/filter.d/nginx-http-auth.conf
```diff
@@ -4,7 +4,8 @@
 [Definition]
 
 
-failregex = ^ \[error\] \d+#\d+: \*\d+ user "\S+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
+failregex = ^ \[error\] \d+#\d+: \*\d+ user \S*:? (password mismatch(: .*)?|(was )?not found(:| in) .*), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
```
sebres: I'm not sure that introducing of such changes is good at all. I mean for example this greedy catch-all here...
It is although not vulnerable, but will be slower thus.
If that would be optional...

I mean someone that need this (catch self generated message) can for example very easy add such regex within jail:
```
[nginx-http-auth]
            ^ \[error\] \d+#\d+: \*\d+ user \S*:? (password mismatch(: .*)? ...
```

@fail2ban/contributors, @yarikoptic Any thoughts about this PR?

benrubson:
> Much too much to do, but...

Thank you for your time @sebres 👍

config/filter.d/nginx-http-auth.conf
benrubson: > this new greedy catch-all

I found a way to remove it :)

> I mean someone that need this (catch self generated message) can for example very easy add such regex

Goal is to have them by default so that application developers adopt them / can use them as their default auth failure messages. Let's say a best-practice :)

benrubson:
See #1646 for an alternative solution ?

Issue #1644, opened 2016-12-23
Apache : log lines with URI containing spaces are not caught


benrubson:
Hello,

Apache URIs can contain spaces.
In this case, log lines are not caught by Fail2Ban.

I'll submit a PR to correct this 👍 

Ben

szepeviktor:
Could you provide some error.log lines?

benrubson:
Some examples showing that URIs can contain spaces :
```
[Sun Jan 01 21:41:10 2017] [error] [client 192.168.0.10] user bob not found: /backup/My Documents/
[Sun Jan 01 21:41:17 2017] [error] [client 192.168.0.10] user ben: authentication failure for "/backup/My Documents/": Password Mismatch
[Sun Jan 01 21:41:35 2017] [error] [client 192.168.0.10] client denied by server configuration: /var/www/archive/My Documents/
```

szepeviktor:
Thanks.
It seem to be Apache 2.2.x

benrubson:
Yep, but same with Apache 2.4 :
```
[Mon Jan 02 16:04:33.775458 2017] [authz_core:error] [pid 91974] [client 192168.0.10:62256] AH01630: client denied by server configuration: /var/www/data/My Documents
```

sebres:
Fixed in #1645

PR #1645, opened 2016-12-23
Apache URIs can contain spaces

benrubson:
Hello,

This PR solves #1644.

_Apache URIs can contain spaces.
In this case, log lines are not caught by Fail2Ban._

Ben


sebres:
Although I like it, but because there are some paranoid of us... :)

@yarikoptic Your decision?

Secondly, I'll appreciate that `user \S*` will be replaced with `\S+`. 
We talked already about it:
> For example if by the handshake, before the client send authentication data, the web-server (or application) will produce 401 (and failure in log in sense of fail2ban), it is also counted. So the client should make `maxretry` of such "failures" - it goes banned.
> This question is thus not really answerable as long as too few info... But rather - no, because makes no sense.

benrubson:
Thank you for your feedback @sebres.
Regarding `\S+` instead of `S*` for user, there are some Apache tests inside `fail2ban/tests/files/logs/apache-auth` that look for an empty user.
This is why I kept the `\S*` version :)

benrubson:
@yarikoptic may we have your feedback regarding this PR please ?

And what about the 3 remaining untouched rules, any idea to shorten them ?

Many thx 👍

sebres:
I've ported it to newest 0.10 (with `prefregex`), with little review and clean-up by the way.
Will merge if test cases passed.
Thanks!

benrubson:
Thank you very much @sebres 👍 
There are still 2 rules containing `(, referer: \S+)?`, you did not manage to get ride of this ?
Thx !

sebres:
Done, see 6dfd080

benrubson:
Perfect thx 👍

sebres:
small amend in 507034c5be77c7894e2eee8afc2ce4ea84002674: joined some similar expressions

PR #1646, opened 2016-12-26
Default message patterns to protect web apps

benrubson:
Hello,

This PR solves #1642, using the following rule :
`user \S* ?(not found|not authorized|password mismatch|sent invalid token|misbehaved|tried to break-in)\b`

Thank you !

Ben


benrubson:
@sebres, do you prefer this solution to #1643 ?
_Should we then close #1643 ?_
Many thanks 👍

sebres:
I can not decide about both, as I already told you my opinion:
> Someone that need this (catch self generated message) can for example very easy add such regex within jail with extending of failregex (`failregex = %(known/failregex)s NL ...`).

For example you've a new expresion with `PHP message` in `apache-auth` filter. This enhancements will burden all the people using apache, but neither use such "default auth failure messages" nor have PHP at all.
I meant already if this would be an option, it would be OK.

I understand your intention:
> Goal is to have them by default so that application developers adopt them / can use them as their default auth failure messages

But IMHO it is currently not a goal but the overload.

ATM no time to go deeper...

benrubson:
Thank you for you feedback @sebres.
_(Just for info the `PHP message` was added after @szepeviktor comment in #1642)_

sebres:
> Just for info the PHP message was added after @szepeviktor comment

It's not about PHP... It is just an additional failregex rule, that will never concern many people (but costs time by checking of each entry in log).

JFYI, if we'll accept each new rule in the wild, all our filters will be in the meantime 200 rules and more long...

So again, I'm not against the enhancement, the reason is just to follow the primary goal - to hold fail2ban fast as possible in every respect.

benrubson:
> JFYI, if we'll accept each new rule in the wild, all our filters will be in the meantime 200 rules and more long...

I think the one-line rule I propose is rather justified :)
Here is an example : https://github.com/fkooman/php-remote-storage/issues/54
Access failure messages are ready to be implemented, but last question is :
_Which messages should we use ? Are there any best practices ? How the others do ?_
So would be nice to be able to say :
_Yes see here, there are some officially defined patterns which will be caught and banned !_

> I meant already if this would be an option, it would be OK.

How to make this rule an option ?
Simply making it a comment (`# ...`) in the configuration file ?
Then we would be able to enable / disable it on our own, why not...

benrubson:
> I meant already if this would be an option, it would be OK.

Technically Serg, how to make this rule an option ?
Simply making it a comment (`# ...`) in the configuration file ?
Then we would be able to enable / disable it on our own, why not...

sebres:
> Simply making it a comment (# ...) in the configuration file ?

As possible example maybe... but it intends to change the stock configuration files. And normally all the changes should be done in .local files only (that are not a copy, so contains customizations only).
> how to make this rule an option ?

For example take a look to #1209 how I made it for sshd-aggressive (https://github.com/fail2ban/fail2ban/pull/1209/files#diff-56cef6a5cfe6e39eceea88d750652307R33).

Alternative as simple solution just to set it as new parameter, like `protect_web_apps`.
```
# apache-auth.conf 
...
extendregex =

failregex = ^... our default regex 1 ...$
            ^... our default regex 2 ...$
            %(extendregex)s

protect_web_apps = ^... your new optional regex ...$
```
So the users have 2 possibilities to use it: either to define `extendregex` in `jail.local` or to overwrite `failregex` in `apache-auth.local`:
```
# jail.local (variant 1):
[my-jail]
filter = apache-auth[extendregex="<protect_web_apps>"]
```
```
# apache-auth.local (variant 2):
[Definition]
            %(protect_web_apps)s
```
```
# apache-auth.local (variant 3):
[Definition]
extendregex = %(protect_web_apps)s
```

config/filter.d/lighttpd-auth.conf
```diff
@@ -4,7 +4,13 @@
 [Definition]
 
 failregex = ^: \(http_auth\.c\.\d+\) (password doesn\'t match .* username: .*|digest: auth failed for .*: wrong password|get_password failed), IP: <HOST>\s*$
+            ^: \(\S+\.c\.\d+\) user \S* ?(not found|not authorized|password mismatch|sent invalid token|misbehaved): .+, IP: <HOST>
```
yarikoptic: Does `.+` Here by any chance captures user/client provided data?

sebres: Yep, good point!

benrubson: `.+` is intended here to capture the additional web-app failure message.
I don't think it will capture any user/client provided data, as this data will not come into the log.
Were you thinking about something specific ?
Thank you ver much @yarikoptic for your feeback 👍

benrubson:
>Alternative as simple solution just to set it as new parameter, like `protect_web_apps`.

Thank you very much @sebres for your explanation.
Do you then definitely prefer this PR as being optional ?
If so, I'll then submit a new commit according to your guideline.
Many thanks 👍

sebres:
As already said I don't like adding the rules to our default filters, that don't affect all the people to the same extent. I. e. the extention looks as option, so it should be done optional.
If someone really need it, he can extend his configs.

config/filter.d/lighttpd-auth.conf
yarikoptic: If there is anything from a client in that message (eg login) and there is no trailing anchor - opens possibility for the injection command

benrubson: We should then advise web-app developers not to put (unverified) data from clients into their failure message, which should simply contain for example the name of their application.
We could use `\S+` but I think it is too restrictive.

benrubson:
Is there any way to ask `failJSON` to enable `protect_web_apps` parameter to test log lines ?
As it will not be enabled by default, but would be nice to keep test log lines...

sebres:
> Is there any way to ask failJSON to enable protect_web_apps parameter to test log lines ?

Unfortunately not via failJSON etc. (would be indeed an idea to do that).
For example it is realized in above mentioned PR within sshd-aggressive (as include), but it expected changes in code-base (in same PR), no way to do it currently.

benrubson:
OK, should be OK according to your previous feedbacks, of course feel free to comment !
Thank you again 👍

benrubson:
I made a test with fail2ban 0.9.6-1 (debian).
Using `filter  = apache-auth[extendregex="<protect_web_apps>"]` does not seem to work, fail2ban starts but eligible lines are not caught.
If I set `extendregex = %(protect_web_apps)s` in the filter file, it works.
Any clue why the first (nicer) solution does not work ?
Many thanks 👍

sebres:
> Using `filter  = apache-auth[extendregex="<protect_web_apps>"]` does not seem to work

v.0.10 only, it is not possible in v.0.9, because interpolation of parameters `<protect_web_apps>` is pure runtime interpolation (but in case of failregex, the params are needed to generation of failregex even on starting of filter).
In v.0.10 all already known parameters will be interpolated implicit before it started.

You can even try another interpolation `filter  = apache-auth[extendregex="%(protect_web_apps)s"]`. 
But I believe it can not work in jail (at least in v.0.9), because `protect_web_apps` is a filter option and just unknown in jail. In this case you'll get an interpolation error.

benrubson:
Yep I already tested this last option (and others) and it does not work.
OK, if we backport to 0.9, I will slightly update to use `extendregex = %(protect_web_apps)s` in the filter file.

Any more work to commit this PR ?
To backport to 0.9, best solution is to open a new PR to 0.9 branch ?

Thanks 👍

config/filter.d/nginx-http-auth.conf
```diff
@@ -3,15 +3,23 @@
...
 [Definition]
 
+extendregex = 
 
 failregex = ^ \[error\] \d+#\d+: \*\d+ user "\S+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
+            %(extendregex)s
 
 ignoreregex = 
 
+protect_web_apps = ^ \[error\] \d+#\d+: \*\d+ user \S* ?(not found|not authorized|password mismatch|sent invalid token|misbehaved): .+, client: <HOST>
```
yarikoptic: could `HOST` be anchored at the end here due to `.+` preceding it?

benrubson: Certainly the same issue as the one you noticed with lighttpd.
Do you want me to append the following to this new rule, as in line 8 ?
`, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$`

config/jail.conf
```diff
@@ -256,6 +256,9 @@ logpath  = %(auditd_log)s
 
 [apache-auth]
 
+# To also protect web apps with default message patterns,
+# use the following filter :
+#filter  = apache-auth[extendregex="<protect_web_apps>"]
```
yarikoptic: I am still wondering @sebres how big of an impact would be just to have those regexes enabled? 
Are we spending most of the time regex matching or doing other logistics?
Making it optional as in here makes those regexes not tested anyhow so potentially making them not operating correctly or even harmful.

config/filter.d/nginx-http-auth.conf
sebres: > could HOST be anchored at the end here due to .+ preceding it?

No, still worse (because configurable in nginx-config and per default contains referer etc. so user input).
Imho, it is better to resolve greedy `.+` with something plausible

benrubson: Just anchored in my last commit.

config/jail.conf
sebres: > impact would be just to have those regexes enabled?

I don't like to enable regex's that not affected most of the people (that also theoretically may be vulnerable)

> Making it optional as in here makes those regexes not tested anyhow so potentially making them not operating correctly or even harmful.

But only as long as it is not covered (see how I made it with sshd-aggressive in #1209, what also merged now, so we can add zzz-protect-web-apps test-case enabled this)

benrubson: >I don't like to enable regex's that not affected most of the people

They could, if most of the web-apps adopt them :)

sebres:
> They could, if most of the web-apps adopt them :)

You don't understand me... What I'm trying to say you all the time - it is no proper way to do that.
For example, how IMHO the proper way may look in nginx:
- create a separate location `web-app-unautorized`:
```
log_format  unautorized_fmt  '[$time_local] $remote_addr - auth-failure for "$remote_user" $status "$request" '
    '$body_bytes_sent "$http_referer" '
    '"$http_user_agent" "$http_x_forwarded_for"';

location @web-app-unautorized {
   internal;
   access_log ...path.../web-app-unautorized.log unautorized_fmt;
}

```
- then internal redirect to location `@web-app-unautorized`, or rewrite there from 401-nginx-location, or using something like `post_action @web-app-unautorized` within and much much more;
- then create a jail with exact one regex: 
```
[nginx-auth-failures]
filter =
logpath = ...path.../web-app-unautorized.log
failregex = ^[] <HOST> - auth-failure
```
And it is only one scenario, and I can imagine still 20 other, where all 20 would be several times better as yours.
To make it one-line configurable, the most of the web-apps may provide an include for nginx-config, where all that already done and another include for fail2ban to ban it.

What you're trying to do is IMHO wrong (at least not performant and possibly vulnerable) and this will navigate the people using this (IMHO wrong) way per default. I don't want it, but if someone really need zero-way config, he can do that using your way (but not per default).

Just to close this discussion - this is the main reason why it should be optional.

benrubson:
Yep @sebres, I thought that it was the fact to make it "per-default" (not optional) that was not the proper way.
But I did not understand that according to you, the whole method I propose it not the proper way.

sebres:
> the whole method I propose it not the proper way.

Because in my architectural understanding, if the web-application will say, that authentication failure occurred, it should:
- either use the default unauthorize-logic of the web-server (nginx/apache/lighty, etc) - so failure will be catched with default fail2ban rules, which the end-user specified for his web-server.
- or write it in separate log-file with own format (then the end-user will made own fail2ban-rule for this).

Mix of both is IMHO the wrong way, and I don't want to popularize it per default

yarikoptic:
If I got it right from my scarce following of the tread, that introduced here "logging approach" is neither standard (i.e. there is a number of ways how logging from web apps could be achieved) nor possibly recommended, I would slide with @sebres on not including it into stock filters/jails. Not even to a degree of making it just optional since it just adds clutter to the stock config/specification, leaves patterns not tested etc... how could we resolve it? not sure really.  If there were no combinatorial explosion (separate filter/jail per each web service) with e.g. a custom log (as @sebres suggested) -- it would be imho better so only a single clean jail+filter could be created/enabled etc as with any other service.

benrubson:
So, 2 voices 👎 1 voice 👍 
^^
What should we do then ? :)
Feel free to close this as "won't resolve" if you feel that's the best thing to do, I'll understand :)
Anyway, thank you very much @sebres & @yarikoptic for your feedback regarding this 👍

sebres:
I've merged it (squashed rebase relative current 0.10, so conflicts resolved).
- small review and adjustment (filter parameter called `extension` now, and syntax changed, thus allow to adjust not the `failregex`, but also `prefregex` etc.):
```diff
[jail]
-filter = nginx-http-auth[extendregex="<protect_web_apps>"]
+filter = nginx-http-auth[extension=protect-web-apps]
+# or fewer flexible:
+#filter = nginx-http-auth[extfailregex="<protect-web-apps>"]
```
- extension `protect-web-apps` test covered now for all 3 filters.

I'm still not sure it should be really part of default fail2ban installation, but just to have ... resp. to make it compatible with newest version + the test coverage.

config/jail.conf
```diff
@@ -369,8 +373,9 @@ logpath = %(suhosin_log)s
 
 
 [lighttpd-auth]
-# Same as above for Apache's mod_auth
-# It catches wrong authentifications
+# Same as above for Apache's mod_auth. It catches wrong authentifications
```
benrubson: Why this line @sebres ?

benrubson:
Thank you very much for your rework @sebres 👍 
Looks nice 👍

config/jail.conf
sebres: Just unwrapped, but indeed we can remove it.

sebres: thus removed now

sebres:
@yarikoptic your conclusion now?
As already said, I'm not sure also, but I see also advantages of this. E. g. it can be used as example "how to make an own extension for default filter's, without making it totally incompatible"...
Of course we've another already available enhancement (known/param), that can be used also to specify own extension of `failregex`, but this thing is more flexible, IMHO.

config/jail.conf
benrubson: OK, thank you Serg, looks perfect 👍

szepeviktor:
In my [WordPress WAF](https://github.com/szepeviktor/wordpress-fail2ban) written in PHP I use

```php
private $prefix = 'Malicious traffic detected: ';
private $prefix_instant = 'Break-in attempt detected: ';
```

as a prefix for the actual messages which look like this:

- bad_request_uri_length
- bad_request_uri_encoding
- wpf2b_404_head

benrubson:
Perhaps you could then add the following prefix to be compatible with a real Fail2Ban installation (using this PR) :
`user misbehaved: `

Which could then give in your code :
```
private $prefix = 'user misbehaved: malicious traffic detected: ';
private $prefix_instant = 'user misbehaved: break-in attempt detected: ';
```

szepeviktor:
Thank you.

I've commented to note that there are events which should be banned **instantly**

benrubson:
Not sure how we could do this here @szepeviktor, a message pattern which is instantly banned :|
The only solution I see is an additional jail with its own `retry` parameter and its instantly banned rules.

benrubson:
Perhaps we could at least extend the current list :
`user \S* ?(not found|not authorized|password mismatch|sent invalid token|misbehaved)`
to :
`user \S* ?(not found|not authorized|password mismatch|sent invalid token|misbehaved|sent malicious traffic|tried to break-in)`

Would this "help" you somewhat @szepeviktor ?

szepeviktor:
Thank you!

> `sent malicious traffic|tried to break-in`

We could skip `sent malicious traffic` as it is the same as `misbehaved`.
So `tried to break-in` would be the *instant* one.

benrubson:
Not sure it is possible, but @sebres could certainly confirm, can we make one of the patterns (or one of the rules) an _instant_ (`retry = 0`) rule, without having to define a new jail ?
Thx 👍

benrubson:
@yarikoptic did you see Serg's question some posts above ?
https://github.com/fail2ban/fail2ban/pull/1646#issuecomment-314523740
Thank you very much :) 👍

benrubson:
Just resolved conflicts here @sebres, if ever you feel we can finally merge this :)
Thank you 👍

sebres:
> if ever you feel we can finally merge this

I'm still not sure as regards this.
I would like the people are still configuring the fail2ban "properly", due to reasons described https://github.com/fail2ban/fail2ban/issues/2378#issuecomment-476140401

No idea how it is exactly implementable for apache, but it is simply for nginx/lighty... anyway I guess it'd be rather a task of apache (so web-server) self, not of the the fail2ban.

Let this still open, may be we'll find later something appropriate.

benrubson:
Thank you for your reply @sebres.

Regarding your linked comment, yes, Apache sends 4xx to access.log, not error.log.
Since a long time as you mentioned there :)

Btw, this PR relies on messages in error.log, as these would popup there.
So this should not be an issue.

Anyway, glad to see you keep this opened :+1:

Issue #1546, opened 2016-09-15
fail2ban CloudFlare v4 API Integration


wpbullet:
### Environment:

- Fail2Ban version (including any possible distribution suffixes):
- OS, including release name/version: 
- [x ] Fail2Ban installed via OS/distribution mechanisms
- [ ] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)

### The issue:

The old CloudFlare API is being deprecated in November, 2016 so migrating to v4 is necessary

### Configuration, dump and another helpful excerpts

I have detailed the method I am using for [integrating fail2ban with CloudFlare v4 API here](https://guides.wp-bullet.com/integrate-fail2ban-cloudflare-api-v4-guide/)

#### Any customizations done to /etc/fail2ban/ configuration

Custom action created following guide

sebres:
Thx for details, but ... 
I'll nevertheless close it as duplicate of #1420.

blindpet:
Mine does not use external dependencies like jq, it only uses curl and awk.

sebres:
In any case, we have already one issue about subject, so...  let's not have duplicates
The PR is any time welcome.

blindpet:
OK, I will find time to make a PR, do you want it on the 0.9.4 and newer branch?

sebres:
0.9 (master) is more desirable for common (version independent) things like configuration, etc.
Because we make periodically merge into 0.10, so all changes occurred in 0.9 apply for both branches.

PR #1651, opened 2016-12-31
Introduce Cloudflare API v4

### LINKED ISSUE #1420: Cloudflare API v1 sunsetting
Cloudflare is sunsetting the version of their API that the fail2ban cloudflare action is built on. See: https://blog.cloudflare.com/sunsetting-api-v1-in-favor-of-cloudflares-current-client-api-api-v4/

Their new v4 API is much expanded with different endpoints and much different semantics. This will break the current f2b cloudflare action starting in Nov 2016.

I've built some bash scripts that use the v4 API and I am testing integrating them with f2b by calling them from the cloudflare action rather than using the built-in actions.

I was going to try to fork and fix the cloudflare action to include my v4 API code, but I am actually using an older (os packaged, v. 8.x) version of fail2ban so I wouldn't actually be able to test adequately. In addition, after looking at what f2b does to execute the code defined in action files, I wasn't sure how much of my bash code would work in that context.

So I published my ban and urban scripts as a gist here: https://gist.github.com/pjv/926ece8549cd45bac4821945f6ad253c

Perhaps someone who has the f2b chops to turn that into a proper action can make use of it to update the cloudflare action before the sunsetting of their v1 API.

Note: the v4 API let's you be much more specific about how and what and where you "ban". My scripts are more or less identical in behavior to what the v1 API action currently does.

szepeviktor:
In the cloudflare action everyone is suggested to use API v4.
And I don't dare to contribute any actual change.

Closes #1420

### DIFF
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -41,11 +41,11 @@ actioncheck =
 # Values:  CMD
 #
 # API v1
-actionban = curl -s -o /dev/null https://www.cloudflare.com/api_json.html -d 'a=ban' -d 'tkn=<cftoken>' -d 'email=<cfuser>' -d 'key=<ip>'
+#actionban = curl -s -o /dev/null https://www.cloudflare.com/api_json.html -d 'a=ban' -d 'tkn=<cftoken>' -d 'email=<cfuser>' -d 'key=<ip>'
 # API v4
-#actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-#            -H 'Content-Type: application/json' -d '{ "mode": "block", "configuration": { "target": "ip", "value": "<ip>" } }' \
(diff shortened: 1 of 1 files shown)


sebres:
Just to link - #1420, #1546

szepeviktor:
@sebres Coud we merge it?

sebres:
Of course and you've although changed comments only, but...
- should it still stay commented? (I heard the old API is depreciated now)
- there are at least 2 issues which provided cloudflare v4 support. Could you take a look about a suggested solutions (to merge it with your, or to close it finally).
- me self have no experience with cloudflare at all (@yarikoptic any suggestions?)

szepeviktor:
I use the commented code myself
https://github.com/szepeviktor/debian-server-tools/blob/master/security/fail2ban-conf/action.d/cloudflare.local#L31

You may close both issues.

I wrote comment as I am not able to add tests nor provide support.

sebres:
Should we then rather uncomment v4 API (and comment v1)?

szepeviktor:
If you agree...

szepeviktor:
Thank you!

PR #1656, opened 2017-01-07
wp-login.php brute force attacks filter suggestions

mitchellkrogza:

   Helps deal with wp-login brute force attacks
   Hope this does not break anything
   Have kept it small, hope I did it right
   Have kept to correct formatting

### DIFF
--- a/config/filter.d/nginx-wplogin.conf
+++ b/config/filter.d/nginx-wplogin.conf
@@ -1,0 +1,15 @@
+# Fail2Ban filter for Nginx Web Server wp-login brute force attacks
+#
+#
+[INCLUDES]
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition] 
+
+failregex = ^<HOST> - .*] "POST /wp-login.php HTTP/1.1" 500
+ignoreregex =
+
+# Author: Mitchell Krog
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -344,7 +344,7 @@ port     = http,https
 logpath  = %(nginx_error_log)s
 maxretry = 2
 
-[wp-login]
+[nginx-wplogin]
 port     = http,https
 logpath = %(nginx_access_log)s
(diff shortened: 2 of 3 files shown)


mitchellkrogza:
Trying my hand at contributing again, I seem to always fail horribly with TRAVIS.

config/filter.d/nginx-wplogin.conf
```diff
@@ -0,0 +1,15 @@
...
+#
+[INCLUDES]
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+[Definition] 
+
+failregex = ^<HOST> - .*] "POST /wp-login.php HTTP/1.1" 500
```
sebres: Never do that. It's very vulnerable (so it would be not anchored at all, neither at end but nor at begin of expression)
Just take a look by another nginx filters how it will be made, e. g. https://github.com/fail2ban/fail2ban/blob/0.10/config/filter.d/nginx-botsearch.conf#L11

Additionally it may be very slow and resources hungry on long lines, that normally not match (because regex will try to follow greedy match `.*` as long as possible).
BTW. It will work only for method "POST" and "HTTP/1.1", so attacker can easy bypass it by using another method or http-protocol.

So your expression would be like:
```diff
- failregex = ^<HOST> - .*] "POST /wp-login.php HTTP/1.1" 500
+ failregex = ^<HOST> \- \S+ \[[^\]]*\] \"(\w+) /wp-login.php \S+\" 500
```

sebres:
I'll close this PR because we already have such filter (but more safer as yours and expandable with the custom parameters). 

See https://github.com/fail2ban/fail2ban/blob/0.10/config/filter.d/nginx-botsearch.conf

mitchellkrogza:
No problem @sebres I am running the [nginx-botsearch] jail (unmodified) however it is not picking up any wp-login brute force attempts. I will have a look at trying to rather customize the nginx-botsearch instead.

Issue #1647, opened 2016-12-26
IPv6 logo


szepeviktor:
Adding the ipv6 logo to the readme would be nice
http://www.worldipv6launch.org/wp-content/themes/ipv6/downloads/World_IPv6_launch_logo.svg

Modifying size parameters: `width="512px" height="512px" viewBox="0 0 512 512" enable-background="new 0 0 512 512"`

sebres:
[Click to see how it looks](https://github.com/fail2ban/fail2ban/blob/ipv6-logo-gh-1647/README.md)

PR #1661, opened 2017-01-11
0.10: readme.md - added IPv6 launch logo for 0.10th branch

sebres:
Closes gh-1647

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ Though Fail2Ban is able to reduce the rate of incorrect authentications
 attempts, it cannot eliminate the risk that weak authentication presents.
 Configure services to use only two factor or public/private authentication
 mechanisms if you really want to protect services.
+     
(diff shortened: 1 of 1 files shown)


sebres:
[Click to see how it looks](https://github.com/fail2ban/fail2ban/blob/ipv6-logo-gh-1647/README.md)

szepeviktor:
Very nice!

szepeviktor:
We could even add World_IPv6_launch_logo.svg to *this* repo.

sebres:
> We could even add World_IPv6_launch_logo.svg to this repo.

I don't know, it would be a good idea - although it is 4 KB only, but it belong later to the each copy, release, etc.
BTW. I think github will replace it in the markdown html-representation with something like "...githubusercontent.com/hash/...", so it is IMHO irrelevant.

sebres:
@yarikoptic Merge?

szepeviktor:
Aren't `/files` deleted when making a deb/rpm etc package?

sebres:
Maybe, but... let us do that, if something goes wrong with current http-reference...

PR #1663, opened 2017-01-12
Introduced netscaler action

jjeziorny:
Hi team,

I've created a new action for fail2ban which uses a netscaler load balancer appliance to block any offending IPs, the action talks to the netsclaer using it's NITRO API and IPs are added/removed to a dataset is then used on a responder policy to RESET or DROP packets from the offending source.
You can find some more info on this action at http://blog.ckzone.eu/2017/01/fail2ban-action-for-citrix-netscaler.html

Cheers


jjeziorny:
What's the next ste here?  If there's anything I should be doing please let me know

sebres:
Looks good so far.
> If there's anything I should be doing please let me know

Yes, if this should belong to 0.9 (master) branch also, you can rebase your branch netscaler-action (or cherry pick this single commit) to master and push it hereafter to github with -force with same branch name. Then you can change the target branch here in PR (using Edit button in head und switch branch in the select-box).
If not, it is ok also.

And please provide a changelog entry (with reference to gh-1663).

jjeziorny:
Hi @sebres ,

I should have mentioned I'm a bit new to using git. Nevertheless, I think I managed to do what you asked, I reset my branch to 0.9 and re-introduced my file plus an update to ChangeLog.
I hope I've done it all right, let me know if it looks OK or if I need to do something else.

Cheers

sebres:
Good job! (You are single one, that did it so far successfully. All others abandoned and then started with new PR 😄)

Issue #1141, opened 2015-07-29
don't re-email after reboot


jepio:
I'm using fail2ban on my ssh server, with the iptables, complain and mail-whois actions. After a reboot all the previous bans get reapplied (definitely good, because they're permanent bans), but fail2ban insists on re-emailing me and the abuse emails. 

I get why it's doing this, but is there a way to prevent this? When googling for "fail2ban persistent bans across reboot" I saw mention of some form of database, but in my case it's just rescanning the log and reapplying all the rules, I don't see any form of memory

yarikoptic:
On Wed, 29 Jul 2015, jepio wrote:

>    I'm using fail2ban on my ssh server, with the iptables, complain and
>    mail-whois actions. After a reboot all the previous bans get reapplied
>    (definitely good, because they're permanent bans), but fail2ban insists on
>    re-emailing me and the abuse emails.

>    I get why it's doing, this but is there a way to prevent this? When
>    googling for "fail2ban persistent bans across reboot" I saw mention of
>    some form of database, but in my case it's just rescanning the log and
>    reapplying all the rules, I don't see any form of memory

which version of fail2ban are you using???

jepio:
0.9.2

sebres:
I think, the last version will do it also, because all of "active" (not expired) bans go banned "again" after restoring from db, and all of actions, including emails also, will be executed again.

jepio:
On Wed, 29 Jul 2015, Serg G. Brester wrote:

> 
> I think, the last version will do it also, because all of "active" (not 
> expired) bans go banned "again" after restoring from db, and all of 
> actions, including emails also, will be executed again.
> 

Is there a way to check from within an action if it's being triggered from 
the db or from the log? That way I could return early and not execute the 
action. Sort of like a "run-once" mode.

sebres:
The problem is currently a banning is action also, and fail2ban makes no difference between all of actions - so each action will be executed by banning of IP (ex.: ban - iptables, mail, anythingelse, etc.)

The solution would be:
* an optional parameter in action (like `exec_for_restored_ban=off`), that will prevent executing it for restored ban (comming from db).
* or just extending notifying (mail, etc.) into separate processing, for example that will send exact one mail with summary and list of all bans after some latency (have partially already realized it in one of my branches, similar issue resp. solution discussed in https://github.com/fail2ban/fail2ban/issues/832#issuecomment-60483697).

yarikoptic:
[...]

Issue #921, opened 2015-01-23
Some actions (e.g. send-mail) should not be ran upon banning on restart


szepeviktor:
```
Lines containing IP:106.187.49.59 in /var/log/apache2/*error.log
(no lines here)
```

And there are no lines because logs are rotated.

``` apache
2015-01-21 15:14:27,823 fail2ban.actions        [29757]: NOTICE  [apache-combined] Ban 106.187.49.59
2015-01-23 01:08:56,059 fail2ban.actions        [2941]: NOTICE  [apache-combined] Ban 106.187.49.59
```

szepeviktor:
There are 38 IPs blocked in this jail but I only get few emails. I do not understand.

szepeviktor:
BTW. Persistent banning list is a very nice feature.

yarikoptic:
so what is the problem really??? that email is empty because logs got rotated or what??? clarify plz

szepeviktor:
1. Restart fai2ban server after a few weeks of operation
2. I get emails of bans. I do not understand why. Is it because of the persistent banning?
3. There are bans whos IP is not in the log anymore because log rotation.

Is it OK to get emails (`banaction`) on server restarts?

szepeviktor:
What if I have 1000 bans and I restart the server?

yarikoptic:
> I get emails of bans. I do not understand why. Is it because of the persistent banning?

yeap.  because how are those IPs get banned?  actionban command is ran, and you have "send email" there.  I guess ideally we should be able to identify which actions should be run always or only on non-restarts... but atm we don't have that feature. So I would edit title here to signal that this is what needed

szepeviktor:
OK.
Thank you.

sebres:
BTW: in #716 object `Ticket` has a property `restored`, can be used to check a ban was restored after restart (to filter/ignore some action).

PR #1669, opened 2017-01-13
Recognize state of restored tickets

sebres:
Recognize restored (from database) tickets after restart (tell action restored state of the ticket).
Prevent executing of several actions (e.g. mail, send-mail etc) on restart (bans were already notified).
Test cases extended (smtp and by restart in ServerReloadTest).
Closes gh-1141
Closes gh-921

Affected actions:
  - complain.conf
  - dshield.conf
  - mail-buffered.conf
  - mail-whois-lines.conf
  - mail-whois.conf
  - mail.conf
  - sendmail-buffered.conf
  - sendmail-geoip-lines.conf
  - sendmail-whois-ipjailmatches.conf
  - sendmail-whois-ipmatches.conf
  - sendmail-whois-lines.conf
  - sendmail-whois-matches.conf
  - sendmail-whois.conf
  - sendmail.conf
  - smtp.py
  - xarf-login-attack.conf

New option `norestored` introduced for actions, to complete avoid of execution ban/unban for such tickets.
Usage: `norestored = true`.

For conditional usage in the shell-based actions an interpolation `<restored>` could be used also. 
E. g. it is enough to add following script-piece at begin of `actionban` (or `actionunban`), to prevent execution for restored tickets.
```bash
if [ '<restored>' = '1' ]; then exit 0; fi;
```
How it can be implemented you can see in https://github.com/fail2ban/fail2ban/pull/1669/commits/ee3c787cc68f6d043c8702c8f53a566277e2ee02#diff-b8000792a7bad4f1b05eeddea812cec5

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -166,11 +166,14 @@ fail2ban-client set loglevel INFO
     (DateDetector.matchTime calls rarer DateTemplate.matchDate now);
   - several standard filters extended with exact prefixed or anchored date templates;
 * Added possibility to recognize restored state of the tickets (see gh-1669).
-  For the shell-based actions it is enough to add following script-piece at begin 
-  of `actionban` (or `actionunban`), to prevent execution for the restored 
-  (after restart) tickets.
+  New option `norestored` introduced, to ignore restored tickets (after restart).
+  To avoid execution of ban/unban for the restored tickets, `norestored = true`
+  could be added in definition section of action.
+  For conditional usage in the shell-based actions an interpolation `<restored>` 
+  could be used also. E. g. it is enough to add following script-piece at begin
+  of `actionban` (or `actionunban`) to prevent execution:
   `if [ '<restored>' = '1' ]; then exit 0; fi;`
-  Several actions extended now:
+  Several actions extended now using `norestored` option:
   - complain.conf
   - dshield.conf
   - mail-buffered.conf
--- a/ChangeLog
+++ b/ChangeLog
@@ -165,6 +165,28 @@ fail2ban-client set loglevel INFO
   - faster match and fewer searching of appropriate templates
     (DateDetector.matchTime calls rarer DateTemplate.matchDate now);
   - several standard filters extended with exact prefixed or anchored date templates;
+* Added possibility to recognize restored state of the tickets (see gh-1669).
+  For the shell-based actions it is enough to add following script-piece at begin 
+  of `actionban` (or `actionunban`), to prevent execution for the restored 
+  (after restart) tickets.
+  `if [ '<restored>' = '1' ]; then exit 0; fi;`
+  Several actions extended now:
+  - complain.conf
+  - dshield.conf
+  - mail-buffered.conf
+  - mail-whois-lines.conf
+  - mail-whois.conf
+  - mail.conf
+  - sendmail-buffered.conf
+  - sendmail-geoip-lines.conf
+  - sendmail-whois-ipjailmatches.conf
+  - sendmail-whois-ipmatches.conf
+  - sendmail-whois-lines.conf
+  - sendmail-whois-matches.conf
+  - sendmail-whois.conf
+  - sendmail.conf
+  - smtp.py
+  - xarf-login-attack.conf
 * fail2ban-testcases:
   - `assertLogged` extended with parameter wait (to wait up to specified timeout,
     before we throw assert exception) + test cases rewritten using that
--- a/config/action.d/complain.conf
+++ b/config/action.d/complain.conf
@@ -34,6 +34,9 @@ before = helpers-common.conf
 
 [Definition]
 
+# bypass ban/unban for restored tickets
+norestored = 1
+
 # Option:  actionstart
 # Notes.:  command executed once at the start of Fail2Ban.
 # Values:  CMD
@@ -58,8 +61,7 @@ actioncheck =
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-actionban = %(_bypass_if_restored)s
-            oifs=${IFS}; 
+actionban = oifs=${IFS}; 
               IFS=.; SEP_IP=( <ip> ); set -- ${SEP_IP}; ADDRESSES=$(dig +short -t txt -q $4.$3.$2.$1.abuse-contacts.abusix.org); 
               IFS=,; ADDRESSES=$(echo $ADDRESSES)
             IFS=${oifs}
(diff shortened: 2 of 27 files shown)


szepeviktor:
It there a use case when we shouldn't bypass restored tickets?

sebres:
@szepeviktor I've currently updated description of PR, see https://github.com/fail2ban/fail2ban/pull/1669#issue-200693370 (or the changelog)

szepeviktor:
I see! Actual ban actions (firewall etc.) should occur but notifications.

sebres:
I've rewritten it with more better solution (using parameter

sebres:
I've rewritten it with much better slution - in all the actions new option `norestored = true`, to complete avoid execution of ban/unban.
See https://github.com/fail2ban/fail2ban/pull/1669/files#diff-8922b37854649fe53c88ec9e262af0a7R9

brianjmurrell:
Does this skip just banning on startup, or does it also prevent bans that expired while stopped from being unbanned?

sebres:
Expired bans were ignored always (also before this PR got implemented).
This PR addresses banning of **still active** tickets (during startup) only.

brianjmurrell:
> Expired bans were ignored always (also before this PR got implemented).

But of course they were if they were assumed to not exist in whatever banning mechanism is being used.  I.e. all `iptables` rules being removed on a system reboot, or bans being removed from `iptables` on a `fail2ban` _shutdown_.

But if you recognise that bans can persist a spontaneous stopping of fail2ban (i.e. the server loses power) , which this feature seems to be doing, then you have to make an allowance for bans that expire while fail2ban is not running, don't you?

Also, I'm not sure I'm familiar with this _ticket_ nomenclature.  Can you expand, or provide a pointer to an explanation of what a _ticket_ is in `fail2ban`?

sebres:
1. The ban-ticket is everything got banned (note that a ticket is not an IP only, since fail2ban can ban a user/e-mail/whatever).
2. The expired ticket is an outdated entry (fail2ban won't read them from database by startup) with the `endOfBan <= now`. Tickets like this are **always ignored**.
3. The restored ticket is the entry (fail2ban read from database by startup) with the `endOfBan > now`.
This tickets will be banned **always**, but fail2ban would bypass executing of `actionban` for the actions having "norestored = true" (or if it is specified as parameter for the action in jail config).

> But if you recognise that bans can persist a spontaneous stopping of fail2ban ...

Fail2ban does not consider that at all - this is matter of banning actions.
The tickets handling is always the same, regardless how fail2ban was stopped, by power failure or by regular stop, which would call `actionflush` (for all tickets) or `actionunban` (individually for each ticket). The re-ban process takes place at startup for every active ticket (considering restored flag).
If you want some different handling either you'd use your own `actionban`/`actionunban` that are able to recognize the state of IP from the netfilter (or it is even enough to check restored flag, for example ignore such IPs).

Again, the check whether the ticket (IP) is banned (e. g. through some "strange" persistent logic) in the action is (and remains) a matter of action. I don't see a necessity to make it a part of the fail2ban's logic (why?). Especially if some new features (like bulk-ban and bulk-unban) will be implemented finally.

If we either speaking about something like restart, so you want just update fail2ban configuration to the run-time - you'd rather use `reload` instead of `restart` (in this case fail2ban doesn't reban the tickets, unless you remove an action for some jail, otherwise fail2ban will flush all the tickets for this action).

brianjmurrell:
> The ban-ticket is everything got banned (note that a ticket is not an IP only, since fail2ban can ban a user/e-mail/whatever).

Interesting.  I'd only ever used it for IP addresses.

> The expired ticket is an outdated entry (fail2ban won't read them from database by startup) with the endOfBan <= now. Tickets like this are always ignored.

Understood.  But the problem is that should they be ignored if `endOfBan > last_time_fail2ban_was_running`?  I.e. `endOfBan` happened while `fail2ban` was stopped and so the `endOfBan` processing didn't happen?

Maybe tickets need a field indicating whether unban processing was completed for them and on startup, any tickets where `endOfBan < now && unbanned != true` need to have the _unban_ action applied to them.

> Fail2ban does not consider that at all - this is matter of banning actions.

Except that the banning action doesn't even get a chance to act on the above cases since they are completely ignored.

> If you want some different handling either you'd use your own actionban/actionunban that are able to recognize the state of IP from the netfilter (or it is even enough to check restored flag, for example ignore such IPs).

But again, that won't work since the above example would not even get an action process applied to it, correct?

>  I don't see a necessity to make it a part of the fail2ban's logic (why?).

Because of the above problem.  But to restate, to process any unbans that should have been processed while fail2ban was stopped.

> Especially if some new features (like bulk-ban and bulk-unban) will be implemented finally.

Yay!  This would be awesome for configurations like mine where each ban/unban is an SSH round-trip and can take literally hours to process many 10s-thousands of bans.

sebres:
> Except that the banning action doesn't even get a chance to act on the above cases since they are completely ignored.

Because it is not interesting for any action of fail2ban at the moment (I speak about default fail2ban action). Either it wants to ignore tickets already banned once (all actions where we already have `norestored = true`, like mail-actions etc), or it would reban it after (re)start, no matter how it was stopped, the state of the action is quasi "empty" after start.

For your case it would be enough to save the state in some file/database in `actionban` and remove it in `actionunban`. And restore all them in `actionstart` for all that are not going unbanned (e. g. we had power failure).
For a database usage example see https://github.com/fail2ban/fail2ban/issues/2013#issuecomment-426940189, although it's not "similar" issue, but similar usage (just not in ignore but by ban/unban/stop/start).
Also you can use fail2ban-database (or its API, `fail2ban.server.database`) to capture a list of all IPs that are currently banned:
```
# get all active IPs of jail from database:
sqlite3 'file:/var/lib/fail2ban/fail2ban.sqlite3?mode=ro' "select distinct ip from bans where jail = <name> and timeofban + <bantime> > datetime('now', 'localtime')"
```
For an example using file see similar PR #2456 (implemented "norestored" for an action of 0.9), but you can extend it for your purposes.

Also because v.0.10 already partially implements "bulk" unban (for all tickets) - `actionflush` (`actionstop` would not avoid individual unbanning for every single IP), you can do that much simple right now:
```ini
[Definition]
cmd = ssh ...
actionban = %(cmd)s ban <name> -single <ip>
actionunban = %(cmd)s unban <name> -single <ip>
actionflush = %(cmd)s unban <name> -all; # this should remove all tickets of jail.
actionstart = %(actionflush)s
```
This configuration would simply remove all known tickets with a single command (by stop using flush and not via unban over the single IPs), but also by the start before it'd begin to restore all the tickets.
So it hold your other side consistent and it does no matter anymore how restart is happened (after power failure as well as after stop/start it would clear the "table" on other side).
Sure it'd reban every single IP after start, but as already said it's matter of bulk-ban and has nothing with restored flag or persistence state.

PR #1678, opened 2017-01-22
An IP address has a PTR record

szepeviktor:
In sendmail-geoip-lines action

### DIFF
--- a/config/action.d/sendmail-geoip-lines.conf
+++ b/config/action.d/sendmail-geoip-lines.conf
@@ -36,7 +36,7 @@ actionban = ( printf %%b "Subject: [Fail2Ban] <name>: banned <ip> from `uname -n
             http://whois.domaintools.com/<ip>\n\n
             Country:`geoiplookup -f /usr/share/GeoIP/GeoIP.dat "<ip>" | cut -d':' -f2-`
             AS:`geoiplookup -f /usr/share/GeoIP/GeoIPASNum.dat "<ip>" | cut -d':' -f2-`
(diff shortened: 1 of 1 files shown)


szepeviktor:
Should it be against 0.9?

szepeviktor:
Would it be possible to merge this one?

sebres:
I've made a little bit better solution (see #1705).
For older version (v.0.9.x) we can use this one.

szepeviktor:
@sebres Thank you.

If there are multiple PTR record which one will get displayed?

sebres:
> If there are multiple PTR records which one will get displayed?

Currently this will return the first only. Of course we can change it to return a list, or still one tag `<ip-hosts>` that in opposition to `<ip-host>` returns a list.
But multiple PTR records (i. e. reverse-DNS) are not recommended (and too rare in the practice).

BTW In opposition to `host -t PTR ...` the new tag `<ip-host>` returns only available host name. I. e. no arpa reversed address (for this purpose use another tag `<ip-rev>`):
```bash
$ host -t PTR "4.4.4.4"
Host 4.4.4.4.in-addr.arpa. not found: 3(NXDOMAIN)

$ # ip-host:
$ fail2ban-python -c "from fail2ban.server.ipdns import IPAddr; print(IPAddr('4.4.4.4').getHost())"
None

$ # ip-rev:
$ fail2ban-python -c "from fail2ban.server.ipdns import IPAddr; print(IPAddr('4.4.4.4').getPTR())"
4.4.4.4.in-addr.arpa.
```

Issue #1684, opened 2017-02-05
Syntax error for action 'complain'


jangrewe:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.9.6
- OS, including release name/version:  Debian stretch
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so):
I replaced `complain.conf` and `helpers-common.conf` with the files from the master branch to see if they fix my issue, but it still persists.

### The issue:

```
2017-02-05 16:57:24,033 fail2ban.action         [3530]: ERROR   oifs=${IFS}; IFS=.;SEP_IP=( 87.200.5.106 ); set -- ${SEP_IP}; ADDRESSES=$(dig +short -t txt -q ${4}.${3}.${2}.${1}.abuse-contacts.abusix.org); IFS=${oifs}
IP=87.200.5.106
if [ ! -z "${ADDRESSES}" ]; then
(printf %b "Dear Sir/Madam,\n\nWe have detected abuse from the IP address $IP, which according to a abusix.com is on your network. We would appreciate if you would investigate and take action as appropriate.\n\nLog lines are given below, but please ask if you require any further information.\n\n(If you are not the correct person to contact about this please accept our apologies - your e-mail address was extracted from the whois record by an automated process.)\n\n This mail was generated by Fail2Ban.\nThe recipient address of this report was provided by the Abuse Contact DB by abusix.com. abusix.com does not maintain the content of the database. All information which we pass out, derives from the RIR databases and is processed for ease of use. If you want to change or report non working abuse contacts please contact the appropriate RIR. If you have any further question, contact abusix.com directly via email (info@abusix.com). Information about the Abuse Contact Database can be found here: https://abusix.com/global-reporting/abuse-contact-db\nabusix.com is neither responsible nor liable for the content or accuracy of this message.\n\n"; date '+Note: Local timezone is %z (%Z)'; grep -E '(^|[^0-9])87.200.5.106([^0-9]|$)' /var/log/mail.log) | mail -s "Abuse from 87.200.5.106"  ${ADDRESSES//,/\" \"}
fi -- stdout: b''
[...]

PR #1685, opened 2017-02-06
WIP: Make Abusix lookup compatible with Dash

jangrewe:

### DIFF
--- a/config/action.d/complain.conf
+++ b/config/action.d/complain.conf
@@ -63,7 +63,7 @@ actioncheck =
 #
 actionban = oifs=${IFS}; 
             REV_IP=$(printf %%s <ip>. | tac -s.)
-            ADDRESSES=$(dig +short -t txt -q ${REV_IP}abuse-contacts.abusix.org)
+            ADDRESSES=$(dig +short -t txt -q ${REV_IP}abuse-contacts.abusix.org | tr -d '"')
             IFS=,; ADDRESSES=$(echo $ADDRESSES)
             IFS=${oifs}
             IP=<ip>
--- a/config/action.d/complain.conf
+++ b/config/action.d/complain.conf
@@ -62,7 +62,7 @@ actioncheck =
 # Values:  CMD
 #
 actionban = oifs=${IFS}; 
-            REV_IP=$(printf %%s ${IP}. | tac -s.)
+            REV_IP=$(printf %%s <ip>. | tac -s.)
             ADDRESSES=$(dig +short -t txt -q ${REV_IP}abuse-contacts.abusix.org)
             IFS=,; ADDRESSES=$(echo $ADDRESSES)
             IFS=${oifs}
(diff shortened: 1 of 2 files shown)


sebres:
So you want resolve one dependency by introducing another one? I mean `tac` (it is always available by any posix-systems?).
Additionally test cases failed, so something is wrong there (maybe, usage of `${IP}` before declared?)...
Some test cases are a little bit artificial (replace something within test-case with output to stdout instead of real work), it could be just a mistake... ATM no time to check it.

jangrewe:
The "WIP:" is supposed to mean "Work In Progress", so it's not done yet.
`tac` is part of `coreutils`, so if you're on a system that has `dig`, or Fail2ban itself, installed, i think it's safe to assume that the `coreutils` packages is already installed.
Regarding the tests: i'll have a look at those when my modifications are completed.

jangrewe:
I had a look why the test fails (after my recent fix), and it's pretty obvious:

```

FAIL: testComplexMailActionMultiLog (fail2ban.tests.servertestcase.ServerConfigReaderTests)

  File "./fail2ban/tests/servertestcase.py", line 1725, in testComplexMailActionMultiLog

    self.assertLogged(*tests['ip4-ban'], all=True)

  File "./fail2ban/tests/utils.py", line 670, in assertLogged

    self.fail("%r was not found in the log: ===\n%s===" % (s_, logged))

AssertionError: 'mail -s Abuse from 87.142.124.10 abuse-1@abuse-test-server abuse-2@abuse-test-server' was not found in the log: ===
```

Yup, that line is not in the log indeed, as the purpose of the script is to get the abuse contact and send the mail there, not to the pre-defined test addresses ;-)

But that part works fine!
```
[...]
 -- stdout: 'Lines containing failures of 87.142.124.10 (max 2)'
 -- stdout: './fail2ban/tests/files/testcase01.log:Dec 31 11:59:59 [sshd] error: PAM: Authentication failure for kevin from 87.142.124.10'
 -- stdout: './fail2ban/tests/files/testcase01a.log:Dec 31 11:55:01 [sshd] error: PAM: Authentication failure for test from 87.142.124.10'
 -- stdout: 'mail -s Abuse from 87.142.124.10 "abuse@telekom.de"'
```

I haven't figured out where that address comes from (or the structure of the tests in general), but that should be easily fixable for somebody who knows this.

Is suspect that /bin/sh in that Vagrant box is also using Bash instead of Dash, so i'm not quite done yet...

jangrewe:
I just noticed that my box is now sending complaint mails, which probably means that Dash only had an issue with Arrays, but subshells are working fine.
Now if anyony could have a look at the tests...

sebres:
PR for #1684

jangrewe:
Thanks @sebres , i've been running my patched version for the last week, and it's working fine so far.

sebres:
I've closed this because of another solution, with better handling using new tag `<ip-rev>` in our 0.10th version, see complain.conf in #1698.

Your solution can be used in previous versions (0.9.x), where tag `<ip-rev>` is not implemented.

Issue #1689, opened 2017-02-14
Make lowest rule number in action.d/bsd-ipfw.conf configurable


chtheis:
As discussed on the user mail list:
Under FreeBSD's ipfw tables are added which then contain the IPs to be blocked. The action starts looking for an empty slot starting with rule number 1 which may be OK. But if someone has fine tuned the firewall settings, or accidentally allowed localhost to be blocked, another rule number to start with may be desired. 
I have changed bsd-ipfw.conf so that it contains another variable lowest_rulenum with which the rule number the action starts looking for a free slot can be configured.

PR #1690, opened 2017-02-14
#1689: Make lowest rule number in action.d/bsd-ipfw.conf configurable

chtheis:

- [X ] **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement
- [X ] **LIST ISSUES** this PR resolves
- [X ] **KEEP PR small** so it could be easily reviewed.
- [X ] **AVOID** making unnecessary stylistic changes in unrelated code


sebres:
Closes #1689

sebres:
Looks good, if someone on bsd can test it and provide a feedback.... I'll merge it then.

St-Ranger:
Thank you!
Now that you've shown this example, I found the relevant portion in the man page.
OK. so than 2) is no longer an issue.

Can someone, please, change the default value, as described in 1)?

St-Ranger:
The numbers that come from default rules in `/etc/rc.firewall` could be different, depending on which default configuration of firewall you choose. But we do not need to worry about that. Moreover, we should not aim at going beyond all the default rules (actually some of those are `allow` rules). 
Rather, we should just place it:
1. behind the loopback rule (to avoid shooting oneself in the foot inadvertently, and
2. before any other "allow" rule.
This way, we are not conflicting with any _default_ rules.

It can be a problem only if someone created a custom "safety" rule ahead of the most of the default, assigning `deny` rules (to prevent an accidental lockout from the server). But then the person consciously assigned a specific number. 
(Note that for those people, this change should not have any negative effect if they already have fail2ban installed and running, since previously, the rule was numbered 0001 already. ) So, for new fail2ban users, it would make sense to add a custom notice (probably on the FreeBSD port/pkg side) that is displayed upon installation of the package), alerting to that fact (like a few packages do).
Something like this:

> When enabled, fail2ban's ipfw action creates a rule with the number 111. 
> If you have a customized firewall that might conflict with this, you can change this number by setting # 
> lowest_rule_num to a different value, e.g. in /usr/local/etc/fail2ban/jail.local as follows:
> `banaction = bsd-ipfw[lowest_rule_num=XXX]`

I suggest that you duplicate this same note in the comments in the file edited here, `config/action.d/bsd-ipfw.conf`

Moreover, in `config/action.d/bsd-ipfw.conf` comments, I would also add something like this:

`This variable is set to 111 to be after the first rule, that allows loopback connection. 
numbered 100 in the default /etc/rc.firewall 
If you are using default /etc/rc.firewall, avoid setting this number to a smaller value.
With an exception for that first loopback rule, you want this number to be smaller than your first rule in your firewall configuration that allows packets that match that rule ("allow", "accept", "pass", "permit") or redirects such packets ("forward", "fwd", "nat", "pipe", "queue", "tee", etc.). Therefore, in most cases, setting this number above all of your customized rules would be a mistake.
`

So, I propose:
-lowest_rule_num = 1
+lowest_rule_num = 111

Can you make this changes, or would you want me to create another commit? (I am not sure if one can create a commit on top of someone else's un-merged pull request without creating a new pull request.)

config/action.d/bsd-ipfw.conf
```diff
@@ -81,3 +81,11 @@ block = ip
 # Values:  STRING
 #
 blocktype = unreach port
+
+# Option:  lowest_rule_num
+# Notes:   When fail2ban starts with action and there is no rule for the given table yet
+#          then fail2ban will start looking for an empty slot starting with this rule number.
+# Values:  NUM
+lowest_rule_num = 1
```
sebres: So how we'll set it now? 
Leave as 1, or as suggested from @St-Ranger:
```diff
- lowest_rule_num = 1
+ lowest_rule_num = 111
```

chtheis:
Am 30.06.2017 um 18:10 schrieb Serg G. Brester:
> *@sebres* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In config/action.d/bsd-ipfw.conf 
> <https://github.com/fail2ban/fail2ban/pull/1690#discussion_r125075554>:
> 
>> @@ -81,3 +81,11 @@ block = ip
>   # Values:  STRING
>   #
>   blocktype = unreach port
> +
> +# Option:  lowest_rule_num
> +# Notes:   When fail2ban starts with action and there is no rule for the given table yet
> +#          then fail2ban will start looking for an empty slot starting with this rule number.
> +# Values:  NUM
> +lowest_rule_num = 1
> 
> So how we'll set it now?
> Leave as 1, or as suggested from @St-Ranger <https://github.com/st-ranger>:
> 
> - lowest_rule_num = 1
> + lowest_rule_num = 111

I suggest following as suggested by st-ranger.
One reason to make this configurable was to avoid accidentally blocking 
localhost connection, effectively locking oneself out of the system.
With the change from st-ranger that couldn't happen.

When I submitted the patch I was conservative, making the old behaviour 
the default, but without a real reason than to keep it as it was.
But those who really administrate the firewalls claim it is not a 
problem to put it at a higher rule number and I don't see a problem 
there, too.


Best regards

Christoph

sebres:
Thus fixed, and merged...

Issue #1693, opened 2017-02-15
Apache : syslog logging method breaks rules


benrubson:
Hello,

Apache can easily log using syslog :
`ErrorLog syslog`

Unfortunately this breaks fail2ban rules.

Here is an example of Apache syslog lines :
```
Feb 15 16:23:00 srv httpd[22034]: [authz_core:error] [pid 22034] [client 1.2.3.4:58585] AH01630: client denied by server configuration: /home/www/
Feb 15 16:23:40 srv httpd/backend1[22034]: [authz_core:error] [pid 22036] [client 1.2.3.4:59392] AH01630: client denied by server configuration: /home/backend1/
Feb 15 16:54:53 tools apache2[18154]: [:error] [pid 18154:tid 140680873617152] [client 1.2.3.4:48154] AH01630: client denied by server configuration: /var/www/
```

Sounds like the following would need to be extended ?
```
_apache_error_client = \[\] \[(:?error|\S+:\S+)\]( \[pid \d+(:\S+ \d+)?\])? \[client <HOST>(:\d{1,5})?\]
```

Thank you very much 👍 

Best regards,

Ben

benrubson:
Solved by #1695 !

benrubson:
Closed, see PR.

PR #1695, opened 2017-02-17
Apache, detect syslog prefix

benrubson:
Hello,

This PR solves #1693.

It detects syslog prefix (` host process`) in log lines, making fail2ban work when Apache logs using syslog.

Thank you 👍 

Ben

### DIFF
--- a/config/filter.d/apache-common.conf
+++ b/config/filter.d/apache-common.conf
@@ -22,6 +22,9 @@ apache-prefix-syslog = %(__prefix_line)s
 apache-prefix-logfile = \[\]\s
 apache-prefix-all = (?:%(apache-prefix-logfile)s|%(apache-prefix-syslog)s)?
 
+# Setting for __prefix_line (only `logging=syslog`):
+_daemon = (?:apache\d*|httpd(?:/\w+)?)
+
 apache-prefix = <apache-prefix-<logging>>
 
 _apache_error_client = <apache-prefix>\[(:?error|\S+:\S+)\]( \[pid \d+(:\S+ \d+)?\])? \[client <HOST>(:\d{1,5})?\]
--- a/config/filter.d/apache-common.conf
+++ b/config/filter.d/apache-common.conf
@@ -1,23 +1,30 @@
 # Generic configuration items (to be used as interpolations) in other
 # apache filters.
 
-# If you use "ErrorLog syslog" in your Apache configuration,
-# you should then create the following apache-common.local file :
-# [INCLUDES]
-# before = common.conf
-# [DEFAULT]
-# _apache_prefix = %(__prefix_line)s
-
 [INCLUDES]
 
+before = common.conf
 # Load customizations if any available
 after = apache-common.local
 
 [DEFAULT]
 
-_apache_prefix = \[\]\s
+# Apache logging mode:
+#   all - universal prefix (logfile, syslog)
+#   logfile - logfile only
+#   syslog - syslog only
+# Use `filter = apache-auth[logging=syslog]` to get more precise regex if apache logs into syslog (ErrorLog syslog).
+# Use `filter = apache-auth[logging=all]` to get universal regex matches both logging variants.
+logging = logfile
 
-_apache_error_client = ^%(_apache_prefix)s\[(:?error|\S+:\S+)\]( \[pid \d+(:\S+ \d+)?\])? \[client <HOST>(:\d{1,5})?\]
+# Apache logging prefixes (date-pattern prefix, server, process etc.):
+apache-prefix-syslog = %(__prefix_line)s
+apache-prefix-logfile = \[\]\s
+apache-prefix-all = (?:%(apache-prefix-logfile)s|%(apache-prefix-syslog)s)?
+
+apache-prefix = <apache-prefix-<logging>>
+
+_apache_error_client = <apache-prefix>\[(:?error|\S+:\S+)\]( \[pid \d+(:\S+ \d+)?\])? \[client <HOST>(:\d{1,5})?\]
 
 datepattern = {^LN-BEG}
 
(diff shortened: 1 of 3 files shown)


benrubson:
Any thoughts @sebres & @yarikoptic ?
Many thx 👍 
Ben

sebres:
Well, if apache will do logging via syslog, you could get more complex log-line.

We've already a proper way for parse a syslog-entries. It's called `__prefix_line` from include `common.conf`.
In this case it can be used like this:
```
[INCLUDES]
before = common.conf
[Definition]
failregex = ^%(__prefix_line)s...
# or rather
failregex = ^(?:\[\] |%(__prefix_line)s)...
```

But I'm not sure it should be default handling.
I mean that fail2ban can be and should be configured (it is rather a tool not a self-learning thing).
I this case if someone do `ErrorLog syslog`, then he can also customize his fail2ban configuration.
Because you can switch it also using systemd-journal, change longging format etc.
All this is IMHO customization and should be viewed as it.

benrubson:
Thank you Serg for your feedback.
It not enabled by default, perhaps we could at least give a working example in the `apache-common.conf` so that users can easily switch to syslog method removing comment characters (`#`) ?
Thank you 👍

sebres:
> perhaps we could at least give a working example

You have not understood: I told about too many possibilities there. A "working example" by you may not work by someone other.
Please take a look in `common.conf`, then you'll possibly get why it may be too different.

benrubson:
I meant a working example with `__prefix_line` :)

sebres:
I say never "no" to little bit more documentation...
BTW possibly a wiki would be also a good place for it (I mean something like "How to add common prefix if service logs via syslog")?

benrubson:
I then simply wrote a syslog use-case in the `apache-common.conf` file following the example you previously gave.
Will be useful for those who use `ErrorLog syslog`, everything they need is ready.
Of course feel free to comment !
Many thanks 👍 
Ben

config/filter.d/apache-common.conf
```diff
@@ -5,10 +5,16 @@
 
 # Load customizations if any available
 after = apache-common.local
+# Uncomment the following line if you use ErrorLog = syslog
+before = common.conf
```
sebres: > Uncomment the following line...

Hmm... Is already uncommented.

Secondly, it's a bad idea to force the user to modify `.conf` (Good practice is "Never touch the stock configuration").
Better would be to write own `apache-common.local` with something like this:
```
[INCLUDES]
before = common.conf
[DEFAULT]
_apache_prefix = %(__prefix_line)s
```

benrubson: My mistake for the 1st point, sorry...
For the second, totally agree, I then just committed a fix.
Thank you 👍

sebres: You don't get it...

Local configs do not belong to the release (It makes no sense, because will overwrite user config). 
The user should get an advice to create `apache-common.local`.

benrubson: So an advice in the .conf file ?
I thought about doing so, and finally created the `.local` file...

benrubson:
Should be OK this time :)
Sorry for all the retries...
Ben

benrubson:
Could we also think about backporting this to 0.9 ? File is identical.
Thank you again.

sebres:
Rebased to current 0.10, a bit reorganized: to switch to the syslog-format used parameter `logging` now.
Example of usage (per jail):
```INI
# jail.local:
[apache-auth]
filter = apache-auth[logging=syslog]
```
Example of usage (for all apache filters, via `apache-common`)
```INI
# apache-common.local:
[Definition]
logging = syslog
```

config/filter.d/apache-common.conf
```diff
@@ -22,6 +22,9 @@ apache-prefix-syslog = %(__prefix_line)s
 apache-prefix-logfile = \[\]\s
 apache-prefix-all = (?:%(apache-prefix-logfile)s|%(apache-prefix-syslog)s)?
 
+# Setting for __prefix_line (only `logging=syslog`):
+_daemon = (?:apache\d*|httpd(?:/\w+)?)
```
benrubson: Apache now has an option to configure the syslog tag :
https://httpd.apache.org/docs/trunk/mod/core.html#errorlog
It may then be something different than `apache|httpd`.

benrubson:
Thank you for your review and rework on this @sebres 👍

config/filter.d/apache-common.conf
```diff
@@ -3,12 +3,31 @@
...
+# Use `filter = apache-auth[logging=all]` to get universal regex matches both logging variants.
+logging = logfile
+
+# Apache logging prefixes (date-pattern prefix, server, process etc.):
+apache-prefix-syslog = %(__prefix_line)s
+apache-prefix-logfile = \[\]\s
+apache-prefix-all = (?:%(apache-prefix-logfile)s|%(apache-prefix-syslog)s)?
+
+# Setting for __prefix_line (only `logging=syslog`):
+_daemon = (?:apache\d*|httpd(?:/\w+)?)
```
benrubson: Apache now has an option to configure the syslog tag :
https://httpd.apache.org/docs/trunk/mod/core.html#errorlog
It may then be something different than `apache|httpd`.

sebres: Apache has many options to configure everything, so whole log-format may be different.
We support only default formats. Another things should be then configured via own/custom filter...
Back to your question, you can set in to other thing (v0.10 only):
```INI
# jail.local
[apache-auth]
filter = apache-auth[logging=syslog, _daemon="my-apache"]
```
Or via apache-common.local:
```INI
[Definition]
_daemon = \S*
```

benrubson: Well, the default regex already authorise things such as `httpd/backend1`, so it's permissive enough :)
Thx 👍

benrubson:
Thx @sebres 👍

PR #1697, opened 2017-02-18
postfix: ban lost connection after <STAGE> from <HOST>

evgeni:
"`lost connection after <STAGE> from <HOST>`" is a common log-filer for Postfix [1]. Analyzing my own logs I did not find any legitimate occurrences of such lines in bulk, see also [2].

The `<STAGE>` could be limited to `AUTH`, as this is the main offender (both according to my logs and "the Internet").

[1] http://www.iredmail.org/forum/topic8172-iredmail-support-howto-protect-against-postfix-auth-dos-attacks.html
[2] https://sourceforge.net/p/fail2ban/mailman/message/31893097/

### DIFF
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -19,6 +19,7 @@ failregex = ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 554 5\.7
             ^%(__prefix_line)sNOQUEUE: reject: VRFY from \S+\[<HOST>\]: 550 5\.1\.1 .*$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.1\.8 <\S*>: Sender address rejected: Domain not found; from=<\S*> to=<\S+> proto=ESMTP helo=<\S*>$
             ^%(__prefix_line)simproper command pipelining after \S+ from [^[]*\[<HOST>\]:?$
+            ^%(__prefix_line)slost connection after \S+ from \S+\[<HOST>\]$
 
 ignoreregex = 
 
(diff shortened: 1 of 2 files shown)


config/filter.d/postfix.conf
```diff
@@ -19,6 +19,7 @@ failregex = ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 554 5\.7
             ^%(__prefix_line)sNOQUEUE: reject: VRFY from \S+\[<HOST>\]: 550 5\.1\.1 .*$
             ^%(__prefix_line)sNOQUEUE: reject: RCPT from \S+\[<HOST>\]: 450 4\.1\.8 <\S*>: Sender address rejected: Domain not found; from=<\S*> to=<\S+> proto=ESMTP helo=<\S*>$
             ^%(__prefix_line)simproper command pipelining after \S+ from [^[]*\[<HOST>\]:?$
+            ^%(__prefix_line)slost connection after \S+ from \S+\[<HOST>\]$
```
sebres: If someone uses this filter like "out-of-the-box", it should be not affected by things, that are not really mean auth-failures / wrong handling. I mean things like port-scanning or DDOS (e. g. like `lost connection`)... 
So if we'll include it in fail2ban, then only optional (see for example the parameter `mode=aggressive` in sshd-filter).

victimsofgaybullying:
Thank you for adding a mode=agressive to postfix's filter. In the wild I've seen this would be a benefit if you see #1813. Up to spammers do 20-40 retries and all I get in the postfix logs is lost connection after AUTH ...

  Cheers, Luke

victimsofgaybullying:
@sebres 
postfix-rbl is used becuase peope want to have different ban times.

Why not merge in the other postfix filters together (postfix-sasl.conf - postfix-rbl.conf) into the same .conf file and then people from multiple jails can invoke the same filter with different arguments. Would that work?

e,g aggressive (inclusive of RBL sasl the kitchen sink), sasl, rbl, auth etc into one filter.

sebres:
Implemented in #1825 via `mode=ddos` or `mode=aggressive`.
Thanks!

evgeni:
:+1: thanks!

Issue #1625, opened 2016-11-26
TODO: config-files - replace static init-param `<param>` with `%(param)s`


sebres:
Substitution `<param>` should be used for dynamic interpolation only after #1622 merged in 0.10
- [ ] review configurations to replace all static parameters with `%(param)s`
- [ ] if done - filter::getCombined can be removed;

sebres:
filter::getCombined in current 0.10 moved to config reader, because now used in actions also.

sebres:
Implemented in newer version of #1622 additionally with #1698:
- substitution `<param>` can be used for dynamic and static interpolation now.
- if parameter `<param>` can be resolved direct in config-reader (static), it will be replaced there, otherwise `<param>` remains (dynamic) and can be substituted later if it becomes value (e. g. by executing of the action);

Issue #1173, opened 2015-08-28
fail2ban failing on lines with many (10.000+) characters


mrosic:
We had obscure database errors with roundcube and the errors were over 10.000 characters in a single line. The resulting file was 150MB and fail2ban never finished reading that file, not even after 650 CPU hours. During those 650 CPU hours fail2ban also didn't ban anybody because apparently it never finished starting.

I used fold -w 100 on the file, restarted fail2ban and it finished reading the file in less than 2 CPU Minutes.

I suggest that fail2ban only reads the first 500 characters (or whatever) of each line and then goes to the next line AND/OR shows an error message when it notices lines with an obscene amount of characters.

PS
It would be interesting to see whether one could force roundcube (or other programs, for that matter) to create such errors on-demand. In that case one could remotely disable fail2ban and brute-force away at the server.

yarikoptic:
On Fri, 28 Aug 2015, mrosic wrote:

>    I suggest that fail2ban only reads the first 500 characters (or whatever)
>    of each line and then goes to the next line AND/OR shows an error message
>    when it notices lines with an obscene amount of characters.

Many regexes require end of the line anchoring, so just trimming the
logline for all jails might lead  to injection attacks -- if those lines
didn't actually contain any relevant to be matched entries (did
they?), we better be just skipping those.  But again I think it should
be an option per jail.

mrosic:
They did not contain relevant entries. And I agree, this should be a per jail option.

sebres:
Additionally me interested a question - why it hangs, because independent of new option exists or not - it should process such lines (possible a little bit slower, but not endless). Possible we have a not optimal regexp here, and I think that are a (non greedy) regexp parts like `.*?`.

Additionally I think this filter resp. jail expect an new `ignoreregex` rule.
Can you provide a pair examples or at least some common begin of such lines?

As regards a new option:
Per jail or not, can be controlled within DEFAULT. If it will be added (uncommented) inside default, so it applies to all jails, otherwise it can be (re)defined for single jail.
So we can add an option `maxchars`, here is a config example:
```
# "maxchars_mode" specified the mode to process lines longer as "maxchars_len" characters:
# crop: truncate line (add a new line) and try to match this line.
# ignore: ignore lines longer as N characters
maxchars_mode = ignore

# A maximum line length "maxchars_len" to be matched (if "ignore") or truncated (if "crop").
maxchars_len = 500

[...]

PR #1698, opened 2017-02-20
0.10 filter captures to actions

sebres:
* Introduced new filter option `prefregex` for pre-filtering using single regular expression;
Closes gh-1173
* Many times faster and fewer CPU-hungry because of parsing with `maxlines=1`, so without line buffering (scrolling of the buffer-window).
Combination of tags `<F-MLFID>` and `<F-NOFAIL>` can be used now to process multi-line logs using single-line expressions:
  - tag `<F-MLFID>`: used to identify resp. store failure info for groups of log-lines with the same identifier (e. g. combined failure-info for the same conn-id by `<F-MLFID>(?:conn-id)</F-MLFID>`, see sshd.conf for example)
  - tag `<F-NOFAIL>`: used as mark for no-failure (helper to accumulate common failure-info, e. g. from lines that contain IP-address);
* Several filters optimized with pre-filtering using new option `prefregex`, and multiline filter using `<F-MLFID>` + `<F-NOFAIL>` combination;
* [filter.d/pam-generic.conf]: grave fix injection on user name to host fixed;
* Exposes filter group captures in actions (non-recursive interpolation of tags `<F-...>`);
Closes gh-1110
* Some filters extended with user name (can be used in gh-1243 to distinguish IP and user, resp. to remove after success login the user-related failures only);
* Safer, more stable and faster replaceTag interpolation (switched from cycle over all tags to re.sub with callable)
* substituteRecursiveTags optimization + moved in helpers facilities (because currently used commonly in server and in client)
* section "Init" no more needed (can be removed later from all action/filter definitions), currently excepting conditional (?family=...), because only expected parameters and section `[init]` supplied to the server from actionreader.
* provides new tag `<ip-rev>` for PTR reversed representation of IP address
* action.d/complain.conf fixed using new tag `<ip-rev>`
Closes gh-1685

Closes gh-1625


sebres:
I've ported many features from my private branches, otherwise doing the further development was always needlessly complicated to me.

If no objections follow, I'll merge it next week.

Small info by the way about this branch:
- ca. 25% faster in banning/unbanning;
- ca. 50% faster and fewer CPU-hungry in failure search by multi-line filters (using `prefregex` resp. combination of tags `<F-MLFID>` and `<F-NOFAIL>`) and up to 20% without it (old multiline syntax);
- no more hits missed in failure-recognizing on busy resp. deeply loaded services by multi-line filters (previously if too many log entries, multiline expression does not find failure, because second line exceeds the `maxlines` value, so out of scanning buffer)

PR #1709, opened 2017-03-10
More specific configuration file to FreeBSD OS and ipfw

njrizzo:
HI,

I wrote this to more specific on FreeBSD OS and IPFW
I created some new variables, because on my system 
when I ran this by first time, create some problems, like
putting a new rule after the last deny rule.

I suggest this variables names to add more functionalities

rulenumber    - rule number under all ban IPs will be
nic                  - nic name on no default  network interface
proto               - to choice a better protocol (tcp, udp or both, or others)

TIA

Nilton J. Rizzo

### DIFF
--- a/config/action.d/freebsd-ipfw.conf
+++ b/config/action.d/freebsd-ipfw.conf
@@ -1,0 +1,99 @@
+# Fail2Ban configuration file
+#
+# This file was created to provide ipfw with a configuration for FreeBSD
+# as flexible as possible
+# Author:	Nilton J. Rizzo
+#
+#
+
+[Definition]
+
+# Option:  actionstart
+# Notes.:  command executed once at the start of Fail2Ban.
+# Values:  CMD
+#
+actionstart = 
+
+
+# Option:  actionstop
+# Notes.:  command executed once at the end of Fail2Ban
+# Values:  CMD
+#
+actionstop =  ipfw delete <rulenumber>
+
+
+# Option:  actioncheck
+# Notes.:  command executed once before each actionban command
+# Values:  CMD
+#
+actioncheck = 
+
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+#actionban = ipfw add <rulenumber> <blocktype> <proto> from <ip> to <localhost> dst-port <port> in via <nic> 
+actionban = ipfw add <rulenumber> <blocktype> <proto> from <ip> to <localhost> <port>
+
+
+# Option:  actionunban
+# Notes.:  command executed when unbanning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionunban = ipfw delete `ipfw list | grep -i "[^0-9]<ip>[^0-9]" | awk '{print $1;}'`
+
+[Init]
+
+# Option:  port
+# Notes.:  specifies port to monitor
+# Values:  [ NUM | STRING ]
+#
+port = ssh
+
+# Option:  localhost
+# Notes.:  the local IP address of the network interface
+# Values:  IP
+#
+#localhost = me # first nic on system
+localhost = 127.0.0.1
+
+
+# Option:  blocktype
+# Notes.:  How to block the traffic. Use a action from man 5 ipfw
+#          Common values: deny, unreach port, reset
+# Values:  STRING
+#
+blocktype = deny
+
+
+# Option:  rulenumber
+# Notes.:  Number of set of rules.  IPFW can manipulate some rules on the same number rule
+#          Common values: 10000
+# Values:  NUM
+#
+rulenumber = 10000
+
+
+
+# Option:  proto
+# Notes.:  Name of protocoll under rule
+#          Common values: tcp, udp, all, ip
+#				see ipfw man page
+# Values:  STRING
+#
+proto = ip   
+
+
+
+# Option:  nic
+# Notes.:  Name of network inteface, see /etc/rc.conf
+#          Common values: em0, re0, ....
+#				see /usr/src/sys/{ARCH}/conf/LINT for all nic names
+# Values:  STRING
+#
+nic = 
--- a/config/action.d/freebsd-ipfw.conf
+++ b/config/action.d/freebsd-ipfw.conf
@@ -19,7 +19,10 @@ actionstart =
 # Notes.:  command executed once at the end of Fail2Ban
 # Values:  CMD
 #
-actionstop =  ipfw delete <rulenumber>
+#actionstop =  ipfw delete <rulenumber>
+actionstop =  a=$((<rulenumber>+<numberofrules>)); e=`ipfw show <rulenumber>-$a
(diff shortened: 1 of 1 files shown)


config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
+# Fail2Ban configuration file
+#
+# This file was created to provide ipfw with a configuration for FreeBSD
+# as flexible as possible
+# Author:	Nilton J. Rizzo
+#
```
sebres: May be some usage example would be good here?
E. g. something like:
```
# Usage example for `jail.local`:
#    banaction = freebsd-ipfw[proto=%(protocol)s, port=%(port)s, rulenumber=10]
```

sebres:
BTW why a new config file?
Additionally can you take a look in #1690. Possibly we can combine this to the single file.

config/action.d/freebsd-ipfw.conf
njrizzo: Yes, it's seem okay

njrizzo:
I'm think that the #1690 is too complex more than necessary
All FreeBSD administrator compile into kernel a ipfw module or
put it in loader.local.conf.  Not is necessary too many complexity

IMHO

config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
...
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+#actionban = ipfw add <rulenumber> <blocktype> <proto> from <ip> to <localhost> dst-port <port> in via <nic> 
+actionban = ipfw add <rulenumber> <blocktype> <proto> from <ip> to <localhost> <port>
+
```
St-Ranger: You did not do any check if there were any pre-existing ipfw rules at this number.
As a result, when you stop or reload fail2ban, your `actionstop` will delete the pre-existing ipfw rule.

You can check how this is implemented in the existing action.d/bsd-ipfw.conf
While the script in there might look a bit convoluted, it works well. In your case, it will require a very complicated logic hook to verify that there were no pre-existing firewall rules, because you are creating new rules and not setting any flags that would allow distinguishing whose rule(s) is sitting at this number.

config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
...
+actionunban = ipfw delete `ipfw list | grep -i "[^0-9]<ip>[^0-9]" | awk '{print $1;}'`
+
+[Init]
+
+# Option:  port
+# Notes.:  specifies port to monitor
+# Values:  [ NUM | STRING ]
+#
+port = ssh
+
```
St-Ranger: Why would you block a particular port and not all of them? If it's a hostile IP, block it completely, on all ports (and all interfaces).
Besides, this would require passing the port parameter for each action. What if a jail configured for the connects on more than one port (e.g. 25, 587, 465)? That would make it impossible (or require some awkward multiple jails).

config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
...
+# Option:  localhost
+# Notes.:  the local IP address of the network interface
+# Values:  IP
+#
+#localhost = me # first nic on system
+localhost = 127.0.0.1
+
```
St-Ranger: Have you tested this on your system? 
I suspect your firewall rules will not work.
`ipfw add <rulenumber> <blocktype> <proto> from <ip> to <localhost> <port>` from `actionban` will result in
`10000           deny ip from <123.456.789.0> to 127.0.0.1 22`
No external host will try to reach you at 127.0.0.1

I see that you have `localhost = me` commented out. (You uncommented it in the subsequent change.)
But even if the host has multiple IPs why would someone need to fiddle with only one IP, instead of blocking all of them (see my comment below regarding the interface name).

config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
...
+blocktype = deny
+
+
+# Option:  rulenumber
+# Notes.:  Number of set of rules.  IPFW can manipulate some rules on the same number rule
+#          Common values: 10000
+# Values:  NUM
+#
+rulenumber = 10000
+
```
St-Ranger: This number is too high! (and is probably **after** the "allow" rules in the standard FreeBSD's /etc/rc.firewall.
You want to have ban rules to be **before** any "allow" rules (except the rule 100, which covers loopback: 
`100      allow ip from any to any via lo0` )

config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
...
+
+# Option:  proto
+# Notes.:  Name of protocoll under rule
+#          Common values: tcp, udp, all, ip
+#				see ipfw man page
+# Values:  STRING
+#
+proto = ip   
+
```
St-Ranger: Why would you want to have anything but `ip` here?
Also, why is the option name abbreviated from the full word "protocol". It is not too long, but it is unambiguous.

config/action.d/freebsd-ipfw.conf
```diff
@@ -0,0 +1,99 @@
...
+
+
+# Option:  nic
+# Notes.:  Name of network inteface, see /etc/rc.conf
+#          Common values: em0, re0, ....
+#				see /usr/src/sys/{ARCH}/conf/LINT for all nic names
+# Values:  STRING
+#
+nic =
```
St-Ranger: Why would you want to block only one interface and not _all_ of them?
Can you provide a realistic use case where this would be needed?

So far, I cannot come up with any **realistic** scenario where this would be needed.

St-Ranger:
1. The proposed action file (`action.d/freebsd-ipfw.conf`) duplicates the function of  the esiting `action.d/bsd-ipfw.conf`.  (see: https://github.com/fail2ban/fail2ban/blob/0.10/config/action.d/bsd-ipfw.conf )
2. While it introduces new parameters, no reason is provided why they are necessary. While I see usefulness of some of them, the rest are not obvious. It seems to me that the parameters here are just for the sake of flexibility of configuration that nobody would really need.
3. Some checks (e.g. for the existence of the rule, to avoid the rule-number collision) that are performed in the existing bsd-ipfw.conf are not performed here, potentially creating some unpredictable behavior.
4. The existing action (bsd-ipfw) creates a single ipfw rule with a table. It is efficient and easy to manage. In contrast, the action proposed here suggests creation of a potentially _huge_ number of rules (all under the same number).
This is unnecessary and, to the best of my knowledge, reduces performance efficiency of the firewall.
(There are different sources demonstrating that table is more efficient that a linear set of rules.
The first that I found in google was this [one](https://lists.freebsd.org/pipermail/freebsd-ipfw/2004-August/001347.html)).
5. I highly suspect that this is not carefully tested by the author. I expect it not working properly, as described in the review comments.

I would recommend that you first look at the already existing `action.d/bsd-ipfw.conf`, which is working very well.
If something is needed to be improved there, you can build on top of that. Or, you have a compelling reason why it has to be re-written from scratch, - you need to explain that.

And just in case, - there is nothing complex in #1690. It is a relatively simple change to the existing `action.d/bsd-ipfw.conf`, - to allow changing the rule number used for the table.

PR #1716, opened 2017-03-13
Prohibit recursive replacement of action info (calling map)

sebres:
* [Important] Prohibit replacement of recursive "tags" in the action info resp. calling map (very bad idea to do this):
  - the calling map contains normally dynamic values only (no recursive tags);
  - recursive replacement can be vulnerable, because can contain foreign (user) input captured from log (will be replaced in the shell arguments);
* [BF] prevents always converting of calling map items in replaceTag (without direct access of item):
substituteRecursiveTags: ignore replacing callable items from calling map - should be converted on demand only (by get)

### DIFF
--- a/fail2ban/client/configreader.py
+++ b/fail2ban/client/configreader.py
@@ -29,22 +29,10 @@ import os
 from ConfigParser import NoOptionError, NoSectionError
 
 from .configparserinc import sys, SafeConfigParserWithIncludes, logLevel
-from ..helpers import getLogger, substituteRecursiveTags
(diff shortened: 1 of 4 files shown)


sebres:
Better to review without space comparison here: https://github.com/fail2ban/fail2ban/pull/1716/files?w=1

sebres:
Because of possible grave vulnerability (theoretical injection on tag `matches`, that can be supplied to the action) the back-porting to v.0.9 is necessary.

PR #1726, opened 2017-03-20
0.10 grave fix escape tags

sebres:
* fixed grave vulnerability by wrong escape of tags by executing of shell actions:
  - now all dynamic data (foreign input) will be escaped (if needed)
  - tag escaping implemented via variables mapped to the parameters of the command line:
`action-script ...<matches>...` will be replaced into:
`'f2bV_matches=$0 ... action-script ...$f2bV_matches...' 'matches-value1\n...\nmatches-valueN'`
* bug fix in the config readers: mixing with the init section should affect only own init options (from init section only bypass default section);

### DIFF
--- a/fail2ban/client/configparserinc.py
+++ b/fail2ban/client/configparserinc.py
@@ -200,13 +200,16 @@ after = 1.conf
 	def get_sections(self):
 		return self._sections
 
-	def options(self, section, onlyOwn=False):
-		"""Return a list of option names for the given section name."""
+	def options(self, section, withDefault=True):
+		"""Return a list of option names for the given section name.
+
+		Parameter `withDefault` controls the include of names from section `[DEFAULT]`
+		"""
 		try:
 			opts = self._sections[section]
 		except KeyError:
 			raise NoSectionError(section)
-		if not onlyOwn:
+		if withDefault:
 			# mix it with defaults:
 			return set(opts.keys()) | set(self._defaults)
 		# only own option names:
--- a/fail2ban/client/configparserinc.py
+++ b/fail2ban/client/configparserinc.py
@@ -32,7 +32,7 @@ from ..helpers import getLogger
 if sys.version_info >= (3,2):
 
 	# SafeConfigParser deprecated from Python 3.2 (renamed to ConfigParser)
-	from configparser import ConfigParser as SafeConfigParser, \
+	from configparser import ConfigParser as SafeConfigParser, NoSectionError, \
 		BasicInterpolation
 
 	# And interpolation of __name__ was simply removed, thus we need to
@@ -60,7 +60,7 @@ if sys.version_info >= (3,2):
 				parser, option, accum, rest, section, map, depth)
 
 else: # pragma: no cover
-	from ConfigParser import SafeConfigParser
+	from ConfigParser import SafeConfigParser, NoSectionError
 
 # Gets the instance of the logger.
 logSys = getLogger(__name__)
@@ -199,6 +199,18 @@ after = 1.conf
 
 	def get_sections(self):
 		return self._sections
+
+	def options(self, section, onlyOwn=False):
+		"""Return a list of option names for the given section name."""
+		try:
+			opts = self._sections[section]
+		except KeyError:
+			raise NoSectionError(section)
+		if not onlyOwn:
+			# mix it with defaults:
+			return set(opts.keys()) | set(self._defaults)
+		# only own option names:
+		return opts.keys()
 
 	def read(self, filenames, get_includes=True):
 		if not isinstance(filenames, list):
--- a/fail2ban/server/utils.py
+++ b/fail2ban/server/utils.py
@@ -118,6 +118,15 @@ class Utils():
 
 	@staticmethod
 	def buildShellCmd(realCmd, varsDict):
+		"""Generates new shell command as array, contains map as variables to
+		arguments statement (varsStat), the command (realCmd) used this variables and
+		the list of the arguments, mapped from varsDict
+
+		Example:
+			buildShellCmd('echo "V2: $v2, V1: $v1"', {"v1": "val 1", "v2": "val 2", "vUnused": "unused var"})
+		returns:
+			['v1=$0 v2=$1 vUnused=$2 \necho "V2: $v2, V1: $v1"', 'val 1', 'val 2', 'unused var']
+		"""
 		# build map as array of vars and command line array:
 		varsStat = ""
 		if not isinstance(realCmd, list):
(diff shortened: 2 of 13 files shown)


sebres:
Because of theoretical vulnerability in 0.9 also should be partially back-ported into 0.9 branch.

fail2ban/server/utils.py
```diff
@@ -116,7 +117,22 @@ def setFBlockMode(fhandle, value):
...
+	def buildShellCmd(realCmd, varsDict):
+		# build map as array of vars and command line array:
+		varsStat = ""
+		if not isinstance(realCmd, list):
+			realCmd = [realCmd]
+		i = len(realCmd)-1
+		for k, v in varsDict.iteritems():
+			varsStat += "%s=$%s " % (k, i)
+			realCmd.append(v)
+			i += 1
```
yarikoptic: did you want to use Python's enumerate?
```python
for i, (k, v) in enumerate(varsDict.iteritems()):
	varsStat += "%s=$%s " % (k, i+1)
	realCmd.append(v)
```

fail2ban/server/utils.py
```diff
@@ -116,7 +117,22 @@ def setFBlockMode(fhandle, value):
 		return flags
 
 	@staticmethod
-	def executeCmd(realCmd, timeout=60, shell=True, output=False, tout_kill_tree=True, success_codes=(0,)):
+	def buildShellCmd(realCmd, varsDict):
```
yarikoptic: what is the purpose of this function? (seems to be "To append to the command list of environment variables from varsDict" or smth like that?)
also if we take the point of "order of items in dict is not guaranteed" in a wider sense of "order might differ between runs of `.*items`" this wouldn't work -- might better first to sort keys or smth like that to get a deterministic order

fail2ban/server/utils.py
```diff
@@ -146,10 +164,18 @@ def executeCmd(realCmd, timeout=60, shell=True, output=False, tout_kill_tree=Tru
...
 		stdout = stderr = None
 		retcode = None
-		popen = None
+		popen = env = None
+		if varsDict:
+			if shell:
+				# build map as array of vars and command line array:
+				realCmd = Utils.buildShellCmd(realCmd, varsDict)
+			else: # pragma: no cover - currently unused
+				env = _merge_dicts(os.environ, varsDict)
```
yarikoptic: - what if there is an overlap? 
should may be all those get some unique prefix e.g. `_F2B_`? (yet to grasp the whole idea here)

- could those vars contain any symbol (e.g. `/`) not appropriate for env variables and thus should be sanitized first?

fail2ban/server/utils.py
```diff
@@ -146,10 +164,18 @@ def executeCmd(realCmd, timeout=60, shell=True, output=False, tout_kill_tree=Tru
...
+		realCmdId = id(realCmd)
+		outCmd = lambda level: logSys.log(level, "%x -- exec: %s", realCmdId, realCmd)
```
yarikoptic: please choose a bit more "proper" name -- e.g. "logCmd".  I was reading code below and got myself asking "what does it output? where to?

fail2ban/tests/actiontestcase.py
```diff
@@ -389,6 +389,51 @@ def testExecuteActionStartEmpty(self):
...
 
+	def testExecuteWithVars(self):
+		self.assertTrue(self.__action.executeCmd(
+			r'''printf %b "foreign input:\n'''
+			r''' -- $f2bV_A --\n'''
+			r''' -- $f2bV_B --\n'''
+			r''' -- $(echo -n $f2bV_C) --''' # echo just replaces \n to test it as single line
+			r'''"''', 
+			varsDict={
+			'f2bV_A': 'I\'m a hacker; && $(echo $f2bV_B)',
```
yarikoptic: shouldn't we just sanitize all the input we are getting from logs right away and replace any possibly dangerous symbol with some placeholder? e.g. `$` -> `+DOLLAR+`, ``````` -> `+BACKTICK+` etc so they could be used in regexps etc and also safely passed into the commands?

fail2ban/server/utils.py
sebres: both made outside, see https://github.com/fail2ban/fail2ban/pull/1726/files#diff-46f1e4875877693c8114acab1b699dddR576

fail2ban/tests/actiontestcase.py
sebres: This was my first thought, but the goal is:
- to rewrite escape tomorrow with "compiled" version, so replace only once and supply always as parameters dictionary (in this case no matter what the "hacker" will provide in the user input).
- myself can better analyze it if not replaced with placeholder;
- in my new ban-time-incr version I've introduced a **factor** for banning (so e. g. we can analyze and increase the factor to ban faster and longer if such unescaped chars present); Of course it is possible with the counting of the placeholder, but it make this confuse-able, bad readable and more complicated by searching for possible complex injections, etc.

IMHO as option it is imaginable.

fail2ban/client/configparserinc.py
```diff
@@ -200,6 +200,18 @@ def get_defaults(self):
 	def get_sections(self):
 		return self._sections
 
+	def options(self, section, onlyOwn=False):
+		"""Return a list of option names for the given section name."""
```
sebres: I believe that by the function that "Return a list of option names for the given section name" the parameter onlyOwn is enough obvious and speak for itself...

fail2ban/server/utils.py
sebres: as for order of items: no matter as long as variable statement (varsStat) and arguments list build in the same function (direct here in buildShellCmd).

fail2ban/client/configparserinc.py
yarikoptic: it might be obvious to you (and today) but it was not obvious to me on what kind of 'ownership' we are talking about ... also "only own" sounds awkward in english afaik... "only owned" would have been better but anyways imho it is not appropriate (once again imho) since ownership is a vague non-described idea here

fail2ban/server/utils.py
yarikoptic: I guess I am yet to comprehend fully... btw -- what happens when a value has a space in it or some other "interesting character"?

sebres: ??
No matter what for characters the value has... It will be appended as argument in the command-tuple (array).
E. g. from 
```python
'echo "failure of <f-user>: <matches>"'
``` 
will be to:
```python
['f2bV_matches=$0 f2bV_matches=$1\necho "failure of $f2bV_f-user: f2bV_matches"', 'unknown user', 'this is the string from tag matches']
```

sebres:
coveralls acts up again 😠 (seems to be really buggy today)

PR #1733, opened 2017-03-24
0.10 repl skiplines

sebres:
* Normalizes replacement of `<SKIPLINES>` (moved to _resolveHostTag, so will be replaced together with another tags);
Regex will be compiled as MULTILINE only if needed (buffering with `maxlines` > 1), that enables:
  - to improve performance by the single line parsing;
  - make regex more precise (because distinguish between anchors `^`/`$` for the begin/end of string and the new-line character '\n', e. g. if coming from filters (like systemd journal) that allow the parsing of log-entries contain new-line chars (as single entry);
  - if multiline regex however expected (by single-line parsing without buffering) - prefix `(?m)` could be used in regex to enable it;
* Removed warning "Mutliline regex set for jail ... but maxlines not greater than 1", because can be expected situation now:
non multi-line entry from systemd-filter containing new-lines (that should be ignored by anchors resp. entry parsed as single string);
* fail2ban-regex: fixed matched output by multi-line (buffered) parsing
* fail2ban-regex: support for multi-line debuggex URL implemented
Closes gh-422

### DIFF
--- a/fail2ban/client/fail2banregex.py
+++ b/fail2ban/client/fail2banregex.py
@@ -55,11 +55,14 @@ from ..helpers import str2LogLevel, getVerbosityFormat, FormatterWithTraceBack, 
 # Gets the instance of the logger.
 logSys = getLogger("fail2ban")
 
-def debuggexURL(sample, regex, useDns="yes"):
-	q = urllib.urlencode({ 're': Regex._resolveHostTag(regex, useDns=useDns),
-							'str': sample,
-							'flavor': 'python' })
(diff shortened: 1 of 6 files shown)


sebres:
Actually I don't recommend to use pure multi-line parsing (with buffering using `maxlines` > 1), so IMHO obsolete if can be implemented using more effective `prefregex`, see #1698.
But nevertheless fixed... 
And this PR has benefits to the single-line parsing also (especially using systemd-backend).

Issue #1738, opened 2017-03-28
python3 setup.py build in branch 0.10


gracinet:
### Environment:

Clean checkout of 0.10 branch:

```
~/divers/fail2ban $ git remote -v
origin	https://github.com/fail2ban/fail2ban (fetch)
origin	https://github.com/fail2ban/fail2ban (push)
~/divers/fail2ban $ git checkout origin/0.10
HEAD is now at c8249535... Update mysqld-auth.conf (#1725)
~/divers/fail2ban $ git status            
HEAD detached at origin/0.10
nothing to commit, working tree clean
```

### The issue:

The derivation introduced in `setup.py` in order not to run 2to3 if the `install` subcommand gets launched from tests (according to comments therein) successfully preserves the normal invocation of `2to3` in `python3 setup.py install`, but breaks `python3 setup.py build`. 

#### Steps to reproduce

```
~/divers/fail2ban $ python3 setup.py build
running build
running build_py
running build_scripts
```

#### Expected behavior

`2to3` runs as part of the build under Python 3. This can be seen, e.g, by using `iteritems()` as a marker. After `2to3`, it appears only in one file (in comments actually)::

```
~/divers/fail2ban $ grep -rl --include '*.py' iteritems build        
build/lib/fail2ban/server/ticket.py
```

#### Observed behavior
For quick display here, I'm using `iteritems()` as a marker that `2to3` didn't run

```
~/divers/fail2ban $ grep -rl --include '*.py' iteritems build | wc -l
18
```

#### Any additional information

``python3 setup.py build`` is used in particular during Debian packaging (`gpg buildpackage` in the `debian` branch). 

I'm trying to build Debian packages for fail2ban 0.10 for my own purposes so that it'll be more convenient for me to deploy on a bunch of servers (and will be transparently upgraded to 0.10.0 final if it lands in backports at some point). The major feature I'm looking forward to, compared to the 0.9 series, is IPv6 support.

Since in that context I'm not interested in building with Python 2, I tried bypassing the exception to `2to3` altogether, and it worked::

```
~/divers/fail2ban $ git checkout  0.10-quickfix-setup 
Switched to branch '0.10-quickfix-setup'
~/divers/fail2ban $ git diff origin/0.10 -- setup.py
diff --git a/setup.py b/setup.py
index 5754db49..eba0ecb5 100755
--- a/setup.py
+++ b/setup.py
@@ -168,7 +168,7 @@ setup(
        license = "GPL",
[...]

PR #1739, opened 2017-03-28
Fixes test_smtp connects to wrong inet (if listening on ::1 instead of 127.0.0.1)

gracinet:
Hi, while running the tests with --no-network, I got errors for `test_smtp`, it seems that it was missing the bit to be skipped under these conditions (although one may find skipping it entirely a bit too harsh)

- [x] **CONSIDER adding a unit test** if your PR resolves an issue (this is just about tests)
- [ ] **LIST ISSUES** this PR resolves (no user issue related, I can make one if you like)
- [x] **MAKE SURE** this PR doesn't break existing tests (checked with py2/py3 with -fng)
      within `fail2ban/tests/files/logs/X` file (there's none)

### DIFF
--- a/fail2ban/tests/action_d/test_smtp.py
+++ b/fail2ban/tests/action_d/test_smtp.py
@@ -67,7 +67,7 @@ class SMTPActionTest(unittest.TestCase):
 		port = self.smtpd.socket.getsockname()[1]
 
 		self.action = customActionModule.Action(
-			self.jail, "test", host="127.0.0.1:%i" % port)
+			self.jail, "test", host="localhost:%i" % port)
 
 		## because of bug in loop (see loop in asyncserver.py) use it's loop instead of asyncore.loop:
 		self._active = True


sebres:
I'm not sure that this would be good idea, because this test case uses mockup (listening on loopback interface, look at TestSMTPServer('localhost')), and thus should normally work without network also.

Sporadic? Should we rewrite it to *nix-file-socket (no idea that is possible)?

gracinet:
Thanks for the comment !

Not really sporadic, but it's clear: this is certainly due to the fact that I don't have an SMTP server binding to localhost:25 on my dev rig by default (because with all the work I'm doing reproducing bugs for various real-life applications it leads too easily to sending lots of unwanted emails and very embarrassing situations).

So what I didn't get is that 'no network' still encompasses the use of localhost.

sebres:
>  I don't have an SMTP server binding to localhost:25

Me too...
You did not understood. As already said, we've there own listener (mockup test smtp-server, that creates own listener on some dynamic port).
Can you please provide the call-stack by error, if not visible start this test-case with heavydebug:
```bash
bin/fail2ban-testcases -f -vv -l 10 test_smtp
```

gracinet:
@sebres ah yes sorry, I read both the code and your comment too fast. 
Moreover it appears that the issue happens with Python 3 only.
At least, I didn't reproduce with Python 2, but the test fails systematically for me with Python 3 (Debian 9's python3.5). Apparently not for you, according to you comments on #1738.


Here's the traceback, it's identical for the 4 tests in error.

```
ERROR: testStop (fail2ban.tests.action_d.test_smtp.SMTPActionTest)
  File "./fail2ban/tests/action_d/test_smtp.py", line 101, in testStop
    self._exec_and_wait(self.action.stop)
  File "./fail2ban/tests/action_d/test_smtp.py", line 89, in _exec_and_wait
    doaction()
  File "config/action.d/smtp.py", line 206, in stop
    messages['stop'] % self.message_values)
  File "config/action.d/smtp.py", line 161, in _sendMessage
    self.host, *smtp.connect(self.host))
  File "/usr/lib/python3.5/smtplib.py", line 335, in connect
    self.sock = self._get_socket(host, port, self.timeout)
  File "/usr/lib/python3.5/smtplib.py", line 306, in _get_socket
    self.source_address)
  File "/usr/lib/python3.5/socket.py", line 712, in create_connection
    raise err
  File "/usr/lib/python3.5/socket.py", line 703, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused

```
That really looks like a client-side error to me, but that's what would happen if the mock binding didn't fully succeed without raising an error, I suppose.

If you don't mind, I'd like to investigate a bit more and propose a better solution (don't know if Unix Domain socket would be possible either btw), especially because my goal is not to waste your time :-)

gracinet:
was suspecting something along those lines… The mock servers listens on ::1 while the client attempts to connect on 127.0.0.1

Here's the evidence:

I put a breakpoint in ``action.d/smtp.py``:

```
(Pdb) l
156  			msg['Date'] = formatdate()
157  	
158  			smtp = smtplib.SMTP()
159  			try:
160  				import pdb; pdb.set_trace()
161  ->				self._logSys.debug("Connected to SMTP '%s', response: %i: %s",
162  					self.host, *smtp.connect(self.host))
163  				if self.user and self.password:
164  					smtp.login(self.user, self.password)
165  				failed_recipients = smtp.sendmail(
166  					self.fromaddr, self.toaddr.split(", "), msg.as_string())
(Pdb) self.host
'127.0.0.1:46247'
```
and here's what netstat has to say:

```
~ $ netstat -ltn | grep 46247 
tcp6       0      0 ::1:46247               :::*                    LISTEN   
```

Double-check that the testing mock answers on IPv6 and not on IPv4:

```
~ $ telnet ::1 46247
Trying ::1...
Connected to ::1.
Escape character is '^]'.
220 bench.test Python SMTP proxy version 0.3
^CConnection closed by foreign host.
~ $ telnet 127.0.0.1 46247
Trying 127.0.0.1...
telnet: Unable to connect to remote host: Connection refused
~ $ telnet ::1 46247 
Trying ::1...
Connected to ::1.
Escape character is '^]'.
220 bench.test Python SMTP proxy version 0.3
^]
telnet> Connection closed.
```

fail2ban/tests/action_d/test_smtp.py
```diff
@@ -67,7 +67,7 @@ def setUp(self):
 		port = self.smtpd.socket.getsockname()[1]
 
 		self.action = customActionModule.Action(
-			self.jail, "test", host="127.0.0.1:%i" % port)
+			self.jail, "test", host="localhost:%i" % port)
```
sebres: 👍

sebres:
Merged. Thx!

PR #1740, opened 2017-03-28
strptime.py: small code review and performance optimization

sebres:
- get some properties on demand, etc.
- important by many parasitic entries with date-time (resp. large log files traffic) 
- and by starting (jail-init with seektotime)


sebres:
For better comparision: [#1740/files?w=1](https://github.com/fail2ban/fail2ban/pull/1740/files?w=1)

PR #1746, opened 2017-04-07
haproxy-http-auth IPv6 (Closes #1745)

### LINKED ISSUE #1745: haproxy-http-auth: wrong IPv6 banned address
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.10.0a1 and current 0.10 branch
- OS, including release name/version: 
- [ ] Fail2Ban installed via OS/distribution mechanisms

### The issue:

Both in configuation (bind) and its standard logs, HAProxy always appends the port to the IP address, using a colon as separator. Most other apps, in particular URLs, avoid the ambiguity created by the colon by enclosing IPv6 addresses into square brackets

In particular, in logs, a client IPv4 socket looks like `10.0.0.1:58340`. In this case the failregex for haproxy-http-auth extracts the IP correctly as 10.0.0.1

But a client IPv6 socket from address, say, `2001:db8::1` with client port `58340` will be interpreted incorrectly by the failregexp as originating from `2001:db8::1:5834`.

Most other applications and standards (e.g., URLs), avoid this problem by enclosing the IPv6 address part in square brackets. such as`[2001:db8:1.2::3.4]:80`, but HAProxy does not.

#### Steps to reproduce

* Enable the haproxy-http-auth jail
* Trigger a banning by getting enough 401 responses from haproxy over IPv6

#### Expected behavior

Correct IPv6 address is banned

#### Observed behavior

The first 4 digits of the port are mistaken for part of the IPv6 address

I've got a fix, finalizing the pull-request right now. This ticket for reference

Noticed this while preparing a custom filter similar to what is asked in #1307. Noticed the provided filter for failed auths would be affected (took it as an example to build upon).

#### Relevant lines from monitored log files in question:

Here's an excerpt from HAProxy httplog. This is not a 401 (more akin to question #1307) , but it shows the thing :

```
Apr  7 06:53:48 myhostname haproxy[9487]: 2002:52d1:1e9e::52d1:1e9e:56834 [07/Apr/2017:06:53:48.107] ft_http-in ft_http-in/<NOSRV> 0/-1/-1/-1/0 301 94 - - LR-- 18/4/0/0/0 0/0 "GET /wp-login.php HTTP/1.1"

```
Sample taken on a Debian system. I believe it'd be the same across all distributions.

gracinet:
Here's the promised fix for #1745.

Two more remarks about this filter.

- I'm not sure the detailed comment about the meaning of `<HOST>` is very usefull. These things tend to go out-of-sync and besides, other filters don't have it (it's easy to see it's just f2b standard stuff)
- One may want also to broaden failregexp : `<NOSRV>` is HAProxy telling it didn't pass it through a backend. One my want also to track the 401 responses that originated from a backend.

Checklist: 

- [x] **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement against 0.9.x series, choose `master` branch
- [x ] **CONSIDER adding a unit test** if your PR resolves an issue
- [x ] **LIST ISSUES** this PR resolves
- [x ] **MAKE SURE** this PR doesn't break existing tests
- [x ] **KEEP PR small** so it could be easily reviewed.
- [x ] **AVOID** making unnecessary stylistic changes in unrelated code

Closes gh-1745


sebres:
Looks good, an interim question: would be possible better to use a small sub-regex `\s+` before catch-all, e. g.:
```diff
-failregex = ^%(__prefix_line)s<HOST>:\d+.*<NOSRV> -1/-1/-1/-1/\+*\d* 401
+failregex = ^%(__prefix_line)s<HOST>(?::\d+)?\s+.*<NOSRV> -1/-1/-1/-1/\+*\d* 401
```
That makes it additionally backwards-compatible, where the port part would be optional (no idea whether it is necessary).

BTW Very weird way to write a port into log, I mean without any kind of "escape" as something like `[ipv6]:port`.

gracinet:
Thanks !

If I understand your `(?::\d+)?` correctly, it works because, although being optional, it's more greedy than the regexps behind `<HOST>`. So it would accomodate these cases:

- IPv4:port
- IPv6:port
- bare IPv4

but still, it would'nt fit a bare IPv6 format, since it'd strip the last segment of some bare IPv6 addresses.

I'm not sure it's worth the small loss of readability. 

As far as I understand it, HAProxy's standard [log formats](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#8.2) (log, tcplog, httplog) are almost non customizable - if one wants something else, one has to go [full custom](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#8.2.4) and then of course, anything is possible, and I guess the local admin is supposed to adapt the regexp anyway.
 
It's true that `httplog` has the `clf` option, which is already rather different and does separate IP address and client port, but this format is not targetted by the failregexp anyway. This CLF format is provided as an example in the section about custom log formats.

Maybe it's just better to explicitely state in the filter conf file that this is for the `httplog` format ?

Also what do you think of broadening to the cases where the response actually came from a backend ? (dropping NOSRV, other numbers than -1 etc)

Should this be a separate issue/PR ?

sebres:
> but still, it would'nt fit a bare IPv6 format

It's impossible for current log-format (at least as long as the port can be 4 digits, otherwise `(?:\d{5})?`)...

Adding of sub-regex `\s+` is more precise anyway (whether with or without optional port)...

> Should this be a separate issue/PR ?

IMHO, not necessarily if anyway affected the same filter... You can extend original issue/PR...

gracinet:
Hi, seems I missed your last reply ! I'm fine with the altered regex anyway, ans thanks for the merge
I will follow-up with a broader one (based on some examples I've got on hand), but not sure I can do it before the week end / next week

PR #1749, opened 2017-04-11
Parameter not needed

petervanderdoes:
The parameter '-s' causes an error for a few actions as the <mailcmd> already has the parameter.


sebres:
Thanks, merge...

Issue #1554, opened 2016-09-22
pynotify still chosen as the backend even when backend specified to polling in jail.local


ashakunt:
f2b version: v0.10.0a2
operating system: Ubuntu 14.04, installed f2b using the tarball for 0.10 and used instructions from the wiki.
- [ ] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

This bug at least concerns: courier-smtp, postfix, postfix-rbl, sendmail-auth, sendmail-reject, courier-auth, postfix-sasl. 

My configuration file looks like this (in breaks, all relevant parts):

Default backend in jail.local is set to polling:

```
 83 # Note: if systemd backend is chosen as the default but you enable a jail
 84 #       for which logs are present only in its own log files, specify some other
 85 #       backend for that jail (e.g. polling) and provide empty value for
 86 #       journalmatch. See https://github.com/fail2ban/fail2ban/issues/959#issuecomment-74901200
 87 backend = polling
```

The above mentioned jails have the following configuration:

```
555 [courier-smtp]
556 enabled = true
557 port     = smtp,465,submission
558 logpath  = %(syslog_mail)s
559 #backend  = %(syslog_backend)s

562 [postfix]
563 enabled = true
564 port     = smtp,465,submission
565 logpath  = %(postfix_log)s
566 #backend  = %(postfix_backend)s

569 [postfix-rbl]
570 enabled = true
571 port     = smtp,465,submission
572 logpath  = %(postfix_log)s
573 #backend  = %(postfix_backend)s
574 maxretry = 1

577 [sendmail-auth]
578 enabled = true
579 port    = submission,465,smtp
580 logpath = %(syslog_mail)s
581 #backend = %(syslog_backend)s


584 [sendmail-reject]
585 enabled = true
586 port     = smtp,465,submission
587 logpath  = %(syslog_mail)s
588 #backend  = %(syslog_backend)s

643 [courier-auth]
644 enabled = true
645 port     = smtp,465,submission,imap3,imaps,pop3,pop3s
646 logpath  = %(syslog_mail)s
647 #backend  = %(syslog_backend)s

650 [postfix-sasl]
651 enabled = true
652 port     = smtp,465,submission,imap3,imaps,pop3,pop3s
653 # You might consider monitoring /var/log/mail.warn instead if you are
654 # running postfix since it would provide the same log lines at the
[...]

PR #1750, opened 2017-04-13
fixes default backend handling (as default used value of `known/backe…

sebres:
* fixes default backend handling (as default used value of `known/backend`, which can now be overridden in default section of jail.local);
* introduces fallback for `known/option`: interpolates missing `known/option` as `option` from default section

Closes issues with `default_backend` introduced in #1225, like #1372, #1554 etc.
Closes gh-1554

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -134,6 +134,13 @@ TODO: implementing of options resp. other tasks from PR #1346
         Hence `%z` currently match literal Z|UTC|GMT only (and offset-based), and `%Exz` - all zone 
 	abbreviations.
 * `filter.d/courier-auth.conf`: support failed logins with method only
+* Config reader's: introduced new syntax `%(section/option)s`, in opposite to extended interpolation of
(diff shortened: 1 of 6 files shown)


sebres:
Rebased to current 0.10 and branch is complete now.

Bonus:
* Config reader's: introduced new syntax `%(section/option)s`, in opposite to extended interpolation of python 3 `${section:option}` work with all supported python version in fail2ban and this syntax is like our another features like `%(known/option)s`, etc.;
* Variable `default_backend` switched to `%(default/backend)s`, so totally backwards compatible now, but now the setting of parameter `backend` in default section of `jail.local` can overwrite default backend also.
Thus both following examples are equivalent for all jails, which backend was not overridden in corresponding paths-distribution.conf:
```INI
[DEFAULT]
default_backend = systemd
```
```INI
[DEFAULT]
backend = systemd
```
In the future versions parameter `default_backend` can be removed (incompatibility, possibly some distributions affected).
* Test cases extended: test targeted section options "section/option" (default and cross sections options);

Issue #796, opened 2014-08-29
Fail2ban restriction and exclude logfiles?


JuSSt1:
Hello,

I want to configure my fail2ban to block brute force logins with Wordpress, but i need to exclude some logs when i turn on fail2ban, i have configured jail.conf like this:

logpath = /var/log/httpd/domains/*[!error].log

With this i exclude all logs that contains "error", but i need to do a second filter that contains "mail", so, if i configure jail.conf like this:

logpath = /var/log/httpd/domains/_[!error].log
                /var/log/httpd/domains/_[!mail].log

Fail2ban only loads domains with ".es" but not the rest,is not logic.

My question is, did i something wrong? or how i need to configure jail.conf to make this exclude?

Kind regards.

hatifnatt:
Sorry for reviving this old topic, but I think it can be useful for anybody who came here in search of a solution "how to exclude some logs". Actually globs with negative part work fine. Sample files:
```bash
# ls -1 /tmp/test
a.error.log
a.log
b.error.log
b.log
```
Let's check in Python
```python
>>> import glob
>>> glob.glob('/tmp/test/*.log')
['/tmp/test/a.error.log', '/tmp/test/b.error.log', '/tmp/test/b.log', '/tmp/test/a.log']
>>> glob.glob('/tmp/test/*[!error].log')
['/tmp/test/b.log', '/tmp/test/a.log']
```
I think everything is  pretty clear from example above. All you need is something like `/path/to/logs/*[!exclude_this].log`

sebres:
Indeed, good point @hatifnatt. Thx.

tgoeg:
For anyone coming here in search for excludes:
The above (https://github.com/fail2ban/fail2ban/issues/796#issuecomment-423580283) is wrong, as the test case is incomplete. `[!error]` matches exactly _one_ character that is not `e`,`r` or `o` (so yes, it matches `a` and `b`).
See https://docs.python.org/3/library/fnmatch.html#module-fnmatch for details.

This feature is planned but not implemented yet. See #1379 for the current state of implementation.

Issue #1379, opened 2016-04-01
logpath issue: move logic "collecting of logfiles" from client to server resp. filter ...


sebres:
Just a reminder for me (will too long provide a PR for this, waiting for 716-cs ... and #1346).

Should fix imho great problem with "collecting of logfiles" client side: current solution uses fixedly list of the log files, found by the moment of the fail2ban starting. 
Variable logfiles of some services (multiple domains, etc.) will not be taken into account, as long as not restart of the fail2ban.
- [X] move logic "collecting of logfiles" from client to server resp. filter ... (* pro rata realized in my own branch);
- [ ] support appearing of new log files (matched) - (for example search in X minutes for new log files, or via pyinotify callback, if supported);
- [ ] better support disappearing of the log files (log-rotate, service down, etc...), and reactivating of jail from idle if log files found again;
- [ ] add `logregex` as additionally extended "replacement" for `logpath`;
- [ ] add `logregex-ignore` (and `logpath-ignore`) for negative filter;

sebres:
should resolve #1407

sebres:
Several issues refer to `Have not found any log file`, in relation to "prevent to start fail2ban service" are partially fixed in #1622

eadthem:
What version was this merged in?  Specificity the 1407 (Related to log files having dates in the name).

Thanks
Eadthem

sebres:
> What version was this merged in?

Nowhere. It's an issue. I should still finish it. Just persistently no time...
And it is a pre-requirement for 1407, that as you can see is still open also.

szepeviktor:
@sebres Should Fail2ban accept multiline logpaths?

```ini
logpath  = %(apache_error_log)s
           /home/*/website/log/*error.log
```

sebres:
Fail2ban self does it... it looks rather as an issue of ConfigReader's and/or ExtractOptions (by supplying to the action "sendmail-geoip-lines").
Currently I don't know which versions are affected (0.9 is too old, so it could be fixed in-between).

The alternative for you which could work - switch to another action (e. g. that using `<matches>` info instead of the log-files).

szepeviktor:
Thank you.

sebres:
yw

> using `<matches>` info instead of the log-files

Oh, forgotten that we had a grave issue with not proper "escaped" `<matches>` - #1726 (fixed for >= 0.10), so be carefully with that in 0.9.

szepeviktor:
Thank you.

[...]

PR #1756, opened 2017-04-20
Add excludepath config option

pschultz:
Add a new jail option "excludepath". Files that match any logpath pattern and also any excludepath pattern
are ignored. We were looking for a way to exclude certain webserver logs. The "create a new directory with a bunch of symlinks" solution doesn't work well for us. We'd much rather have a blacklist than a whitelist.

```ini
[my-jail]
enabled = true
...
logpath = /var/log/nginx/*-access.log
excludepath = /var/log/nginx/api*
```

The PR does not break compatibility; only the client's jail reader is affected.

Mentioning #796 because it shows up on Google when searching "fail2ban exclude logfiles".

### DIFF
--- a/fail2ban/client/jailreader.py
+++ b/fail2ban/client/jailreader.py
@@ -24,6 +24,7 @@ __author__ = "Cyril Jaquier"
 __copyright__ = "Copyright (c) 2004 Cyril Jaquier"
 __license__ = "GPL"
 
+import fnmatch
 import glob
 import json
 import os.path
@@ -83,13 +84,21 @@ class JailReader(ConfigReader):
 			self.__opts and self.__opts.get("enabled", False))
 
 	@staticmethod
-	def _glob(path):
-		"""Given a path for glob return list of files to be passed to server.
+	def _glob(path, excludeGlobs):
+		"""Given a path for glob, and a list of exclude patterns, return list of files to be passed to server.
 
-		Dangling symlinks are warned about and not returned
+		Dangling symlinks are warned about and not returned.
 		"""
 		pathList = []
 		for p in glob.glob(path):
+			exclude = False
+			for g in excludeGlobs:
+				if fnmatch.fnmatch(p, g):
+					exclude = True
+					break
+			if exclude:
+				continue
+
 			if os.path.exists(p):
 				pathList.append(p)
 			else:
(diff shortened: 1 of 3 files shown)


sebres:
I've planned a redesign of the log-file handling (in #1379), so the whole logic round about wildcards should be moved to the server/jail-backend (allows dynamically "filtering" during the runtime, and not as now by the start only).
But I'll take a look here later to see how it can be useful together with planned arrangements.
At least as interim solution it would be good. Thanks!

fail2ban/client/jailreader.py
```diff
@@ -223,20 +235,10 @@ def convert(self, allow_no_files=False):
 		for opt, value in self.__opts.iteritems():
 			if opt == "logpath":
 				if self.__opts.get('backend', None).startswith("systemd"): continue
-				found_files = 0
-				for path in value.split("\n"):
-					path = path.rsplit(" ", 1)
-					path, tail = path if len(path) > 1 else (path[0], "head")
-					pathList = JailReader._glob(path)
```
sebres: Why not right here?
```python
if excludeGlobs is None:
  excludeGlobs = self.__opts.get("excludepath", "").split("\n")
pathList = JailReader._glob(path, excludeGlobs)
```

and of course above `excludeGlobs = None`

fail2ban/client/jailreader.py
```diff
@@ -261,6 +263,21 @@ def convert(self, allow_no_files=False):
 				stream.extend(action.convert())
 			else:
 				stream.append(action)
+
+		if len(pathGlobs) > 0:
```
sebres: see comment above...

pschultz:
> and not as now by the start only

Does that imply that newly created files are not picked up by fail2ban unless it is restarted?

pschultz:
I guess it can't pick up new files, since the server never sees the globs. That didn't occur to me yet. Quite unexpected I must say.

fail2ban/client/jailreader.py
```diff
@@ -83,13 +84,21 @@ def isEnabled(self):
 			self.__opts and self.__opts.get("enabled", False))
 
 	@staticmethod
-	def _glob(path):
-		"""Given a path for glob return list of files to be passed to server.
+	def _glob(path, excludeGlobs):
```
sebres: small change for backwards compatibility (+ optional, little performance fix by too many entries)
```diff
-	def _glob(path, excludeGlobs):
+	def _glob(path, excludeGlobs=[]):
...
		pathList = []
+		excludeGlobs = set(excludeGlobs)
		for p in glob.glob(path):
```

sebres:
> Quite unexpected I must say.

You're right, that's why I want to "fix" it.

PR #1764, opened 2017-05-05
Feature/postfix postscreen

petervanderdoes:
Add a filter for when Postfix does a RBL check during the postscreen process.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,7 @@ TODO: implementing of options resp. other tasks from PR #1346
 * New Actions:
 
 * New Filters:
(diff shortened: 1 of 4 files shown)


sebres:
Why it should be separate filter? I mean why not a part of postfix-rbl?
The only difference I see ATM are 4.7.1 vs 5.7.1 and possibly the service name (postfix/smtpd[pid] vs. postfix/postscreen[pid])...
If so, may be it should be combined to the postfix-rbl...

sebres:
Implemented in #1825.
Thanks!

PR #1770, opened 2017-05-08
don't call setLogTarget from setSyslogSocket when holding the lock

tjikkun:
If we do, setLogTarget will wait for the lock, but it will never be released


sebres:
Thx, indeed it looks like lock in lock. 
We should possibly cover this logging type.
Merge.

PR #1778, opened 2017-05-15
Fix pyinotify directory-based log-rotate

### LINKED ISSUE #1379: logpath issue: move logic "collecting of logfiles" from client to server resp. filter ...

### LINKED ISSUE #1769: ERROR Unable to get failures in
Hello. 
Periodically, an error occurs:

[~] tail -1 /logs/fail2ban.filter 
May  7 06:01:00 fail2ban.filter[14160]: ERROR Unable to get failures in /data/logs/linux/current

After the occurrence of this error, the processing of rules is terminated, until the fail2ban-client restarted. 

It seems that the filter tries to get the directory instead of the log file ('/data/logs/linux/current/sshd'), and why fail2ban does not try to open the file again and just stops the processing?

Thanks

***INFO***

[~] fail2ban-client -V
Fail2Ban v0.9.6

[~] cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 6.8 (Santiago)

[~] uname -a
Linux 2.6.32-642.el6.x86_64 #1 SMP Wed 

[~] fail2ban-client -d
['set', 'syslogsocket', 'auto']
['set', 'loglevel', 'INFO']
['set', 'logtarget', 'SYSLOG']
['set', 'dbfile', '/var/lib/fail2ban/fail2ban.sqlite3']
['set', 'dbpurgeage', 86400]
['add', 'sshd', 'auto']
['set', 'sshd', 'usedns', 'warn']
['set', 'sshd', 'addlogpath', '/data/logs/linux/current/sshd', 'head']
['set', 'sshd', 'maxretry', 8]
['set', 'sshd', 'addignoreip', '127.0.0.1/8']
['set', 'sshd', 'logencoding', 'auto']
['set', 'sshd', 'bantime', 1801]
['set', 'sshd', 'ignorecommand', '']
['set', 'sshd', 'findtime', 600]
['set', 'sshd', 'maxlines', '10']

sebres:
* Fixed detection of directory-based log-rotation of pyinotify backend.
If directory moved and the target is not watched path, so the monitoring of it could not be continued.
Now fixed with pending files await a monitoring if there (resp. its directories) appears again (respawn).

* Differentiate between watched directories and files (refreshing monitoring of files/dirs expected different flags for watcher).

* Switch from ThreadedNotifier to Notifier:
  - instance of Filter is already a thread;
  - avoid stop of pyinotify processing if an interim error occurs (and breaks main-loop, e. g. during multi-threaded processing by add/remove log-files)

This logic (pending restoration) can be used in observer-thread by the implementation to fix #1379

Closes #1769

### DIFF
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -895,7 +895,8 @@ class FileFilter(Filter):
 			# see http://python.org/dev/peps/pep-3151/
 			except IOError as e:
 				logSys.error("Unable to open %s", filename)
-				logSys.exception(e)
+				if e.errno != 2:
+					logSys.exception(e)
 				return False
 			except OSError as e: # pragma: no cover - requires race condition to tigger this
 				logSys.error("Error opening %s", filename)
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -895,7 +895,7 @@ class FileFilter(Filter):
 			# see http://python.org/dev/peps/pep-3151/
 			except IOError as e:
 				logSys.error("Unable to open %s", filename)
-				if e.errno != 2:
+				if e.errno != 2: # errno.ENOENT
 					logSys.exception(e)
 				return False
 			except OSError as e: # pragma: no cover - requires race condition to tigger this
--- a/fail2ban/server/filterpyinotify.py
+++ b/fail2ban/server/filterpyinotify.py
@@ -89,23 +89,12 @@ class FilterPyinotify(FileFilter):
 		logSys.log(7, "[%s] %sCallback for Event: %s", self.jailName, origin, event)
 		path = event.pathname
 		# check watching of this path:
-		isWF = isWD = False
-		if path in self.__watchDirs:
-			isWD = True
-		elif path in self.__watchFiles:
+		isWF = False
+		isWD = path in self.__watchDirs
+		if not isWD and path in self.__watchFiles:
 			isWF = True
-		# fix pyinotify behavior with '-unknown-path' (if target not watched also):
-		if (event.mask & pyinotify.IN_MOVE_SELF and 
-				path.endswith('-unknown-path') and not isWF and not isWD
-		):
-			path = path[:-len('-unknown-path')]
-			isWD = path in self.__watchDirs
 		assumeNoDir = False
 		if event.mask & ( pyinotify.IN_CREATE | pyinotify.IN_MOVED_TO ):
-			# refresh watched dir (may be expected):
-			if isWD:
-				self._refreshWatcher(path, isDir=True)
-				return
 			# skip directories altogether
 			if event.mask & pyinotify.IN_ISDIR:
 				logSys.debug("Ignoring creation of directory %s", path)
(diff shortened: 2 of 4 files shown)


sebres:
I've no interest to back-port this to 0.9 (because branches just too differently), if someone wants - welcome with PR.

fail2ban/server/filter.py
```diff
@@ -895,7 +895,8 @@ def getFailures(self, filename):
 			# see http://python.org/dev/peps/pep-3151/
 			except IOError as e:
 				logSys.error("Unable to open %s", filename)
-				logSys.exception(e)
+				if e.errno != 2:
```
yarikoptic: What is special about 2? Please leave a comment describing

fail2ban/server/filterpyinotify.py
```diff
@@ -113,14 +151,81 @@ def _process_file(self, path):
 			self.failManager.cleanup(MyTime.time())
 		self.__modified = False
 
+	def _addPending(self, path, reason, isDir=False):
+		if path not in self.__pending:
+			self.__pending[path] = [self.sleeptime / 10, isDir];
```
yarikoptic: Why by 10? Now that in some obscure case could become 0, so much be worth ceiling float division instead

fail2ban/server/filterpyinotify.py
```diff
@@ -113,14 +151,81 @@ def _process_file(self, path):
...
+			self.__pendingNextTime = 0
+			if isinstance(reason, pyinotify.Event):
+				reason = [reason.maskname, reason.pathname]
+			logSys.log(logging.MSG, "Log absence detected (possibly rotation) for %s, reason: %s of %s",
+				path, *reason)
```
yarikoptic: reason could be that event passed in... Is it a named tuple of guaranteed length 2?

fail2ban/server/filterpyinotify.py
```diff
@@ -113,14 +151,81 @@ def _process_file(self, path):
...
+
+	def _delPending(self, path):
+		try:
+			del self.__pending[path]
+		except KeyError: pass
+
+	def _checkPending(self):
+		if self.__pending:
```
yarikoptic: A matter of style of cause but I would have just checked  for not and returned if so to not cause this massive indented block

fail2ban/server/filter.py
sebres: 2 - no file, but OK :)

fail2ban/server/filterpyinotify.py
sebres: although sleeptime is float, but to be sure - yes, I'll do.

fail2ban/server/filterpyinotify.py
sebres: either event (then transformed above) or array with 2 elms.

fail2ban/server/filterpyinotify.py
sebres: No problem.

sebres:
I found another more or less grave error there, that may sporadic occur within notifyier (by processing some events and modifying of watcher at the same time in another thread).
This should be fixed now within [62e580b7cf647a2f66c19a833e5aa703ad1c19f8].

PR #1783, opened 2017-05-19
Added a new filter/jail for roundcubemail auth-failures

### LINKED ISSUE #1159: Roundcube filter v1.1.2
Hello i would like to note that existing\* filter did not work for my config

*https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/roundcube-auth.conf //checked 17-8-2015

I made a few testings from the logfile in /var/log/roundcubemail/userlogins with the format:
[14-Aug-2015 17:10:42 +0300]: <gigtq175> Failed login for sdsssdsd from xxx.xxx.xxx.xxx in session gigtq175h7bi66676toau3m9n4 (error: 0)

and i came up with the Working rule:
failregex = ._.Failed login for ._? from <HOST>.*$
PS. in the editor the rule appears ok but in the preview and in the github skips two \* between .. in the start and .? in the middle.

Specs
Centos 7.1
apache 2.4
latest fail2ban via yum
Roundcube Webmail IMAP Client | Version 1.1.2

weberhofer:
The jail should fix #1159 when logging the authentication errors to journald instead to a local file.


sebres:
Thx, Johannes!

I've rebased it with my amend using the same filter and jail (`roundcube-auth`), because the same regex can be used for both (logging into journal or logfile).
Additionally fixes more complex injections on username.
Test cases extended.

I will merge this, if it gets ready in travis.

weberhofer:
Thank you for improving the filter. What I'm missing is the "journalmatch = SYSLOG_IDENTIFIER=roundcube" in the [init] section of the filter.

sebres:
You are right, seems to be lost by merge together with the deletion of new filter (`config/filter.d/roundcube.conf`).
I'll restore it now. BTW Init section is not required anymore in v.0.10.
```bash
# filter.d/rouncube-auth.conf
[Definition]
journalmatch = SYSLOG_IDENTIFIER=roundcube
...

# jail.local
[roundcube-auth]
backend = systemd
enabled = true

$ sudo fail2ban-client -d | grep 'roundcube.*journalmatch'
['set', 'roundcube-auth', 'addjournalmatch', 'SYSLOG_IDENTIFIER=roundcube']

# jail.local
[roundcube-auth]
backend = systemd
enabled = true
filter = roundcube-auth[journalmatch="_SYSTEMD_UNIT=test.service"]

$ sudo fail2ban-client -d | grep 'roundcube.*journalmatch'
['set', 'roundcube-auth', 'addjournalmatch', '_SYSTEMD_UNIT=test.service']
```

sebres:
Restored,
Thx!

weberhofer:
Thank you! Looking forward to see v0.10, as the filters improve and are no longer compatible to 0.9

sebres:
Well, you can very easy make it compatible to 0.9 (not so fast and not the same handling, but compatible):

Just remove ` <F-CONTENT>.+</F-CONTENT>$` to make parameter `prefregex` interpolate-able as `%(prefregex)s`. And extend all `failregex` with it as prefix.

```diff
-prefregex = ^\s*(\[\])?(%(__hostname)s\s*(?:roundcube(?:\[(\d*)\])?:)?\s*(<[\w]+>)? IMAP Error)?: <F-CONTENT>.+</F-CONTENT>$
+prefregex = ^\s*(\[\])?(%(__hostname)s\s*(?:roundcube(?:\[(\d*)\])?:)?\s*(<[\w]+>)? IMAP Error)?:

-failregex = ^(?:FAILED login|Login failed) for <F-USER>.*</F-USER> from <HOST>(\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
-            ^(?:<[\w]+> )?Failed login for <F-USER>.*</F-USER> from <HOST> in session \w+( \(error: \d\))?$
+failregex = %(prefregex)s (?:FAILED login|Login failed) for <F-USER>.*</F-USER> from <HOST>(\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
+            %(prefregex)s (?:<[\w]+> )?Failed login for <F-USER>.*</F-USER> from <HOST> in session \w+( \(error: \d\))?$

+[Init]
+journalmatch = SYSLOG_IDENTIFIER=roundcube
```

weberhofer:
But also the F-USER and P=user macros are not available, are they?

sebres:
Yes, the replacement for it:
```diff
-<F-USER>.*</F-USER>
+(?P<user>.*)
```
BTW. For 2nd line takes nothing into account (v.0.9 is not able to save a failure info, so can be easy replaced with `.*`).

weberhofer:
Thanks! I'm updated the roundcube-auth.conf for 0.9.7 with the following code. Seems to work nicely now. Thanks for your support!
```php
prefregex = ^\s*(\[\])?(%(__hostname)s\s*(?:roundcube(?:\[(\d*)\])?:)?\s*(<[\w]+>)? IMAP Error)?:

failregex = %(prefregex)s (?:FAILED login|Login failed) for (?P<user>.*) from <HOST>(\. (?:(?! from ).)*(?: .*)? in \S+\.php on line \d+ \(\S+ \S+\))?$
        %(prefregex)s (?:<[\w]+> )?Failed login for (?P<user>.*) from <HOST> in session \w+( \(error: \d\))?$

ignoreregex =

[Init]
journalmatch = SYSLOG_IDENTIFIER=roundcube

**Comment by sebres** [comment]
Your first failregex seems to be no precise enough, and is affected by the injection (try to run the new test-cases, it will fails, IMHO).
```diff
-failregex = %(prefregex)s (?:FAILED login|Login failed) for (?P<user>.*) from <HOST>(\. (?:(?! from ).)*(?: .*)? in \S+\.php on line \d+ \(\S+ \S+\))?$
+failregex = %(prefregex)s (?:FAILED login|Login failed) for (?P<user>.*) from <HOST>(\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
```
BTW.  In my comment above, I meant "2nd failregex" not the second occurrence of user-capture.
The pair `(?P<user>.*)`...`user=(?P=user)` ptotects the regex as a result of dual availability of user name (and dual match) in the log-line. It's few vulnerable for possible injections.

weberhofer:
Thanks again for the update. I have updated the code above - in case someone likes to copy it. I'll include this update in the openSUSE distribution - if you have no objection :-)

sebres:
> if you have no objection

Why I should? :)
It would be possibly better if you'll provide a PR for master-branch (0.9) with cherry-picked test-case and your newest config. But I don't need it resp. don't see it as pre-requirements, because development of 0.9 is discontinued to me. Also, I plan no new releases of 0.9.
So if you want...

weberhofer:
The old filter no longer works with the latest distribution so I'll update this filter for the openSUSE package. If there is no new release planned, other users could update the rule this thread - it makes no sense to prepare a PR.

PR #1788, opened 2017-05-22
Block recidive attackers on all protocols

szepeviktor:
Not just the default protocol, tcp

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -761,6 +761,7 @@ logpath  = /var/log/fail2ban.log
 banaction = %(banaction_allports)s
 bantime  = 1w
 findtime = 1d
+protocol = all
 
 
 # Generic filter for PAM. Has to be used with action which bans all


szepeviktor:
I do not know non-iptables-based firewalls.
Please advise.

szepeviktor:
...and yes I know, the ChangeLog

szepeviktor:
OOPS! Was https://github.com/fail2ban/fail2ban/pull/442/files but somehow got removed.

szepeviktor:
Q: Shouldn't we change everywhere where there is banaction_allports??

config/jail.conf
```diff
@@ -761,6 +761,7 @@ logpath  = /var/log/fail2ban.log
 banaction = %(banaction_allports)s
 bantime  = 1w
 findtime = 1d
+protocol = all
```
sebres: This will enable also connection-less protocol like UDP, I don't think it would be good idea per default,
see waning in https://github.com/fail2ban/fail2ban/blob/0.10/config/jail.conf#L668

PR #1792, opened 2017-05-23
New logtimezone jail option, zone abbreviations, new date-pattern tokens %Exz, %ExZ

### LINKED ISSUE #1773: Enhancement ? default time zones
Fail2ban needs to understand properly the time stamps from matching log lines, and in particular, the timezones must be correctly interpreted, but there are some cases in which the log lines do not specify in which timezones they are to be understood. 

As far as I understand it, in that case, fail2ban interprets the time zone as its own default one, and there is no way to specify otherwise (at least, I couldn't find one), yet it can actually differ.

While I agree it's better for the applications to be very explicit about time zones in logs, even in the most frequent case of logging in UTC, it can be outside the scope of fail2ban users to fix that, or simply a long process, during which they still would like the protection fail2ban offers.

So, am I right that there's no way to specify a default timezone ?
In that case, I believe that a default timezone option alongside ``datepattern`` would be useful, and I can see myself working on a PR towards that. 

Thanks,

Note: in my immediate use-case, the logs are emitted in a LXC guest container, and watched from the host. Actually the container's time zone is identical to the host's but the application logs are in UTC with no indication of that (this is hardcoded) and changing that won't be immediate.

gracinet:
This new option allows to force the time zone on log lines
that don't bear a time zone indication (GitHub issue #1773), so it behaves
actually with respect to log line contents as a default time zone.

For the time being, only fixed offset timezones (UTC or UTC[+-]hhmm) are
supported, but the implementation is designed to later on treat the case
of logical timezones with DST, e.g., Europe/Paris etc.

In particular, the timezone name gets passed all the way to the strptime
module, and the resulting offset is computed for the given log line, even
though for now, it doesn't actually depend on it.

Also, the DateTemplate subclass gets to choose whether to use it or not.
For instance, it doesn't make sense to apply a time zone offset to
Unix timestamps.

The drawback is to introduce an API change for DateTemplate. I hope it's
internal enough for that not being a problem.

(ok, so that was the commit the message).

I tried to respect the surrounding style, hope I got it right (noticed methods with docstrings, which I tend to prefer).
Did not check where to add it in doc and/or changelogs at all, yet.

For future plans involving DST timezones, I suppose a simple conditional import used in validateTimeZone and zone2offset would do the trick.

closes #1773

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -98,7 +98,7 @@ TODO: implementing of options resp. other tasks from PR #1346
   e. g. by unban all, reload with removing action, stop, shutdown the system (gh-1743),
   the actions having `actionflush` do not execute `actionunban` for each single ticket
 * add new command `actionflush` default for several iptables/iptables-ipset actions (and common include);
-
+* add new jail option `logtimezone` to force the timezone on log lines that don't have an explicit one (gh-1773)
 
 ver. 0.10.0-alpha-1 (2016/07/14) - ipv6-support-etc
 -----------
--- a/ChangeLog
+++ b/ChangeLog
@@ -88,17 +88,32 @@ TODO: implementing of options resp. other tasks from PR #1346
     the parsing of log-entries contain new-line chars (as single entry);
   - if multiline regex however expected (by single-line parsing without buffering) - prefix `(?m)` 
     could be used in regex to enable it;
-* implemented execution of `actionstart` on demand (conditional), if action depends on `family` (gh-1742):
+* Implemented execution of `actionstart` on demand (conditional), if action depends on `family` (gh-1742):
   - new action parameter `actionstart_on_demand` (bool) can be set to prevent/allow starting action
     on demand (default retrieved automatically, if some conditional parameter `param?family=...` 
     presents in action properties), see `action.d/pf.conf` for example;
   - additionally `actionstop` will be executed only for families previously executing `actionstart`
     (starting on demand only)
-* introduced new command `actionflush`: executed in order to flush all bans at once 
+* Introduced new command `actionflush`: executed in order to flush all bans at once 
   e. g. by unban all, reload with removing action, stop, shutdown the system (gh-1743),
   the actions having `actionflush` do not execute `actionunban` for each single ticket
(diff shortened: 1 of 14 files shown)


sebres:
I made a brief review + small optimizations and enchancements.

What I still miss here, is the entry in [man/jail.conf.5](https://github.com/fail2ban/fail2ban/blob/0.10/man/jail.conf.5) and the change-log entry :)
So if you can append it, would be nice.

I'll merge if travis-ci says it's good.

gracinet:
@sebres sounds great ! I'll look into it in one hour or two

gracinet:
I saw your latest commit, but I have no time left to amend the man page accordingly.
I have to run, will follow up if needed tomorrow.

Cheers

sebres:
* Implemented zone abbreviations (like CET, CEST, etc.) and abbr+-offset functionality (accept zones like 'CET+0100'), for the list of abbreviations see strptime.TZ_STR;
* Tokens `%z` and `%Z` are more precise now;
* Introduced new tokens `%Exz` and `%ExZ` that fully support zone abbreviations and/or offset-based zones (implemented as enhancement using custom `datepattern`, because may be too dangerous for default patterns and tokens like `%z`, see todo below);
Note: the extended tokens supported zone abbreviations, but it can parse 1 or 3-5 char(s) in lowercase,
see todo below. Don't use them in default date-patterns (if not anchored, few precise resp. optional).

TODO: because python currently does not support mixing of case-sensitive with case-insensitive matching, check how TZ (in uppercase) can be combined with `%a`/`%b` etc. (that are currently case-insensitive), to avoid invalid date-time recognition in strings like '11-Aug-2013 03:36:11.372 error ...' with wrong TZ "error", which is at least not backwards compatible.
Hence `%z` currently match literal Z|UTC|GMT only (and offset-based), and `%Exz` - all zone abbreviations.

sebres:
I leave it a bit open for review and possible objections from @fail2ban/developers and co.
At the beginning of next week I'll merge it as is.
Thx, again.

gracinet:
@sebres awesome, many thanks !

PR #1794, opened 2017-05-30
IPv6 addresses may overlap

szepeviktor:
**This is just my first thought.**

Could it be that there is a colon before/after an IPv4 address in the log?

IPv6 addresses may overlap because of compression.

- 2001:db8:a0b::1
- 2001:db8:a0b::1:12f0


sebres:
I've extended it a little bit (more strict now).

BTW usage of the `:`-character as boundary is not so good (at least at end, where it can be used for the port-separator), but ATM I don't see better solution for it.

sebres:
No objections, thus merge... Thx.

szepeviktor:
Thank you. 

-- 
+36204242498
Ezen a készüléken sok az elütés. 
Elnézést!

PR #1796, opened 2017-05-31
dovecot: Match also when user field is empty

peternowee:
Commit 5678d08 of 2016-11-26 changed:

    ( user=<\S*>,)?

to:

    ( user=<[^>]+>,)?

The change from `*` (zero or more times) to `+` (one or more times) may not have been intended. It will miss lines containing, for example:

    Aborted login (tried to use disallowed plaintext auth): user=<>

This commit reverts the `+` back to `*`.

Sorry, I did not have time to perform any tests, etc. Hope this helps nonetheless.

Filing against the `master` branch, but applies to both `0.9` and `0.10` as well.


peternowee:
I don't know. Unfortunately, I also do not really have time to dive into this more. I just noticed in passing what looked like an unintended regression and wanted to share a quick fix. Feel free to close this PR. Hope you can find some other way to still let those lines match. Thanks for all your great work!

sebres:
Good, it looks like the case with handshake not really affected by this regex.
So I'll merge it as long as someone don't say (with an example-line) where it should ignore empty user-name...

peternowee:
Great, thanks!

PR #1803, opened 2017-06-14
Update iptables actions to cooperate with and give precedence to iptables-persistent

TrooperT:
The reasoning for this PR is a solution to a problem I ran across and figured it could be useful to others. I use iptables-persistent to manage static iptables entries and allow fail2ban to manage dynamic iptables entries.

The issue I ran across was that the ssh jail i had running was inadvertently banning valid users who, like all humans, fat fingered their credentials once too many. I traced it down to the actionstart for the included files Inserting the jumps into the INPUT chain AFTER iptables-persistent had Inserted rules, and thus causing the potential for a user to unintentionally ban themself if they flubbed credentials and were targetted by fail2ban.

Observed:
```
iptables -nL INPUT
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
f2b-recidive  tcp  --  0.0.0.0/0            0.0.0.0/0           
f2b-sshd   tcp  --  0.0.0.0/0            0.0.0.0/0            multiport dports 22         
ACCEPT     all  --  172.32.0.0/11        0.0.0.0/0           
ACCEPT     all  --  208.54.128.0/19      0.0.0.0/0           
ACCEPT     all  --  208.54.0.0/17        0.0.0.0/0           
ACCEPT     all  --  206.180.38.20        0.0.0.0/0   
```
Desired (and accomplished):
```
iptables -nL INPUT
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
ACCEPT     all  --  172.32.0.0/11        0.0.0.0/0           
ACCEPT     all  --  208.54.128.0/19      0.0.0.0/0           
ACCEPT     all  --  208.54.0.0/17        0.0.0.0/0           
ACCEPT     all  --  206.180.38.20        0.0.0.0/0   
f2b-recidive  tcp  --  0.0.0.0/0            0.0.0.0/0           
f2b-sshd   tcp  --  0.0.0.0/0            0.0.0.0/0            multiport dports 22         
```

### DIFF
--- a/config/action.d/iptables-allports.conf
+++ b/config/action.d/iptables-allports.conf
@@ -19,7 +19,7 @@ before = iptables-common.conf
 #
 actionstart = <iptables> -N f2b-<name>
               <iptables> -A f2b-<name> -j <returntype>
-              <iptables> -I <chain> -p <protocol> -j f2b-<name>
+              <iptables> -A <chain> -p <protocol> -j f2b-<name>
 
 # Option:  actionstop
 # Notes.:  command executed once at the end of Fail2Ban
--- a/config/action.d/iptables-multiport-log.conf
+++ b/config/action.d/iptables-multiport-log.conf
@@ -21,7 +21,7 @@ before = iptables-common.conf
 #
 actionstart = <iptables> -N f2b-<name>
               <iptables> -A f2b-<name> -j <returntype>
-              <iptables> -I <chain> 1 -p <protocol> -m multiport --dports <port> -j f2b-<name>
+              <iptables> -A <chain> 1 -p <protocol> -m multiport --dports <port> -j f2b-<name>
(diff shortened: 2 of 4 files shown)


sebres:
I can imagine many situations where "append" could be not desired.

So think, we can provide it as an option like `<iptables-pos>` (open for better names), that can be used like here:
```bash
# jail.local:
[DEFAULT]
banaction = %(banaction)s[iptables-pos="-A"]
banaction_allports = %(banaction_allports)s[iptables-pos="-A"]
```
or in `action.d/iptables-common.local`
```
[Init]
iptables-pos = -A
```

sebres:
BTW If your white-list is only to prevent fail2ban to ban this IPs, you can add all this to fail2ban using option `ignoreip`.

Issue #1806, opened 2017-06-15
Not rebanning after restart if persistent ban-time set (bantime=-1)


ksmacd:
Hi i have updated the version of fail2ban to the new release 0.10
The problem i have is when ever i restart the service it unbans all ip's that have been banned and does not reban them.

When i was using the old fail2ban when ever i restarted the service or rebooted persistent recidive bans got reinstated which i want let me explain why i want how i got to v0.10 and give you some of the jails i moved over from v0.9.

Ok right i want my bans to stick as this is a closed system for my self and have set up a strict policy 1 failed attempt perm ban as its only me who will have access to this system on v.09 this worked on reboot or service restart

Im using ubuntu 16.04
i purged fail2ban then downloaded from git v0.10 then installed.
i moved my own jails over from my old setup and changed nothing just added them with say example like below the new jail setups for v0.10 all i did is change the following.
```
enabled = true
findtime = 100
maxretry = 1
bantime = -1
banaction = iptables-allports 
```

```
[sshd]
mode    = aggressive
filter  = sshd[mode=%(mode)s]
port    = ssh
logpath = %(sshd_log)s
backend = %(sshd_backend)s
enabled = true
findtime = 100
maxretry = 1
bantime = -1
banaction = iptables-allports
```
on the old version when i rebooted or restarted fail2ban my recidive would stay which i want as far as im concerned im the only user who will have backend or service access to this cluster so if they try to hack or etc they can be perm banned for good as far as im concerned but after upgrading the ban only lasts now until reboot or service restart.

Any input would be great

sebres:
Thanks!
Possibly a bug by calculating of rest of the ban-time (within restore process). I'll check it.

If you can try it in the meantime without persistent ban (anyway bad idea in the long term). 
Thus just try `bantime = 1y` (for 1 year).
Just as a workaround if it works (as long as not fixed).

ksmacd:
ok yes this is indeed a bug
changing to 1y enables the bans on reboot and service restart
just to help out on the bug as-well
if on -1 if you do a search on jails after service restart it shows no bans
if you change it to 1y restart service then search jail for banned all previous bans are implied too which was banned under the -1 ban time

PR #1807, opened 2017-06-15
bug-fix: restoring of tickets from database for jails with persistent ban

sebres:
Fixes restoring of tickets from database for jails with persistent ban (if `bantime = -1`)
Closes gh-1806

### DIFF
--- a/fail2ban/server/database.py
+++ b/fail2ban/server/database.py
@@ -593,7 +593,7 @@ class Fail2BanDb(object):
 		if ip is not None:
 			query += " AND ip=?"
 			queryArgs.append(ip)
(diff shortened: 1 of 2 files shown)


sebres:
BTW. Unfortunately it introduces another bad thing: If `bantime` of jail will be changed to -1 (and restarted), all tickets (also very old) from the database will be persistent banned after restart.

This going to 2 issues:
- because at the moment 0.10 does not hold current ban-time of ticket (only time of ban)
- and `purge` will be currently never called (no tickets will be cleaned up at all)

Branch ban-time-incr (PR #1460) fixes both things already, so not affected from the problem.

PR #1812, opened 2017-06-25
Replace port imap3 with imap

jpotter:
imap was missing from the list of ports managed by courier-auth, preventing fail2ban from blocking connections on standard IMAP port 143.


sebres:
@fail2ban/contributors Any objections to extend another mail-jails with port `imap` also (e. g. postfix-sasl, perdition, etc.)? 
Some jails (for example dovecot, kerio etc) already have it.

sebres:
Extended in order to fix #1942.
Thus port `imap3` removed everywhere (replaced with `imap`, in sence of original PR)

Issue #993, opened 2015-03-18
ejabberd filter doesn't work


digininja:
I've got my ejabberd log level set to info and these are the lines printed after a failed login attempt:

```
=INFO REPORT==== 2015-03-18 22:44:59 ===
I(<0.376.0>:ejabberd_listener:281) : (#Port<0.3791>) Accepted connection {{192,168,0,2},58651} -> {{192,168,0,50},5222}

=INFO REPORT==== 2015-03-18 22:44:59 ===
I(<0.380.0>:ejabberd_c2s:802) : ({socket_state,tls,{tlssock,#Port<0.3791>,#Port<0.3819>},<0.379.0>}) Failed authentication for robin@test.int
```

The filter rule is expecting a "from IP <HOST>" on the end of the failed auth line but that isn't there, it is in the connection line above it. Even with log level set to debug the IP/host isn't included on the same line.

digininja:
Just realised, I assumed I was on a relatively up-to-date version of ejabberd as I installed it as a package from Debian Wheezy but I've just checked and it looks to be quite old so I'll do an upgrade in the morning and close the ticket if it is my side that is at fault.

digininja:
I've upgraded to the latest ejabberd directly from github and the log entry is different but the matching lines still don't work. This is a line from the new log file

2015-03-19 13:57:35.805 [info] <0.585.0>@ejabberd_c2s:wait_for_sasl_response:965 ({socket_state,p1_tls,{tlssock,#Port<0.6434>,#Port<0.6436>},<0.584.0>}) Failed authentication for robin@test.int from 192.168.0.2

The match is looking for "wait_for_feature_request" in the middle but I've got the "wait_for_sasl_response" line instead. It also seems to expect IP before the host name at the end which isn't there.

I tried creating a new entry with sasl and no IP and that didn't work so for now I've gone with this that seems to work OK.

^.* \[info\] .* Failed authentication for \S+ from <HOST>$

MonolithProjects:
This is what works for me (ejabberd 17.01, fail2ban 0.9.5):

^.* Failed authentication for \S+ from <HOST>.*$

sebres:
`^.*` makes no sense at all;
`^.* [info] .*` makes no realy sense (except it checks that entry contains `info` somewhere);
The provided expression corresponds almost to ahead-unanchored one:
```
Failed authentication for \S+ from <HOST>$
```

MonolithProjects:
[...]

PR #1850, opened 2017-07-29
Support ejabberd 17.06 log format

MarcelWaldvogel:
Current log output looks as follows:
`2017-07-29 08:24:04.773 [info] <0.6668.0>@ejabberd_c2s:handle_auth_failure:433 (http_bind|ejabberd_bosh) Failed c2s PLAIN authentication for test@example.ch from ::FFFF:192.0.2.3: Invalid username or password`

I have not added an entry in the Changelog, as I believe it still matches the description and title from #993 ("new" is relative)


sebres:
Thx!
No objections, so merge...

PR #1856, opened 2017-08-05
Apache <HOST> field can contain several hosts

benrubson:
Hello,

Apache let us configure the `ErrorLogFormat` :
https://httpd.apache.org/docs/2.4/mod/core.html#errorlogformat
In case of a server behind a `ProxyPass` directive, it is then very easy to log the original client IP replacing `%a` by `%{X-Forwarded-For}i` or `%{HTTP_X_FORWARDED_FOR}e`.
This situation can however produce log lines with several addresses :
`[Sat Jun 01 02:17:42 2013] [error] [client 192.168.0.2, 1.2.3.4] user root not found`

This PR then detects this and works on the left-most address, the original client one.

Thank you 👍 

Ben

### DIFF
--- a/config/filter.d/apache-common.conf
+++ b/config/filter.d/apache-common.conf
@@ -27,7 +27,7 @@ _daemon = (?:apache\d*|httpd(?:/\w+)?)
 
 apache-prefix = <apache-prefix-<logging>>
 
-_apache_error_client = <apache-prefix>\[(:?error|\S+:\S+)\]( \[pid \d+(:\S+ \d+)?\])? \[client <HOST>(:\d{1,5})?\]
+_apache_error_client = <apache-prefix>\[(:?error|\S+:\S+)\]( \[pid \d+(:\S+ \d+)?\])? \[client <HOST>(:\d{1,5})?(,[ \d\.]+)*\]
 
 datepattern = {^LN-BEG}
 
--- a/fail2ban/tests/files/logs/apache-auth
+++ b/fail2ban/tests/files/logs/apache-auth
@@ -122,6 +122,9 @@
 # failJSON: { "time": "2013-06-01T02:17:42", "match": true , "host": "192.168.0.2" }
 [Sat Jun 01 02:17:42 2013] [error] [client 192.168.0.2] user root not found
 
+# failJSON: { "time": "2013-06-01T02:17:42", "match": true , "host": "192.168.0.2" }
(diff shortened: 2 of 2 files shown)


sebres:
Thx for the PR, but as already several times said, we support only default configuration.
If you change logging format of apache, this is IMHO customization and should be viewed as it.
Thus despite very short (and backwards-compatible) enhancement, I tend to close it nevertheless.

If someone changes the log-format, he can change also the config-parameter `_apache_error_client` etc in the fail2ban-configuration.

benrubson:
Thank you for your feedback Serg !
Yes this is not the default Apache configuration, but as the change is very minor (and backward compatible as you emphasise), I thought that perhaps it would be possible to include it.
Perhaps we could add a comment line above saying that it is compatible with both default `%a` and ProxyPass `%{X-Forwarded-For}i` options ?
This would direct users toward / make them think about the best configuration to used based on their setup, and perhaps will convince you to adopt this PR :)
Thank you Serg 👍

sebres:
The question is what should happens there:
- either you'll ban the whole proxy-IP?
- or you want to write another filter and another custom action, that will forbid the original IP via some apache/nginx auth-preprocessor (not via firewall like iptables)?
- or you want create new jail with custom filter and larger `maxretry` value, that will match such lines?

That's why I don't see it as simple PR.

benrubson:
Currently the PR bans the left-most address, which is the IP of the client itself in `X-Forwarded-For`, where the request began.
This is the less-risky situation, as we won't ban the whole proxy itself.

Think about situation where you manage both the proxy and targeted servers.
Of course what you want is to ban the client which initiated the request, not your proxy :)

This behaviour is very closed to the current one, in addition it keeps the modification simple.

sebres:
> This is the less-risky situation, as we won't ban the whole proxy itself.

No, then this does nothing, because it bans the IP that never build connection (directly). Thus despite the ban you'll get further requests from this client via same proxy (and multiple entries "Already banned" in fail2ban.log).

Via default firewall actions (like iptables) you can only ban the tcp-ip connection (and it is the proxy resp. forwarder IP, not the original one). If you want to do it for original, you should reject such connections via special custom actions (mostly directly in apache resp. nginx servers)...

Thus closed, sorry.

benrubson:
Yep you're right, the IPs detected by the targeted end-servers should then have to be shared with the proxy itself. This is what I do.
A situation where this also works flawlessly is then you have both proxy and end-server(s) on the same host.
Thank you for your time @sebres 👍

Issue #1849, opened 2017-07-29
Extend dovecot auth regex with sql


yamlfd:
### Environment:

_Fill out and check (`[x]`) the boxes which apply. If your Fail2Ban version is outdated, 
and you can't verify that the issue persists in the recent release, better seek support 
from the distribution you obtained Fail2Ban from_

- Fail2Ban version (including any possible distribution suffixes): 0.9.6-2 all
- OS, including release name/version: Debian 9.0
- [X] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

### The issue:

Dovecot rules are missing auth via SQL. They only look for `PAM` and `passwd-file`

#### Steps to reproduce
Try to authenticate with invalid user against dovecot imap.

#### Expected behavior
Connection should be blocked as this are only tries to find valid users.

#### Observed behavior
User is not blocked as rexex only expects `(?:pam|passwd-file)`, but mysql auth is used and this changes the string to `sql`

### Configuration, dump and another helpful excerpts

Original:
`
^%(__prefix_line)s(?:auth|auth-worker\(\d+\)): (?:pam|passwd-file)\(\S+,<HOST>\): unknown user\s*$
`

Patched:
`
^%(__prefix_line)s(?:auth|auth-worker\(\d+\)): (?:pam|passwd-file|sql)\(\S+,<HOST>\): unknown user\s*$
`

#### Any customizations done to /etc/fail2ban/ configuration
```
[dovecot]
enabled = true
```

#### Relevant parts of /var/log/fail2ban.log file:
_preferably obtained while running fail2ban with `loglevel = 4`_

```
not available
```

#### Relevant lines from monitored log files in question:

```
Jul 29 13:54:06 auth-worker(22401): Info: sql(haine@example.de,70.165.93.163,<n4JLdHNVngZGpV2j>): unknown user
Jul 29 13:54:09 imap-login: Info: Disconnected (auth failed, 1 attempts in 5 secs): user=<haine@example.de>, method=PLAIN, rip=70.165.93.163, lip=62.75.123.123, TLS: Disconnected, session=<n4JLdHNVngZGpV2j>
```

sebres:
Already fixed in current 0.9 and 0.10 (Note the second line from your excerpt will be recognized as failure).

```bash
$ git checkout 0.9
Switched to branch '0.9'

[...]

PR #1857, opened 2017-08-07
filter.d/dovecot: extend failregex to accept another methods as pam/pwd-file

sebres:
- Recognizes "unknown user" for additional auth-methods (pam, passwd-file, ldap, sql, etc)
- Simplifying regular expressions (put "unknown user" and "invalid credentials" together as one regex).

Currently the `failregex` of dovecot filter will generate two failures for methods like pam or passwd-file (see gh-1849), once after log-entry `unknown user` and once after `Disconnected (auth failed ...)`.
This PR extends it similar for another methods also... But only for 2 cases ("unknown user" and "invalid credentials"), so I think in this cases the attacker can be banned early (`maxretry / 2`), because of the brute-force resp. penetration tests. 

BTW Made for both versions (0.9/0.10)...


yarikoptic:
FWIW, looks kosher to me

Issue #1406, opened 2016-05-04
Monit files are moved to /etc/monit/conf-available


szepeviktor:
/etc/monitrc.d -> /etc/monit/conf-available
http://anonscm.debian.org/cgit/collab-maint/monit.git/commit/?id=0bcc9189cb4f43863f08b2423cc54097a3e124ef

Was #896

sebres:
Could you then add it as comment to the PR #896 (or just change the first comment there).
If you mean we should change something (docu, files/monit/fail2ban, changelog, etc), please PR.
If you want, you can also add a wiki-page with howto.
Thx!

szepeviktor:
@yarikoptic It needs to be changed in the Debian package.

szepeviktor:
Here I think https://github.com/fail2ban/fail2ban/blob/debian/debian/rules#L37

sebres:
Just make a PR for debian branch as target...

szepeviktor:
Seems to be outdated
https://github.com/fail2ban/fail2ban/commits/debian
https://anonscm.debian.org/cgit/collab-maint/monit.git/log/

PR #1858, opened 2017-08-10
Monit files are moved

szepeviktor:
/etc/monitrc.d -> /etc/monit/conf-available
http://anonscm.debian.org/cgit/collab-maint/monit.git/commit/?id=0bcc9189cb4f43863f08b2423cc54097a3e124ef

Was #1406

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -34,7 +34,7 @@ override_dh_install:
 	: # take care about it anyways
 	rm -rf $(DESTDIR)/var/run/ $(DESTDIR)/run/
 	: # Install monit configuration
-	install -d $(DESTDIR)/etc/monit/monitrc.d
+	install -d $(DESTDIR)/etc/monit/conf-available
 	install -m 644 files/monit/fail2ban $(DESTDIR)/etc/monit/monitrc.d/fail2ban
 	: # Install bash completion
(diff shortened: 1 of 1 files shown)


szepeviktor:
@yarikoptic Could you say a *yes* please?

szepeviktor:
Done.

debian/rules
```diff
@@ -34,7 +34,11 @@ override_dh_install:
 	: # take care about it anyways
 	rm -rf $(DESTDIR)/var/run/ $(DESTDIR)/run/
 	: # Install monit configuration
-	install -d $(DESTDIR)/etc/monit/monitrc.d
+	if dpkg --compare-versions "$(dpkg-parsechangelog --show-field Version)" lt "1:5.15-2" 2> /dev/null; then
```
yarikoptic: this will parse our changelog... we should check which monit is available on the system, but that involves parsing apt-cache policy output etc... probably too much for the rules... 

so I just merged your original commit, added debian/changelog entry and pushed.

yarikoptic:
Thanks, merged in the first version, closing

debian/rules
szepeviktor: Oh! Sorry.
I've got confused.

PR #1862, opened 2017-08-13
Updated Horde filter for Horde 5.2

wb7:
Updated filter from my environment; I am using Horde 5.2 and this works for me pretty good.

### DIFF
--- a/config/filter.d/horde.conf
+++ b/config/filter.d/horde.conf
@@ -4,7 +4,7 @@
 [Definition]
 
 
-failregex = ^ HORDE \[error\] \[(horde|imp)\] FAILED LOGIN for \S+ \[<HOST>\](\(forwarded for \[\S+\]\))? to (Horde|{[^}]+}) \[(pid \d+ )?on line \d+ of \S+\]$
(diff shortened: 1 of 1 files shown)


config/filter.d/horde.conf
```diff
@@ -4,7 +4,7 @@
 [Definition]
 
 
-failregex = ^ HORDE \[error\] \[(horde|imp)\] FAILED LOGIN for \S+ \[<HOST>\](\(forwarded for \[\S+\]\))? to (Horde|{[^}]+}) \[(pid \d+ )?on line \d+ of \S+\]$
+failregex = ^ ERR: HORDE \[(horde|imp)\] FAILED LOGIN for \S+ to (horde|{[^}]+}) \((<HOST>)\) \[(pid \d+ )?on line \d+ of \S+\]$
```
sebres: Please add second failregex after your new one (for the backwards compatibility to previous versions)...
```INI
failregex = ^ ERR: HORDE ... (1st regex)
            ^ HORDE ... (2nd regex)
```

sebres:
Still interested to complete this?
1. Both regex (old also for the purpose of backwards compatibility)
2. extend https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/horde with log-message in new format (with match-JSON string).

At least please provide here a short excerpt of the log file with new log format.

PR #1867, opened 2017-08-17
Tweaks to action.d/pf.conf

distler:
Document recent changes.
Add an option to customize the pf block rule (surely, what the user
really wants, here, is "block quick").

### DIFF
--- a/config/action.d/pf.conf
+++ b/config/action.d/pf.conf
@@ -15,8 +15,17 @@
 #
 # we don't enable PF automatically; to enable run pfctl -e 
 # or add `pf_enable="YES"` to /etc/rc.conf (tested on FreeBSD)
+# also, these rulesets are loaded into (nested) anchors
+# to enable them, add
+#     anchor f2b {
+#        name1
+#        name2
+#        ...
+#     }
+# to your main pf ruleset, where "namei" are the names of the jails
+# which invoke this action
 actionstart = echo "table <<tablename>-<name>> persist counters" | <pfctl> -f- 
-              echo "block proto <protocol> from <<tablename>-<name>> to <actiontype>" | <pfctl> -f-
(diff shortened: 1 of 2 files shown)


sebres:
Thanks for PR.
Would be the following piece not better?
```diff
- block quick
+ block in quick
```
Are incoming only...

distler:
The difference being whether *outgoing* packets to the offending ip address are allowed?

Is there a scenario where you'd want to allow outgoing packets to the offending ip, but block incoming ones?

sebres:
> Is there a scenario where you'd want to allow outgoing packets

No. 
But perhabs it could burden on the firewall system, if anyway not allowed... I don't know - just as question.

distler:
Every packet (incoming or outgoing) is examined by the firewall system, to see whether it matches any of the rules. It doesn't "cost" more to `block` a packet than it does to `pass` it.

The advantage of the `quick` keyword is that it says, "stop checking this packet against all the *remaining* rules." If a subsequent rule would have changed the disposition of the packet, we never get to check.

I suppose you could think of that as an optimization (fewer rules to check), but the more important reason I suggested it is that it makes the disposition of the packet independent of the ordering of the rules (which is a little tricky to figure out, when you have nested anchors like this).

sebres:
> It doesn't "cost" more to block a packet than it does to pass it.

I would be not so sure: if the rule burden two **separate** blocks within internal firewall structures (in this case both lists grow insted of single one). But...

Nevertheless, it looks good. Thus merge (in 0.11 as selected from you, I'll look to cherry-pick it later in 0.10).

distler:
Actually, I think you were right: `block in quick` is *slightly* better than `block quick`. [This link](http://www.undeadly.org/cgi?action=article&sid=20060927091645) explains how PF does its ruleset optimization, and the processing of outbound packets is streamlined in your version.

PR #1891, opened 2017-09-10
Fix Gentoo init script's shebang

sbraz:
Use openrc-run instead of runscript.
https://github.com/OpenRC/openrc/commit/5d5856c193768d24f11d5f0533e48c39526aef5c

The current ebuild does it with sed so we might as well fix it upstream:
https://github.com/gentoo/gentoo/commit/48dca3295fe1d9353ecff4e155139aff7deb2351#diff-bebdfc44a2067a248deb151d3547f3c3R44

I hope 0.10 is the right branch to apply my PR to.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,7 +35,7 @@ ver. 0.10.1-dev-1 (2016/??/??) - development edition
 -----------
 
 ### Fixes
-* fix Gentoo init script's shebang to use openrc-run instead of runscript
+* fix Gentoo init script's shebang to use openrc-run instead of runscript (gh-1891)
 * jail "pass2allow-ftp" supply blocktype and returntype parameters to the action (gh-1884)
 * avoid using "ANSI_X3.4-1968" as preferred encoding (if missing environment variables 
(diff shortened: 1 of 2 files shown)


sbraz:
@sebres Here you go, hope this is clear enough.

sbraz:
That ChangeLog looks a lot like MarkDown, shouldn't it be renamed so that GitHub displays it nicely?

sebres:
> That ChangeLog looks a lot like Markdown, shouldn't it be renamed so that GitHub displays it nicely?

@yarikoptic What do you think?

sbraz:
@sebres @yarikoptic Wow it looks really good :) https://github.com/sbraz/fail2ban/blob/changelog/ChangeLog.md

yarikoptic:
Sure, why not... (I already forgot why we didn't rename it awhile back, so let's just make it happen)

sebres:
> I already forgot why we didn't rename it awhile back

Maybe because this will affect all the open PRs? (with changelog entries :)

yarikoptic:
I don't think so... if rename is done in a single commit without any other changes, git is smart enough to figure it out AFAIK ;)

PR #1894, opened 2017-09-11
Fix nftables actions for IPv6 addresses, fixes #1893

### LINKED ISSUE #1893: nftables action does not support ipv4 and ipv6 simultaneously
Hello,
I am running Fail2ban v0.10.0 and I have created a jail which uses `nftables-multiport` as `banaction`. This works fine for banning IPv4 addresses, however ipv6 addresses trigger a failure:
```
[17793]: Level 39 7fc76c084e40 -- exec: nft add element inet filter f2b-sshd \{ 2001:xxx \}
[17793]: ERROR   7fc76c084e40 -- stderr: '<cmdline>:1:36-60: Error: Could not resolve hostname: Address family for hostname not supported'                                                                            
```
I understand that this is because the nftables set is configured to contain  `ipv4_addr` as set [here](https://github.com/fail2ban/fail2ban/blob/fbd46f29f2d47954ab3507a9d2be993a86302379/config/action.d/nftables-common.conf#L73) but I don't see how I could make my action work for both IPv4 and IPv6.

I guess something can be worked out with `[Init?family=inet6]` but I am not exactly sure how. Perhaps @albel727 will be interesting in adding this feature?

sbraz:
* add `[Init?family=inet6]` to `nftables-common.conf` and make nftable
  expressions more modular
* change `ip protocol` to `meta l4proto` in `nftables-allports.conf`
  since the former only works for IPv4

Although `meta protocol` and `ip protocol` do not seem to be synonyms to each other, they seem to both work. I've [posted on the netfilter mailing list](https://marc.info/?l=netfilter&m=150516423319129&w=2) to know more about this difference.

@albel727 feel free to review my changes since you wrote the action files.

Closes gh-1893

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,6 +35,7 @@ ver. 0.10.1-dev-1 (2016/??/??) - development edition
 -----------
 
 ### Fixes
+* fix nftables actions for IPv6 addresses (gh-1893)
 * fix Gentoo init script's shebang to use openrc-run instead of runscript (gh-1891)
 * jail "pass2allow-ftp" supply blocktype and returntype parameters to the action (gh-1884)
 * avoid using "ANSI_X3.4-1968" as preferred encoding (if missing environment variables 
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,7 +35,6 @@ ver. 0.10.1-dev-1 (2016/??/??) - development edition
 -----------
 
 ### Fixes
-* fix nftables actions for IPv6 addresses (gh-1893)
 * fix Gentoo init script's shebang to use openrc-run instead of runscript (gh-1891)
 * jail "pass2allow-ftp" supply blocktype and returntype parameters to the action (gh-1884)
 * avoid using "ANSI_X3.4-1968" as preferred encoding (if missing environment variables 
@@ -55,6 +54,7 @@ ver. 0.10.1-dev-1 (2016/??/??) - development edition
   human readable representation (opposite to `-d`);
 
 ### Enhancements
+* nftables actions are IPv6-capable now (gh-1893)
 * filter.d/dovecot.conf: introduced mode `aggressive` for cases like "disconnected before auth was ready" (gh-1880)
 
 
--- a/config/action.d/nftables-allports.conf
+++ b/config/action.d/nftables-allports.conf
@@ -17,6 +17,6 @@ before = nftables-common.conf
 # Notes.:  additional expressions for nftables filter rule
 # Values:  nftables expressions
 #
-nftables_mode = ip protocol <protocol>
+nftables_mode = meta l4proto <protocol>
 
 [Init]
--- a/config/action.d/nftables-common.conf
+++ b/config/action.d/nftables-common.conf
@@ -28,11 +28,11 @@ nftables_mode = <protocol> dport \{ <port> \}
 # Notes.:  command executed once at the start of Fail2Ban.
 # Values:  CMD
 #
-actionstart = <nftables> add set <nftables_family> <nftables_table> f2b-<name> \{ type <nftables_type>\; \}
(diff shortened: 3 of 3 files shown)


ChangeLog
```diff
@@ -35,6 +35,7 @@ ver. 0.10.1-dev-1 (2016/??/??) - development edition
 -----------
 
 ### Fixes
+* fix nftables actions for IPv6 addresses (gh-1893)
```
sebres: is not really a fix, rather an enhancement...

config/action.d/nftables-allports.conf
```diff
@@ -17,6 +17,6 @@ before = nftables-common.conf
 # Notes.:  additional expressions for nftables filter rule
 # Values:  nftables expressions
 #
-nftables_mode = ip protocol <protocol>
+nftables_mode = meta l4proto <protocol>
```
sebres: > change `ip protocol` to `meta l4proto` in `nftables-allports.conf` since the former only works for IPv4

I don't think so (ATM I don't have nft to hand to check), just try following:
```INI
nftables_mode = ip protocol <protocol>
[Init?family=inet6]
nftables_mode = ip6 protocol <protocol>
```

sebres: Or much better, because you already have parameter `address_family`, just try to rewrite (without new section):
```diff
-nftables_mode = ip protocol <protocol>
+nftables_mode = <address_family> protocol <protocol>
```

ChangeLog
sbraz: Well with the current actions, I get errors. When I installed fail2ban 0.10.0, I got no warning that these actions didn't support IPv6 so I consider it to be a bug. However I can change it if you prefer.

config/action.d/nftables-allports.conf
sbraz: Yeah I thought it would work but it doesn't (I wouldn't have changed the code otherwise).
IPv6 doesn't have a protocol header so `ip6 protocol` makes no sense. See [the reply I got on the netfilter ML](https://marc.info/?l=netfilter&m=150519799600765&w=2).

ChangeLog
sebres: we've still many actions that are not IPv6-capable, but Ok...

config/action.d/nftables-allports.conf
sebres: Understood

sbraz:
Thanks! Do you plan on releasing a minor release any time soon? I'm wondering if I should add that patch to the downstream Gentoo package instead of waiting for a new release.

sebres:
I'll try, but I can't promise (just too busy in nearby future)

sbraz:
>  I'll try to cover it inside our test-cases, to check at least the result for IPv4 is still the same...

I don't even see any tests for nftables at the moment :/

sebres:
> I don't even see any tests for nftables at the moment

As already said, because these were not yet written. :)
The correct place for it would be [testCheckStockCommandActions](https://github.com/fail2ban/fail2ban/blob/3f83b22de26ac0ef2643a81827194b996c4d185c/fail2ban/tests/servertestcase.py#L1202).

But ATM it doesn't matter, because they will be anyway different to previous version (because of change on `nftables_mode` to `meta l4proto`)...

I'm still waiting for any objections from someone of @fail2ban/contributors, otherwise I'll merge it tomorrow as it is.

albel727:
>@albel727 feel free to review my changes since you wrote the action files.

I don't have inbound ipv6 to test things on, so I welcome your changes however they are. They look fine to me though.

sebres:
@albel727 Thank you for the review.
> I don't have inbound ipv6 to test things on

You could test it with ipv4 (since it's affected also)...

Issue #1900, opened 2017-09-25
addIgnoreIP should not produce duplicates


szepeviktor:
Adding an IP to ignore list could produce duplicates.

Could we loop through the current list to check for the IP being added?

sebres:
> Adding an IP to ignore list could produce duplicates.

Yes, but only correct way would be like [here](https://github.com/fail2ban/fail2ban/blob/b5dd5adb08f9fd3496f8c63b1d53b0e3c6aabbf1/fail2ban/server/filter.py#L501) - using `isInNet` to recognize the sub-nets... Something like:
```python
if ip.isInNet(net):
  continue
if net.isInNet(ip):
  # replace:
  self.__ignoreIpList[neti] = ip
```
Just, it can be too risky for dyn-dns entries and similar...

Or did you meant exact duplicates?

szepeviktor:
> Or did you meant exact duplicates?

Of course!

PR #1904, opened 2017-10-02
Avoid exact duplicates by addIgnoreIP (closes gh-1900)

sebres:
Avoid exact duplicates by addIgnoreIP (closes gh-1900)


szepeviktor:
Thank you @sebres

Issue #1230, opened 2015-10-25
fail2ban and owncloud


CuoreSportivo:
Hello everyone,

I have installed on a CentOS 7 machine ownCloud 8.2 (upgraded from 8.1.3) and latest fail2ban.
I have also installed ownCloud 8.2 and fail2ban on RaspberryPi (Rasbian). Even though the configuration files are similar on both systems, fail2ban has a weird behavior on my CentOS machine, but on the Rasbian everything is working as it should.

To be more specific about this weird behavior, if I go to https://myip/owncloud and try to enter wrong usernames/passwords, nothing will happen. I can type as many passwords as I want. But If I close and reopen my browser (any browser), then I will get a connection time out, cause apparently I am banned.

On my raspberry, after 4 failed attempts, I automatically get banned and face the connection time out.

I installed fail2ban via epel repos, but also (and after a yum purge) by downloading fail2ban-master.zip from github. 

I made a thread on CentOS forums about this and the same behavior was reported by another user, after he upgraded to latest fail2ban version. So, I'd firstly like to report this issue, and ask if anyone knows if there is a fix about this, and if not, how I could downgrade to another version.

Any help is appreciated.

szepeviktor:
What filter do you use?

CuoreSportivo:
Hi szepeviktor,
I am using the one for owncloud 8.2, proposed at owncloud forums.


[Definition]
failregex={"reqId":".*","remoteAddr":".*","app":"core","message":"Login failed: '.*' \(Remote IP: '<HOST>'\)","level":2,"time":".*"}

ignoreregex =

(https://forum.owncloud.org/viewtopic.php?f=8&t=28678)

szepeviktor:
See if it's matching with `fail2ban-regex` (built-in tool).

CuoreSportivo:
....
"Lines: 213 lines, 0 ignored, 19 matched, 194 missed [processed in 0.05 sec]"
....

I think the problem is not the filter itself, cause it would have been reported to owncloud forums. Plus that I had the same problem with fail2ban and owncloud 8.1.3 (different filter), plus that on my Raspberry the same filter is operating as it should.

The user gets banned, but you have to close the browser and reopen it, in order to get connection timed out. After every failed authentication attempt, even though the browser refreshes, I will not get a connection timed out, so I can try as many username/password combinations as I want. 

[...]

PR #1913, opened 2017-10-11
Create nextcloud-auth.conf

avi9526:
Filter to detect nextcloud (fork of ownCloud) failed auth attempts. Require nextcloud to use syslog (journalctl)

Example of log: journalctl --since '-20m' -o cat  |  grep "Login failed"
  {core} Login failed: 'user_name' (Remote IP: '192.168.0.100')

### DIFF
--- a/config/filter.d/nextcloud-auth.conf
+++ b/config/filter.d/nextcloud-auth.conf
@@ -1,0 +1,15 @@
+#
+
+[INCLUDES]
+
+before = common.conf
+
+[Definition]
+
+prefregex = ^%(__prefix_line)s<F-CONTENT>.+</F-CONTENT>$
+
+failregex = \{core\} Login failed: '.+' \(Remote IP: '<HOST>'\)
+
+ignoreregex =
+
+journalmatch = SYSLOG_IDENTIFIER=ownCloud


config/filter.d/nextcloud-auth.conf
```diff
@@ -0,0 +1,15 @@
...
+
+[INCLUDES]
+
+before = common.conf
+
+[Definition]
+
+prefregex = ^%(__prefix_line)s<F-CONTENT>.+</F-CONTENT>$
+
+failregex = \{core\} Login failed: '.+' \(Remote IP: '<HOST>'\)
```
sebres: The anchors should extend the regex, something like:
```diff
- failregex = \{core\} Login failed: '.+' \(Remote IP: '<HOST>'\)
+ failregex = ^\s*\{core\} Login failed: '.+' \(Remote IP: '<HOST>'\)$
```

config/filter.d/nextcloud-auth.conf
```diff
@@ -0,0 +1,15 @@
+#
+
+[INCLUDES]
+
+before = common.conf
+
+[Definition]
+
+prefregex = ^%(__prefix_line)s<F-CONTENT>.+</F-CONTENT>$
```
sebres: The `prefregex` seems to be unneeded here.

benrubson:
This is what #1646 tries to help with.
Instead of having one Fail2ban filter configuration per web application, web-apps should use default message patterns, as proposed in #1646.
Same as with #1230 :)

benrubson:
#1646 helps with this proposing default patterns, you can add more details to the messages at the end of them, they still will trigger Fail2ban.
Goal is also to avoid having too many different regex.
I remember we added together the `tried to break-in` pattern to the rule, which is rather generic.
And this could of course be updated 👍

PR #1919, opened 2017-10-15
Update pf.conf commet to fix syntax error

IdahoPL:
There is syntax error in achnor definition in pf.conf:

cat /etc/pf.conf
anchor f2b {
  sshd
}

$ sudo service pf reload
Reloading pf rules.
/etc/pf.conf:2: syntax error

New version:

cat /etc/pf.conf
anchor f2b {
  anchor sshd
}

$ sudo service pf reload
Reloading pf rules.


sebres:
I've rebased it to 0.10th branch...

PR #1954, opened 2017-11-03
Add some path workarounds to Arch path configuration

felixonmars:
This avoids errors about missing variables in jail configuration. See downstream report at https://bugs.archlinux.org/task/48540

### DIFF
--- a/config/paths-arch.conf
+++ b/config/paths-arch.conf
@@ -20,6 +20,15 @@ mysql_log = /var/log/mariadb/mariadb.log
 
 roundcube_errors_log = /var/log/roundcubemail/errors
 
+# Workaround missing variables in jail paths
+syslog_authpriv =
+syslog_mail =
+syslog_mail_warn =
+syslog_user =
+syslog_ftp =
+syslog_daemon =
+syslog_local0 =
+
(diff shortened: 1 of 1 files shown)


sebres:
Thanks for the PR!

But let us make regular solution (not a workaround :)). For the current versions I would move this as initial default values into patch-common.conf (what is a only correct place for it).

Please note also that this will work only if the systemd-backend used (where the logpath doesn't matter). Possibly do you know the real paths for the older versions of Arch (still without systemd)? Similar some other distribution we already have the paths? Just to provide non-empty default values for these also...

For previous versions like 0.9 (although no more developing but still actual for many distributions) - the master branch is a good place for it... (you've selected 0.11th branch ATM).
Can you rebase (cherry-pick) this commit and push (--force) to felixonmars:patch-1 branch, and hereafter change the target base to master (using Edit button above)?
Let us know if you are not familiar with (I will do it then)...

sebres:
I've fixed it in #1955, thus close. Thanks again!
If you can provide a real paths for the config/paths-arch.conf (for usage with file-related backend), welcome with new PR :)

PR #1960, opened 2017-11-10
https, "Fail2Ban", other language improvements

comradekingu:
I tried to write "Fail2Ban", and happened upon some of the documentation, and here we are.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,45 +6,45 @@
 
 ## Fail2Ban: ban hosts that cause multiple authentication errors
 
-Fail2Ban scans log files like `/var/log/auth.log` and bans IP addresses having
+Fail2Ban scans log files like `/var/log/auth.log` and bans IP addresses conducting
 too many failed login attempts. It does this by updating system firewall rules
 to reject new connections from those IP addresses, for a configurable amount
 of time. Fail2Ban comes out-of-the-box ready to read many standard log files,
-such as those for sshd and Apache, and is easy to configure to read any log
-file you choose, for any error you choose.
+such as those for sshd and Apache, and is easily configured to read any log
+file of your choosing, for any error you wish.
 
-Though Fail2Ban is able to reduce the rate of incorrect authentications
-attempts, it cannot eliminate the risk that weak authentication presents.
-Configure services to use only two factor or public/private authentication
+Though Fail2Ban is able to reduce the rate of incorrect authentication
+attempts, it cannot eliminate the risk presented by weak authentication.
+Set up services to use only two factor, or public/private authentication
 mechanisms if you really want to protect services.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,45 +6,45 @@
...
-Configure services to use only two factor or public/private authentication
+Though Fail2Ban is able to reduce the rate of incorrect authentication
+attempts, it cannot eliminate the risk presented by weak authentication.
+Set up services to use only two factor, or public/private authentication
 mechanisms if you really want to protect services.
 
-This README is a quick introduction to Fail2ban. More documentation, FAQ, HOWTOs
-are available in fail2ban(1) manpage and on the website http://www.fail2ban.org
+This README is a quick introduction to Fail2Ban. More documentation, FAQ, and HOWTOs
+available in the fail2ban(1) manpage and on the website https://www.fail2ban.org
```
sebres: > in the fail2ban(1) manpage

Hmm, for me it sounds better without the before artificial or even "in the manpage fail2ban(1)". 
Note that `fail2ban(1)` meant really an artificial and this is the typically spelling for the man-page.

README.md
```diff
@@ -6,45 +6,45 @@
...
 -------------
 
-**It is possible that Fail2ban is already packaged for your distribution.  In
-this case, you should use it instead.**
+**It is possible that Fail2Ban is already packaged for your distribution.  In
+this case, you should use that instead.**
 
 Required:
-- [Python2 >= 2.6 or Python >= 3.2](http://www.python.org) or [PyPy](http://pypy.org)
+- [Python2 >= 2.6 or Python >= 3.2](https://www.python.org) or [PyPy](https://pypy.org)
```
sebres: http \-\> https? What is the reason for?

comradekingu: With full HSTS it prevents some MITM attacks. Doesn't otherwise hurt anything but bandwidth and time, in return for some authentication. They have it, so maybe some service on there needs the encryption.

README.md
```diff
@@ -6,45 +6,45 @@
...
 
 Optional:
 - [pyinotify >= 0.8.3](https://github.com/seb-m/pyinotify)
-  - Linux >= 2.6.13
-- [gamin >= 0.0.21](http://www.gnome.org/~veillard/gamin)
-- [systemd >= 204](http://www.freedesktop.org/wiki/Software/systemd)
+- Linux >= 2.6.13
```
yarikoptic: indentation was on purpose to signal that it is needed only for Linux over that version... ATM that ancient specific is probably a non-issue and might be just removed

comradekingu: I'd argue avoiding systemD is still a relevant issue. Not touching.

PR #1961, opened 2017-11-11
New filter suggested: apache-zerotolerance.conf

rafaces:
This filter has rules against:
- Bruteforce attacks (e.g.: phpmyadmin, mysql, pma, xampp)
- Port scanners
- Vulnerability scanners
- PHP, CGI exploiters
- Shellshock

### DIFF
--- a/config/filter.d/apache-zerotolerance.conf
+++ b/config/filter.d/apache-zerotolerance.conf
@@ -1,0 +1,49 @@
+# Fail2Ban filter for:
+# - Bruteforce attacks (e.g.: phpmyadmin, mysql, pma, xampp)
+# - Port scanners
+# - Vulnerability scanners
+# - PHP, CGI exploiters
+# - Shellshock
+#
+# -------------------------------------------------------
+# HOW TO APPLY THIS RULE: JAIL.LOCAL
+#
+# make sure to use a smaller name, so you don't exceed the limit
+# [apache-zerotol]
+#
+# enabled  = true
+# port     = all
+# port     = http,https,ssh
+# filter   = apache-zerotolerance
+# banaction = iptables-allports
+# logpath  = /var/log/apache*/*access*.log
+# maxretry = 0 #zero retries, so the ip gets blocked immediately
+#
+# -------------------------------------------------------
+#
+
+[Definition]
+
+failregex = ^<HOST> -.*\"(GET|POST)\s*(\/admin)?\/(php)?((m|M)y(sql)?|db)?((a|A)dmin\d?)?\/scripts\/(s|S)etup\d*\.php\s*HTTP.*404.*$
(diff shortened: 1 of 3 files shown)


config/filter.d/apache-zerotolerance.conf
```diff
@@ -0,0 +1,48 @@
...
+# banaction = iptables-allports
+# logpath  = /var/log/apache*/*access*.log
+# maxretry = 0 #zero retries, so the ip gets blocked immediately
+#
+# -------------------------------------------------------
+#
+
+[Definition]
+
+failregex = ^<HOST> -.*\"(GET|POST)\s*(\/admin)?\/(php)?((m|M)y(sql)?|db)?((a|A)dmin\d?)?\/scripts\/(s|S)etup\d*\.php\s*HTTP.*404.*$
```
sebres: Well, I known we've already some apache-filters with similar "vulnerable" regex, but let us don't make more of which.
And no capturing groups please, thus `(?: ... )` instead of `( ... )`.
In this case:
```diff
- failregex = ^<HOST> -.*\"(GET|POST)\s* ... \s*HTTP.*404.*$
+failregex = ^<HOST> -[^"]*\"(?:GET|POST)\s* ... \s*HTTP/[^"]+" 404 \d+ "[^\"]" "[^\"]"\s*$ 
```
**We'll not accept unjustified catch-alls any more.**

Additionally you've made this PR for v.0.11, where (effective since v.0.10) the fail2ban accept pre-filtering regex `prefregex` to catch failures like rows at once (resp. to not write the same regexp-parts for each line) and hereafter to parse with failregex only content between tag `<F-CONTENT>...</F-CONTENT>`.
Because the URL in the middle this can be written more effective (without greedy catch-alls in-between), you don't need to parse referrer part after error code at all.

So you can try something like this:
```INI
# req_method = (?:GET|POST|HEAD|...)
req_method = [A-Z]{3,}
prefregex = ^<HOST> -[^"]*\"%(req_method)s\s+<F-CONTENT>\S+</F-CONTENT> HTTP/[^"]+" 40\d\s\d+

failregex = ^(?:\/admin)?\/(?:php)?(?:(?:m|M)y(?:sql)?|db)?(?:(?:a|A)dmin\d?)?\/scripts\/(?:s|S)etup\d*\.php$
            ...
```

And last but not least - don't use apache direct for this purposes:
- deficiency resp. lacking of proper escaping in the apache access-log can bring you many surprises;
- please note the https://github.com/fail2ban/fail2ban/pull/1070#issuecomment-113232258
- may be too heavy for load servers (the access-log may be really large ;)
- etc.

Try nginx with special location for all such failures (e. g. in case of 403, 404 etc), that logs in extra log-file, which can be then observed from fail2ban.

PR #1969, opened 2017-11-26
Update nginx-limit-req filter.

RaidForums:
Nginx log file marks a rate limit as a warn not an error.
At least for my version of nginx, 1.12.2

### DIFF
--- a/config/filter.d/nginx-limit-req.conf
+++ b/config/filter.d/nginx-limit-req.conf
@@ -36,10 +36,10 @@ ngx_limit_req_zones = [^"]+
 # Use following full expression if you should range limit request to specified 
 # servers, requests, referrers etc. only :
 #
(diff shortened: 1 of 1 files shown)


sebres:
You are wrong a bit here - default log-level of message from limit-req-module is still `error`, but you can change it with directive `limit_req_log_level`...
See the nginx-documentation [Default: limit_req_log_level error](http://nginx.org/en/docs/http/ngx_http_limit_req_module.html);

I'm not against to affect the failure for any log-level (inclusive warn), then the regex should be rewritten:
```diff
-failregex = ^\s*\[error\] ...
+failregex = ^\s*\[[a-z]+\] ...
```

sebres:
I've rebased it to master (because unchanged across all versions).

PR #1976, opened 2017-11-29
Update apache-botsearch.conf

ikxdf:


sebres:
The merge of master branch to major branches will done automatically from us (by next change of 0.10 resp. 0.11). Especially if it affects only slight change (a comment in this case).

PR #1977, opened 2017-11-29
Update botsearch-common.conf

ikxdf:
founded lot of phpmyadmin request with a number before or after phpmyadmin string

### DIFF
--- a/config/filter.d/botsearch-common.conf
+++ b/config/filter.d/botsearch-common.conf
@@ -9,7 +9,7 @@ block = \/?(<webmail>|<phpmyadmin>|<wordpress>|cgi-bin|mysqladmin)[^,]*
 # isn't installed
 webmail = roundcube|(ext)?mail|horde|(v-?)?webmail
 
-phpmyadmin = (typo3/|xampp/|admin/|)(pma|(php)?[Mm]y[Aa]dmin)
+phpmyadmin = (typo3/|xampp/|admin/|)(pma|([0-9])?(php)?[Mm]y[Aa]dmin([0-9])?)
 
 wordpress = wp-(login|signup|admin)\.php
 


config/filter.d/botsearch-common.conf
```diff
@@ -9,7 +9,7 @@ block = \/?(<webmail>|<phpmyadmin>|<wordpress>|cgi-bin|mysqladmin)[^,]*
 # isn't installed
 webmail = roundcube|(ext)?mail|horde|(v-?)?webmail
 
-phpmyadmin = (typo3/|xampp/|admin/|)(pma|(php)?[Mm]y[Aa]dmin)
+phpmyadmin = (typo3/|xampp/|admin/|)(pma|([0-9])?(php)?[Mm]y[Aa]dmin([0-9])?)
```
sebres: The construct `([0-9])?` is very strange:
- capturing group, that anyway unexpected by single token, thus `([0-9])?` does the same as `[0-9]?` (or better `\d?`) here.
- I believe this first optional single digit is not really at correct place.

Therefore please something like this:
```diff
-phpmyadmin = (typo3/|xampp/|admin/|)(pma|(php)?[Mm]y[Aa]dmin)
+phpmyadmin = (typo3/|xampp/|admin/|)(pma|(\d?php)?[Mm]y[Aa]dmin\d?)
```
Or still better without captured groups at all:
```diff
-phpmyadmin = (typo3/|xampp/|admin/|)(pma|(php)?[Mm]y[Aa]dmin)
+phpmyadmin = (?:typo3/|xampp/|admin/|)(?:pma|(?:\d?php)?[Mm]y[Aa]dmin\d?)
```
> founded lot of phpmyadmin request with a number before or after phpmyadmin string

Additionally I'll not merge it as long as I don't see any example excerpt from you.
Still better would be to extend the test-case [apache-botsearch](https://github.com/fail2ban/fail2ban/blob/0.11/fail2ban/tests/files/logs/apache-botsearch) resp. nginx-botsearch with matching entries to cover this.
But I can make it self, if you'll provide the log-excerpt.

sebres: ping

PR #1978, opened 2017-11-29
Update apache-botsearch.conf

ikxdf:
webroot init don't consider multi ip or virtualhosts configuration.
scan attack is blocked only for default apache document root.


sebres:
Sorry, but I'll reject this, because:
- it is too vulnerable (as it currently implemented by you, allows "injecting" on url's)
- not the default case (affects virtualhosts configuration only).
- each filter-parameter can be overwritten in jail.local, thus if needed it can be easy configured personally without any modification of the filter, for example:
```INI
# jail.local:
[apache-botsearch]
filter = apache-botsearch[webroot="/var/www/(?:[^/]+/)?"]
```

PR #1979, opened 2017-11-29
Exim failregex: Include lower/mixed case AUTH

peternowee:
When reporting the error `AUTH command used when not advertised`, Exim starts with `SMTP protocol error in "........."`. Here, [Exim logs the SMTP command as it was provided by the connecting client](https://github.com/Exim/exim/blob/exim-4_89+fixes/src/src/smtp_in.c#L2850).

According to [RFC 5321 (SMTP)](https://tools.ietf.org/html/rfc5321#section-2.4) "[..] a command verb [..] MAY be encoded in upper case, lower case, or any mixture of upper and lower case with no impact on its meaning."

Lower case `auth login` brute-force attempts were seen in the wild and were not caught by the current failregex.

This commit makes the failregex case-insensitive for the `AUTH` command, so that lower case (`auth`) or mixed case (`aUtH`) now also match. The failregex was already case-insensitive for the command
arguments (e.g. `AUTH login` already matched).

Additional notes:
- I have tested the new filter and log samples. Everything seems to work correctly and not cause any regressions.
- Unfortunately, `[Aa][Uu][Tt][Hh]` makes the expression less readable. Alternative solutions:
  - Flag the entire failregex as case-insensitive using the flag `(?i)`. This matches as well, but the flag affects the entire regex, making it less restrictive than it currently is. Please let me know if you prefer this solution.
  - Python 3.6 supports setting the ignore case flag for only a part of the regular expression by using `(?i:)`. I have not tested this. See https://stackoverflow.com/questions/1455160/how-to-set-ignorecase-flag-for-part-of-regular-expression-in-python and https://docs.python.org/3/library/re.html.

### DIFF
--- a/config/filter.d/exim.conf
+++ b/config/filter.d/exim.conf
@@ -18,7 +18,7 @@ failregex = ^%(pid)s %(host_info)ssender verify fail for <\S+>: (?:Unknown user|
             ^%(pid)s %(host_info)srejected RCPT [^@]+@\S+: (?:relay not permitted|Sender verify failed|Unknown user|Unrouteable address)\s*$
             ^%(pid)s SMTP protocol synchronization error \([^)]*\): rejected (?:connection from|"\S+") %(host_info)s(?:next )?input=".*"\s*$
             ^%(pid)s SMTP call from \S+ %(host_info)sdropped: too many nonmail commands \(last was "\S+"\)\s*$
-            ^%(pid)s SMTP protocol error in "[^"]+" %(host_info)sAUTH command used when not advertised\s*$
+            ^%(pid)s SMTP protocol error in "[^"]+(?:"+[^"]*(?="))*?" %(host_info)sAUTH command used when not advertised\s*$
             ^%(pid)s no MAIL in SMTP connection from (?:[^\[\( ]* )?(?:\(\S*\) )?%(host_info)sD=\d\S*s(?: C=\S*)?\s*$
             ^%(pid)s (?:[\w\-]+ )?SMTP connection from (?:[^\[\( ]* )?(?:\(\S*\) )?%(host_info)sclosed by DROP in ACL\s*$
 
--- a/config/filter.d/exim.conf
+++ b/config/filter.d/exim.conf
@@ -18,7 +18,7 @@ failregex = ^%(pid)s %(host_info)ssender verify fail for <\S+>: (?:Unknown user|
             ^%(pid)s %(host_info)srejected RCPT [^@]+@\S+: (?:relay not permitted|Sender verify failed|Unknown user|Unrouteable address)\s*$
             ^%(pid)s SMTP protocol synchronization error \([^)]*\): rejected (?:connection from|"\S+") %(host_info)s(?:next )?input=".*"\s*$
             ^%(pid)s SMTP call from \S+ %(host_info)sdropped: too many nonmail commands \(last was "\S+"\)\s*$
-            ^%(pid)s SMTP protocol error in "[Aa][Uu][Tt][Hh] \S*(?: \S*)?" %(host_info)sAUTH command used when not advertised\s*$
+            ^%(pid)s SMTP protocol error in "[^"]+" %(host_info)sAUTH command used when not advertised\s*$
             ^%(pid)s no MAIL in SMTP connection from (?:[^\[\( ]* )?(?:\(\S*\) )?%(host_info)sD=\d\S*s(?: C=\S*)?\s*$
             ^%(pid)s (?:[\w\-]+ )?SMTP connection from (?:[^\[\( ]* )?(?:\(\S*\) )?%(host_info)sclosed by DROP in ACL\s*$
 
--- a/config/filter.d/exim.conf
+++ b/config/filter.d/exim.conf
@@ -18,7 +18,7 @@ failregex = ^%(pid)s %(host_info)ssender verify fail for <\S+>: (?:Unknown user|
             ^%(pid)s %(host_info)srejected RCPT [^@]+@\S+: (?:relay not permitted|Sender verify failed|Unknown user|Unrouteable address)\s*$
             ^%(pid)s SMTP protocol synchronization error \([^)]*\): rejected (?:connection from|"\S+") %(host_info)s(?:next )?input=".*"\s*$
             ^%(pid)s SMTP call from \S+ %(host_info)sdropped: too many nonmail commands \(last was "\S+"\)\s*$
-            ^%(pid)s SMTP protocol error in "AUTH \S*(?: \S*)?" %(host_info)sAUTH command used when not advertised\s*$
+            ^%(pid)s SMTP protocol error in "[Aa][Uu][Tt][Hh] \S*(?: \S*)?" %(host_info)sAUTH command used when not advertised\s*$
             ^%(pid)s no MAIL in SMTP connection from (?:[^\[\( ]* )?(?:\(\S*\) )?%(host_info)sD=\d\S*s(?: C=\S*)?\s*$
             ^%(pid)s (?:[\w\-]+ )?SMTP connection from (?:[^\[\( ]* )?(?:\(\S*\) )?%(host_info)sclosed by DROP in ACL\s*$
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -33,7 +33,10 @@ releases.
 * filter.d/dovecot.conf:
     - fixed failregex, see gh-1879 (partially cherry-picked from gh-1880)
     - extended to match pam_authenticate failures with "Permission denied" (gh-1897)
-* filter.d/exim.conf - fixed failregex for case of flood attempts with `D=0s` (gh-1887)
+* filter.d/exim.conf
+    - fixed failregex for case of flood attempts with `D=0s` (gh-1887)
+    - fixed failregex of "AUTH command used when not advertised" to better handle the foreign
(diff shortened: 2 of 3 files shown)


config/filter.d/exim.conf
```diff
@@ -18,7 +18,7 @@ failregex = ^%(pid)s %(host_info)ssender verify fail for <\S+>: (?:Unknown user|
             ^%(pid)s %(host_info)srejected RCPT [^@]+@\S+: (?:relay not permitted|Sender verify failed|Unknown user|Unrouteable address)\s*$
             ^%(pid)s SMTP protocol synchronization error \([^)]*\): rejected (?:connection from|"\S+") %(host_info)s(?:next )?input=".*"\s*$
             ^%(pid)s SMTP call from \S+ %(host_info)sdropped: too many nonmail commands \(last was "\S+"\)\s*$
-            ^%(pid)s SMTP protocol error in "AUTH \S*(?: \S*)?" %(host_info)sAUTH command used when not advertised\s*$
+            ^%(pid)s SMTP protocol error in "[Aa][Uu][Tt][Hh] \S*(?: \S*)?" %(host_info)sAUTH command used when not advertised\s*$
```
sebres: Please don't do that, just use `(?:AUTH|auth)` or even case-insensitive `(?:(?i)AUTH)` instead of.

peternowee: `(?:AUTH|auth)` won't match mixed case (`AuTh`), which is also valid SMTP.

`(?:(?i)AUTH)` is ok with me, but it results in the entire expression (not limited by the parentheses) becoming case-insensitive, thus making it less restrictive than it was. If you are ok with that, please confirm and I will change it to this.

sebres: You are right (I forgot the non-branching `(?i)` PCRE-variant that used in python...). 
Then leave it as is...
Or what do you think about:
```diff
- SMTP protocol error in "AUTH \S*(?: \S*)?" ... AUTH command used when not advertised
+ SMTP protocol error in "[^"]+" ... AUTH command used when not advertised
```
IMHO, the anchor "AUTH command used when not advertised" is precise enough here, and it will prevent by any user data here (because really forign-input).

I'll merge it later if no other objections follow.

peternowee: Sorry for the late reply, but yes, your proposal is probably best. I agree that the final part is probably specific enough. I did some manual tests and ran `fail2ban-testcases -n` and everything seems to work ok, no regressions.

sebres: BTW only one thing I take care would be potential "injection" like `AUTH login "test"` (with quotes), but currently no time to check this.
If it produces an unescaped output in exim-log, then the expression should be rewritten as more precise...

peternowee: Ouch, they do go through unescaped:

    2017-11-30 06:11:27 SMTP protocol error in "AUTH "LOGIN"" H=localhost (peternowee.com) [127.0.0.1] AUTH command used when not advertised
    2017-11-30 06:12:04 SMTP protocol error in "AUTH LOGIN "test"" H=localhost (peternowee.com) [127.0.0.1] AUTH command used when not advertised

These lines get matched by the current regex `"AUTH \S*(?: \S*)?"`, but not by the new one `"[^"]+"`. The client can enter any long string and Exim will just copy it there, although it does seem to strip trailing spaces. Commands with more than two arguments (e.g. `"AUTH LOGIN " ""`) and just the command `"AUTH"` do not get matched by `"AUTH \S*(?: \S*)?"` either, though these would probably not be used in login attempts anyway.

Perhaps match any string in the middle `".*"` then, and trust the anchors at the start and the end? Here is [a Debuggex using `".*"` in which I let a malicious client (192.0.2.6) try to inject a match for a third-party victim (192.0.2.5)](https://www.debuggex.com/r/tUSSaffQyEfN6pzO). This correctly fails because of the `$` anchoring at the end. I also checked with `fail2ban-testcases`: It sees host 192.0.2.6. Looking at the [Exim source code for these error messages](https://github.com/Exim/exim/blob/exim-4_89+fixes/src/src/smtp_in.c#L2850), I don't expect Exim will produce log lines starting with "SMTP protocol error in" and ending in foreign input.

sebres: > Perhaps match any string in the middle ".*" ....

Nope. Although I don't think that more complex injection could go through (because of the end-anchor). Also with the second foreign input from `helo` logged here, etc. But...
Just "pure" catch-all is always bad, just because the string can be long (thus it slows down the whole handling). I saw too often how the attackers paralyze the whole service with pair lines affected by such expressions.
But because we are forced, I'll add one but not greedy as alternating fallback (no chance otherwise).

config/filter.d/exim.conf
```diff
@@ -18,7 +18,7 @@ failregex = ^%(pid)s %(host_info)ssender verify fail for <\S+>: (?:Unknown user|
             ^%(pid)s %(host_info)srejected RCPT [^@]+@\S+: (?:relay not permitted|Sender verify failed|Unknown user|Unrouteable address)\s*$
             ^%(pid)s SMTP protocol synchronization error \([^)]*\): rejected (?:connection from|"\S+") %(host_info)s(?:next )?input=".*"\s*$
             ^%(pid)s SMTP call from \S+ %(host_info)sdropped: too many nonmail commands \(last was "\S+"\)\s*$
-            ^%(pid)s SMTP protocol error in "[^"]+" %(host_info)sAUTH command used when not advertised\s*$
+            ^%(pid)s SMTP protocol error in "[^"]+(?:"+[^"]*(?="))*?" %(host_info)sAUTH command used when not advertised\s*$
```
peternowee: This is the first time I meet lookarounds and I also do not have much experience with greedy/non-greedy, so this is a bit difficult for me to grasp. It seems the `(?=")` is there to make sure that there is always at least one `"` left later in the string while the engine keeps matching occurrences of `"+[^"]*`. Just some comments:
- I do not have experience with regex performance, but I can imagine there is some benefit in this 'narrower' approach over a catch-all.
- I did wonder if things could go wrong if `" [victimIP]` was injected in a later foreign input field. However, I checked that Exim does not accept `"` in `EHLO` and assume that `I=` (interface) is not foreign input and `U=` (user) will not be filled because we have not passed `AUTH`.
- Tests results are ok.

sebres:
The Noun here is "possessive"...
For more details, see the message of my [commit](https://github.com/fail2ban/fail2ban/pull/1979/commits/4f631806110ee8bdb760b78da5e52632cff3aed8) about this.

> It seems the `(?=")` is there to make sure that there is always at least one `"` left

Not really, because the char `"` anyway expected hereafter. But...
Possessive handling will be used to be greedy (`*`) in non greedy (`*?`) subexpression. Normally possessive syntax is `(RE)++` (the second `+`), but python does not know such syntax.
So lookahead `[^"]*(?=")` used here instead of `[^"]*+` even to force possessive handling.

Example:
```python
# not possessive (match 12 because 123 will range to "test" and it is not allowed):
print(re.search(r"(\d*)(?!test)", "123test").group())
12
# "possessive" variants (here don't match at all, because forbidden to end the number in-between):
print(re.search(r"(\d*(?=[a-z]))(?!test)", "123test").group())
print(re.search(r"(\d*(?!\d))(?!test)", "123test").group())
# only here we'll find match:
print(re.search(r"(\d*(?=[a-z]))(?!test)", "123xxxx").group())
123
print(re.search(r"(\d*(?!\d))(?!test)", "123xxxx").group())
123
```

Thus the fallback `(?:"+[^"]*(?="))*?...$` looks like:
- search non-greedy until `...$` not reached,
- thereby bypass any count of `"` with `"+`;
- and search **possessive** up to next `"`;
- repeat until non-greedy match (corresponding the rest `...$`) does not took place.

As regards the performance, the possessive handling inside non-greedy expressions is barely visible on short strings, but sometimes very much faster on very long strings (they could be used by the attackers just to slow down fail2ban processing).

sebres:
Can you please add a changelog entry (e. g. [ChangeLog-Edit](https://github.com/peternowee/fail2ban/edit/fix-exim-lowercase-auth/ChangeLog?pr=/fail2ban/fail2ban/pull/1979))
Then I'll merge it. Thx for your contribution!

peternowee:
Ok, I think I am starting to understand it a little bit now. I found this [example of emulating possessive quantifiers](https://stackoverflow.com/questions/5537513/emulating-possessive-quantifiers/5545615#5545615) and was wondering why you did not retrieve the lookahead to consume it (`\1` in the example). As I now understand it, that is because in this case the lookahead value `"` is consumed at either the next loop of the non-greedy group or by the closing `"` that follows the group.

For future reference, here is some test code that helped me understand better. It uses escaped quotes `\"` and a capturing group retrieved by `group(1)`:

    >>> print(re.search(r"^\"[^\"]+(\"+[^\"]*(?=\"))*?\"$", "\"AUTH LOGIN\"abc\"\"").group(1))
    "
    >>> print(re.search(r"^\"[^\"]+(\"+[^\"]*(?=\"))*?\"$", "\"AUTH LOGIN\"\"abc\"").group(1))
    ""abc

Thanks for the advanced regex lesson :-) I will edit the changelog now.

sebres:
> For future reference, here is some test code that helped me understand better.

I don't wanted to confuse you with so many quotation marks in the example expressions 😄

sebres:
Thus merged, thanks!

peternowee:
My pleasure, thank you too!

Issue #1980, opened 2017-11-30
allow logging to a different syslog facility


anarcat:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.9.6-2
- OS, including release name/version: Debian GNU/Linux stretch 9.1 x86_64
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

### The issue:

It would be nice to be able to change the syslog facility used by fail2ban. Right now, it is hardcoded to `DAEMON` in the source code, which is a fine default for most daemons. But considering that fail2ban often (always?) deals with authentication systems, I feel that `AUTH` is a more appropriate value here. Indeed, here is the documentation of facilities from [syslog(3)](https://manpages.debian.org/syslog.3)

       LOG_AUTH       security/authorization messages

       LOG_AUTHPRIV   security/authorization messages (private)

       LOG_CRON       clock daemon (cron and at)

       LOG_DAEMON     system daemons without separate facility value

It would seem more relevant to use `LOG_AUTH`. Not wanting to create a huge bikeshed discussion about such a cosmetic change, I figured it was better to just make that customizable. Some users, for example, may *only* use fail2ban for mail purposes so they may prefer to see the logs end up in `LOG_MAIL`. Or maybe they want to use a fail2ban-specific facility to filter logs on the server, using `LOG_LOCAL0`.

Having that flexibility would be very useful in some configurations.

#### Steps to reproduce

I was looking for entries in /var/log/auth.log, found them in /var/log/daemon.log instead, which makes it hard to correlate with `sshd` logs.

#### Expected behavior

I would expect to see a `logfacility = DAEMON` line in the configuration, it could look something like this:

```
# Option: logfacility
# Notes.: Sets the syslog "facility". 
# syslog facilities allow the syslog daemon to sort output to different logfiles or servers.
# Allowed values are defined by the Python syslog handler, define here:
# https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SysLogHandler.encodePriority
# Values: [ AUTH | DAEMON | LOCAL0 | ... ]  Default: DAEMON
#
logfacility = DAEMON
```

#### Observed behavior

No such configuration exists, obviously.

#### Any customizations done to /etc/fail2ban/ configuration

This is obviously relevant only with the following configuration:

```
logtarget = SYSLOG
```

Thanks!

[...]

PR #1989, opened 2017-12-05
New server logging options

sebres:
Introduced new parameters for logging within fail2ban-server.
Usage `logtarget = target[facility=..., datetime=on|off, format="..."]`:
  - `facility` - specify syslog facility (default `daemon`, see https://docs.python.org/2/library/logging.handlers.html#sysloghandler for the list of facilities);
  - `datetime` - add date-time to the message (default on, ignored if `format` specified);
  - `format` - specify own format how it will be logged, for example for short-log into STDOUT:
```bash
fail2ban-server -f --logtarget 'stdout[format="%(relativeCreated)5d | %(message)s"]' start
```
Closes gh-1980

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -68,6 +68,13 @@ ver. 0.10.2-dev-1 (2017/??/??) - development edition
 ### Enhancements
 * jail.conf: extended with new parameter `mode` for the filters supporting it (gh-1988);
 * action.d/pf.conf: extended with bulk-unban, command `actionflush` in order to flush all bans at once.
+* Introduced new parameters for logging within fail2ban-server (gh-1980).
+  Usage `logtarget = target[facility=..., datetime=on|off, format="..."]`:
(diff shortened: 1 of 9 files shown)


sebres:
If someone needs more "pretty" option settings, e. g. some distribution uses `syslog` as default target, the following enhancement can be done in `fail2ban.conf` (to see like suggested in #1980):
```INI
# Option: logfacility
# Notes.: Sets the syslog "facility". 
# syslog facilities allow the syslog daemon to sort output to different logfiles or servers.
# Allowed values are defined by the Python syslog handler, define here:
# https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SysLogHandler.encodePriority
# Values: [ AUTH | DAEMON | LOCAL0 | ... ]  Default: DAEMON
#
logfacility = DAEMON

# Option: logtarget ...
logtarget = SYSLOG[facility=%(logfacility)s]
```

PR #2011, opened 2018-01-03
 Fix filter not catching asterisk requests with quote character in username (fixes #2010)

### LINKED ISSUE #2010: fail2ban doesn't catch asterisk log entrys with the quote character in the username
### Environment:
Debian stretch

### The issue:
`filter.d/asterisk.conf` doesn't catch attacks with the `'` (quote) character in the username.
```

Lines: 3 lines, 0 ignored, 0 matched, 3 missed
[processed in 0.01 sec]

|- Missed line(s):
|  [2017-12-14 22:18:00] NOTICE[1943] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:'or''='@31.18.206.92>' failed for '185.107.94.10:43678' (callid: UmOkE9yQPGOsF3Az24YTRe..) - No matching endpoint found
|  [2017-12-14 22:21:01] NOTICE[1943] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:a'or'3=3--@31.18.206.92>' failed for '185.107.94.10:19379' (callid: DGiug8PmgZ352viPITGcVP..) - No matching endpoint found
|  [2017-12-14 22:23:35] NOTICE[1943] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:a'or'3=3--@31.18.206.92>' failed for '185.107.94.10:25673' (callid: V1XIdd5xBn3xt14E3gJ9ZQ..) - No matching endpoint found
```
#### Steps to reproduce
Use fail2ban-regex on the above lines for the asterisk filter
#### Expected behavior
The lines should be caught.

Yannik:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,7 @@ releases.
     - fixed failregex AMI Asterisk authentification failed (see gh-1302)
     - removed invalid (vulnerable) regex blocking IPs using forign data (from header "from")
       thus not the IP-address that really originates the request (see gh-1927)
+    - fixed failregex for the SQL-injection attempts with single-quotes in connect-string (see gh-2011)
 * filter.d/dovecot.conf:
     - fixed failregex, see gh-1879 (partially cherry-picked from gh-1880)
     - extended to match pam_authenticate failures with "Permission denied" (gh-1897)
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -25,7 +25,7 @@ failregex = ^%(__prefix_line)s%(log_prefix)s Registration from '[^']*' failed fo
             ^%(__prefix_line)s%(log_prefix)s hacking attempt detected '<HOST>'$
             ^%(__prefix_line)s%(log_prefix)s SecurityEvent="(?:FailedACL|InvalidAccountID|ChallengeResponseFailed|InvalidPassword)"(?:(?:,(?!RemoteAddress=)\w+="[^"]*")*|.*?),RemoteAddress="IPV[46]/(UDP|TCP|WS)/<HOST>/\d+"(?:,(?!RemoteAddress=)\w+="[^"]*")*$
             ^%(__prefix_line)s%(log_prefix)s "Rejecting unknown SIP connection from <HOST>"$
-            ^%(__prefix_line)s%(log_prefix)s Request (?:'[^']*' )?from '[^']*' failed for '<HOST>(?::\d+)?'\s\(callid: [^\)]*\) - (?:No matching endpoint found|Not match Endpoint(?: Contact)? ACL|(?:Failed|Error) to authenticate)\s*$
+            ^%(__prefix_line)s%(log_prefix)s Request (?:'[^']*' )?from '.*' failed for '<HOST>(?::\d+)?'\s\(callid: [^\)]*\) - (?:No matching endpoint found|Not match Endpoint(?: Contact)? ACL|(?:Failed|Error) to authenticate)\s*$
 
 # FreePBX (todo: make optional in v.0.10):
 #            ^(%(__prefix_line)s|\[\]\s*WARNING%(__pid_re)s:?(?:\[C-[\da-f]*\])? )[^:]+: Friendly Scanner from <HOST>$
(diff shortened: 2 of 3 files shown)


sebres:
Don't like catch-alls in regexp, but interesting is the question: what does this "attacker"?
Can be the user name `'or''='` correct? Trying sql-injection? Or it was just your own test?

Yannik:
It is a SQL injection attempt.

It could possibly also be a correct user name, but it's not used as such in this case.

What do you recommend? We definitely need to fix this.

sebres:
> What do you recommend? We definitely need to fix this.

Well, if the space would be impossible, then `'\S*'`...
But I think, it will affect us again later... 
So, please change it in following:
```diff
-'[^']*'
+'(?:[^']*|.*?)'
```
This would be a bit safer (at least it allows faster parse of 99% all wrong requests without single quote).

Yannik:
@sebres I have modified my PR according to your suggestion.

Just out of curiosity: Does this really have a significant performance impact?

sebres:
> Just out of curiosity: Does this really have a significant performance impact?


Well, greedy catch-all would be unsafe here, and non-greedy tends to be slower if the string will grow...

Here is a result of this interpolated regexp with simple test case:
```python
for sub in (r"(?:[^']*|.*?)", r".*?"):
  print('%20s : %r' % (sub, timeit.timeit('''m = r.search("[] NOTICE[32497] res_pjsip/pjsip_distributor.c: Request from '317 <sip:317@1.2.3.4>' failed for '2.3.4.5:5089' (callid: 206f178f-896564cb-57573f49@1.2.3.4) - No matching endpoint found");''', setup='''import re; fr = r"""^(?:\[\])?\s*(?:<[^.]+\.[^.]+>\s+)?(?:\S+\s+)?(?:kernel: \[ *\d+\.\d+\]\s+)?(?:@vserver_\S+\s+)?(?:(?:(?:\s*\[\d+\])?:\s+[\[\(]?asterisk(?:\(\S+\))?[\]\)]?:?|[\[\(]?asterisk(?:\(\S+\))?[\]\)]?:?(?:\s*\[\d+\])?:?)\s+)?(?:\[ID \d+ \S+\]\s+)?(?:NOTICE|SECURITY|WARNING)(?:\s*\[\d+\]):?(?:\[C-[\da-f]*\])? [^:]+:\d*(?:(?: in)? \w+:)? Request (?:'[^']*' )?from '%s' failed for '(?P<host>(?:(?:::f{4,6}:)?(?P<ip4>(?:\d{1,3}\.){3}\d{1,3})|\[?(?P<ip6>(?:[0-9a-fA-F]{1,4}::?|::){1,7}(?:[0-9a-fA-F]{1,4}|(?<=:):))\]?|(?P<dns>[\w\-.^_]*\w)))(?::\d+)?'\s\(callid: [^\)]*\) - (?:No matching endpoint found|Not match Endpoint(?: Contact)? ACL|(?:Failed|Error) to authenticate)\s*$"""; r = re.compile(fr)''' % (sub,), number=1000000)))
```
```
       (?:[^']*|.*?) : 6.898367166519165
                 .*? : 7.472841024398804
```
But, as already said, it will be still worse if the string becomes larger, e. g. attacker can make multiple connection to flood fail2ban with "bad" (long) messages (manipulated foreign data to produce larger info, logged from asterisk), which could cause extremely slow parsing using non-greedy regexp's.

Yannik:
@sebres Thanks alot for that explanation, it's certainly interesting. 

Can you merge this change now? Somehow, implementing your change made the tests fail on python 3.3 and 3.4 - I don't know if this is related to this or has another reason.

sebres:
> Can you merge this change now? 

Yes, but:
- I'll let it rest a bit for possible objections from @fail2ban/contributors...
- You've deleted one test, please restore it (I can do it later by merge also).

> Somehow, implementing your change made the tests fail on python 3.3 and 3.4 - I don't know if this is related to this or has another reason.

This was again a [BadIPs-test cases](https://travis-ci.org/fail2ban/fail2ban/jobs/326371681#L767) (BadIPs-server overloaded, timeout, etc).
This occurs often and I want long time disable this test-cases... Still because of possible coverage decrease remained.

Yannik:
@sebres I removed that test-case because it was an exact duplicate. Why wouuld you keep it?

sebres:
> because it was an exact duplicate. Why would you keep it?

So you cover now the obscure case (`'or''=`), but not the "normal" case.
Additionally this PR affects one special case, therefore in the best case I'll see only this test case affected in the test-cases.

> Additional question: Will this be applied to previous releases (as it is highly security related) or only be released in an upcoming version?

I'll not release master branch anymore (development for 0.9 is discontinued), thus only 0.10 or 0.11 will be released.
But the maintainers of corresponding distribution can make self a patch/update for each own release.
We're not responsible what they do.

Yannik:
@sebres Which test cases are you talking about? I deleted 2 test cases and added one, so I agree with keeping the normal case, but I would still delete one, because the 2 were duplicates already.

Yannik:
This commit keeps the normal test case.

PR #2015, opened 2018-01-11
nginx-http-auth: match usernames with spaces

BenediktSeidl:
> - [ ] **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement
>      against 0.9.x series, choose `master` branch

not sure about this one. this could also be backported to other versions, the rest of your checklist should be ok.

I'm not a regex expert, I hope `.+` is not a dangerous operation in this context?

### DIFF
--- a/config/filter.d/nginx-http-auth.conf
+++ b/config/filter.d/nginx-http-auth.conf
@@ -4,7 +4,7 @@
 [Definition]
 
 
-failregex = ^ \[error\] \d+#\d+: \*\d+ user "\S+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
+failregex = ^ \[error\] \d+#\d+: \*\d+ user "[^"]+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
 
 ignoreregex = 
 
--- a/config/filter.d/nginx-http-auth.conf
+++ b/config/filter.d/nginx-http-auth.conf
@@ -4,7 +4,7 @@
 [Definition]
 
 
-failregex = ^ \[error\] \d+#\d+: \*\d+ user "[^"]+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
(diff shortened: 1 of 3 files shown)


config/filter.d/nginx-http-auth.conf
```diff
@@ -4,7 +4,7 @@
 [Definition]
 
 
-failregex = ^ \[error\] \d+#\d+: \*\d+ user "\S+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
+failregex = ^ \[error\] \d+#\d+: \*\d+ user ".+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
```
sebres: > I'm not a regex expert, I hope `.+` is not a dangerous operation in this context?

Unfortunately - yes it is, although it is anchored both sides, but because it takes place before `<HOST>` and because of another catch-all's (e. g. `was not found in ".*"`), it is a bit vulnerable and I'll not accept it, so as you have built it currently.

Thus, please change:
```diff
- ".+"
+ "[^"]+"
```

BenediktSeidl: unfortunately if you input `asd"asd` as a user name in the basicauth dialog, the nginx error log will contain:

```
2018/01/11 19:45:26 [error] 1120#1120: *1597 user "asd"asd" was not found in "xxx.htpasswd", client: 127.0.0.1, server: example.com, request: "GET / HTTP/1.1", host: "example.com"
```

Should I change it anyway?

sebres: Could you really imagine a user name with quote inside, like `asd"asd`?

sebres: A bit thinking about this, if we could agree, that the file-name with passwords not really usable with quote, then we could change it like:
```diff
-failregex = ^ \[error\] \d+#\d+: \*\d+ user "\S+":? (password mismatch|was not found in ".*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
+failregex = ^ \[error\] \d+#\d+: \*\d+ user "(?:[^"]+|.+)":? (password mismatch|was not found in "[^"]*"), client: <HOST>, server: \S*, request: "\S+ \S+ HTTP/\d+\.\d+", host: "\S+"(, referrer: "\S+")?\s*$
```

sebres:
I've extended it a bit and rebased to the master-branch.

PR #2022, opened 2018-01-19
0.11: fix get current bans (after upgrade DB)

sebres:
- upgrade database: update new created table `bips` (combined bad-ips) with entries from table `bans` (allows restore current bans after upgrade from version <= 0.10);
- algorithm of restore current bans after restart changed: update the restored ban-time (and therefore end of ban) of the ticket with ban-time of jail (as maximum), for all tickets with ban-time greater (or persistent); not affected if configured ban-time of the jail is unchanged between stop/start;
- test-cases extended and fixed.


sebres:
extended within f342ecb2a5bae8e764241d03e525b214218f29f4:
- differentiate between unknown ban-time after upgrade (-2) and persistent ban-time (-1);
- better handling to correct current ban-time, uses `bantime.maxtime` (if increment allowed) or `bantime` of jail;
- extended test-cases in order to cover the change of `bantime.maxtime`.

Issue #2024, opened 2018-01-21
A collection of tests failing while testing during debian package build


yarikoptic:
yet to look in detail and most likely "my fault" one way or another. Just wanted to report first in case @sebres has an immediate clue. Cheers

```shell
..
cd build && LC_ALL=C.UTF-8 FAIL2BAN_CONFIG_DIR="/home/yoh/deb/gits/fail2ban/config" PYTHONPATH="/home/yoh/deb/gits/fail2ban/.pybuild/pythonX.Y_3.6/build/" scripts-*/fail2ban-testcases --no-network
......................s.........................................................................sssss.........................s.................................ss....................................................F...sss.............s......sssssssssssssss...............................................................................................EFail2ban 0.10.2 test suite. Python 3.6.4 (default, Jan  5 2018, 02:13:53) [GCC 7.2.0]. Please wait...
........
..Exception in thread _TestCaseWorker:  
Traceback (most recent call last):
  File "/home/yoh/deb/gits/fail2ban/.pybuild/pythonX.Y_3.6/build/fail2ban/client/fail2banserver.py", line 186, in start
    raise ServerExecutionException('Async configuration of server failed')
fail2ban.client.fail2bancmdline.ServerExecutionException: Async configuration of server failed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/yoh/deb/gits/fail2ban/.pybuild/pythonX.Y_3.6/build/fail2ban/tests/fail2banclienttestcase.py", line 410, in _testStartForeground
    self.execCmd(SUCCESS, ("-f",) + startparams, "start")
  File "/home/yoh/deb/gits/fail2ban/.pybuild/pythonX.Y_3.6/build/fail2ban/tests/fail2banclienttestcase.py", line 401, in execCmd
    (self.exec_command_line[1:] + startparams + args))
  File "/usr/lib/python3.6/unittest/case.py", line 733, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/lib/python3.6/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
[...]

PR #2025, opened 2018-01-22
A number of fixes toward making tests pass while building Debian pkg for 0.10.2

yarikoptic:

- [x] **LIST ISSUES** this PR resolves:
       I am going after the ones I listed in #2024:
       - Allow again to test "out of source" by specifying `FAIL2BAN_CONFIG_DIR` env var
- [x] **MAKE SURE** this PR doesn't break existing tests - that is what I am doing by submitting a PR


yarikoptic:
Replied in https://github.com/fail2ban/fail2ban/issues/2024#issuecomment-359452823 -- I am testing against "our" test-configs AFAIK.

setup.py
```diff
@@ -208,8 +208,10 @@ def run(self):
 	license = "GPL",
 	platforms = "Posix",
 	cmdclass = {
-		'build_py': build_py, 'build_scripts': build_scripts, 
-		'install_scripts': install_scripts_f2b, 'install': install_command_f2b
+		'build_py': build_py_2to3 if _2to3 else build_py,
```
sebres: I'm against this change, sorry:
- before install the caller can easy execute 2to3 to do the same;
- explicit execution of 2to3 during test-cases may be unwanted, e. g. because...
- I use incremental 2to3 in my own test-cases (only changed files), so for me it is very worse enhancement as regards the performance. Just compare:
```
fail2ban-testcases -n '[Ss]etup'
- Fail2ban 0.10.3.dev1 test suite. Python 2.7.12+ (default, Sep 17 2016, 12:08:02) [GCC 6.2.0 20160914]. Please wait...
+ Fail2ban 0.10.3.dev1 test suite. Python 3.5.2+ (default, Sep 22 2016, 12:18:14) [GCC 6.2.0 20160927]. Please wait...
- Ran 2 tests in 1.939s
+ Ran 2 tests in 23.589s
```
And think about it does nothing (already converted via my incremental 2to3).

I ask for permission to revert it. :)

yarikoptic: sshure... since I was the only one whining about this one, I think not that much harm would be done and for my own (debian pkg) purposes I could patch it up.  Ideally though, someone should just make fail2ban natively support both 2 and 3 without this old-school 2to3 conversion.  In a year or two it could just switch to python3 overall I think ;-)

sebres: > I was the only one whining about this one

I would not say it's a large win: all the same it is just `2to3; ...` before setup/fail2ban-testcases call's :)

sebres: > someone should just make fail2ban natively support both 2 and 3 without this old-school 2to3 conversion

It is very different on some places, and unfortunately some code-pieces cannot be written backwards-compatible to both version and cannot be `if`-ed around (you'll get compile errors either in py-2x or in py-3x).

But we can create new read-only branch's for 3x and I'll put there the result of 2to3 (resp. merge this changes later periodically).

yarikoptic: hm... could you point me to any of such pieces?

sebres: > hm... could you point me to any of such pieces?

Too many, but OK - here you go:
```python
if sys.version_info < (3,):
  class ActionBase(object):
    __metaclass__ = ABCMeta
    pass
else:
  class ActionBase(object, metaclass=ABCMeta):
    pass
```
You cannot write such manner, because then in python2.x you'll get "SyntaxError: invalid syntax" during compile-time (not run-time error).
And unfortunately, as already said, it is valid for too many syntax-variation's between both versions (some things going in py2x, some only in py3x).

This is really annoying (I mean not the changes in syntax, but such error's by loading) and we can say "Thanks, Guido!" for it. As I criticized it at some point (too long ago), the discussion was just closed at that time (how many times before based on the motto "no discussions").
I really hate this in python (and e. g. therefore love TCL).

sebres: BTW you/we can leave it (I mean `build_py_2to3`) in your or other distribution-related branches, like `debian`.
`setup.py` is very rarely changed (so I think it will not produce the conflicts often).

yarikoptic: yeah, don't sweat over build_py_2to3 -- I will keep it in a patch within debian branch, as it is now: https://github.com/fail2ban/fail2ban/blob/debian/debian/patches/0004-BF-use-build_-_2to3-if-in-_2to3.patch  (I will just will minimize it and make "permanent" by renaming away from cherry-picked 00*.patch)

Re python 2 and 3 compatible code -- the easiest way is to use a helper library, such as six (my regular) or future (not familiar much).  As for metaclasses -- here is the recipe http://python-future.org/compatible_idioms.html#metaclasses . Overall, in my experience, I haven't yet heard about any particular python2 construct forbidding maintaining compatibility between python 2 and 3.

BTW -- found some `debian-0.10` branch in the main repo -- seems to be not needed any longer, right? ;)

sebres: Well back to the thema "backwards-compatibility", now it will be impossible to use some previously allowed variable names, see the [last broken build here](https://travis-ci.org/fail2ban/fail2ban/jobs/332666245#L777).

yarikoptic:
Well, more reserved keywords. Annoying but not a showstopper

PR #2027, opened 2018-01-23
Minor spelling typos etc

yarikoptic:
Very minor changes (only 1 functional so far) while skimming through codebase

- [x] typos and minor rewording in few sentences
- [x] replaced dict with a tuple for `COND_FAMILIES` since no dict anyhow needed
- [x] Also dos2unix'ed config/action.d/helpers-common.conf


fail2ban/server/action.py
```diff
@@ -383,7 +383,7 @@ def _executeOperation(self, tag, operation, family=[]):
 		except ValueError as e:
 			raise RuntimeError("Error %s action %s/%s: %r" % (operation, self._jail, self._name, e))
 
-	COND_FAMILIES = {'inet4':1, 'inet6':1}
+	COND_FAMILIES = ('inet4', 'inet6')
```
sebres: I had although some branch where `if c in COND_FAMILIES: ...`, but not find it anymore. I can change it back, if then will be need.

yarikoptic: hm, I thought I have listened to the talk by some pythonista that for little sequences lie this, list or tuple should outperform for `in` operation the fancier 'dict' or 'set' since no hashing would be needed... upon a quick try in ipython (for both python2 and 3) -- not really.  well -- it does only if there is a direct hit in the first element (i.e. if doing `'inet4' in COND_FAMILIES`). good to know.

sebres: > ... only if there is a direct hit in the first element

Depend on length of compared strings. Also dict-object uses numeric hashes in comparison (which will be calculated once, and persists as long the objects exists).

Issue #2053, opened 2018-02-25
Several badips.py actions, but ban only one time


benrubson:
Hi,

Let's imagine the following configuration :
```
action = %(action_)s
         badips.py[actname="b1", category="ssh", banaction="%(banaction)s", score="3", age="24h"]
         badips.py[actname="b2", category="ssh", banaction="%(banaction)s", score="0", age="48h", bankey="01234ABCDE"]
```

Goal is to retrieve and ban "community" bans with score 3, and retrieve and ban _all_ my bans.
It works perfectly.
However, when action is launched to ban an IP newly detected by fail2ban, 2 banips API calls are made to report the IP, which is useless and counter-productive.

I did not found a way to keep both badips.py actions, but report only once.

Any way ?

Thank you very much 👍 

Ben

benrubson:
I went with the following workaround :
- a jail which bans and report ;
- a _dummy_ jail with no filter, but which retrieves all "community" bans (see #2056) ;
- a second _dummy_ jail with no filter, which retrieves my bans.

So feel free to close this is there is no other solution.

Thx 👍

sebres:
I believe specifying of 2 actions here is wrong.
If you really need it, rather `badips.py` should be extended to act with different set of parameters on local/community bans.
I mean something like:
```INI
action = %(action_)s
         badips.py[category="ssh", banaction="%(banaction)s", score="3", age="24h", local-score="0", local-age="48h" ....]
```

But, I don't understand at all, why do you need to "retrieve and ban all my bans" via badips.
Fail2ban has own mechanism to hold (e. g. restore after restart) all active bans (in local sqlite database), which works well.
Could you explain please for which purposes do you need it exactly?

benrubson:
I share the same `bankey` on several servers.
As soon as one server bans an IP, the IP is retrieved by all the other servers thanks too the `bankey` and `score=0`, which is really convenient 👍 
However, I don't want to retrieve all other "community" with score 0, this is why I use 2 badips.py actions.

sebres:
Admittedly I don't think the preventive banning (without any attempt on the other servers) is a good idea at all (because it just clutters your ip-sets/ip-tables with many unneeded values, which will possible never do any attempt on all this servers)... 
Surely do you know, that this makes your whole network sub-system slower?

But I'll check later, what we can do here.

benrubson:
Don't bother with this and feel free to close if I'm the only requester, as I implemented solution described in my second post.

[...]

PR #2054, opened 2018-02-25
(Free)BSD IPFW does not allow 2 identical rules

benrubson:
Hi,

On Linux, `iptables` allows to have same rule added multiple times.
However, on (Free)BSD, `ipfw` does not allow this.

This PR then handles this.
It will make `actionban` to succeed if rule already exists.
And as a result, `actionunban` to succeed if rule does not exist anymore.

Thank you 👍 

Ben

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,7 @@ ver. 0.10.3-dev-1 (20??/??/??) - development edition
 
 ### Fixes
 * `filter.d/exim.conf`: failregex extended - SMTP call dropped: too many syntax or protocol errors (gh-2048);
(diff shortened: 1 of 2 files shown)


benrubson:
I encountered the need for this PR with the following configuration :
```
action = %(action_)s
         badips.py[category="ssh", banaction="%(banaction)s"]
```
As here there will be 2 bans, one by `action_`, which is `banaction`, the other by `badips.py`.

I've come with a workaround explained in #2053, so not sure the configuration above is good and then this PR is required, or the configuration above is not the way to go and then this PR should be ignored (but perhaps other configurations could lead to have the same rule added at least 2 times).

benrubson:
Perfect thank you Sergey 👍

PR #2056, opened 2018-02-26
Add 'any' badips.py bancategory

benrubson:
Hi,

This PR adds the `any` badips bancategory, which can be useful, for example to add all IPs of a desired score, in a global firewall lookup table.

Thank you 👍 

Ben


benrubson:
Just added changelog entry.

benrubson:
I just resolved conflits.
Of course you can squash-merge if you want to have a clean commit log 👍 
Thx !

sebres:
I assume it is still not completely correct so.
The wrong case you still have with this PR is - it will fail by ban:

Just try:
```bash
$ curl -s http://www.badips.com/add/any/192.0.2.1
{"err":"invalid category selected: any","suc":""}
```
So you can get list for `any`, but it will immediatelly fail by ban (cannot be used for ban purposes).
Thus I cannot accept the PR as it is.

benrubson:
You can't add to `any` category.
However you can retry `any` bancategory.
This is why this PR adds `any` only to the `bancategory` option.
However `any` will not be accepted as a `category` option.
So this is OK :)

benrubson:
You can then use for example `category=ssh`, `bancategory=any`.

sebres:
Yeah, right.

benrubson:
Thx 👍

PR #2058, opened 2018-02-26
iptables-allports should not care about protocol

benrubson:
Hi,

As with other firewall actions on all ports, `iptables` should not care about protocol when banning on all ports.

Thank you 👍 

Ben


sebres:
I don't think it is a good idea:

- connection-less protocols provide a simple way for spoofing of IP and/or imitation of illegal actions;
- banning of UDP (as well as other connection-less protocols) should be configured very careful (therefore I'm definitely against this per default);
- if you want this nevertheless for some jails, you can just specify this:
```INI
protocol = all
# or :
banaction = iptables-allports[protocol=all]
```

If some older versions of iptables do not support `--protocol all`, we can extend action `iptables-allports` to add `-p` only if protocol specified (not empty).

benrubson:
`banaction = iptables-allports[protocol=all]` is definitely the way to go, without having to modify current stock actions.
Let's then close this, thx for your feedback 👍

PR #2059, opened 2018-02-26
badips.py, solve a str() issue - expected string, IPAddr found

benrubson:
Hi,

When banning, `badips.py` leads to the following error under FreeBSD :
```
'ActionInfo(...)': sequence item 3: expected string or Unicode, IPAddr found
  File "/usr/local/lib/python2.7/site-packages/fail2ban/server/actions.py", line 399, in __checkBan
    action.ban(aInfo)
  File "/usr/local/etc/fail2ban/action.d/badips2.py", line 371, in ban
    url = "/".join([self._badips, "add", self.category, aInfo['ip']])
```

This PR then solves this issue.

Thx 👍 

Ben

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -44,6 +44,7 @@ ver. 0.10.3-dev-1 (20??/??/??) - development edition
 * possibility to specify own regex-pattern to match epoch date-time, e. g. `^\[{EPOCH}\]` or `^\[{LEPOCH}\]` (gh-2038);
   the epoch-pattern similar to `{DATE}` patterns does the capture and cuts out the match of whole pattern from the log-line,
   e. g. date-pattern `^\[{LEPOCH}\]\s+:` will match and cut out `[1516469849551000] :` from begin of the log-line.
+* `action.d/badips.py`: implicit convert IPAddr to str, solves an issue "expected string, IPAddr found" (gh-2059);
 
 
 ver. 0.10.2 (2018/01/18) - nothing-burns-like-the-cold
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -368,7 +368,7 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 			Any issues with badips.com request.
 		"""
 		try:
-			url = "/".join([self._badips, "add", self.category, aInfo['ip']])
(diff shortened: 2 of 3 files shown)


ChangeLog
```diff
@@ -36,6 +36,7 @@ ver. 0.10.3-dev-1 (20??/??/??) - development edition
 
 ### Fixes
 * `filter.d/exim.conf`: failregex extended - SMTP call dropped: too many syntax or protocol errors (gh-2048);
+* `action.d/badips.py`: solve an issue which made ban to fail under FreeBSD (gh-2059);
```
sebres: Good catch, but ...

something more precise, like:
```diff
-* `action.d/badips.py`: solve an issue which made ban to fail under FreeBSD (gh-2059);
+* `action.d/badips.py`: implicit convert IPAddr to str, solves an issue "expected string, IPAddr found" (gh-2059);
```
would be appreciated here.

sebres:
This is not really FreeBSD issue (introduced as we implement IPv6-support, so with IPAddr instead of str), 
just wondering why it has not affect the test-cases previously (I thought badips is test-covered).

benrubson:
I don't have this issue on Linux.
Linux : Python 2.7.13
FreeBSD : Python 2.7.14

ChangeLog
benrubson: Done 👍

benrubson:
Mmmh but I run  0.9.6-2 on Linux Debian 9, and 0.10.1 on FreeBSD.

sebres:
Yeap, v. 0.9 does not have support of IPv6.

I assume you'll get the same error on linux (with F2B >= 0.10), if you'll execute this one:
```bash
cd /tmp/test-fail2ban-0.10
PYTHONPATH=. python -c 'from fail2ban.server.ipdns import IPAddr; ip = IPAddr("1.2.3.4"); print(":".join(["test", ip]))'
```

The question is why this is not occurred in our CI previously?

benrubson:
You're right, on a Linux machine with 0.10.1 :
```
  File "<string>", line 1, in <module>
TypeError: sequence item 1: expected string, IPAddr found
```

benrubson:
Thx 👍

PR #2061, opened 2018-03-02
fix sshd rule

MatthieuBarbu:
just remove the space before ":11" line 52 because don't match on my Debian 9 stretch with version 0.10.3 of fail2ban
I don't know if this is wrong on all OS

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,6 +37,7 @@ ver. 0.10.3-dev-1 (20??/??/??) - development edition
 ### Fixes
 * `filter.d/asterisk.conf`: fixed failregex prefix by log over remote syslog server (gh-2060);
 * `filter.d/exim.conf`: failregex extended - SMTP call dropped: too many syntax or protocol errors (gh-2048);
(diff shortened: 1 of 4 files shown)


sebres:
Thank you!
I've rebased it to 0.10 and fixed backward-compatibility the same way (with optional space) similar to other cases (`\s*3:`, `\s*14:`).
I'll merge it if the tests have been passed.

benrubson:
Perhaps we could think about backporting this to Debian package v0.9.6 so that this issue would be corrected in the next update of the stable Debian package ?
@yarikoptic ?
Many thx 👍

sebres:
Perhaps better would be upgrade the package in debian stretch to 0.10th too? :)
Version 0.9 is really buggy and (whisper) even a bit vulnerable in one or other point...

benrubson:
That would be a good thing yes 👍 
But will Debian stable branch allow a jump from 0.9.6 to 0.10 ? :-/

PR #2064, opened 2018-03-04
Sendmail logs IPv6 addresses with the prefix 'IPv6:'.  Changed <HOST> to (?:IPv6:<IP6>|<IP4>)

mgrant0:
… before all <HOST> regexes to match the IPv6 address (but not the prefix).

Tested on Debian using 0.10.2-1 and sendmail 8.15.2-10
Problem still seems to exist on 0.11 from reading the filter.d/sendmail*.conf files

I admit that I only tested this on Debian using both sendmail and fail2ban installed from the debian package repo.  I don't know if sendmail on other platforms also logs addresses with this prefix in front of it, but I made it an optional regex using (IPv6:)? so it should not hurt if this prefix isn't there on other systems.

Here is a sample line that it now matches:
```
Mar  3 18:35:19 strange sm-mta[14956]: w23NZFZk014956: rejecting commands from example.com [IPv6:2a00:0000:0000:5000:abcd:abcd:abcd:abcd] due to pre-greeting traffic after 3 seconds
```

### DIFF
--- a/config/filter.d/sendmail-auth.conf
+++ b/config/filter.d/sendmail-auth.conf
@@ -9,7 +9,7 @@ before = common.conf
 
 _daemon = (?:sendmail|sm-(?:mta|acceptingconnections))
 
-failregex = ^%(__prefix_line)s\w{14}: (\S+ )?\[(IPv6:)?<HOST>\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
+failregex = ^%(__prefix_line)s\w{14}: (\S+ )?\[(?:IPv6:<IP6>|<IP4>)\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
 
 ignoreregex =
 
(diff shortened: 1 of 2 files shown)


mgrant0:
In retrospect maybe the HOST variable itself should have been modified for sendmail.

sebres:
> In retrospect maybe the HOST variable itself should have been modified for sendmail.

You cannot change `HOST`...
But since version 0.10 we've other constructions to get IP-addresses (HOST is just a combined construct), e. g. `<IP4>` and `<IP6>`, so in this case one can replace `<HOST>` like:
```diff
-failregex = ...<HOST>...
+failregex = ...(?:IPv6:<IP6>|<IP4>)...
```

sebres:
Additionally please rebase your branch to 0.10 (only your commits) and after push please change the base of this PR to 0.10 (see Edit button above).

mgrant0:
I have verified that your idea of using `(?:IPv6:<IP6>|<IP4>)` works.  

I will redo my commit and rebase on 0.10.

sebres:
I rebased it to 0.10 locally and reintegrated back for both version.
Thx!

Issue #2044, opened 2018-02-09
Fail2ban with AbuseIPBD reports , get a  curl (59) cypher error 


stratege1401:
Tried to use fail2ban with automatic abuseIPBD reports , get a  curl (59) cypher error when curl script is run.
Followed the https://wiki.shaunc.com/wikka.php?wakka=ReportingToAbuseIPDBWithFail2Ban

### Environment:
Debian 9.3 ( Stable + Sid MIX ) 4.9.0-5-amd64 Debian 4.9.65-3+deb9u2
php  1:7.0+49 
mariadb-client and mariadb-core 10.1   10.1.26-0+deb9u1
Fail2ban 0.10.2  --> apt-get install from SID source.list
apache  2.4.29-2
openssl   1.1.0f-3+deb9u1
curl   7.52.1-5+deb9u4

 [X] You have not applied any additional foreign patches to the codebase
 [X] No customizations were done to the configuration 

### The issue:
```
curl --fail --ciphers ecdhe_ecdsa_aes_256_sha --data 'key=<myAPI>' --data-urlencode 'comment=<matches>' --data 'ip=<127.0.0.1>' --data 'category=<4>' "https://www.abuseipdb.com/report/json"

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (59) failed setting cipher list: ecdhe_ecdsa_aes_256_sha
```

#### Steps to reproduce
run by daemon fail2ban OR manualy, same results

#### Expected behavior
should have a connection, but the cypher error prohibid connection

#### Observed behavior
no reports done

#### Any additional information
none

### Configuration, dump and another helpful excerpts
```
/etc/jail.local  ( TESTING CONF )
[INCLUDES]
before = paths-debian.conf

[DEFAULT]
ignoreip = 127.0.0.1/8 192.168.1.1/24
ignorecommand =
bantime  = 3600
findtime  = 36000
maxretry = 5
backend = auto
usedns = warn
logencoding = auto
enabled = false
filter = %(__name__)s
destemail = root@localhost
sender = root@localhost
mta = sendmail
protocol = tcp
chain = INPUT
port = 0:65535
fail2ban_agent = Fail2Ban/%(fail2ban_version)s

banaction = iptables-multiport
banaction_allports = iptables-allports
[...]

Issue #2039, opened 2018-02-03
F2B v0.10.2 fails resolving the "matches" variable properly when passing to actions


NoTeefy:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): **0.10.2**
- OS, including release name/version: **Debian GNU/Linux unstable (sid)** - **Linux 4.14.0-3-amd64**
- [X] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase
- [X] Some customizations were done to the configuration (provide details below if so)

### The issue:

After updating to the latest version an error occured within the resolve of variables (e.g: \<matches\>) which get passed to specific filters (blocklist_de in this case).
I already checked everything and didn't find the root cause which leads to that behaviour. As a result all generated reports get denied by blocklist.de since they miss correct log files and only pass a `$f2bV_matches`.

Another little addition to the checkbox about customizations above: I added my blocklist-API key in the [DEFAULT] section of my jail.config as described in the commented area from the corresponding jail.

#### Steps to reproduce
Activate the 'blocklist_de' action in the jail.config / jail.local file and wait till F2B gets 3 failed attempts in its authentification log. Check the fail2ban log after that and inspect the same error I described above.

#### Expected behavior
F2B should send a curl request containing sender adress, api key, bantime, service category and the corresponding logs in a url-encoded text format.

#### Observed behavior
F2B returns an error stating that the request got denied, if I remove the --fail parameter from the action file I get the following text back from blocklist.de: logs:

> Please insert the Logfiles, and more than 20 Signs.<br />status: error<br />

#### Any additional information
It seems like the variable `$f2bV_matches` gets passed at it is instead of resolving it into the log excerpt.
Other webmasters are experiencing those issues too, as you can see right here: https://forum.blocklist.de/viewtopic.php?f=4&t=678 (it's in German though).

### Configuration, dump and another helpful excerpts
_jail.config_
```
...
[sshd]

# To use more aggressive sshd modes set filter parameter "mode" in jail.local:
# normal (default), ddos, extra or aggressive (combines all).
# See "tests/files/logs/sshd" or "filter.d/sshd.conf" for usage example and details.
enable  = true
mode    = aggressive
port    = ssh
logpath = %(sshd_log)s
backend = %(sshd_backend)s
...
```

#### Any customizations done to /etc/fail2ban/ configuration
```
```

#### Relevant parts of /var/log/fail2ban.log file:
[...]

PR #2101, opened 2018-04-02
Fixes abuseipdb curl cypher error and comment $f2bV_matches

mercurytoxic:
Fixes issue #2044 curl cypher error  
and used #2039 to fix comment=<matches> resulting in abuseipdb.com only showing $f2bV_matches on the comment line.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -47,6 +47,7 @@ ver. 0.10.3-dev-1 (20??/??/??) - development edition
   - amend to gh-1263 with better handling of multiple attempts (failures for different user-names recognized immediatelly);
   - mode `ddos` (and `aggressive`) extended to catch `Connection closed by ... [preauth]`, so in DDOS mode
     it counts failure on closing connection within preauth-stage (gh-2085);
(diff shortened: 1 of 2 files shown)


sebres:
Thanks for PR!
Rebased to current 0.10 and added changelog entry.

sebres:
Thus merged in 0.10 and 0.11.

stratege1401:
does that mean this is also fixed in the apt-get packages version (0.10.2-1) for debian ?

sebres:
> does that mean this is also fixed in the apt-get packages version (0.10.2-1) for debian ?

Nope, 
if you'll look at commit https://github.com/fail2ban/fail2ban/pull/2101/commits/d9525ad3aafb3a3a3387e402fcb965a730cf6a97 you'll see versions `0.10.3.1`, `0.10.3` only.

stratege1401:
ok, any clue about date for apt-get packages updates ??

sebres:
No idea...
@yarikoptic ?

In between you could simply update your /etc/fail2ban/action.d/abuseipdb.conf manually, like in commit https://github.com/fail2ban/fail2ban/commit/fc76ccf19218f7eb14afcbf916fe47c3c9ccaa95

PR #2112, opened 2018-04-14
Create filter for Dante SOCKS server

al42and:
Creates filter for [Dante](https://www.inet.no/dante/index.html) server.

Limitations: I only tested it with "username" authentication mode and Dante 1.4.1 (from Debian 9), as this is the configuration I have running.

      for dev-edition use `master` branch: selected **master**
- [X] **CONSIDER adding a unit test** if your PR resolves an issue: does not resolve any issues
- [X] **LIST ISSUES** this PR resolves: does not resolve any issues
- [X] **MAKE SURE** this PR doesn't break existing tests: tests pass on my machine

### DIFF
--- a/config/filter.d/dante.conf
+++ b/config/filter.d/dante.conf
@@ -9,7 +9,7 @@ before = common.conf
 [Definition]
 _daemon = danted
 
-failregex = ^%(__prefix_line)sinfo: block\(1\): tcp/accept \]: <HOST>\.\d+ [\d.]+: error after reading \d+ bytes in \d+ seconds: (could not access user "\w+"'s records in the system password file: no system error|system password authentication failed for user "\w+")$
+failregex = ^%(__prefix_line)sinfo: block\(1\): tcp/accept \]: <HOST>\.\d+ [\d.]+: error after reading \d+ bytes? in \d+ seconds?: (could not access user "\w+"'s records in the system password file: no system error|system password authentication failed for user "\w+")$
 
 [Init]
 journalmatch = _SYSTEMD_UNIT=danted.service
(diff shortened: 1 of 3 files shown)


Loriowar:
You forgot about a follow case: "error after reading 1 byte~~s~~ in 1 second~~s~~". This may be fixed in this way: `error after reading \d+ byte(s)? in \d+ second(s)?`

sebres:
No idea it will be logged this way, but... please don't use capturing groups here, so because each letter is a token in regex, it is enough to change:
```diff
-error after reading \d+ bytes in \d+ seconds
+error after reading \d+ bytes? in \d+ seconds?
```

al42and:
Nice catch, thanks! It is indeed [a possible logging output](https://github.com/notpeter/dante/blob/master/dante/sockd/sockd_negotiate.c#L674).

Fixed and added a test for this particular case.

Also rebased the whole branch to avoid unrelated failed tests.

maximmasiutin:
Would you please consider accepting this pull request?

al42and:
@maximmasiutin, I resolved the conflicts, but I don't have dante instance readily available. Could you check if the log format used here is still relevant?

sebres:
done

maximmasiutin:
> @maximmasiutin, I resolved the conflicts, but I don't have dante instance readily available. Could you check if the log format used here is still relevant?

@al42and - I will check now.

maximmasiutin:
I have installed fail2ban from source and configured it to monitor Dante. Will simulate attacks and let you know how it worked.

maximmasiutin:
@al42and - Thank you, however, the regex was incomplete, I have addressed this issue in the pull request to update the regex to work on a real Dante server instance - please see https://github.com/fail2ban/fail2ban/pull/3410 #3410

maximmasiutin:
> done

@sebres Thank you, however, the regex was incomplete, I have addressed this issue in the pull request to update the regex to work on a real Dante server instance - please see https://github.com/fail2ban/fail2ban/pull/3410

PR #2097, opened 2018-03-29
Detect Apache SNI error / misredirect attempts

benrubson:
Hi,

This PR closes #2017 adding detection of Apache SNI error messages.
Based on this Apache commit :
https://bz.apache.org/bugzilla/show_bug.cgi?id=62003

Thank you 👍 

Ben

PR #2116, opened 2018-04-22
Merge Apache SNI error / misredirect attempts rules

benrubson:
Hi,

This PR follows #2097, it merges Apache SNI related filter rules.

Thank you 👍 

Ben

_See #2151 for a PR to 0.10._


benrubson:
Mmmh I should have proposed this PR against 0.10 branch.
Will however be very easy to backport as concerned file is the same in both 0.10 and 0.11.

PR #2125, opened 2018-05-08
Fix data type for journal flags (systemd backend)

jodlajodla:
List of changes:
- Added test to prove bug when specifying journal flags to systemd backend
- Fixed data type of journal flags from str to int (which then applies to python-systemd)
- Added Dockerfile for running tests

---

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,0 +1,11 @@
+FROM python:alpine
+
+RUN apk --update add bash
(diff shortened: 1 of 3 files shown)


sebres:
I've rebased it to 0.10 and merged partially (excepting docker test file, no sure ATM, let stay as separate PR, and may be more info).
Thank you for contributing!

sebres:
merged

PR #2131, opened 2018-05-18
added apf-multiport action

b2cc:
The existing APF action bans IPs without the possibility to only block access to specific ports. Therefore I would like to propose to add a new action `apf-multiport` which takes care of ports too. It also doesn't rely on the apf command line binary at all which seems to get installed in varying location depending on the linux distribution in use which seems to have led to problems in the past.

This action works by:
 - on start: creating the chain "f2b-apf-multiport"
 - inserting jump-to rule at the start of the INPUT chain to divert all traffic to the chain "f2b-apf-multiport"
 - adding blocked IPs with "-s <ip> -m multiport --dport <port>" to the chain "f2b-apf-multiport" (defaults to tcp if not specified otherwise)
 - removing unblocked IPs in the same manner
 - on stop: delete the chain "f2b-apf-multiport" and remove the jump-to rule from INPUT

regards && thx

### DIFF
--- a/config/action.d/apf-multiport.conf
+++ b/config/action.d/apf-multiport.conf
@@ -1,0 +1,40 @@
+# Fail2Ban configuration file
+#
+# multiport ban action for APF
+#  - https://www.rfxn.com/projects/advanced-policy-firewall/
+#
+# Author: David Gabriel <b2c@dest-unreachable.net>
+#
+# apf config hint: add something like this to apf/postroute.rules 
+#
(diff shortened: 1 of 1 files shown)


sebres:
> not sure why travis isn't passing, can someone give me hint what's wrong?

It's not your problem, this broken test-case was fixed in current branches (>= 0.10), but still broken/incompatible in master.
So just ignore this error.

b2cc:
@sebres  ok thanks for the info

PR #2140, opened 2018-05-27
CloudFlare started to indent their API responses

szepeviktor:
We need to use https://github.com/stedolan/jq to parse it.

### DIFF
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -60,7 +60,7 @@ actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-
 # API v4
 actionunban = curl -s -o /dev/null -X DELETE -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
               https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | cut -d'"' -f6)
+              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | tr -d '\n' | cut -d'"' -f6)
 
 [Init]
 
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -5,7 +5,7 @@
 #
 # Please set jail.local's permission to 640 because it contains your CF API key.
 #
-# This action depends on curl.
+# This action depends on curl and jq.
 # Referenced from http://www.normyee.net/blog/2012/02/02/adding-cloudflare-support-to-fail2ban by NORM YEE
 #
 # To get your CloudFlare API Key: https://www.cloudflare.com/a/account/my-account
@@ -60,7 +60,7 @@ actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-
 # API v4
 actionunban = curl -s -o /dev/null -X DELETE -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
               https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | tr -d '\n' | sed -nE 's/^.*"result"\s*:\s*\[\s*\{\s*"id"\s*:\s*"([^"]+)".*$/\1/p' )
+              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | jq -r '.result[0].configuration.value')
 
 [Init]
 
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -59,13 +59,13 @@ actionban = curl -s -o /dev/null -X POST <_cf_api_prms> \
 #actionunban = curl -s -o /dev/null https://www.cloudflare.com/api_json.html -d 'a=nul' -d 'tkn=<cftoken>' -d 'email=<cfuser>' -d 'key=<ip>'
 # API v4
 actionunban = id=$(curl -s -X GET <_cf_api_prms> \
-                   "<_cf_api_url>?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1&notes=Fail2Ban%20<name>" \
+                   "<_cf_api_url>?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1&notes=Fail2Ban%%20<name>" \
                    | jq -r '.result[0].id')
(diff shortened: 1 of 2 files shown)


sebres:
What is currently the output of second curl from `actionunban` (if not piped to `jq`)?
Could we write a simple "json-parser" without using of `jq` (is the usage of extern tool eligible if the output is very short and can be implemented using posix-tools)?
Python has JSON support, do we should possibly reimplement this as pythonic action?

szepeviktor:
Basically I ran into using posix tools :)
I usually pares JSON with cut and grep but that is a home-grown solution!

szepeviktor:
> What is currently the output of second curl from actionunban (if not piped to jq)?

Oh!!! That would be the access rule ID (a CF internal ID)
But I've parsed the IP :)
Sorry

szepeviktor:
It is `jq -r '.result[0].scope.id'`

@sebres Please decide on the paring tool!

sebres:
> jq is 200k, python3.4 is 4.4M

But we're already in python. What I meant was the python-action, see for example https://github.com/fail2ban/fail2ban/blob/8069eef50c30f01f0d21ed23027e0ea40446ff0f/config/action.d/badips.py#L37

> sebres Please decide on the parsing tool!

I can, if you give me the output in JSON (as requested) or give me any constellation (or description) how to test it, because I've not the CloudFlare, so no key (cfuser/cftoken) to hand.

szepeviktor:
> python -c "import sys, json; print json.load(sys.stdin)['result'][0]['id'];"

is really better.

wpkc:
Yes, a more complete answer would be that I use these for actionban and actionunban...
```
actionban = curl -s -X POST https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules \
	-H "X-Auth-Email: <cfuser>" -H "X-Auth-Key: <cfkey>" -H "Content-Type: application/json" \
        --data '{"mode":"block","configuration":{"target":"ip","value":"<ip>"},"notes":"Banned by Fail2Ban"}'

actionunban = curl -s -X GET -H "X-Auth-Email: <cfuser>" -H "X-Auth-Key: <cfkey>" -H "Content-Type: application/json" \
	"https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?page=1&per_page=5&mode=block&configuration.target=ip&configuration.value=<ip>&notes=Banned by Fail2Ban&match=all&order=configuration.value&direction=desc" | \
	python -c "import sys, json; print json.load(sys.stdin)['result'][0]['id'];" | \
	xargs -I@@ curl -s -X DELETE -H "X-Auth-Email: <cfuser>" -H "X-Auth-Key: <cfkey>" -H "Content-Type: application/json" https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/@@
```

config/action.d/cloudflare.conf
```diff
@@ -60,7 +60,7 @@ actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-
 # API v4
 actionunban = curl -s -o /dev/null -X DELETE -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
               https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | cut -d'"' -f6)
+              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | jq -r '.result[0].configuration.value')
```
sebres: Is the value returned from json via `jq -r '.result[0].configuration.value'` not the same as `<ip>`?
Because if yes, I don't know why we need `jq` here...

szepeviktor: YEs it is.
But cut does not work anymore.

sebres: What I don't understand, if you read the IP (with the second `$(curl... | jq...)`), then why simply not:
```diff
 actionunban = curl -s -o /dev/null -X DELETE -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | jq -r '.result[0].configuration.value')
+              https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/<ip>
```
Or (if I take a look to discussion, compared to solution of @wpkc) possibly should it be:
```diff
-... | jq -r '.result[0].configuration.value'
+... | jq -r '.result[0].id'
```
Then is it not the IP what API expects by DELETE, but some internal id of cloudflare?
Can you show me which value exactly needs this call?

wpkc: Yes, CloudFlare API v4 expects and access rule ID for the delete call.  The reason jq is not used is because you can rely on it being available. But you can always count on python being available if fail2ban is working. So...
 python -c "import sys, json; print json.load(sys.stdin)['result'][0]['id'];" 
...should always work.

sebres:
Tried to fix this PR (still using jq), + small review/normalization.
Additionally notes gets a jail-name now (should be possible to differentiate the same IP across several jails).
Please test.

I'll try to provide later a common solution (as pythonic action, without jq).

szepeviktor:
@sebres Thank you.

szepeviktor:
The optimal solution would be all-in-python which does not start a new python process.

sebres:
> The optimal solution would be all-in-python which does not start a new python process.

Yes, this is the plan (and I meant as "pythonic action").

config/action.d/cloudflare.conf
```diff
@@ -58,9 +58,14 @@ actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-
...
-              https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | cut -d'"' -f6)
+actionunban = id=$(curl -s -X GET <_cf_api_prms> \
+                   "<_cf_api_url>?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1&notes=Fail2Ban%20<name>" \
+                   | jq -r '.result[0].id')
+              if [ -z "$id" ]; then echo "<name>: id for <ip> cannot be found"; exit 0; fi;
+              curl -s -o /dev/null -X DELETE <_cf_api_prms> "<_cf_api_url>/$id"
+
+_cf_api_url = https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules
+_cf_api_prms = -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cfkey>' -H 'Content-Type: application/json'
```
akadusei:
```suggestion
_cf_api_prms = -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' -H 'Content-Type: application/json'
```
`cfkey` does not seem to be defined anywhere, which causes action to not work

akadusei:
```suggestion
_cf_api_prms = -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' -H 'Content-Type: application/json'
```
`cfkey` does not seem to be defined anywhere, which causes action to not work

sebres:
@szepeviktor I get still no time to write pythonic action for that...

I extended it (to be POSIX-compatible / for people without `jq`).

sebres:
Rebased to 0.10. Will be merged to 0.11 / 1.0 later. Thx!

szepeviktor:
You're welcome.

PR #2168, opened 2018-06-29
dovecot: collect F-USER and variants 

dpavlin:

### DIFF
--- a/config/filter.d/dovecot.conf
+++ b/config/filter.d/dovecot.conf
@@ -12,8 +12,8 @@ _daemon = (?:dovecot(?:-auth)?|auth)
 
 prefregex = ^%(__prefix_line)s(?:%(_auth_worker)s(?:\([^\)]+\))?: )?(?:%(__pam_auth)s(?:\(dovecot:auth\))?: |(?:pop3|imap)-login: )?(?:Info: )?<F-CONTENT>.+</F-CONTENT>$
 
-failregex = ^authentication failure; logname=\S* uid=\S* euid=\S* tty=dovecot ruser=\S* rhost=<HOST>(?:\s+user=\S*)?\s*$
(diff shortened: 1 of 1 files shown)


config/filter.d/dovecot.conf
```diff
@@ -12,8 +12,8 @@ _daemon = (?:dovecot(?:-auth)?|auth)
 
 prefregex = ^%(__prefix_line)s(?:%(_auth_worker)s(?:\([^\)]+\))?: )?(?:%(__pam_auth)s(?:\(dovecot:auth\))?: |(?:pop3|imap)-login: )?(?:Info: )?<F-CONTENT>.+</F-CONTENT>$
 
-failregex = ^authentication failure; logname=\S* uid=\S* euid=\S* tty=dovecot ruser=\S* rhost=<HOST>(?:\s+user=\S*)?\s*$
-            ^(?:Aborted login|Disconnected)(?::(?: [^ \(]+)+)? \((?:auth failed, \d+ attempts(?: in \d+ secs)?|tried to use (?:disabled|disallowed) \S+ auth)\):(?: user=<[^>]*>,)?(?: method=\S+,)? rip=<HOST>(?:[^>]*(?:, session=<\S+>)?)\s*$
+failregex = ^authentication failure; logname=\S* uid=\S* euid=\S* tty=dovecot ruser=<F-USER>\S*</F-USER> rhost=<HOST>(?:\s+user=\S*)?\s*$
```
sebres: Are you sure that the interesting value is always logged as `ruser` (which should be taken into account as remote-user)? I mean there is also 2 other values to check (`user=` and `logname=`)...

BTW. Since #2090 got merged, we've possibility to provide more as one user-group (so alternate groups)...
Example: 
`login=<F-ALT_USER1>\S*</F-ALT_USER1> ruser=<F-ALT_USER>\S*</F-ALT_USER>, user=<F-USER>\S*</F-USER>`
First non-empty value wins there (starting with USER, ALT_USER, ALT_USER1, ALT_USER2 ...)

dpavlin: My understanding is that ruser= are credentials of connection as opposed to anything which dovecot does to map to final user. I think it's more interesting to see original credentials which attacker tried to use. So I would expect to see it in F-USER and user= in ALT_USER.

Does this make sense?

dpavlin:
It seems that very useful comment about ALT_USER variables dissipated when I push -f new version. Also rationale for ruser as primary USER var is now included in comment. Is there anything else I should do?

sebres:
Merged. Thx!

PR #2173, opened 2018-07-05
Load logpath only after findtime is configured

mattsta:
### Commit Message
When new log paths are configured, their start offset is immediately determined
by a filter searching for (now - findTime).
But, since findTime is configured *after* the log is loaded and
searched, logs are only searched back by the default 10 minute findTime,
regardless of user configuration of jail settings.

So, findTime must be configured before logpath or else the default findtime
is used, which ignores any findtime time defined by the user.

This fixes new reads on startup for actual log files. The systemd filter
always performed as expected due to being setup after the jail's
findtime config submission.

### Story Time!
It took me probably 10 hours to track down this problem. Spent a long time thinking regexes were just broken for half my jails because lookback-on-load worked reliably on some jails, but not others. Turns out the reliable jails were all systemd and loading from files was kinda broken. Then took a while to figure out how configs actually get loaded.

### Additional Fixes Maybe
The newer config options `logtimezone` and `logencoding` are also loaded after the logfile is initially processed. If those are used to read the log, they should also be configured before `logpath` gets sent to the transmitter, otherwise default class values will always be used instead of configured values.

and, one minor change that may have helped cut hours off these debuggles: surfacing the initial lookback results in INFO mode instead of deep debug mode ([here](https://github.com/fail2ban/fail2ban/blob/ddff67acb3f529e6835c66954d64f8fd322eefa6/fail2ban/server/filter.py#L1039-L1041) and [here](https://github.com/fail2ban/fail2ban/blob/ddff67acb3f529e6835c66954d64f8fd322eefa6/fail2ban/server/filter.py#L1118-L1120)) would be nice for users to always know (since it's the start point for _actually_ searching for bad actions on startup).

### DIFF
--- a/fail2ban/client/jailreader.py
+++ b/fail2ban/client/jailreader.py
@@ -90,12 +90,12 @@ class JailReader(ConfigReader):
 		opts1st = [["bool", "enabled", False],
 				["string", "filter", ""]]
 		opts = [["bool", "enabled", False],
-				["string", "logpath", None],
-				["string", "logtimezone", None],
-				["string", "logencoding", None],
 				["string", "backend", "auto"],
 				["int",    "maxretry", None],
 				["string", "findtime", None],
+				["string", "logpath", None],
+				["string", "logtimezone", None],
+				["string", "logencoding", None],
 				["string", "bantime", None],
 				["string", "usedns", None], # be sure usedns is before all regex(s) in stream
 				["string", "failregex", None],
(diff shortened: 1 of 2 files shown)


sebres:
I've rebased this to 0.10 (because affected also) and made a small amendment to cover all possible constellations.
Will merge hereafter if all test-cases passed.
Thank you for your contribution!

PR #2181, opened 2018-07-15
is remove unused modules

inztar:
I found unused imports, i can remove it?


yarikoptic:
unless @sebres has objections, could be done I guess, but might cause merge conflicts etc. so better be cleaned up "upon encounter".  in any case - probably would need a refresh (merge of 0.11 or rebase) to guarantee that it is still adequate

PR #2182, opened 2018-07-15
OpenRC service script improvements

orlitzky:
I basically rewrote the whole OpenRC (Gentoo, mainly) service script. Individual commits should provide details. This is only lightly tested on my desktop machine, but if you like the idea, I can push it out to our servers and run it by the Gentoo maintainers.


thesamesam:
@sebres I'm the maintainer of fail2ban downstream in Gentoo and @orlitzky is one of my fellow developers. We're probably the main consumer of OpenRC (and are extremely closely linked with its maintainers), possibly other than Alpine. We're quite happy with these changes on our end.

Would it be possible for you to pull these in if that's alright? This shouldn't have an affect on any other users.

thesamesam:
@sebres I'm the maintainer of fail2ban downstream in Gentoo and @orlitzky is one of my fellow developers. We're probably the main consumer of OpenRC (and are extremely closely linked with its maintainers), possibly other than Alpine. We're quite happy with these changes on our end.

Would it be possible for you to pull these in if that's alright? This shouldn't have an effect on any other users.

sebres:
> Would it be possible for you to pull these in if that's alright?

Sure, I'll try to review and merge it to 0.11th branch soon. Just my intention was either to prepare and release 1.0 (master).

> This shouldn't have an effect on any other users.

I don't think so, because there are some removed files (like `gentoo-initd`), therefore it is not fully backwards "compatible", e. g. some scripts people may use wouldn't work anymore with the next release (0.11.3).
Not really a blocker (such files are rather matter of maintainers), just a small note from my side.

thesamesam:
FWIW, we've also pushed these bits into Gentoo now and no complaints.

sebres:
thus merged

orlitzky:
Thank you!

PR #2189, opened 2018-07-24
debian-initd: exit with non-0 if fail, account that 255 is "Ok" exit code, use 255 explicitly instead of -1

### LINKED ISSUE #2187: Failing to start service returns 0
### Environment:
- Fail2Ban version (including any possible distribution suffixes): Fail2Ban v0.10.2
- OS, including release name/version: Ubuntu 18.04

### The issue:
When starting the service fails (using `service fail2ban start` or `/etc/init.d/fail2ban start`, the error code is still 0. This is quite a problem for automation, for example using Ansible, as failing programs are meant to return non-zero codes. The program properly compains over `stderr`, but `stderr` is not a proper way to check for errors in the sense that it can be reliably used by other programs (it is normal for programs to succeed but return debugging info or warnings over `stderr` for example). It would be nice if fail2ban could properly return a non-zero exit code.

#### Steps to reproduce
Start a Docker container using the `ubuntu:18.04` image, and install fail2ban using the package manager. Now run `service fail2ban start` or `/etc/init.d/fail2ban start`.

#### Expected behavior
It expects to find a file at `/var/log/auth.log`, so we should get some error and a non-zero exit code.

#### Observed behavior
We do get an error:
```
 * Starting authentication failure monitor fail2ban
 No file(s) found for glob /var/log/auth.log
 Failed during configuration: Have not found any log file for sshd jail
   ...fail!
```

But the exit code (`echo $?`) is `0`.

yarikoptic:
Proposed changes fixes #2187 (hopefully, yet to test in the docker)

Note that, after spending some time grepping for `exit.*255` decided to actually make it possible ;)

### DIFF
--- a/files/debian-initd
+++ b/files/debian-initd
@@ -180,11 +180,17 @@ do_reload() {
 #
 log_end_msg_wrapper()
 {
-	[ $1 -lt $2 ] && value=0 || value=1
+	if [ $1 != 0 ] && [ $1 != $2 ]; then
+		value=1
+	else
+		value=0
+	fi
 	if [ "$3" != "no" ]; then
 		log_end_msg $value
 	fi
-	exit $value
+	if [ $code != 0 ]; then
+		exit $1
+	fi
 }
 
 command="$1"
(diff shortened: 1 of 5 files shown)


files/debian-initd
```diff
@@ -180,24 +180,31 @@ do_reload() {
...
+	if [ $1 != 0 ] && [ $1 != $2 ]; then
+		value=1
+	else
+		value=0
+	fi
 	if [ "$3" != "no" ]; then
-		[ $1 -lt $2 ] && value=0 || value=1
 		log_end_msg $value
 	fi
+	if [ $code != 0 ]; then
```
Uwila: This gives an error for me: `[: !=: unexpected operator`. Quoting the args fixes it for me, though:
```
if [ "$code" != "0" ]; then
```

yarikoptic: oh oh... Should have been "$value" not "$code" (was thinking about "exit code" too much I guess)
(since then I usually `set -u` in my scripts, but didn't want to do that here and here you go -- can't even see what I am typing), but should be ok to wrap "0" as well

thanks @Uwila , pushing the fix...

Uwila:
Seems to work for me now

PR #2205, opened 2018-08-19
Add loglevel option to badips.py

benrubson:
Hi,

This PR adds a new yes/no `log` option to `badips.py`.
It allows to enable / disable the log line produced when an IP is (un)banned.
One may want to define this per jail, thanks to this new option.

Thank you 👍

### DIFF
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -343,13 +343,16 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 				ips = self.getList(
 					self.bancategory, self.score, self.age, self.bankey)
 				# Remove old IPs no longer listed
-				self._unbanIPs(self._bannedips - ips)
+				s = self._bannedips - ips
+				m = len(s)
+				self._unbanIPs(s)
 				# Add new IPs which are now listed
-				self._banIPs(ips - self._bannedips)
-
-				self._logSys.debug(
-					"Updated IPs for jail '%s'. Update again in %i seconds",
-					self._jail.name, self.updateperiod)
+				s = ips - self._bannedips
+				p = len(s)
+				self._banIPs(s)
+				self._logSys.info(
+					"Updated IPs for jail '%s' (-%d/+%d). Update again in %i seconds",
+					self._jail.name, m, p, self.updateperiod)
 			finally:
 				self._timer = threading.Timer(self.updateperiod, self.update)
 				self._timer.start()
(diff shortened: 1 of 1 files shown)


benrubson:
You do not find this relevant @sebres ?
Thank you !

benrubson:
I find it rather useful to easily track badips activity :)

sebres:
I don't closed it, this looks like a github BUG (several PRs went closed - I pushed something in 0.10 (and branch 0.10 disappeared for some reason, together with all PRs for this branch)...

benrubson:
What about merging this tiny PR @sebres ?
Many thanks !

config/action.d/badips.py
```diff
@@ -99,6 +104,7 @@ def __init__(self, jail, name, category, score=3, age="24h", key=None,
 		self.banaction = banaction
 		self.bancategory = bancategory or category
 		self.bankey = bankey
+		self.loglevel = str2LogLevel(loglevel) if isinstance(val, basestring) else loglevel
```
benrubson: Typo here with `val` instead of `loglevel` ?

sebres:
I rewrote this a bit.

So usage for logging with level NOTICE:
```ini
action = badips.py[loglevel=notice]
```

BTW: you've used operator `is` (`if log is 'yes'`), this is wrong in python - it is not an operator equal `==` (`is` compares objects are equals not the values, so just meant the same object).
You can use it with ``None` (or similar global singletons)...
Just compare: 
```python
>>> 'Yes'.lower() == 'yes'` 
True
>>> 'Yes'.lower() is 'yes'
False
```

config/action.d/badips.py
sebres: Thx, I saw it already by latest common review.

benrubson:
Nice rewrite, even more generic, perfect, thank you 👍

> you've used operator `is` (`if log is 'yes'`), this is wrong in python

Yes you're right, thank you for your explanation.
Did work with an empty value, but would not have worked with the _yes_ value...

benrubson:
Thank you @sebres 👍

config/action.d/badips.py
```diff
@@ -350,7 +350,7 @@ def update(self):
 				s = ips - self._bannedips
 				p = len(s)
 				self._banIPs(s)
-				self._logSys.info(
+				self._logSys.log(self.loglevel,
 					"Updated IPs for jail '%s' (-%d/+%d). Update again in %i seconds",
```
benrubson: @sebres, would have been interesting to keep `INFO` loglevel here.
So that we can have the individual _Banned IP_ messages in `DEBUG` level (so not logged if we use for example the global configuration `loglevel = INFO`), and this summary message in `INFO` level.
This was one of the purposes of this PR :)

benrubson: See #2245 for a quick fix. Thx !

sebres: answered in PR

Issue #2213, opened 2018-09-08
Passive waiting to solve defunct proc and to dramatically improve performances


benrubson:
Hi,

From time to time I have alerts on my servers regarding _defunct_ processes.
I investigated and found that these are fail2ban processes, launched to execute system commands.
As fail2ban does not use the `wait()` call to wait for its children, there is a short moment where the children are seen as _defunct_.

I then tried the following, in `fail2ban/fail2ban/server/utils.py`, in `executeCmd()` function.

I replaced the first :
`retcode = popen.poll()`
With :
`retcode = popen.wait()`

In addition to solve the defunct issue, this passive waiting dra-ma-ti-ca-lly improves performances.

On Linux, using iptables, 3350 rules are added in 33 seconds instead of 11.5 minutes !

I know that the `wait()` call is a blocking call, but perhaps there's a workaround.
I tried to put it in a `try` bock using `signal.alarm(timeout)`, but this can't be used in threads.

Thank you very much,

Ben

benrubson:
See #2215 for a proposal to solve this !

benrubson:
See #2215 for some tryouts & solutions.

sebres:
Hmm... `popen.wait()` is noting else as try `popen.poll()` (or rather internal call of `os.waitpid` used in both) while the child process running. Just the normal call of poll will use a very short inbetween sleep's additionally (but this does not explain why this should take 11 min instead of 33 seconds).
So I don't know how you tested this (by the way which python version is used?) to reach values like this (or something is just wrong on your system).

Because if I test 5 second simple ban+unban the original version vs. your version, I get:
```diff
- # original 0.10:
- 6.078ms/# 823 # (5.002s total)
+ # PR gh-2215:
+ 7.662ms/# 653 # (5.004s total)
```

So your variant seems to be 1.6ms slower per iteration (ban/unban).

This means that if I test both on my machine 33 seconds long, I get:
- 5430 ban/unban invocations, so ca. 11000 rules processed **with original version**;
- 4310 ban/unban invocations, so ca. 8600 rules processed **with your version**;

And I don't really understand how such "dra-ma-ti-ca-lly" performance increase can be reached, neither by usage of wait, nor by usage of additional command (`timeout`).

Can you please describe a scenario how you tested this? 
Additionally please provide:
1. which fail2ban and python version exactly?
2. which banaction is used in your case?

sebres:
[...]

PR #2215, opened 2018-09-08
Passive wait children to improve performances

benrubson:
Hi,

Here is a proposal / a tryout to solve #2213.
Goal is to rely on the `timeout` binary to launch the commands and then passively wait for our children.

Ben

### DIFF
--- a/config/action.d/iptables-common.conf
+++ b/config/action.d/iptables-common.conf
@@ -73,7 +73,7 @@ lockingopt = -w
 # Option:  iptables
 # Notes.:  Actual command to be executed, including common to all calls options
 # Values:  STRING
-iptables = iptables <lockingopt>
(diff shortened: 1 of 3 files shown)


benrubson:
So I tested several things.

First commit tries to prefix every command with `timeout -k Utils.DEFAULT_SLEEP_INTERVAL timeout`.
However, as some commands are rather complex, it fails on them.
So Travis fails here.

Second commit detects `timeout -k` in the command itself, and if found, passively waits for the command to finish.
_iptables_ is taken as an example here.
_Travis fails just because iptables command name has changed in the log, should be easy to correct._

I think second example commit could be the way to go, activating timeout on a per command basis.

benrubson:
Closing this in favor of discussion in #2213.

PR #2221, opened 2018-09-14
Use secure URI in Vcs control header.

jelmer:
Use secure URI in Vcs control header.

Fixes lintian: vcs-field-uses-insecure-uri
See https://lintian.debian.org/tags/vcs-field-uses-insecure-uri.html for more details.


sebres:
Merged. Thx.

PR #2245, opened 2018-10-02
Add a shortloglevel badips.py option

benrubson:
Hi,

This follows #2205 to fix the summary message loglevel.

Thank you 👍 

Ben

### DIFF
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -32,7 +32,7 @@ else: # pragma: 3.x no cover
 	from urllib import urlencode
 
 from fail2ban.server.actions import ActionBase
-from fail2ban.helpers import str2LogLevel
+from fail2ban.helpers import splitwords, str2LogLevel
 
 
 
@@ -75,9 +75,9 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 	loglevel : int/str, optional
 		Log level of the message when an IP is (un)banned.
 		Default `DEBUG`.
-	shortloglevel : int/str, optional
-		Log level of the summary message when a group of IPs is (un)banned.
-		Default `DEBUG`.
+		Can be also supplied as two-value list (comma- or space separated) to
+		provide level of the summary message when a group of IPs is (un)banned.
+		Example `DEBUG,INFO`.
 	agent : str, optional
 		User agent transmitted to server.
 		Default `Fail2Ban/ver.`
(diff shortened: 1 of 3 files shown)


config/action.d/badips.py
```diff
@@ -350,7 +350,7 @@ def update(self):
 				s = ips - self._bannedips
 				p = len(s)
 				self._banIPs(s)
-				self._logSys.log(self.loglevel,
+				self._logSys.info(
```
sebres: Please change:
```diff
- logSys.log(loglevel,
+ logSys.log(logging.MSG,
```

benrubson:
> I don't use badips, but cannot imagine for which reasons this should be logged after (successful) IPs-merge by the successful start of action (separated from other things, which are rather a debug purposes).

Think about merging sometimes hundreds of IPs.
You don't want an individual message for every processed IP, but rather a summary message with the number of removed/added IPs.
I agree that the _next update in X seconds_ part of the message is for sure debug.
What about then the commit I just pushed ? Would it be suitable ? We would not have "duplicate" messages.

> Anyway I can promise you, that tomorrow (as many times earlier) we'll get a request to disable this info, because it bothers someone (does not saw this message previously).

Sure, I understand.

benrubson:
> So then why you have extended the logging in #2205 for "individual message for every processed IP" instead of doing this for summary message only?

The other reason is that you may want to have these individual messages for some of your badips jails, but not for some others (you can define several badips jails, so with different log levels).

> You still did not understand, I am not against this message, but against the **default** level (`INFO`) for it (for the reasons I already mentioned above).

Understood, let me then rework this correctly.

benrubson:
I think this rework should be OK @sebres 👍 
We then have the `loglevel` option for the detailed (un)ban message.
And the `shortloglevel` option for the summary message.
We are then able to output both or only one of these 2 messages, independently, and at a jail level.

Thank you 👍

benrubson:
@sebres what about merging this small PR ?
Many thanks again !

benrubson:
Let's go with this if you're OK @sebres ?
Many thanks !

benrubson:
Hi @sebres, may I ask you to review this tiny change please when you'll have some time ?
Many thanks !

sebres:
It has been rewritten a bit: `loglevel` and `shortloglevel` combined now to single parameter `loglevel`.

Here an example logging summary with NOTICE and rest with DEBUG log-levels:
```ini
action = badips.py[loglevel="debug, notice"]
```
Parameter `shortloglevel` does not exists anymore.

benrubson:
Really nice idea, many thx @sebres 👍

PR #2247, opened 2018-10-03
Update mysql logging command and fix mod_security filter

Holston5:

I couldn't find any issues related to this change so there are none to close.
The  "log-warning = 2" option in both files should read "log-warnings = 2".  Mysql demon fails to start when incorrectly set to "log-warning = 2"

[Commit 2](https://github.com/fail2ban/fail2ban/pull/2247#issuecomment-427043957):

Filter for mod_security doesn't work out of the box and needs tweaking.

### DIFF
--- a/config/filter.d/apache-modsecurity.conf
+++ b/config/filter.d/apache-modsecurity.conf
@@ -10,7 +10,7 @@ before = apache-common.conf
 [Definition]
 
 
-failregex = ^%(_apache_error_client)s *ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
+failregex = ^%(_apache_error_client)s(?: \[client [^\]]+\])? ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
 
 ignoreregex = 
 
--- a/config/filter.d/apache-modsecurity.conf
+++ b/config/filter.d/apache-modsecurity.conf
@@ -10,7 +10,7 @@ before = apache-common.conf
 [Definition]
 
 
-failregex = ^%(_apache_error_client)s ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
+failregex = ^%(_apache_error_client)s *ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
 
 ignoreregex = 
 
--- a/config/filter.d/mysqld-auth.conf
+++ b/config/filter.d/mysqld-auth.conf
@@ -3,7 +3,7 @@
 #
 # To log wrong MySQL access attempts add to /etc/my.cnf in [mysqld]:
 # log-error=/var/log/mysqld.log
-# log-warning = 2
+# log-warnings = 2
 #
 # If using mysql syslog [mysql_safe] has syslog in /etc/my.cnf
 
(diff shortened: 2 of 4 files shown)


Holston5:
Filter for mod_security also needed amending as currently fail2ban doesn't work out of the box.
As mentioned [here](https://serverfault.com/questions/845270/fail2ban-and-modsecurity-not-working) the default filter is:
`
failregex = ^%(_apache_error_client)s ModSecurity: etc...
`
But that doesn't match the actual log pattern: (A real entry taken from my log)

> [Thu Oct 04 08:05:49.983217 2018] [:error] [pid <pid>] [client XXX.XXX.XXX.XXX:XXXX] [client XXX.XXX.XXX.XXX] ModSecurity: Warning. Pattern match "^[\\\\d.:]+$" at REQUEST_HEADERS:Host. [file "/usr/share/modsecurity-crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] <snip>

The changed filter resolves this

config/filter.d/apache-modsecurity.conf
```diff
@@ -10,7 +10,7 @@ before = apache-common.conf
 [Definition]
 
 
-failregex = ^%(_apache_error_client)s ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
+failregex = ^%(_apache_error_client)s *ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
```
sebres: Hmm...
Note that quantifier `*` applied to the token after it is placed, so in this case for space ` ` only.

May be you want something like this...
```diff
- failregex = ^%(_apache_error_client)s ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
+ failregex = ^%(_apache_error_client)s(?: \[client [^\]]+\])? ModSecurity:\s+(?:\[(?:\w+ \"[^\"]*\"|[^\]]*)\]\s*)*Access denied with code [45]\d\d
```

Holston5: There are two client instances next to each other in some versions of mod_security (see my log entry).  The * would catch both of them

sebres: Nope, it would not - the `*` is after space, so it matches 0..N spaces only.
I try to explain this - you see, the second case does not match:
```bash
$ python -c 'import re; m = re.match(r"^C * mod", "C mod"); print(m.group(0) if m else m)'
C mod
$ python -c 'import re; m = re.match(r"^C * mod", "C C mod"); print(m.group(0) if m else m)'
None
$ python -c 'import re; m = re.match(r"^C(?: C)? mod", "C C mod"); print(m.group(0) if m else m)'
C C mod
```
Possibly you've tried with `*` directly after `%(_apache_error_client)s` (so again, not after space), then it can work, but it is too aggressive...

Holston5: Yes you are right, thank you

sebres:
Are the IPs in the both `[client ...]` equal? (and if not, then which from both is a real IP of intruder?)

Holston5:
> Are the IPs in the both `[client ...]` equal? (and if not, then which from both is a real IP of intruder?)

The first client field is IP:SOURCEPORT and the second is just IP.  Both IP's appear to be the same but I have a relatively small sample pool (<100) to check from.

crazy4chrissi:
Regarding modsecurity:
It is true that the current rule does not match current modsecurity logs.
See these links for explanations and modified rules:
https://serverfault.com/questions/845270/fail2ban-and-modsecurity-not-working
https://www.kazimer.com/update-fail2ban-apache-security-conf-filter-for-security2-module/

sebres:
Branch is rebased to 0.10, merge to both major branches (0.10/0.11).
Thx!

PR #2257, opened 2018-10-12
Safer, nicer, uniform Debian initd script

szepeviktor:
Almost *shellcheck OK*


szepeviktor:
@yarikoptic Could you take a look at this?

files/debian-initd
```diff
@@ -68,7 +69,7 @@ log_daemon_msg () {
 #
 report_bug()
 {
-	echo $*
+	echo "$*"
```
yarikoptic: is it actually helpful?

files/debian-initd
```diff
@@ -136,8 +137,8 @@ do_start()
 #
 do_status()
 {
-	$DAEMON ping > /dev/null 2>&1
-	return $?
+	$DAEMON ping >/dev/null 2>&1
+	return "$?"
```
yarikoptic: is there advantage?

files/debian-initd
```diff
@@ -206,41 +208,55 @@ case "$command" in
...
+						;;
+					2)
+						log_failure_msg " $SOCKFILE not readable, status of $NAME is unknown" && exit 3
+						;;
+					3)
+						log_failure_msg " $SOCKFILE exists but not a socket, status of $NAME is unknown" && exit 3
+						;;
+					*)
+						report_bug "Unknown return code from $NAME:check_socket." && exit 4
+						;;
```
yarikoptic: I definitely like compressed, not blown up version more here -- easier to do visual matching of messages, exit codes etc

files/debian-initd
```diff
@@ -22,36 +22,37 @@
...
+SOCKFILE="$(grep -h '^[^#]*socket *=' "/etc/$NAME/$NAME.conf" "/etc/$NAME/$NAME.local" 2>/dev/null \
+  	| tail -n 1 | sed -e 's/.*socket *= *//g' -e 's/ *$//g')"
+[ -z "$SOCKFILE" ] && SOCKFILE="/var/run/fail2ban.sock"
 
 # Exit if the package is not installed
 [ -x "$DAEMON" ] || exit 0
 
 # Run as root by default.
-FAIL2BAN_USER=root
+FAIL2BAN_USER="root"
```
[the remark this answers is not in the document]
szepeviktor: Uniformity.
Coding without thinking :)

files/debian-initd
```diff
@@ -206,41 +208,55 @@ case "$command" in
...
 				;;
-			*)  report_bug "Unknown $NAME status code" && exit 4
+			*)
+				report_bug "Unknown $NAME status code" && exit 4
+				;;
 		esac
 		;;
 	*)
-		echo "Usage: $SCRIPTNAME {start|force-start|stop|restart|force-reload|status}" >&2
+		echo "Usage: $SCRIPTNAME {start|force-start|stop|restart|force-reload|status}" 1>&2
```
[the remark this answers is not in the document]
szepeviktor: Coding explicitly.

Don't expect the reader/reviewer to know what is being redirected.

files/debian-initd
szepeviktor: $1, $2 = "pear" "apple"

"$@" becomes "pear" "apple"

"$*" becomes "pear apple"

`$*` is not protected

files/debian-initd
szepeviktor: Okay.

I code explicitly and in a uniform fashion, learned from bash(1)

gyles19:
Some of those changes match debian standards for initial scripts, like the
bit about testing for the executable before proceeding.

Quoting strings is a scripting best practice.     Quoting parameters and
echo outputs is a security best practice for any shell script.

I'd accept every one of those suggested changes if it were my project. I
follow those rules in the scripts I write.  If nothing else it makes
syntax-aware editors and IDEs happier

files/debian-initd
yarikoptic: agreed -- would expand globs. not that we pass any in our strings afaik but indeed doesn't hurt

files/debian-initd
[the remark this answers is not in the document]
yarikoptic: sure thing "explicit better than implicit", but then should we may be start `declare`ing each variable in bash?
IMHO knowing that echo outputs to 1 is quite an explicit knowledge. Reader should know that. but ok  - doesn't hurt.

yarikoptic:
> like the bit about testing for the executable before proceeding.

that already was there, so this change and my questions were not about that

Just for the sake of my own education:

> Quoting strings is a scripting best practice. 

I was asking about  how uniform quoting with `"` is better over e.g. using `'` quotes where it was desired? if anything - would go against the practice where I would use single quotes for the strings I expect to not be altered

files/debian-initd
yarikoptic:
```
$> man 1 bash | grep uniform; echo $?
1
```
;)

files/debian-initd
szepeviktor: :) actually my editor (Midnight Commander) only highlights `1>&2` not `>&2` :)

files/debian-initd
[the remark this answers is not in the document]
szepeviktor: :) actually my editor (Midnight Commander) only highlights `1>&2` not `>&2` :)

yarikoptic:
and why then not to finish it up and add this to travis run?
```
$> shellcheck -s bash -e SC1090,SC1091  files/debian-initd 

In files/debian-initd line 127:
		$DAEMON_ARGS start >/dev/null \
                ^-- SC2086: Double quote to prevent globbing and word splitting
```

szepeviktor:
> SC2086: Double quote to prevent globbing and word splitting

shellcheck is not able to think with my head, `$DAEMON_ARGS` may be empty, so it could not be quoted

yarikoptic:
pushed few commits for that purpose - but now see the conflict, need to run. Feel welcome to do whatever (rebase/merge) to resolve conflicts

szepeviktor:
I hope I haven't made big trouble!

szepeviktor:
> But I definitely do not like expanded `case`

Should I revert full-form `case`?

yarikoptic:
d'oh -- only seeing merge message realized that you've submitted it against debian branch.
It should be filed against `master` -- debian branch containers only `debian/` folder. This file is under `files/` so could be installed straight from master manually.   Could you resubmit against master?

Re `case` -- in the big realm of things, I do not care/object. Let's not waste more mind cycles reverting

szepeviktor:
Closed with success!

szepeviktor:
@yarikoptic Something not right with branches:

- 4 years old: https://github.com/fail2ban/fail2ban/commits/master/files/debian-initd
- 1/2 year old: https://github.com/fail2ban/fail2ban/commits/debian/files/debian-initd

What do you think? master or debian?

yarikoptic:
@sebres  It seems that "master" is somewhat an abandoned concept (you primarily use "major" release branches such as 0.10 and 0.11) or what is its purpose?  if not really used I guess could be removed, and the main development release branch (0.11) would take its place?

I guess that this particular fix could as well go on top of 0.10.  0.11 receives merges of 0.10 so it wouldn't be lost ;-)

sebres:
This is my intention too, just I wanted take a look about all open PRs targeting master in order to merge or even reset target branch to 0.9.
But time is endless... I'll try to push the prio of it higher in my to-do list.

yarikoptic:
so, probably 0.10 would be the right branch then @szepeviktor ;)

PR #2272, opened 2018-11-11
Add firehol.service to systemd unit PartOf and After directives

nodiscc:
References https://github.com/firehol/firehol/wiki/fail2ban, https://github.com/firehol/firehol/issues/272, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881669

[FireHOL](https://github.com/firehol/firehol/) is a tool to create and managed stateful iptables firewall configurations. It is packaged for major Linux distributions. On startup/restart, FireHOL overwrites/clears all configured iptables rules, including fail2ban-generated rules. This often causes fail2ban iptables rules to be silently lost when restarting the `firehol`  service. The wiki page describes a way to add custom `PartOf`  and `After` directives to the fail2ban systemd unit file so that fail2ban is always stopped _before_ restarting `firehol.service`  (allowing it to remove its rules), and (re)started _after_ the firewall has finished starting up.

This patch removes the need to create a systemd override file, correctly removing/restoring fail2ban rules on firewall stop/start/restart.

- [x] **LIST ISSUES** this PR resolves: proper fix for https://github.com/firehol/firehol/issues/272, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881669

### DIFF
--- a/files/fail2ban.service.in
+++ b/files/fail2ban.service.in
@@ -1,8 +1,8 @@
 [Unit]
 Description=Fail2Ban Service
 Documentation=man:fail2ban(1)
-After=network.target iptables.service firewalld.service ip6tables.service ipset.service
-PartOf=iptables.service firewalld.service ip6tables.service ipset.service
+After=network.target iptables.service firewalld.service ip6tables.service ipset.service firehol.service
+PartOf=iptables.service firewalld.service ip6tables.service ipset.service firehol.service
 
 [Service]
 Type=simple


nodiscc:
Yes, this will only work when starting/restarting/stopping firehol through systemctl (firehol.service does not have a reload action). Issuing `firehol start,restart,stop,try...` manually can still lead to losing fail2ban rules.

But at least this ensures all rules are properly loaded on system boot, and when managing fail2ban/firehol through systemd service management.

sebres:
Well in my opinion the possible handling with `postprocess2`, for example in order to reload fail2ban within, would be a bit wrong (too broad)... Just imagine a re-banning of set of 10000 IP addresses.
I think, FireHOL handling should affect only iptable-entries created from himself or provide a way to remain the chains and entries of other services like fail2ban.
For example all chains in iptables create by fail2ban are prefixed with `f2b-`, so if there would be a way to ignore/retain this on FireHOL reload processes, it could be nice.

nodiscc:
> Just imagine a re-banning of set of 10000 IP addresses.
>I think, FireHOL handling should affect only iptable-entries created from himself or provide a way to remain the chains and entries of other services like fail2ban.

I don't know the impact of unloading/loading a large set of bans (my ban list is modest and I use ipset-based ban actions, which load very fast). This fix does the job for me, but I understand your concerns regarding performance.

I also like to know my firewall rules depend on a single service (firehol started: all firewall rules loaded, including fail2ban's. firehol stopped: iptables chains fully cleared).


I have asked about this in https://github.com/firehol/firehol/issues/398.

PR #2284, opened 2018-11-23
Add link to https://fail2ban.readthedocs.io in README

jedie:
It's hard to find https://fail2ban.readthedocs.io because it's not linked somewhere, isn't it?

IMHO it's also good to link it on https://www.fail2ban.org/ in a **well visible, prominent place**!

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,8 @@ mechanisms if you really want to protect services.
 ------|------
 
 This README is a quick introduction to Fail2Ban. More documentation, FAQ, and HOWTOs
-to be found on fail2ban(1) manpage, [Documentation](https://fail2ban.readthedocs.io/), [Wiki](https://github.com/fail2ban/fail2ban/wiki)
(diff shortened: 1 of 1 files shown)


sebres:
I don't know we should use this "unofficial" documentation directly as link in readme.
So I'm not sure.
@yarikoptic, @fail2ban/contributors Any ideas about that?
I you would like just to reference it "somewhere", you could add it to our [wiki](https://github.com/fail2ban/fail2ban/wiki) as link on the main-page "Home".

yarikoptic:
> I don't know we should use this "unofficial" documentation directly as link in readme.

What non official in it? ;) seems to be as official as it get.

> So I'm not sure.
> @yarikoptic, @fail2ban/contributors Any ideas about that?
> I you would like just to reference it "somewhere", you could add it to our [wiki](https://github.com/fail2ban/fail2ban/wiki) as link on the main-page "Home".

I would have mentioned it in Wiki as "Developers documentation" and also add a pointer in README under the same title.

sebres:
Well, then link title changed and merged.
I added also the entry into wiki main page.
Thx!

PR #2298, opened 2018-12-10
grep command to include loglines in mail could not deal with Apache's error log

chtheis:
The error log from Apache appends the port number to the IP address, e.g. 
```
[Fri Dec 07 06:18:24.826187 2018] [php7:error] [pid 87678] [client 127.0.0.1:5196] script '/usr/local/www/apache24/data/test.php' not found or unable to stat
```
The grep command to search the log files for the given IP address could not deal with it and thus the mail did not include the lines.

### DIFF
--- a/config/action.d/helpers-common.conf
+++ b/config/action.d/helpers-common.conf
@@ -4,8 +4,9 @@
 #   _grep_logs_args = 'test'
 #   (printf %%b "Log-excerpt contains 'test':\n"; %(_grep_logs)s; printf %%b "Log-excerpt contains 'test':\n") | mail ...
 #
-_grep_logs = logpath="<logpath>"; grep <grepopts> -E %(_grep_logs_args)s $logpath | <greplimit>
-_grep_logs_args = "(^|[^0-9a-fA-F:])$(echo '<ip>' | sed 's/\./\\./g')([^0-9a-fA-F:]|$)"
+_grep_logs = logpath="<logpath>"; grep <grepopts> %(_grep_logs_args)s $logpath | <greplimit>
+# options `-wF` used to match only whole words and fixed string (not as pattern)
+_grep_logs_args = -wF "<ip>"
 
 # Used for actions, that should not by executed if ticket was restored:
 _bypass_if_restored = if [ '<restored>' = '1' ]; then exit 0; fi;
--- a/ChangeLog
+++ b/ChangeLog
@@ -65,6 +65,8 @@ ver. 0.10.5-dev-1 (20??/??/??) - development edition
 ### Fixes
 * fixed read of included config-files (`.local` overwrites options of `.conf` for config-files 
   included with before/after)
+* `action.d/helpers-common.conf`: rewritten grep arguments, now options `-wF` used to match only
+  whole words and fixed string (not as pattern), gh-2298
 * `filter.d/sshd.conf`:
   - captures `Disconnecting ...: Change of username or service not allowed` (gh-2239, gh-2279)
   - captures `Disconnected from ... [preauth]` (`extra`/`aggressive` mode and preauth phase only, gh-2239, gh-2279)


chtheis:
I'm starting to understand GitHub ...
I opened another issue, #2299 instead of commenting on this one.

config/action.d/helpers-common.conf
```diff
@@ -5,7 +5,7 @@
 #   (printf %%b "Log-excerpt contains 'test':\n"; %(_grep_logs)s; printf %%b "Log-excerpt contains 'test':\n") | mail ...
 #
 _grep_logs = logpath="<logpath>"; grep <grepopts> -E %(_grep_logs_args)s $logpath | <greplimit>
-_grep_logs_args = "(^|[^0-9a-fA-F:])$(echo '<ip>' | sed 's/\./\\./g')([^0-9a-fA-F:]|$)"
+_grep_logs_args = "(^|[^0-9a-fA-F:])$(echo '<ip>' | sed 's/\./\\./g')(:\d{1,5})?([^0-9a-fA-F:]|$)"
```
sebres: The problem is the IPv6 group separator `::`, so e.g. `::1:1234` where `1234` is port and `::1:1234` as a pure IP are obviously not the same.
The anchoring `([^0-9a-fA-F:]|$)` has previously prevent confusions like this.

chtheis:
Am 11.12.2018 um 12:38 schrieb Sergey G. Brester:
> *@sebres* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In config/action.d/helpers-common.conf 
> <https://github.com/fail2ban/fail2ban/pull/2298#discussion_r240574815>:
> 
>> @@ -5,7 +5,7 @@
>   #   (printf %%b "Log-excerpt contains 'test':\n"; %(_grep_logs)s; printf %%b "Log-excerpt contains 'test':\n") | mail ...
>   #
>   _grep_logs = logpath="<logpath>"; grep <grepopts> -E %(_grep_logs_args)s $logpath | <greplimit>
> -_grep_logs_args = "(^|[^0-9a-fA-F:])$(echo '<ip>' | sed 's/\./\\./g')([^0-9a-fA-F:]|$)"
> +_grep_logs_args = "(^|[^0-9a-fA-F:])$(echo '<ip>' | sed 's/\./\\./g')(:\d{1,5})?([^0-9a-fA-F:]|$)"
> 
> The problem is the IPv6 group separator |::|, so e.g. |::1:1234| where 
> |1234| is port and |::1:1234| as a pure IP are obviously not the same.
> The anchoring |([^0-9a-fA-F:]|$)| has previously prevent confusions like 
> this.

OK, but what can be done?
Apache doesn't allow to have an IP address without port number in the 
error log, so I can't change the format to fit my purpose.

I don't have IPv6 so it would help me if I could put the stuff into a 
helpers-common.local file, but that didn't work.

Or: in the error log the IP address is in a format [client <ip>:<port>]
Would it help to change the optional port number to "(:\d{1-5}])?", i.e. 
include the "]"?

Or is it possible to configure the _grep_log_args per jail?

Or should I have my own specialised mail-with-lines action for Apache's 
error log?


Christoph

sebres:
> OK, but what can be done?

Well, I could try to encapsulate it using conditional init-section (e. g. `[Init?family=inet6]`). Just will expect several changes here. I'll try it later.

> Or is it possible to configure the _grep_log_args per jail?

I'm not sure the substitution of `%(_grep_log_args)s` will be affected if it will be used as init-parameter `mail_action[_grep_log_args="..."]` directly in jail. You could try it this way.
This would be definitely possible if run-time notation `<_grep_log_args>` (instead of `%(_grep_log_args)s`) would be used (or if this would belong to the init-section, not like now in the definition-section).

BTW, I don't like actions like this - IMHO parsing of log-files (via grep) to find matched lines containing failures, that is already available in the ticket self as well as within the action (with tags `<matches>` and `<ipmatches>`), looks very strange.
There are already several mail-actions that are using this tags.

chtheis:
Am 11.12.2018 um 16:09 schrieb Sergey G. Brester:
> BTW, I don't like actions like this - IMHO parsing of log-files (via 
> grep) to find matched lines containing failures, that is already 
> available in the ticket self as well as within the action (with tags 
> |<matches>| and |<ipmatches>|), looks very strange.
> There are already several mail-actions that are using this tags.

That is a good idea, I wasn't aware of these variables.
And there are even suitable actions doing this.
I'll try ...


Christoph

sebres:
@chtheis is this https://github.com/fail2ban/fail2ban/pull/2317#issuecomment-451762918 is also valid for your PR, Christoph? 
I mean `\d{5}` instead of `\d{1,5}`, since `\d{1,4}` is very untypical for dynamic port.
So this would resolve the IPv6 question here...

chtheis:
Am 07.01.2019 um 01:42 schrieb Sergey G. Brester:
> @chtheis <https://github.com/chtheis> is this #2317 (comment) 
> <https://github.com/fail2ban/fail2ban/pull/2317#issuecomment-451762918> 
> is also valid for your PR, Christoph?
> I mean |\d{5}| instead of |\d{1,5}|, since |\d{1,4}| is very untypical 
> for dynamic port.
> So this would resolve the IPv6 question here...

Yes, that would do. Though, there is a small risk missing an offending 
entry if a client uses a less than 5 digit dynamic port.

Maybe even \d{3-5} would do? In IPv6 the parts are always 2 digits. And 
we can almost definitely exclude 2 digit dynamic ports.

sebres:
> In IPv6 the parts are always 2 digits.

Hmm... what do you mean with 2 digits?
```
2001:db8::1
2001:db8::12
2001:db8::123
2001:db8::1234
```
all are valid IPv6 addresses.
So the last both (in spite of 123 and 1234 mean real hex) will definitely cross `\d{3,5}` (so this way confusing with the port), so in my opinion only `\d{5}` could except the IPv6 part.

Normally the programs logging IPv6 as `[2001:db8::1234]:55555`, just in case of this obscure action it should provide some universal solution (which is unfortunately not so simple).

chtheis:
Am 07.01.2019 um 03:12 schrieb Sergey G. Brester:
>     In IPv6 the parts are always 2 digits.
> 
> Hmm... what do you mean with 2 digits?
> 
> 2001:db8::1
> 2001:db8::12
> 2001:db8::123
> 2001:db8::1234 
 >
 > all are valid IPv6 addresses.

Sorry, my mistake. Of course they can be up to 4 digit.
I don't know which pattern I had in mind when I wrote my reply ...

sebres:
I rebased it to 0.10 (and merged with 0.11 in your branch), additionally grep arguments are rewritten now: using options `-wF` to match only whole words and fixed string (not as pattern).

PR #2302, opened 2018-12-14
Upgrade to AbuseIPDB API v2

acs-ferreira:
The IP value should be encoded because IPv6 addresses have colons.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -70,6 +70,7 @@ ver. 0.10.5-dev-1 (20??/??/??) - development edition
   message, see gh-2431)
 * fixed read of included config-files (`.local` overwrites options of `.conf` for config-files 
   included with before/after)
+* `action.d/abuseipdb.conf`: switched to use AbuseIPDB API v2 (gh-2302)
 * `action.d/helpers-common.conf`: rewritten grep arguments, now options `-wF` used to match only
   whole words and fixed string (not as pattern), gh-2298
 * `filter.d/sshd.conf`:
--- a/config/action.d/abuseipdb.conf
+++ b/config/action.d/abuseipdb.conf
@@ -83,13 +83,10 @@ actioncheck =
 #          wherever you install the helper script. For the PHP helper script, see
 #          <https://wiki.shaunc.com/wikka.php?wakka=ReportingToAbuseIPDBWithFail2Ban>
 #
-#          --ciphers ecdhe_ecdsa_aes_256_sha is used to workaround a
-#          "NSS error -12286" from curl as it attempts to connect using
-#          SSLv3. See https://www.centos.org/forums/viewtopic.php?t=52732
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-actionban = lgm=$(printf '%%s\n...' "<matches>"); curl --fail --tlsv1.1 --data "key=<abuseipdb_apikey>" --data-urlencode "comment=$lgm" --data "ip=<ip>" --data "category=<abuseipdb_category>" "https://www.abuseipdb.com/report/json"
+actionban = lgm=$(printf '%%.1000s\n...' "<matches>"); curl -sSf "https://api.abuseipdb.com/api/v2/report" -H "Accept: application/json" -H "Key: <abuseipdb_apikey>" --data-urlencode "comment=$lgm" --data-urlencode "ip=<ip>" --data "categories=<abuseipdb_category>"
 
 # Option:  actionunban
 # Notes.:  command executed when unbanning an IP. Take care that the


AndCycle:
I got something for you to make some improvement,

`$(printf '%%s\n...' "<matches>");`
to
`$(printf '%%.1000s\n...' "<matches>");`

as official document state comment have limit to 1024 bytes, so you need to cut it under the limit.

AndCycle:
in addition,
`curl --silent --show-error` might produce better stderr when failed, as it turn off the progress meter.

sebres:
extended as suggested by @AndCycle

lrosenman:
This works great for me on FreeBSD.  Can this be merged, please?  Using the V1 endpoint garnered 429's for *EVERY* request.

sebres:
Rebased over 0.10 branch and merged now in both major branches.
Thanks all contributing!

feloxxx:
Hello everyone, I downloaded and installed the latest version of fail2ban v1.0.1.dev1 here.

I only have a jail for ssh, it works perfectly. however, no action is submitted to abuseipdb. i don't find any mistakes. for exactly this reason I have to update the fail2ban packet provided by my debian to this version.
after the mentioned action to report to abuseipdb didn't work for me either, i also tried the version from the manufacturer's homepage. https://www.abuseipdb.com/fail2ban.html, but also without success.
the modification of the action in abuseipdb.conf, as on the homepage  was also unsuccessful.

Does somebody has any idea? Thank you so much!

sebres:
> however, no action is submitted to abuseipdb. 

If you see some errors in fail2ban.log please provide that here.
If not, it could be some mistake in your configuration, please provide the dump generated with `fail2ban-client -d | grep sshd`.
Don't forget to blacken your abuseipdb API-key.

feloxxx:
**thank you for your quick reply.** 

I only find these lines in the log file when I search for an error:

[logfile.txt](https://github.com/fail2ban/fail2ban/files/4818896/logfile.txt)

and here the output from : `fail2ban-client -d | grep sshd`
[grepssh.txt](https://github.com/fail2ban/fail2ban/files/4818914/grepssh.txt)

**Many thanks for the support!** Felix

sebres:
> `['set', 'sshd', 'addaction', 'iptables-multiport']`

You have not configured abuseipdb action at all (only `iptables-multiport` which is probably default banning action for sshd jail on your distribution). 
Fail2ban has hundred of actions, and since it cannot read your mind, how you think it would choice abuseipdb?

Per default only banaction is configured in fail2ban (`action = %(action_)s` and `action_ = %(banaction)s[...params...]`).

Please configure it for the jail or default section and restart fail2ban or the jail. See [action.d/abuseipdb.conf](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/abuseipdb.conf#L22-L24) for usage example and details.
Possible configuration in your jail.local for sshd jail would be something like this:
```ini
[sshd]
action = %(action_)s
         abuseipdb[abuseipdb_apikey="your-api-key", abuseipdb_category="18,22"]
```

feloxxx:
thanks again ... I'm already very enthusiastic about fail2ban, I'm experimenting and abuseipdb is a useful extension for me.

here is my excerpt from jail.local

`#
# See action.d/abuseipdb.conf for usage example and details.
#
action_abuseipdb = abuseipdb

# Choose default action.  To change, just override value of 'action' with the
# interpolation to the chosen action shortcut (e.g.  action_mw, action_mwl, etc) in jail.local
# globally (section [DEFAULT]) or per specific section
#action = %(action_)s
action = %(action_)s%(action_abuseipdb)s[abuseipdb_apikey="theapikay", abuseipdb_category="18"]`

and here the ssh jail

`[sshd]

# To use more aggressive sshd modes set filter parameter "mode" in jail.local:
# normal (default), ddos, extra or aggressive (combines all).
# See "tests/files/logs/sshd" or "filter.d/sshd.conf" for usage example and details.
mode   = normal
enabled  = true
port    = ssh
filter = sshd
logpath = %(sshd_log)s
backend = %(sshd_backend)s

action = %(action_)s%(action_abuseipdb)s[abuseipdb_apikey="mykey",abuseipdb_category="18,22"]
`

i think it's okay ;)

sebres:
> i think it's okay ;)
> `action = %(action_)s%(action_abuseipdb)s...`

No, it is not... As you can see in my example above there is a NL+SPACE(s) chars before the second action (and in your excerpt it is in single line).
Multiple actions must be separated by new-line (and ini-config expects at least one space to join next-line to parameter).

feloxxx:
oh my god ..... many thanks and sorry for my questions! it works!

1000 thanks!

PR #2306, opened 2018-12-18
Update DEVELOP reg. ipdns.py as successor for dnsutils.py

koeppea:
Just spotted the mentioning of dnsutils.py which is not valid anymore.
There is not much to explain for this pull req.


sebres:
Rebased to 0.10 and merged in both branches, thx!

PR #2317, opened 2019-01-03
Fix asterisk filter not catching attackers when port is logged (Fixes…

### LINKED ISSUE #2316: asterisk filter missing attackers
The filters defined for asterisk unfortunately miss some attacks with the latest asterisk.
These are the log-lines generated by asterisk:
```
[2019-01-03 23:38:17] WARNING[8447][C-00000244] Ext. s: "Rejecting unknown SIP connection from 212.129.3.61:52126"
```
The filter `^%(__prefix_line)s%(log_prefix)s "Rejecting unknown SIP connection from <HOST>"$` (https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/asterisk.conf#L27) SHOULD catch this, but doesn't.

As this is one of the most common attacks on asterisk, this seems like a pretty big issue to me.

Yannik:
This fixes #2316.

### DIFF
--- a/config/filter.d/asterisk.conf
+++ b/config/filter.d/asterisk.conf
@@ -24,7 +24,7 @@ failregex = ^%(__prefix_line)s%(log_prefix)s Registration from '[^']*' failed fo
             ^%(__prefix_line)s%(log_prefix)s No registration for peer '[^']*' \(from <HOST>\)$
             ^%(__prefix_line)s%(log_prefix)s hacking attempt detected '<HOST>'$
             ^%(__prefix_line)s%(log_prefix)s SecurityEvent="(?:FailedACL|InvalidAccountID|ChallengeResponseFailed|InvalidPassword)"(?:(?:,(?!RemoteAddress=)\w+="[^"]*")*|.*?),RemoteAddress="IPV[46]/(UDP|TCP|WS)/<HOST>/\d+"(?:,(?!RemoteAddress=)\w+="[^"]*")*$
-            ^%(__prefix_line)s%(log_prefix)s "Rejecting unknown SIP connection from <HOST>"$
+            ^%(__prefix_line)s%(log_prefix)s "Rejecting unknown SIP connection from <HOST>(?::\d+)?"$
             ^%(__prefix_line)s%(log_prefix)s Request (?:'[^']*' )?from '(?:[^']*|.*?)' failed for '<HOST>(?::\d+)?'\s\(callid: [^\)]*\) - (?:No matching endpoint found|Not match Endpoint(?: Contact)? ACL|(?:Failed|Error) to authenticate)\s*$
 
 # FreePBX (todo: make optional in v.0.10):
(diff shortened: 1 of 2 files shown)


sebres:
Basically because RE is greedy, it does no matter whether it'd be anchored or not in case of short form of IPv6 (with optional port):
```python
def test(RE, str):
  return re.search(Regex._resolveHostTag(RE), str).groupdict()

# short form of IPv6 (with optional port):
test(r'^"<ADDR>(?::\d+)?"', '"2001:db8::1:1234"')      #	==> {'ip6': '2001:db8::1:1234', 'ip4': None}
# short form of IPv6 (with mandatory port)
test(r'^"<ADDR>(?::\d+)"', '"2001:db8::1:1234"')       #	==> {'ip6': '2001:db8::1', 'ip4': None}
# written-out full form of IPv6:
test(r'^"<ADDR>(?::\d+)?"', '"2001:db8:0:0:0:0:0:1:1234"')#	==> {'ip6': '2001:db8:0:0:0:0:0:1', 'ip4': None}
# short form of IPv6 (enclosed in square brackets):
test(r'^"<ADDR>(?::\d+)?"', '"[2001:db8::1]:1234"')    #	==> {'ip6': '2001:db8::1', 'ip4': None}
```
So for the possible confusing cases, it'd be better to check the output of IPv6 (if simple "wrong" short form like first example showing) and:
- either to fix the log-output in pjsip module (e. g. written-out full form of IPv6 or square brackets or port written as `, port 1234`)
- or switch locally the port-part to mandatory (remove `?`) if port is within log-output.

Anyway it also does no matter if the 5-digits port used, because our IP-RE-part expects up to four digits, and here it's anchored on end of word (by closing quote), thus `:12345` would be never recognized as part of the IPv6 (e. g. neither as `:1` nor as `:1234`), here as PoC:
```
# short form of IPv6 (with optional port), five-digit port:
test(r'^"<ADDR>(?::\d+)?"', '"2001:db8::1:12345"')     #	==> {'ip6': '2001:db8::1', 'ip4': None}
```
This means also - the sentence "an attacker could just use a 4-digit port to circumvent this failregex" is valid for the IPv6 short form too (without brackets). Still worse, indirectly it'd cause an injection on foreign IP (ban another IPs as intruder uses).
But I don't know at the moment what we could do to fix it.

So I'll merge it now as is. 
Thanks.

Yannik:
Hey @sebres,
I just jumped through some hoops to test this with IPv6. The log *does* use brackets for IPv6 addresses, when a port is specified!

Example:
```
[2019-02-21 23:42:26] WARNING[12124][C-00000001] Ext. s: "Rejecting unknown SIP connection from [2003:e7:aaaa:aaaa:aaaa:aaaa:aaaa:aaaa]:5060"
```

sebres:
> Do we perhaps have to modify the regex to consider this syntax?

Nope, the regexp for IPv6 is safe in 0.10 (automatically recognized brackets), see my tests above, I used fail2ban API there.
And 0.9 does not support IPv6 officially.

So it looks alright in my opinion.

PR #2319, opened 2019-01-08
Added support for Request tracker login failures

egroeg:

### DIFF
--- a/config/filter.d/request-tracker
+++ b/config/filter.d/request-tracker
@@ -1,0 +1,9 @@
+#  Fail2Ban filter for Request tracker login failures
+
+#
+
+[Definition]
+
+failregex = RT: (?:\[\d+\]) FAILED LOGIN for .* from <HOST>
+
+ignoreregex =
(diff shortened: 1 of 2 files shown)


config/filter.d/request-tracker
```diff
@@ -0,0 +1,9 @@
+#  Fail2Ban filter for Request tracker login failures
+
+#
+
+[Definition]
+
+failregex = RT: (?:\[\d+\]) FAILED LOGIN for .* from <HOST>
```
sebres: This regex is neither anchored (should be at least start-anchored `^`), nor the catch-all (`.*`) will be accepted from us.
I don't see the test log-messages (BTW this is mandatory, otherwise test-suite will fail), so I don't know how exactly it must be rewritten, but somehow like this:
```diff
-failregex = RT: (?:\[\d+\]) FAILED LOGIN for .* from <HOST>
+failregex = ^\s*RT: (?:\[\d+\]) FAILED LOGIN for \S+ from <HOST>
```

sebres: Additionally this filter-file should be renamed to `request-tracker.conf`

PR #2328, opened 2019-01-29
Adjusted default exim main log path in Arch

timosmit:
Arch uses a slightly different path for Exim's main log file. This fix prevents fail2ban from failing during startup.

### DIFF
--- a/config/paths-arch.conf
+++ b/config/paths-arch.conf
@@ -13,7 +13,7 @@ apache_error_log = /var/log/httpd/*error_log
 
 apache_access_log = /var/log/httpd/*access_log
 
-exim_main_log = /var/log/exim/main.log
+exim_main_log = /var/log/exim/mainlog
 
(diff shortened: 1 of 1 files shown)


sebres:
No idea where it really is and whether it affects all versions (of arch/exim).

@felixonmars, @fail2ban/maintainers:
1. is it always the case there?
2. should this PR better target a 0.10 branch possibly? (what is actually released there).

felixonmars:
Since I did not do anything in exim packaging (I am the maintainer), I confirmed that the "mainlog" path is actually the upstream default:

https://github.com/Exim/exim/blob/afdb5e9cf07fa49e26e128d8d5d2e3cab7a5fe42/src/src/eximon.src#L106

I have tried to trace the history and the name was chosen since the very beginning.

So I'm afraid the "main.log" version is either a downstream patch of other distribution or a typo.

sebres:
Thx! 
Thus I'll rebase it to 0.10th.

PR #2339, opened 2019-02-08
Add override for dovecot failed logins on debian

cFire:
On Debian 9 (stretch), by default failed login attempts for Dovecot are logged in `/var/log/auth.log` rather than `/var/log/mail.warn`. This override corrects for this behavior.

This behavior should probably be independently verified for:
 - Other Debian releases (both older and newer if possible)
 - Independent Debian installations to entirely ensure this **really** is default behavior for Dovecot.

(Time permitting, I will do both of these things soon. But if any other Debian users have noticed this behavior too, please let us know.)

### DIFF
--- a/config/paths-debian.conf
+++ b/config/paths-debian.conf
@@ -31,6 +31,8 @@ apache_error_log = /var/log/apache2/*error.log
 apache_access_log = /var/log/apache2/*access.log
 
 
+dovecot_log = /var/log/mail.log
+
 # was in debian squeezy but not in wheezy
 # /etc/proftpd/proftpd.conf (SystemLog)
 proftpd_log = /var/log/proftpd/proftpd.log
--- a/config/paths-debian.conf
+++ b/config/paths-debian.conf
@@ -31,7 +31,7 @@ apache_error_log = /var/log/apache2/*error.log
 apache_access_log = /var/log/apache2/*access.log
 
 
-dovecot_log = /var/log/mail.log
+dovecot_log = %(syslog_mail)s
 
 # was in debian squeezy but not in wheezy
 # /etc/proftpd/proftpd.conf (SystemLog)


sebres:
Hmm... for 0.10 it was rewritten (57ea38c342b2caa17b61a8cd17f142c218fd0742, together with Postfix logfile), so it is `/var/log/mail.log` there now.

Still the question BTW, as I get updated some of my VPS to stretch, I don't saw it get changed. Must check.

cFire:
I've tested on Debian 7, 8, 9, and 10 and they all behave consistently under the default dovecot configuration;
Failed logins are logged to `/var/log/auth.log` and `/var/log/mail.log`, but not to `/var/log/mail.warn`.
(The log format for `auth.log` and `mail.log` are different, but both are correctly matched by the default dovecot regex set.)

So I think it probably would make the most sense to have the debian default set to `/var/log/mail.log` just as it is in 0.10?

sebres:
Yes, is is better (`auth.log` is more common log, so contains messages from other services too, as well as PAM-failures, etc).
You can change it in your PR, and I'd also merge it, just it is fixed in newer version and v.0.9 is outdated (so would not get new releases anymore).

cFire:
Changed the MR to override `dovecot_log` with the existing `syslog_mail` variable instead of a static path.

cFire:
Somehow it took me until now to realize this PR is entirely pointless since https://github.com/fail2ban/fail2ban/commit/57ea38c342b2caa17b61a8cd17f142c218fd0742

sebres:
Well this commit targeting version 0.10, your PR - 0.9, so this is totally correct amend (and may be helpful for people still using 0.9th version).
Thx anyway.

cFire:
Ah, you are entirely correct.

PR #2353, opened 2019-02-21
Add asterisk ipv6 test cases with and without port (related to #2317)

Yannik:
This is related to #2317


sebres:
I rebased it to 0.10 (0.9 does not support IPv6) and rewritten it a bit to cover short form of IPv6 (see https://github.com/fail2ban/fail2ban/pull/2317#issuecomment-466126000), because written-out full form is safe, no matter with or without square brackets.

sebres:
Tests are succeed, thus merging.
Thx,

PR #2357, opened 2019-02-25
Sent sender for %(mta)s-whois for asterisk and freeswitch jails

opoplawski:
In response to Fedora bug report - https://bugzilla.redhat.com/show_bug.cgi?id=1666855

Provided configuration for Asterisk jail does not use the proper sender. In my installation this is causing the messages to be queued and never delivered as the default sender it is using is a bare username with no domain.

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -701,7 +701,7 @@ logpath = /var/log/nsd.log
 port     = 5060,5061
 action   = %(banaction)s[name=%(__name__)s-tcp, port="%(port)s", protocol="tcp", chain="%(chain)s", actname=%(banaction)s-tcp]
            %(banaction)s[name=%(__name__)s-udp, port="%(port)s", protocol="udp", chain="%(chain)s", actname=%(banaction)s-udp]
-           %(mta)s-whois[name=%(__name__)s, dest="%(destemail)s"]
+           %(mta)s-whois[name=%(__name__)s, sender="%(sender)s", dest="%(destemail)s", chain="%(chain)s"]
 logpath  = /var/log/asterisk/messages
 maxretry = 10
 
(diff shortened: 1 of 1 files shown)


sebres:
Hmm... Normally if this tags were not substituted in action it should be then interpolated in jail.
Although I don't know all actions (affected by `%(mta)s`) would result in this, so it were indeed better to do it explicitly.
Just I don't know the reason why it is at all integrated in both jails per default. Basically it is not the case for any jail.
My intention right now is to remove it resp. to replace it with something like:
```diff
-action   = %(banaction)s[name=%(__name__)s-tcp, port="%(port)s", protocol="tcp", chain="%(chain)s", actname=%(banaction)s-tcp]
-           %(banaction)s[name=%(__name__)s-udp, port="%(port)s", protocol="udp", chain="%(chain)s", actname=%(banaction)s-udp]
+action_  = %(banaction)s[name=%(__name__)s-tcp, port="%(port)s", protocol="tcp", chain="%(chain)s", actname=%(banaction)s-tcp]
+           %(banaction)s[name=%(__name__)s-udp, port="%(port)s", protocol="udp", chain="%(chain)s", actname=%(banaction)s-udp]
-           %(mta)s-whois[name=%(__name__)s, dest="%(destemail)s"]
```
So the admin would be able to decide self and could enable mail in own `jail.local`, using something like:
```ini
[asterisk]
action = %(action_mw)s
```
I must check it is possible, my assumption is - we need to rewrite default substitutions of all action_mw's to something similar this:
```ini
action_mw = %(action_)s
            %(mta)s-whois[name=%(__name__)s, sender="%(sender)s", dest="%(destemail)s", protocol="%(protocol)s", chain="%(chain)s"]
```
BTW, as for `sender="%(sender)s"` - the interpolation would fail if sender address contains quote, (and usage of single quotes, so `sender='%(sender)s'`, is not good too, because regarding RFC 5322 single quote may be a part of e-mail.
We should check what would happen here:
```ini
[DEFAULT]
sender = Brian O'Conner <"brian o'conner"@example.com>
```
Possibly we'd need still another way to supply parameters.

opoplawski:
Any progress on this?

sebres:
As I fixed #2686 I partially rewrote `action` interpolation in `jail.conf` (see 06b46e92eb3fc11c29974fea1753a96edaf8ec20) that also affecting jails mentioned in this PR, for the instance:
- as main rule: don't specify `action` directly in jails (use `action_` or `banaction` instead), this allows to specify own actions enhancements (like it is possible for other jails);
- no mails-actions added per default anymore (e. g. to allow that `action = %(action_mw)s` should be specified per jail or in default section in `jail.local`);
  this basically closes this PR;

So after upgrade the only thing one should do here, would be to specify the mail-action (if it is not done already), for example:
```ini
[DEFAULT]
action = %(action_mw)s
```

PR #2366, opened 2019-03-04
add reading of local config for firewallcmd-common

rgessner:
firewallcmd-common.conf is missing the [INCLUDES] section as seen in other action.d files (e.g. nftables-common.conf or iptables-common.conf).


rgessner:
disregard

PR #2367, opened 2019-03-04
add reading of firewallcmd-common.local

rgessner:
config/action.d/firewallcmd-common.conf is missing the [INCLUDES] section known from other action.d files. (e.g. iptables-common.conf / nftables-common.conf ) to read a .local file for modification of the defaults.

### DIFF
--- a/config/action.d/firewallcmd-common.conf
+++ b/config/action.d/firewallcmd-common.conf
@@ -2,6 +2,12 @@
 #
 # Author: Donald Yandt
 #
+# The user can override the defaults in firewallcmd-common.local
+#
+
+[INCLUDES]
+
+after = firewallcmd-common.local
 
 [Init]
 


sebres:
Thx for PR...
But it is not expected - `x.local` file will be automatically applied after corresponding `x.conf` file.

If it does not work on your system, it could be still a known bug in your version, which was already fixed in c1ccabc1f922506a8f01fbe9551762e15f965068.

---
Just for the record...
PoC with f2b test suite (replace `INPUT_direct` with `INPUT_XXX` in `firewallcmd-common.local`) on current 0.10th version:
```
$ cat ./config/action.d/firewallcmd-common.local

[Init]
chain = INPUT_XXX

$ bin/fail2ban-testcases -fn testCheckStockCommandActions

Fail2ban 0.10.5.dev1 test suite. Python 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]. Please wait...
F
FAIL: testCheckStockCommandActions (fail2ban.tests.servertestcase.ServerConfigReaderTests)
...
AssertionError: '`firewall-cmd --direct --add-rule ipv4 filter INPUT_direct 0 -m conntrack --ctstate NEW -p tcp -m multiport --dports http,https -j f2b-j-w-fwcmd-mp`' was not found in the log: ===
===== # === ban-ipv4 === =====
exec-cmd: `firewall-cmd --direct --add-chain ipv4 filter f2b-j-w-fwcmd-mp`
exec-cmd: `firewall-cmd --direct --add-rule ipv4 filter f2b-j-w-fwcmd-mp 1000 -j RETURN`
exec-cmd: `firewall-cmd --direct --add-rule ipv4 filter INPUT_XXX 0 -m conntrack --ctstate NEW -p tcp -m multiport --dports http,https -j f2b-j-w-fwcmd-mp`
exec-cmd: `firewall-cmd --direct --get-chains ipv4 filter | sed -e 's, ,\n,g' | grep -q '^f2b-j-w-fwcmd-mp$'`
exec-cmd: `firewall-cmd --direct --add-rule ipv4 filter f2b-j-w-fwcmd-mp 0 -s 192.0.2.1 -j REJECT --reject-with icmp-port-unreachable`
===
```
So it failed as expected here because does not find `... INPUT_direct ...` anymore.

PR #2388, opened 2019-03-29
Update sendmail-reject with TLSMTA & MSA port IDs

cepheid666:
Updates `sendmail-reject` filter and associated test files to accommodate port IDs of TLSMTA and MSA, which are default for ports 465 and 587 (respectively) on some distros (e.g., RHEL/CentOS).


cepheid666:
(Not sure what to do about the above checks... complaining about environment variables...)

Issue #1284, opened 2015-12-29
How to block Bad protocol version?


twixi:
I have more 500 in auth.log (ssh port change to 23):
I have more:
sshd[20620]: Bad protocol version identification 'guest' from 124.244.54.245 port 39405
sshd[16472]: Did not receive identification string from 87.76.240.167
How to block this in rules!!?

sebres:
This are no failures in sense of authentication (because login does not take place).
But if you will that yet, just copy `filter.d/sshd.conf` into `filter.d/sshd.local` and add following to the `failregex`:
```
^%(__prefix_line)sBad protocol version identification '.*' from <HOST> port \d+\s*$
^%(__prefix_line)sDid not receive identification string from <HOST>\s*$
```

benrubson:
Hi @sebres,
Could we think about adding the `Bad protocol version` rule to the SSHd filter please ?
I'm also facing these messages...
`Bad protocol version identification '.........' from A.B.C.D port 1234`
Thank you very much 👍

benrubson:
I've submitted PR #2404 for this :)

mhf-ir:
Simple script to generate such a log:
```
 watch -n 0.01 curl -m 1 127.0.0.1:22
```

Potential dos/ddos for sshd :warning: I thinks it's urgent

sebres:
Fix (#2404) has been merged into 0.10/0.11 for mode `ddos` or `aggressive`.

> Potential dos/ddos for sshd warning I thinks it's urgent

Well, DDOS is potentially possible for any service (sshd is not an exception here), and it is also effectively solvable using other tools, for example special rules like limiting the connection rate (total request per IP/subnet per minutes, SYN attempts on sshd or all ports), prohibiting of the port-scanning, etc.

Primary goals of Fail2ban are authentication issues, therefore enclosed in a mentioned modes only, so this is now enough as configuration to enable it (if the filter/fail2ban becomes upgraded to 0.10.5):
```ini
[sshd]
mode = aggressive
enabled = true
```
For fail2ban \<= 0.10.5 (but \>= 0.10):
```ini
[sshd]
mode = aggressive
failregex = %(known/failregex)s
            ^Bad protocol version identification '.*' from <HOST>
enabled = true
```
For 0.9 and below see my comment above.

PR #2404, opened 2019-04-21
SSHd add Bad protocol version message

benrubson:
Hi,

This PR adds the `Bad protocol version identification '.........' from A.B.C.D port 1234` message to the SSHd filter, as discussed in #1284.

Thank you very much 👍

### DIFF
--- a/config/filter.d/sshd.conf
+++ b/config/filter.d/sshd.conf
@@ -64,6 +64,7 @@ mdre-normal =
 mdrp-normal-suff-onclosed = (?:%(__suff)s|\s*)$
 
 mdre-ddos = ^Did not receive identification string from <HOST>
+            ^Bad protocol version identification '.*' from <HOST>
             ^Connection <F-MLFFORGET>reset</F-MLFFORGET> by <HOST>
             ^Connection <F-MLFFORGET>closed</F-MLFFORGET> by%(__authng_user)s <HOST>%(__on_port_opt)s\s+\[preauth\]\s*$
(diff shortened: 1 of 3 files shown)


benrubson:
I don't really know why the test fails...
Any hint ?
Thank you 👍

sebres:
There is still another test filter (zzz-sshd-obsolete-multiline) which covers sshd test file (but you've extended a sshd filter only).
But you don't need to fix this - we've still another tasks for sshd, so I'll cherry-pick your commit and merge all later.

benrubson:
OK, perfect, many thanks @sebres 👍

benrubson:
While here, some news about this perhaps @sebres ? :)
Many thanks :)

benrubson:
To help I had a look at `zzz-sshd-obsolete-multiline` but sounds like it references `sshd`, so I'm not really sure how to deal with it :)

benrubson:
Should be OK with my last commit 👍

benrubson:
Let's merge this @sebres as it's now OK ?
Thank you again 👍

sebres:
> as it's now OK ?

Well, I don't really like your regexp, because is not anchored from right and a catch-all in the middle before the host IP, and although it does not allow a direct injection (illustrated in eb308d0fc86709dbb638618c76d10f1bd23e57f2) because it's greedy, and it is not applicable on other messages due to the anchor left, it remains a catch-all (with all cons of it).

Of course the prefilter with `prefregex` eliminates several disadvantages of that (the anchored part is fixed), so at least in new filter it's OK...

Well, the inject is covered now, let us merge it.

benrubson:
Many thanks @sebres for your detailed explanation, and finally for your merge 👍

Issue #2392, opened 2019-04-01
Set journalflags to 6 (SYSTEM_ONLY + RUNTIME_ONLY) by default


opoplawski:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.10.4
- OS, including release name/version: EL7
- [X ] Fail2Ban installed via OS/distribution mechanisms
- [X] You have not applied any additional foreign patches to the codebase

### The issue:

By default, fail2ban opens a journal Reader with no flags.  This open every file under /var/log/journal/MACHINE-ID, which can include a *lot* of user session files (my mail server has almost 500 for instance).  This can lead to "Too many open files" errors like #2208 

Since it is *highly* unlikely that any services that fail2ban is interested in are running under user sessions, and we start reading from Now - FindTime (probably at most 1d for recidive), I propose that we make journalflags=6 (SYSTEM_ONLY + RUNTIME_ONLY) the default.  At least on my test machine this does not open any files in /var/log/journal, presumably streaming just from the running journald process.

opoplawski:
Except for the fact that fail2ban appears to not find anything at all with journalflags=6, at least on EL7.  Seems to work with journalflags=4.  I wonder what is up with that.

sebres:
Hmm... possibly a flag `RUNTIME_ONLY(2)` wins over `SYSTEM_ONLY(4)`.
But related to documentation:
>  a set of constants which maybe bitwise OR-ed together and passed to the Journal constructor.

So I don't really know which combination is proper or better to use per default.

BTW, there is also another flag `LOCAL_ONLY(1)`
> `LOCAL_ONLY(1)` - Only open journal files generated on the local machine.

PR #2444, opened 2019-06-11
systemd-backend: switched default flags to SYSTEM_ONLY(4)

sebres:
Default `journalflags` is set to SYSTEM_ONLY(4). This conforms to following settings:
```ini
backend = systemd[journalflags=4]
```
This default setting should avoid to open the user session files, so could prevent "Too many open files" errors (like #2208) on a lot of user sessions;
(following @opoplawski's proposal in #2392).

Since default journal reader flag seems to be LOCAL_ONLY(1), this is not backwards-compatible change, to switch back to original value use:
```ini
backend = systemd[journalflags=1]
```

Note also the source code of python-systemd module, e. g.:
https://github.com/systemd/python-systemd/blob/5d3be8ccba5fe48b09e2a4e816f21c39038adfd0/systemd/_reader.c#L227-L232

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,6 +35,8 @@ ver. 0.10.5-dev-1 (20??/??/??) - development edition
 -----------
 
 ### Fixes
+* [compatibility] systemd backend: default flags changed to SYSTEM_ONLY(4), fixed in gh-2444 in order to ignore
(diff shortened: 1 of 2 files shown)


sebres:
@fail2ban/maintainers, @fail2ban/contributors any objections?

szepeviktor:
I am a still-SysVinit-guy.

PR #2452, opened 2019-06-26
Badips key is only used to retrieve list

benrubson:
Hi,

This PR corrects the badips action script.
The bankey is only used to retrieve the list of banned IPs.

The IP of the server adding a new ban has first been associated with the bankey, thanks to the badips API : https://www.badips.com/documentation
No need then to use it when banning, as per the documentation.

Of course, has been tested :)

Many thx 👍

### DIFF
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -54,9 +54,6 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 	age : str, optional
 		Age of last report for bad IPs, per badips.com syntax.
 		Default "24h" (24 hours)
-	key : str, optional
-		Key issued by badips.com to report bans, for later retrieval
-		of personalised content.
 	banaction : str, optional
 		Name of banaction to use for blacklisting bad IPs. If `None`,
 		no blacklist of IPs will take place.
@@ -67,8 +64,8 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 		"postfix", but want to use whole "mail" category for blacklist.
 		Default `category`.
 	bankey : str, optional
-		Key issued by badips.com to blacklist IPs reported with the
-		associated key.
+		Key issued by badips.com to retrieve personal list
+		of blacklist IPs.
 	updateperiod : int, optional
 		Time in seconds between updating bad IPs blacklist.
 		Default 900 (15 minutes)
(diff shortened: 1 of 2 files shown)


sebres:
> The bankey is only used to retrieve the list of banned IPs.
> No need then to use it when banning, as per the documentation.

Unless you want to use different keys to retrieve IPs (e. g. reported from another action) and to ban in this action. Like something cascaded or nested jails.

benrubson:
You can't ban specifying key in you request.
Or you could add IPs to others' lists.
Key is associated to your IP once for all, thanks to the `GET /set/key/<key>` call.
Look at the `GET /add` call in the doc, it confirms there is no key option to it :)

sebres:
I don't use badips, so don't know exactly how one can use that (just loud thoughts)...

> Key is associated to your IP once for all, thanks to the GET /set/key/<key> call.

Ok, the documentation is a bit confusing here, see:
> ... or just append the "?key=yourkey" parameters to the URLs

So the question is - can you obtain more as one key?
And another question: if the key is associated with reporter IP, why then the "bankey" is still expected calling getList function (`/get/list/...` request)?
And related: can one use a bankey there is not associated with the caller IP address?

benrubson:
> I don't use badips, so don't know exactly how one can use that (just loud thoughts)...
> 
> > Key is associated to your IP once for all, thanks to the GET /set/key/ call.
> 
> Ok, the documentation is a bit confusing here, see:
> 
> > ... or just append the "?key=yourkey" parameters to the URLs

Yep, this is simply for getting lists, not for appending to them...

> So the question is - can you obtain more as one key?

It will then erase the previous one associated to your reporter IP.
You'll then have to use `?force` to confirm you replace your existing key.

> And another question: if the key is associated with reporter IP, why then the "bankey" is still expected calling getList function (`/get/list/...` request)?

Because you could request IPs banned by everybody :)
For example, you could request SSH banned IPs in the last hour all over the world that have a very high ban score.

> And related: can one use a bankey there is not associated with the caller IP address?

Yes, you can retrieve the list from an IP which is not associated with the key (just re-tested).
I also re-tested to ban an IP with `?key=...`, request correctly returns but the IP is added to the key associated with the reporter IP, not to the one given as option.

benrubson:
@sebres let's merge this if you're OK ?
Many thanks 👍

benrubson:
Thank you Sergey !

Issue #2129, opened 2018-05-14
Stop messages when IP's are re-banned on startup


thatbobby:
_We will be very grateful, if your problem was described as completely as possible,
enclosing excerpts from logs (if possible within DEBUG mode, if no errors evident
within INFO mode), and configuration in particular of effected relevant settings
(e.g., with ` fail2ban-client -d | grep 'affected-jail-name' ` for a particular
jail troubleshooting).
Thank you in advance for the details, because such issues like "It does not work" 
alone could not help to resolve anything!
Thanks! (remove this paragraph and other comments upon reading)_

### Environment:

_Fill out and check (`[x]`) the boxes which apply. If your Fail2Ban version is outdated, 
and you can't verify that the issue persists in the recent release, better seek support 
from the distribution you obtained Fail2Ban from_

- Fail2Ban version (including any possible distribution suffixes):   0.9.6
- OS, including release name/version:    Raspbian version 9
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)
 
### The issue:
Everything works great; bad IP's are detected for a variety of reasons, bad IP's get banned, and I get IFTTT notifications of the banned IP's. However, on system reboot, the bad IP's are re-banned (which is great) and I get another set of IFTTT messages (which I don't want). Is there a way to configure fail2ban so I don't get IFTTT messages when the IP's are re-banned on a reboot/restart (but still get IFTTT messages on regular banning)?

#### Expected behavior
Can fail2ban be configured to NOT send notifications of re-banned IP's when service is restarted?

### Configuration, dump and another helpful excerpts

------------------------PART OF JAIL.LOCAL WITH IFTTT ---------------------------------
#Report ban via IFTTT
action_ifttt = %(banaction)s[name=%(__name__)s, port="%(port)s", protocol="%(protocol)s", chain="%(chain)s"]
               ifttt[name=%(__name__)s, port="%(port)s", protocol="%(protocol)s", chain="%(chain)s"]

#Choose default action.  To change, just override value of 'action' with the
#interpolation to the chosen action shortcut (e.g.  action_mw, action_mwl, etc) in jail.local
#globally (section [DEFAULT]) or per specific section
#action = %(action_)s
action = %(action_ifttt)s

#JAILS

#SSH servers

[sshd]
enabled = true
port    = ssh
logpath = %(sshd_log)s
[...]

PR #2455, opened 2019-06-28
Improved blocklist_de action to not resend bans that were already reported

Thermi:
This works fine except when two bans have the exact same time stamp. Solving that corner case would require a significantly more complex solution.
This PR solves the problem that fail2ban will report already reported attacks again when fail2ban is restarted and the bans are still active.

### DIFF
--- a/config/action.d/blocklist_de.conf
+++ b/config/action.d/blocklist_de.conf
@@ -57,7 +57,6 @@ actioncheck =
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-
 actionban = curl --fail --data-urlencode "server=<email>" --data "apikey=<apikey>" --data "service=<service>" --data "ip=<ip>" --data-urlencode "logs=<matches><br>" --data 'format=text' --user-agent "<agent>" "https://www.blocklist.de/en/httpreports.html"
 
 # Option:  actionunban
--- a/config/action.d/blocklist_de.conf
+++ b/config/action.d/blocklist_de.conf
@@ -30,6 +30,9 @@
 
 [Definition]
 
+# bypass reporting of restored (already reported) tickets:
+norestored = 1
+
 # Option:  actionstart
 # Notes.:  command executed on demand at the first ban (or at the start of Fail2Ban if actionstart_on_demand is set to false).
 # Values:  CMD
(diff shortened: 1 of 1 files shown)


sebres:
> except when two bans have the exact same time stamp. 

Although I understand the intention, I'd like to know, what exactly (error?) happens in this case.

Thermi:
A time stamp file is maintained per jail. The file's time stamp is the last time a ban occured. fail2ban processes the bans starting from the oldest to the newest (due to how log files are structured). The code checks if the time stamp on the file is less than the ban time. That means that this ban was not processed yet. If that is the case, the time stamp is updated and the relevant logs are sent to blocklist.de using the command.

If two offenses (bans) have the same time stamp, then the one occuring after the first one will not trigger a reporting of the offense via the command.

sebres:
No, I see what the PR does... I meant without the PR.

Anyway I guess your original issue could be fixed easier using parameter `norestored = 1`. See https://github.com/fail2ban/fail2ban/issues/2129#issuecomment-391684114 for example.

Thermi:
Without the PR, the bans are reported multiple times which causes the service (blocklist.de) to send out abuse emails once again.

sebres:
Ok, then I suggest to rewrite the PR adding parameter `norestored = 1`. This was designated exactly for such purposes.

Thermi:
Do you still maintain older releases? If yes, this might be a good fix to add to those.

sebres:
> Do you still maintain older releases? If yes, this might be a good fix to add to those.

No, I don't think @fail2ban/maintainers still hold on older releases... and you've anyway addressed 0.11 branch (what can be simply rebased to 0.10).
And development of 0.9 is discontinued already several years.
But if you want it, just open a new PR targeting 0.9 branch with your initial solution and I'll merge both PRs regarding the target version.

sebres:
Anyway I'll provide here the possible patch for 0.9 (a bit modified for simplification):
```diff
-actionban = curl --fail --data-urlencode "server=<email>" --data "apikey=<apikey>" --data "service=<service>" --data "ip=<ip>" --data-urlencode "logs=<matches><br>" --data 'format=text' --user-agent "<agent>" "https://www.blocklist.de/en/httpreports.html"
+# If the time of the offense is later than the last ban, actually report it to blocklist.de
+actionban = [ -e "<tmpfile>" ] && [ $(stat -c %%X "<tmpfile>") -ge <time> ] && exit 0
+            curl --fail --data-urlencode "server=<email>" --data "apikey=<apikey>" --data "service=<service>" --data "ip=<ip>" --data-urlencode "logs=<matches><br>" --data 'format=text' --user-agent "<agent>" "https://www.blocklist.de/en/httpreports.html"
+            touch -d @<time> "<tmpfile>"
+
+tmpfile = /var/run/fail2ban/last-log-<name>.time
```

Thermi:
TBH I think that code is already a little bit too compacted to be easily readable. I'll make a new PR with slightly easier to read code.

BrainStone:
Any reason this hasn't been merged?

Also would a backport to 0.10 be possible?

NodoBoricua:
Any help whit this

 (systemd[1]: fail2ban.service: Failed with result 'exit-co>
lines 1-10/10 (END)
× fail2ban.service - Fail2Ban Service
     Loaded: loaded (/lib/systemd/system/fail2ban.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Sat 2024-04-27 04:28:14 AST; 21s ago
   Duration: 106ms
       Docs: man:fail2ban(1)
    Process: 12348 ExecStart=/usr/bin/fail2ban-server -xf start (code=exited, status=255/EXCEPTION)
   Main PID: 12348 (code=exited, status=255/EXCEPTION)
        CPU: 104ms)

Soy nuevo en esto y lo que no entiendo es por que días antes todo estaba funcionando bien y veía los registros como ban las ip y apague el servidor por un día y al encender de nuevo me encuentro con esta falla.

Issue #2509, opened 2019-08-20
Wrong filter for 3proxy


al-indigo:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.9.7, 1el7, noarch
- OS, including release name/version: CentOS7
- [ x] Fail2Ban installed via OS/distribution mechanisms
- [ x] You have not applied any additional foreign patches to the codebase
- [ -] Some customizations were done to the configuration (provide details below is so)
- 3proxy 0.8.13 (built from sources)

### The issue:

3proxy.conf seems to have deprecated log format for parsing 3proxy. All authorization failures are ignored by fail2ban

#### Steps to reproduce

1. Build 3proxy from the sources (make -f Makefile.Linux), run with any configuration expecting auth.
2. Install fail2ban, enable filter [3proxy]
3. Look for wrong auth connections and see that fail2ban ignores any.

#### Expected behavior

Fail2ban bans connections with wrong auth.

#### Observed behavior

Status for the jail: 3proxy
|- Filter
|  |- Currently failed:	0
|  |- Total failed:	0
|  `- File list:	/opt/3proxy/log/3proxy.log
`- Actions
   |- Currently banned:	0
   |- Total banned:	0
   `- Banned IP list:

#### Any additional information

Default format string for 3proxy changed to this:
```190820201129.773 9051 00004 - 85.26.233.216:40143 149.154.167.91:80 0 0 0 CONNECT 149.154.167.91:80```

Log format is described here: https://3proxy.ru/doc/howtoe.html#LOGFORMAT
Error format is described here: https://3proxy.ru/doc/howtoe.html#ERRORS

### Configuration, dump and another helpful excerpts

Enabled the single jail:
```[3proxy]
enabled = true
port    = 9051
logpath = /opt/3proxy/log/3proxy.log
```

#### Any customizations done to /etc/fail2ban/ configuration

None

#### Relevant parts of /var/log/fail2ban.log file:

```
2019-08-20 20:20:44,385 fail2ban.filterpoll     [20510]: DEBUG   /opt/3proxy/log/3proxy.log has been modified
2019-08-20 20:20:44,390 fail2ban.datedetector   [20510]: DEBUG   Sorting the template list
2019-08-20 20:20:44,391 fail2ban.datedetector   [20510]: DEBUG   Winning template: (?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)? with 0 hits
[...]

Issue #2193, opened 2018-08-03
Problem with nonstandard timestamp in freeswitch log


oleg-shcherbatykh:
Good afternoon.
I have problem with writing filter rules for blocking illegal attempts to non standard freeswitch installation.
I tried to use two version of fail2ban - provided by centos distro (0.9.7) and downloaded from here (v0.11.0.dev2) with absolutely the same result.
As I understand the problem is with non standard timestamp in freeswitch log. 
The typical illegal attempt in log looks like this:
`08-03 08:10:21.026299 [WARN] [SOFIA] [sofia_reg.c:2248] SIP auth failure (INVITE) on sofia profile 'external' for [41801148436701961@148.251.134.154] from ip 176.31.105.122`

And standard filter just didn't catch anything. So I've corrected filter pules for my condition:
`_pref_line = ^%(__prefix_line)s(?:\d+-\d+ \d+:\d+:\d+\.\d+)?`

`failregex = %(_pref_line)s \[WARN\] \[SOFIA\] \[sofia_reg\.c:\d+\] SIP auth (failure|challenge) \((REGISTER|INVITE)\) on sofia profile \'[^']+\' for \[[^\]]*\] from ip <HOST>$
            %(_pref_line)s \[WARN\] \[SOFIA\] \[sofia_reg\.c:\d+\] Can't find user \[[^@]+@[^\]]+\] from <HOST>$`

And when I tested with fail2ban-regex it found matched lines. But now I'm getting another error in fail2ban.log:
`2018-08-03 10:56:53,037 fail2ban.filter         [16311]: WARNING Found a match for u"08-03 07:56:53.026292 [WARN] [SOFIA] [sofia_reg.c:4130] Can't find user [101@148.251.134.154] from 176.31.105.122" but no valid date/time found for u"08-03 07:56:53.026292 [WARN] [SOFIA] [sofia_reg.c:4130] Can't find user [101@148.251.134.154] from 176.31.105.122". Please try setting a custom date pattern (see man page jail.conf(5)). If format is complex, please file a detailed issue on https://github.com/fail2ban/fail2ban/issues in order to get support for this format.`

I've tried put datepattern into jail.conf too but no success.

Can you give me right direction how to write correct filter rules for my condition please.
Thank you in advance!

sebres:
I don't know how it could work for you in fail2ban-regex without proper datepattern (it looks like it's not specified, and exactly this definitely does not belong to default date-patterns of fail2ban)...
[...]

PR #2513, opened 2019-08-26
Updated failregex and datepattern for 3proxy for 0.10+

al-indigo:
Closes: #2509

Doesn't work for 0.9 due to #2193


sebres:
You can not use 0.10 as target branch without rebase your PR branch to 0.10 too, thus I changed the base of PR to 0.11.
It can be rebased if get merged later.

PR #2537, opened 2019-10-04
Improve systemd service

viiru-:
Minor improvements for the included systemd service file.
* Stop setting PIDFile (not needed since type=simple, not forking)
* Set RuntimeDirectory instead of using both ExecStartPre and tmpfiles.d
* Set StateDirectory

In combination these help make it trivial to run fail2ban without root privileges (like it was with the initscript), just add an override with User=whatever and off you go.


BerkhanBerkdemir:
Hi @viiru-,

I just want to point out that be careful with `RuntimeDirectory=` [0] on CentOS 7. Also, old version of `systemd(1)`s don't have `RuntimDirectory=` [1].

\[0]: https://www.centos.org/forums/viewtopic.php?p=299659&sid=5cdc21cb8744c5d64b63461e2ab15910#p299659
\[1]: https://serverfault.com/questions/779634/create-a-directory-under-var-run-at-boot#comment1134983_779635

egberts:
Copied from #2838 

I think this is the direction that many distros are taking when creating the PID and socket files which is to put them under `/run/<daemon>` subdirectory.  Debian has converted nearly 90% of their daemons; RedHat has started.  Gentoo too.  Suse, as well.

One downside of using tmpfile.d approach with `/run/fail2ban` would get deleted after a reboot.    

So, [FreeDesktop](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html) has this to say:
> System daemons frequently require private runtime directories below /run to store communication sockets and similar. For these, it is better to use `RuntimeDirectory=` in their unit files (see systemd.exec(5) for details), if the flexibility provided by `tmpfiles.d` is not required.

So, there is this systemd's `RuntimeDirectory=` option.  As long as we do not have complicated sub-subdirectory under `/run/fail2ban`, we can use systemd's `RuntimeDirectory=fail2ban` setting to easily auto-create this `/run/fail2ban` subdirectory (as needed) with proper file permissions, but this would still require separate packaging for non-systemd platform.

The new `fail2ban.service.in` would have:
```ini
[Service]

Type=simple

StandardInput=null
StandardOutput=journal
StandardError=journal
EnvironmentFile=-/etc/default/fail2ban

# Relaxed security settings
Group=fail2ban
RuntimeDirectory=fail2ban
RuntimeDirectoryMode=0755

ExecStart=@SBINDIR@/fail2ban-server ${FAIL2BAN_OPT} -xf start
# if should be logged in systemd journal, use following line or set logtarget to sysout in fail2ban.local
# ExecStart=@SBINDIR@/fail2ban-server -${FAIL2BAN_OPT} xf --logtarget=sysout start

ExecStop=@BINDIR@/fail2ban-client stop
ExecReload=@BINDIR@/fail2ban-client reload

PIDFile=/run/fail2ban/fail2ban.pid

Restart=on-failure
RestartPreventExitStatus=0 255
```

### Alternative Security Settings
Assuming that there exist a user with a user ID of `fail2ban` (we block that  via /bin/nologin in /etc/passwd), 

* this user ID (either by file permission or supplemental group ID) cannot have write privilege to any of its files, so all config files and script are user-root-owned and 
* that `fail2ban` group ID is used only for socket communication or read-only of configuration files...

#### Permission by File Permission Only
```ini
Group=fail2ban
RuntimeDirectory=fail2ban
#RuntimeDirectoryMode=0755
```

#### Permission by Group Supplementary ID
For user having `fail2ban` group supplementary ID (adding a user to /etc/group):
```
Group=fail2ban
RuntimeDirectory=fail2ban
RuntimeDirectoryMode=0750  # major change here
```
-----
Note: Assigning fail2ban user to Python script isn't all that beneficial. Group, yes.
```
User=
```

egberts:
`RuntimeDirectory` was introduced in systemd v246

https://github.com/systemd/systemd/commit/e66cf1a3f94fff48a572f6dbd19b43c9bcf7b8c7

viiru-:
> Note: Assigning fail2ban user to Python script isn't all that beneficial. Group, yes.
> 
> ```
> User=
> ```

@egberts  Huh? Running any daemon with less privileges (especially without root privileges) is very beneficial for security. I'm currently using the following override for version 0.10.2:
```
[Service]
User=fail2ban
Group=systemd-journal
RuntimeDirectory=fail2ban
StateDirectory=fail2ban
ExecStartPre=
ExecStartPre=+/usr/bin/find /proc/net/xt_recent -name 'f2b-*' -exec chown fail2ban {} \;

# Hardening
NoNewPrivileges=yes
PrivateTmp=yes
PrivateDevices=yes
ProtectHome=yes
ProtectSystem=strict
ProtectKernelTunables=yes
ProtectControlGroups=yes
ProtectKernelModules=yes
SystemCallArchitectures=native
RestrictRealtime=yes
ReadWriteDirectories=-/run/fail2ban
ReadWriteDirectories=-/var/lib/fail2ban
ReadWriteDirectories=-/var/log
```
I'm sure more could be done to harden this daemon, but this should have the most obvious low-hanging fruit. This MR is about setting RuntimeDirectory and StateDirectory by default, which would make the non-root part much easier for users to apply.

egberts:
> > Note: Assigning fail2ban user to Python script isn't all that beneficial. Group, yes.
> > ```
> > User=
> > ```
> 
> @egberts Huh? Running any daemon with less privileges (especially without root privileges) is very beneficial for security. 

Not disagreeing with you there.  

1. systemd 264 is relatively new so package maintainer needs to peg its dependency by fail2ban to systemd

2. Fail2ban is an interesting design.  It’s useful for embedded system where file permission is not strictly enforced.  Again, tightening of security, at this point, is the mandate of package manager.  

3.  Also might need to investigate splitting fail2ban-server into two daemon, one for main-to-client and other for interaction with OS components.

viiru-:
>     1. systemd 264 is relatively new so package maintainer needs to peg its dependency by fail2ban to systemd

Huh again. RuntimeDirectory was introduced in version 211, in 2014: https://github.com/systemd/systemd/blob/master/NEWS#L8163

StateDirectory is more recent, introduced in version 235, in 2017: https://github.com/systemd/systemd/blob/master/NEWS#L3614

egberts:
Perhaps I should have instead stated last bug got fixed in that version.

egberts:
> > ```
> > 1. systemd 264 is relatively new so package maintainer needs to peg its dependency by fail2ban to systemd
> > ```
> 
> Huh again. RuntimeDirectory was introduced in version 211, in 2014: https://github.com/systemd/systemd/blob/master/NEWS#L8163

Perhaps I should have instead stated last bug got fixed in that version for that option. In other word, a reliable use of that option.

egberts:
> > Note: Assigning fail2ban user to Python script isn't all that beneficial. Group, yes.
> > ```
> > User=
> > ```
> 
> @egberts Huh? Running any daemon with less privileges (especially without root privileges) is very beneficial for security. I'm currently using the following override for version 0.10.2:
> 
> ```
> [Service]
> User=fail2ban
> Group=systemd-journal
> RuntimeDirectory=fail2ban
> StateDirectory=fail2ban
> ExecStartPre=
> ExecStartPre=+/usr/bin/find /proc/net/xt_recent -name 'f2b-*' -exec chown fail2ban {} \;
> 
> # Hardening
> NoNewPrivileges=yes
> PrivateTmp=yes
> PrivateDevices=yes
> ProtectHome=yes
> ProtectSystem=strict
> ProtectKernelTunables=yes
> ProtectControlGroups=yes
> ProtectKernelModules=yes
> SystemCallArchitectures=native
> RestrictRealtime=yes
> ReadWriteDirectories=-/run/fail2ban
> ReadWriteDirectories=-/var/lib/fail2ban
> ReadWriteDirectories=-/var/log
> ```
> 
> I'm sure more could be done to harden this daemon, but this should have the most obvious low-hanging fruit. This MR is about setting RuntimeDirectory and StateDirectory by default, which would make the non-root part much easier for users to apply.

I ran your latest `/etc/systemd/system/fail2ban.service` against the analyzer and there is an "invalid unit":
```console
root:/etc/systemd/system# systemd-analyze verify fail2ban.service 
Unit fail2ban.service has a bad unit file setting.
root:/etc/systemd/system# 
```

Did I miss something?   Running systemd 241-7~deb10u4

viiru-:
> > > Note: Assigning fail2ban user to Python script isn't all that beneficial. Group, yes.
> > > ```
> > > User=
> > > ```
> > 
> > 
> > @egberts Huh? Running any daemon with less privileges (especially without root privileges) is very beneficial for security. I'm currently using the following override for version 0.10.2:
> > ```
> > [Service]
> > User=fail2ban
> > Group=systemd-journal
> > RuntimeDirectory=fail2ban
> > StateDirectory=fail2ban
> > ExecStartPre=
> > ExecStartPre=+/usr/bin/find /proc/net/xt_recent -name 'f2b-*' -exec chown fail2ban {} \;
> > 
> > # Hardening
> > NoNewPrivileges=yes
> > PrivateTmp=yes
> > PrivateDevices=yes
> > ProtectHome=yes
> > ProtectSystem=strict
> > ProtectKernelTunables=yes
> > ProtectControlGroups=yes
> > ProtectKernelModules=yes
> > SystemCallArchitectures=native
> > RestrictRealtime=yes
> > ReadWriteDirectories=-/run/fail2ban
> > ReadWriteDirectories=-/var/lib/fail2ban
> > ReadWriteDirectories=-/var/log
> > ```
> > 
> > 
> > I'm sure more could be done to harden this daemon, but this should have the most obvious low-hanging fruit. This MR is about setting RuntimeDirectory and StateDirectory by default, which would make the non-root part much easier for users to apply.
> 
> I ran your latest `/etc/systemd/system/fail2ban.service` against the analyzer and there is an "invalid unit":
> 
> ```
> root:/etc/systemd/system# systemd-analyze verify fail2ban.service 
> Unit fail2ban.service has a bad unit file setting.
> root:/etc/systemd/system# 
> ```
> 
> Did I miss something? Running systemd 241-7~deb10u4

What you have quoted above is an override file, not a service file. It is incomplete alone, and requires a specific version of the actual service file to work (so that it overrides the exact right things). The first few lines were meant as an example of how to run the daemon without root privileges, and the rest as an example of what can be done in addition to that.

egberts:
Any reason why not to use `EnvironmentFile=` or something to look at the `/etc/default/fail2ban`?

Not needed?

sebres:
Related git-tree it is based on master (1.0), but you want to merge it to 0.11 branch (why not 0.10?)...
So either it should be rebased to 0.11 (or even 0.10) or we've to switch target to master here.
Because it is rather a matter of maintainers I don't really know which branch would be better suitable for this changes.

viiru-:
@sebres I noticed that after rebasing. After a year I have absolutely no recollection why this was pointed at that branch. Perhaps it was the default at that time? In any case, master should be the right target. Thanks for switching that.

sebres:
> Is there a reason why the systemd service calls directly fail2ban-server at startup ? 

Yes. It is pretty simply, just compare:
```diff
- fail2ban-client: read configuration, merge cmd-line parameters, fork process which will be server if it is possible otherwise start server process, supply configuration stream to the server process, **exit client**, server pid != client (starter) pid.
+ fail2ban-server: start cfg-thread, read configuration and merge cmd-line parameters in this thread, start server in main thread, supply configuration stream from cfg-thread to server object in main thread, stop cfg-thread, remain process (no exit).
```
Starting the service with `fail2ban-server` simply uses in-place configuration (in extra thread), no fork, no additional process, its pid is retained, etc. It is more performant and can be better controlled within the service unit.

> The man page of fail2ban-server clearly says to use fail2ban-client instead, and this what is done in the others startup scripts.

Yes, because:
- this means mainly the start of client to tell the server something (reload, restart, stop, etc), which is primary role of the client (which is done by communicating with the server);
- this feature (in-place start of server) was implemented later;
- but the usage of server directly is primarily the intern function, so the sentence is still correct.

Since #1413 the difference between `fail2ban-client` and `fail2ban-server` is marginal (one could use the server or client for almost everything, so theoretically we can combine it to `fail2ban`), but it is historical, so the commands remaining as is for backwards compatibility, and at least it looks better (in lists of top, ps, etc) if the process called `fail2ban-server` means the server and `fail2ban-client` the client.

As for the other init scripts, they can be rewritten also, if the init subsystem allows to retain a starter process, without to fork or create new process (so `service start` would not expect that the scripts ends after start), or using `fail2ban-server ... &` to start the process in the background. More or less this is the matter of maintainers of related distribution.

MichaIng (approved):

MichaIng:
Something missing here? This enhancement makes much sense. If one really uses a such old systemd, then likely because of an old distro which ships an old fail2ban as well then. In the very rare cases of old systemd (>4 years old) + current fail2ban, note that systemd ignores invalid options and prints a warning instead. User who manage to install up-to-date fail2ban on distro with >4 years old systemd will likely be able to figure out that they need to pre-create or change the directory for the PID file.

Issue #2540, opened 2019-10-10
pure-ftpd filter conf file contains french insults


dricoco:
the filter in  config/filter.d/pure-ftpd.conf contains the wording :
"c'est un batard, il connait pas son code"

This is probably not issued by a well configured FTP server and shall be removed from the filter as it is an insult

dricoco:
ok sorry for the noise, apparently pure-ftpd is shipping with a "funny" french translation that could be found in 
https://github.com/jedisct1/pure-ftpd/blob/master/src/messages_fr_funny.h

Closing this issue and PR

PR #2541, opened 2019-10-10
Update pure-ftpd.conf

dricoco:
The removed sentence " c'est un batard, il connait pas son code" is a french insult to the user and should be removed from that file. Could'nt be a serious message

this PR resolves : #2540


dricoco:
closing this PR, text shall be left here for pure-ftd compatibility

PR #2543, opened 2019-10-10
Add jail-filter functionality to query subset of all jails based on n…

Appleman1234:
…ame filter. Document specific jail functionality in usage

### DIFF
--- a/files/nagios/check_fail2ban
+++ b/files/nagios/check_fail2ban
@@ -109,15 +109,15 @@ my $fail2ban_client_path = '/usr/bin/fail2ban-client';
 my $fail2ban_socket = '';
 my $jail_specific = '';
 my $jail_name = '';
-my $jail_filter = '';
+my $use_regex = '';
 
 GetOptions (
     'P=s' => \ $fail2ban_client_path,
     'path-fail2ban_client=s' => \ $fail2ban_client_path,
     'j=s' => \ $jail_specific,
     'jail=s' => \ $jail_specific,
-    'f=s' => \ $jail_filter,
-    'jailfilter=s' => \ $jail_filter,
+    'r' => \ $use_regex,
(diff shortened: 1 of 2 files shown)


files/nagios/check_fail2ban
```diff
@@ -267,6 +274,10 @@ Options:
     default is unset
  -p, --perfdata
     If you want to activate the perfdata output
+ -j, --jail
+    Specific jail to get information for, if not specified uses all jails
+ -f, --jailfilter
+    Filter to match a subset of all jails based on their name, if not specified shows all jails
```
sebres: A bit confusing name for option (since fail2ban operates with `filter` too and `jailfilter` could mean filter of jail)...
Since jails are sections in fail2ban configs, the names with wildcard symbols (`*`, `?`) are not allowed, so why you need this new option, if one could simply use `-j` (or `--jail`) and enable a filtering if specified name contains wildcards?

Appleman1234: Happy to rename to a less confusing name. This isn't about wildcards, the use case was only wanting Nagios to monitor a subset of jails, but not wanting to make a separate check for each jail.
 
For the specific use case I was looking at I have 24 enabled jails but I don't want one check for all 24, or 12 different specific checks, I just wanted 1 check the only monitors a subset of the jails.

sebres: So it's RE-based and you use it quasi as list of jails, something like `jail1|jail2|...|jailN`?
Whether wildcards or RE, the parameters `-j` and `-f` are mutually exclusive, so may be recognition like specificjail contains `[^\w\-_]` or simply a switch like `-r` or something else, which could signal `-j` contains a RE-filter, not whole jail name?

Appleman1234: Changed to use the syntax you suggested.

PR #2550, opened 2019-10-21
Add Centreon jail

CPbN:
Hi,

Here is a new jail, for Centreon Web.
Based on authentication messages here : https://github.com/centreon/centreon/pull/7943

Thank you 👍

### DIFF
--- a/config/filter.d/centreon.conf
+++ b/config/filter.d/centreon.conf
@@ -1,0 +1,9 @@
+# Fail2Ban filter for Centreon Web
+# Detecting unauthorized access to the Centreon Web portal
+# typically logged in /var/log/centreon/login.log
+
+[Init]
+datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
+
+[Definition]
+failregex = ^\[0-9-]*\|[0-9-]*\|\[0-9-]*|\[[^]]*\] \[<HOST>\] Authentication failed for '.+' :
--- a/config/filter.d/centreon.conf
+++ b/config/filter.d/centreon.conf
@@ -6,4 +6,4 @@
 datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
 
 [Definition]
-failregex = ^\[0-9-]*\|[0-9-]*\|\[0-9-]*|\[[^]]*\] \[<HOST>\] Authentication failed for '.+' :
+failregex = ^(?:\|-?\d+){3}\|\[[^\]]*\] \[<HOST>\] Authentication failed for '<F-USER>[^']+</F-USER>'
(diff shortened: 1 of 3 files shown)


CPbN:
Sounds like tests passed, but coverage decreased, might not be this PR's fault 😕

sebres:
> Sounds like tests passed, but coverage decreased

It is not your issue... just timing or else event-related problems causing sporadic avoidance of some code pieces.

CPbN:
OK perfect, thank you @sebres.

config/filter.d/centreon.conf
```diff
@@ -0,0 +1,9 @@
+# Fail2Ban filter for Centreon Web
+# Detecting unauthorized access to the Centreon Web portal
+# typically logged in /var/log/centreon/login.log
+
+[Init]
+datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
+
+[Definition]
+failregex = ^\[0-9-]*\|[0-9-]*\|\[0-9-]*|\[[^]]*\] \[<HOST>\] Authentication failed for '.+' :
```
sebres: Firstly your anchored prefix `^\[0-9-]*...` does not work (because it searching open bracket `[`, because it's escaped `\[`)...
So the anchor does no matter and it's just the weird constellation (e. g. your forgotten escape by minus `-` char in sequence `[...]`, that means all chars starting with `9-`) what makes possible the RE matches by accident.
Also greedy catch-all `.+` at end is completely unneeded there (and one could use user name capturing RE).
So better to replace it with:
```diff
- failregex = ^\[0-9-]*\|[0-9-]*\|\[0-9-]*|\[[^]]*\] \[<HOST>\] Authentication failed for '.+' :
+ failregex = ^(?:\|-?\d+){3}\|\[[^\]]*\] \[<HOST>\] Authentication failed for '<F-USER>[^']+</F-USER>'
```
doing the same job, but it is anchored now.

config/jail.conf
```diff
@@ -820,6 +820,9 @@ udpport = 1200,27000,27001,27002,27003,27004,27005,27006,27007,27008,27009,27010
 action  = %(banaction)s[name=%(__name__)s-tcp, port="%(tcpport)s", protocol="tcp", chain="%(chain)s", actname=%(banaction)s-tcp]
            %(banaction)s[name=%(__name__)s-udp, port="%(udpport)s", protocol="udp", chain="%(chain)s", actname=%(banaction)s-udp]
 
+[centreon]
+logpath = /var/log/centreon/login.log
```
sebres: Does not the jail expected more here? At least default `port` would be nice.

config/filter.d/centreon.conf
CPbN: Thank you very much for your feedback 👍
I then just updated the rule according to your advices.

config/jail.conf
CPbN: You're right, I've added concerned ports.
Regarding ban times, I've left default values, as done with `apache-auth` for example.
Thank you 👍

CPbN:
Thank you very much @sebres 👍

Issue #2562, opened 2019-11-06
Sendmail filter regexes fail for PIDs > 999999


hensema:
In config/filter.d/sendmail-auth.conf the following line will fail:
```
failregex = ^%(__prefix_line)s\w{14}: (\S+ )?\[(?:IPv6:<IP6>|<IP4>)\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
```
It limits sendmail queue IDs to max 14 chars. However when sendmail as a pid > 999999 then the queue id will get longer and therefore the line will never match.

I recommend supporting at least 20 characters for queue ids.

hvanmegen:
Added a pull request for this issue: https://github.com/fail2ban/fail2ban/pull/2563

PR #2563, opened 2019-11-06
Updated sendmail-auth.conf for longer mail IDs

hvanmegen:
Updated the failregex in sendmail-auth.conf to support mail IDs up to 20 characters long as mentioned in https://github.com/fail2ban/fail2ban/issues/2562

### DIFF
--- a/config/filter.d/sendmail-auth.conf
+++ b/config/filter.d/sendmail-auth.conf
@@ -9,7 +9,8 @@ before = common.conf
 
 _daemon = (?:sendmail|sm-(?:mta|acceptingconnections))
 
-failregex = ^%(__prefix_line)s\w{14}: (\S+ )?\[(?:IPv6:<IP6>|<IP4>)\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
+# "w{14,20}" will give support for IDs from 14 up to 20 characters long
+failregex = ^%(__prefix_line)s\w{14,20}: (\S+ )?\[(?:IPv6:<IP6>|<IP4>)\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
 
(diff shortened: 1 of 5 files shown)


config/filter.d/sendmail-auth.conf
```diff
@@ -9,7 +9,8 @@ before = common.conf
 
 _daemon = (?:sendmail|sm-(?:mta|acceptingconnections))
 
-failregex = ^%(__prefix_line)s\w{14}: (\S+ )?\[(?:IPv6:<IP6>|<IP4>)\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
+# "w{20}" will give support for IDs up to 20 characters long
+failregex = ^%(__prefix_line)s\w{20}: (\S+ )?\[(?:IPv6:<IP6>|<IP4>)\]( \(may be forged\))?: possible SMTP attack: command=AUTH, count=\d+$
```
sebres: > up to 20 characters long...

Nope, it will match exactly 20 chars (so it will fail on all shorter IDs, and also in our test suite)... 
Something like `\w{14,20}` would be correct here.

sebres: Also note that test cases are must, so please provide an example of log-line this should cover, or better extend this PR with a new test in [fail2ban/tests/files/logs/sendmail-auth](https://github.com/hvanmegen/fail2ban/edit/patch-1/fail2ban/tests/files/logs/sendmail-auth?pr=%2Ffail2ban%2Ffail2ban%2Fpull%2F2563) with JSON and line containing such ID.

hvanmegen: I've added a longline in the tests as well

sebres:
Done. I'll merge this later if no objections follow.
Thx again.

PR #2565, opened 2019-11-10
Add Apprise Support (50+ Notifications)

caronc:

## Details
[Apprise](https://github.com/caronc/apprise) supports more then 50+ push notification services.  This just extends fail2ban so that it can notify you through more end-points when an action takes place.

It's based on [this ticket](https://github.com/caronc/apprise/issues/171) created not to long ago.

### Installation
```bash
# dependencies:
pip install apprise
```

**FILE:** `/etc/fail2ban/jail.conf`
```bash
#... look for the action_ definitions and add this one:
action_apprise = %(banaction)s[name=%(__name__)s, bantime="%(bantime)s", port="%(port)s", protocol="%(protocol)s", chain="%(chain)s"]
                 apprise[name=%(__name__)s]

# Scroll down a little bit farther so you can find the `action` default (comment it out)
#action = %(action_)s

# and add a new default instead:
action = %(action_apprise)s
```

An apprise configuration file allows you to identify 1 or more services you want to notify.  They're all [well documented here](https://github.com/caronc/apprise/wiki). An example of one might look like:
```bash
# Default config location: /etc/fail2ban/apprise.cfg

# Place your Apprise URLs here you want to be notified:
slack://TokenA/TokenB/TokenC

# There is no limit...
mailto://user:pass@gmail.com

kodi://my.kodi.server
#...
```

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -3,6 +3,7 @@ bin/fail2ban-regex
 bin/fail2ban-server
 bin/fail2ban-testcases
 ChangeLog
+config/action.d/apprise.conf
 config/action.d/abuseipdb.conf
 config/action.d/apf.conf
 config/action.d/badips.conf
--- a/config/action.d/apprise.conf
+++ b/config/action.d/apprise.conf
@@ -10,13 +10,13 @@
 # Notes.:  command executed once at the start of Fail2Ban.
 # Values:  CMD
 #
-actionstart = printf %%b "The jail <name> as been started successfully." |apprise -t "[Fail2Ban] <name>: started on `uname -n`" -c /etc/fail2ban/apprise.conf
+actionstart = printf %%b "The jail <name> as been started successfully." | <apprise> -t "[Fail2Ban] <name>: started on `uname -n`"
 
 # Option:  actionstop
 # Notes.:  command executed once at the end of Fail2Ban
 # Values:  CMD
 #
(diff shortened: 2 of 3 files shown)


caronc:
For some reason when I do a commit you lose test coverage on a file I didn't even touch. :man_shrugging:

sebres:
> when I do a commit you lose test coverage on a file I didn't even touch

This is not a problem, rather a timing issue (some code due to async handling did not touch) or it is some issue of coverage tool (as often).

GeorgeGedox:
Any word on this? I'm looking forward to see it integrated

sebres:
Looks good to me. Excepting possibly that I don't like the many enhancements parameters `action_*` in jail.conf.
Sure we have already many of that, but it does not necessary say that we should do that further.
Just because every new parameter is impeding configuration interpolation and it is really enough to write following config in jail.local to activate this action:
```ini
action = %(action_)s
         apprise
```
I don't think it is more complicated as:
```ini
action = %(action_apprise)s
```

Also I noticed that parameter `config` is not used (it is hard-coded path) in action at the moment.

I'll try to amend that here.

BTW, you can use it right now without to wait for release: create action file (`/etc/fail2ban/action.d/apprise.conf `) and config file and specify action as aforementioned in jail or default section.

caronc:
Thanks for the constructive reply! I really appreciate it! Your code changes and comments helped explain a few things i didn't quite understand at the time. :+1:

PR #2567, opened 2019-11-25
New jail, Bitwarden

Mart124:
Hi,

Here's a new jail, for Bitwarden.
Failed login attempts logging has just been [implemented](https://github.com/bitwarden/server/issues/604).

Thank you !

### DIFF
--- a/config/filter.d/bitwarden.conf
+++ b/config/filter.d/bitwarden.conf
@@ -1,0 +1,9 @@
+# Fail2Ban filter for Bitwarden
+# Detecting failed login attempts
+# Logged in bwdata/logs/identity/Identity/log.txt
+
+[Init]
+datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
+
+[Definition]
+failregex = ^.*Failed login attempt(, 2FA invalid)\. <HOST>$
--- a/config/filter.d/bitwarden.conf
+++ b/config/filter.d/bitwarden.conf
@@ -2,8 +2,5 @@
 # Detecting failed login attempts
 # Logged in bwdata/logs/identity/Identity/log.txt
 
-[Init]
-datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
-
 [Definition]
-failregex = ^.*Failed login attempt(, 2FA invalid)?\. <HOST>$
+failregex = ^\s*\[\w+\]\s+Failed login attempt(?:, 2FA invalid)?\. <HOST>$
--- a/config/filter.d/bitwarden.conf
+++ b/config/filter.d/bitwarden.conf
@@ -6,4 +6,4 @@
 datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
 
 [Definition]
-failregex = ^.*Failed login attempt(, 2FA invalid)\. <HOST>$
+failregex = ^.*Failed login attempt(, 2FA invalid)?\. <HOST>$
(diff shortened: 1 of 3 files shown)


config/filter.d/bitwarden.conf
```diff
@@ -0,0 +1,9 @@
+# Fail2Ban filter for Bitwarden
+# Detecting failed login attempts
+# Logged in bwdata/logs/identity/Identity/log.txt
+
+[Init]
+datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S
```
sebres: I don't think it is valid pattern, since the log-lines seem to have timezones.
Also I don't see it is expected at all, since our default patterns (in 0.10) completely cover this dates.
So please remove it.

sebres: Also note that verbose output of fail2ban-regex in normal case (considering correct TZ) would return following times for your test-lines from `tests/files/logs/bitwarden`:
```
Failregex: 2 total
|   1) [2] ^\s*\[WRN\]\s+Failed login attempt(, 2FA invalid)?\. <HOST>$
|      192.168.0.16  Mon Nov 25 18:04:49 2019
|      192.168.0.21  Mon Nov 25 21:39:58 2019
`-
```

config/filter.d/bitwarden.conf
```diff
@@ -0,0 +1,9 @@
...
+
+[Definition]
+failregex = ^.*Failed login attempt(, 2FA invalid)?\. <HOST>$
```
sebres: Regex `^.*` is not really an anchor (it works same way as it were not there). And after cut out of string matched default datepattern it has pretty simple prefix, so better would be to use something like this:
```diff
- failregex = ^.*Failed login attempt(, 2FA invalid)?\. <HOST>$
+ failregex = ^\s*\[WRN\]\s+Failed login attempt(?:, 2FA invalid)?\. <HOST>$
```
Also note it uses non capturing group `(?: ...)`.

Mart124: Perfect, many thanks for your review and feedback.
In the updated the rule accordingly, with just a little modification :
`failregex = ^\s*\[\w+\]\s+Failed login attempt(?:, 2FA invalid)?\. <HOST>$`
As I'm not sure `WRN` will remain, but for sure we'll have a word there.

config/filter.d/bitwarden.conf
Mart124: Thank you, let's then use the default date/time pattern.
I also updated the test-lines according to their TZ.

config/filter.d/bitwarden.conf
sebres: > As I'm not sure `WRN` will remain

Hmm... why it should not? The same would be valid if `[WRN]` get rewritten without brackets, or completely removed together with brackets. Or if IP will be moved to begin of log.
You cannot absolutely protect the filter against any log-format modifications that devs could do at some point...
Anyway having WRT could speedup it a bit (for example if some other messages logged also, e. g. with something like `[INFO]` or so).
I'm for WRN... But you could define it as parameter:
```ini
[Definition]
failloglev = WRN
failregex = ^\s*\[%(failloglev)s\]\s+Failed login attempt(?:, 2FA invalid)?\. <HOST>$
```
so it'd be possible to overwrite it in jail.local:
```ini
[bitwarden]
filter = bitwarden[failloglev='\w+']
```

Mart124: Let's then get back to `WRN`, the easiest solution, I'm OK with this :)
We'll update the jail whenever log format changes.
Thank you again 👍

Mart124:
@sebres let's merge this PR if it's OK for you ?
Anyway, thank you very much for having reviewed it!

Mart124:
Great thank you again @sebres.

PR #2578, opened 2019-12-23
Fix actions in ufw.conf

xdch47:
ban bug/security issue using ufw:
--> `-A ufw-before-forward -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT` 
is *before* the ufw users rules. Therefore blocking only takes place *after* a disconnect. (this is a problem for web pages, e.g. nginx-http-auth).

Commit Summary:

- ufw does "quickly process packets for which we already have a connection" in before.rules,
  therefore all IPv4/IPv6 sockets should be closed, otherwise the action
  inserted by fail2ban will be by passed. This is done by the `ss` utility
  from the iproute2 package.

- ufw application can contain spaces (e.g. WWW Full)

- add ufw comment to easily identify rules added by fail2ban

### DIFF
--- a/config/action.d/ufw.conf
+++ b/config/action.d/ufw.conf
@@ -1,6 +1,8 @@
 # Fail2Ban action configuration file for ufw
 #
 # You are required to run "ufw enable" before this will have any effect.
+# Please ensure `ss` program from the iproute2 package is installed
+# to close already established connections.
 #
 # The insert position should be appropriate to block the required traffic.
 # A number after an allow rule to the application won't be of much use.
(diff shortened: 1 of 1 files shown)


Klaas-:
I tested it with the default sshd jail on debian buster, lgtm. Without this patch (pre-)existing connections remain open even when sourceip is banned

sebres:
Just for the record - see https://github.com/fail2ban/fail2ban/pull/2607#issuecomment-575942656 for similar issue.
Drop of all connections regardless its port may be a bit "dangerous" (at least for multi-port actions), so possibly it should be extended with ports like in example from comment above (or it should be optional, so user can decide about it).

rlaager:
For UDP, there are no sockets held open, so this needs: `conntrack -D -s <ip>`

sebres:
superseded by #3018

xdch47:
did you like my PR or what was the point here - the look quite identical ?

sebres:
Neither this nor that...
Some PRs can gets crossed in parts, and regarding "superseded", #3018 looks indeed similar, but provides kill facility as an option, and has another stuff (like `prepend`) solving other option.
I don't hold any code of every request in my head, so it can happen... But it is indeed strange that both look really equal here and there... so it looks like #3018 would be continuation of your pull request, therefore I'll merge your commits in order to point to you as an originator of that.

xdch47:
No worries - everything fine. It was just if I did sth. wrong here - and yes  #3018 is an improvement (maybe because it got also review input too), so I happy if this one get merged.

xdch47:
by the way, for some reason (i think to get completion also for older version and maybe because this PR has been on hold for a long time) -  I decided to contribute the zsh-completion to  [zsh-user](https://github.com/zsh-users/zsh-completions/blob/master/src/_fail2ban-client)-repo - not sure if you want to pull this into the fail2ban project as well…

PR #2591, opened 2020-01-10
Enhancements: Update ipfw.conf to better control insertion ob blocking rules.

TheHolm:
Allows to specify where to inset blocking rules to ipfw.  
Does not change default behaviour.


TheHolm:
The Travis CI build fails due to some other problems  not related to this patch.

PR #2057, opened 2018-02-26
Use httpS with badips

benrubson:
Hi,

This PR use https when requesting badips.com.

It then closes #2051.

Thx 👍 

Ben

PR #2601, opened 2020-01-13
Move to https for badips

sylvestre:
Reported https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=800708

### DIFF
--- a/config/action.d/badips.conf
+++ b/config/action.d/badips.conf
@@ -6,14 +6,14 @@
 # Set the category to the appropriate value before use.
 #
 # To get see register and optional key to get personalised graphs see:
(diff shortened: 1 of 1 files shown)


benrubson:
Note that action script was moved from HTTP to HTTPS thanks to https://github.com/fail2ban/fail2ban/pull/2057.
However Debian 10 does not provide the last version of `badips.py`, at least not the one containing this PR.

sebres:
> Note that action script was moved from HTTP to HTTPS thanks to #2057.

There are 2 badips actions in fail2ban (pythonic and not), this PR targeting not pythonic action.

sebres:
badips seems to be not available anymore, see #2889, this closed

PR #2603, opened 2020-01-15
Make badips.py more robust

### LINKED ISSUE #2600: Badips can hang at startup
Hi,

```
# cat /etc/centos-release
CentOS Linux release 7.7.1908 (Core)

# yum list installed | grep fail2ban
fail2ban-server.noarch            0.10.4-1.el7           @epel

# fail2ban-server --version
Fail2Ban v0.10.4
```

Sometimes, badips times-out, which leads to the following into the logfile :
`2020-01-13 19:01:35,422 fail2ban.BadIPsAction   [21220]: ERROR   Failed to fetch bad IP list. badips.com response: 'Unknown'`

Sometimes, timeout occurs at the very first time, when fail2ban starts, leading to this :
```
2020-01-13 20:30:47,476 fail2ban.transmitter    [19560]: WARNING Command ['set', 'badips', 'addaction', 'badips', '/etc/fail2ban/action.d/badips.py', '{"category": "unknown", "bancategory": "any", "banaction": "iptables-allports", "score": "0", "age": "24h", "updateperiod": "60", "loglevel": "INFO,DEBUG", "bankey": "xxxxxxxxxxxx"}'] has failed. Received timeout('The read operation timed out')
2020-01-13 20:30:47,477 fail2ban                [19560]: ERROR   NOK: ('The read operation timed out',)
```

However, unfortunately, it will not retry, as expected in my example above, 60 seconds later.
The only solution is then to restart fail2ban.
Perhaps there's a better workaround ?

Thank you again 👍

### LINKED ISSUE #2602: Badips tries to unban when stopping
Hi,

Some firewalls, such as iptables, bsd-ipfw... have a stop action to flush chains when Fail2Ban stops.
However then, Badips tries to unban every banned IP, leading to the following log :

```
2020-01-15 19:19:25,578 fail2ban.actions        [18862]: NOTICE  [badips] Flush ticket(s) with iptables-allports
2020-01-15 19:19:25,677 fail2ban.utils          [18862]: #39-Lev. 7f02a1d32298 -- exec: iptables -w -D f2b-badips -s A.B.C.D -j DROP
2020-01-15 19:19:25,678 fail2ban.utils          [18862]: ERROR   7f02a1d32298 -- stderr: 'iptables: Bad rule (does a matching rule exist in that chain?).'
2020-01-15 19:19:25,678 fail2ban.utils          [18862]: ERROR   7f02a1d32298 -- returned 1
(...)
```

I assume this is due to the following `badips.py` code :
```
        def stop(self):
                """If `banaction` set, clears blacklisted IPs.
                """
                if self.banaction is not None:
                        if self._timer:
                                self._timer.cancel()
                                self._timer = None
                        self._unbanIPs(self._bannedips.copy())
```

But then I'm not sure how to keep it compatible with both firewalls having and not having stop action.

Thank you 👍

benrubson:
Hi,

This PR makes `badips.py` more robust when badips API does not answer.
It then closes #2600.
We are now then sure the script will correctly run, even if for example badips API is not reachable when fail2ban starts.

Ban categories are retrieved only one time now, instead of 4 times initially... speeding-up the startup time (and also making it more reliable).

It also adds a new option, `unbanonstop`, which allow to enable/disable unban of every IP individually.
It then closes #2602.

Thank you 👍

### DIFF
--- a/fail2ban/tests/action_d/test_badips.py
+++ b/fail2ban/tests/action_d/test_badips.py
@@ -153,5 +153,5 @@ if sys.version_info >= (2,7): # pragma: no cover - may be unavailable
 			# produce an error using wrong category/IP:
 			self.action._category = 'f2b-this-category-dont-available-test-suite-only'
 			aInfo['ip'] = ''
-			self.assertRaises(BadIPsActionTest.pythonModule.HTTPError, self.action.ban, aInfo)
-			self.assertLogged('IP is invalid', 'invalid category', wait=True, all=False)
+			self.action.ban(aInfo)
+			self.assertLogged('IP is invalid', 'Failed to ban', wait=True, all=False)
--- a/ChangeLog
+++ b/ChangeLog
@@ -46,6 +46,8 @@ ver. 0.10.6-dev (20??/??/??) - development edition
   so would bother the action interpolation
 * `action.d/*-ipset*.conf`: several ipset actions fixed (no timeout per default anymore), so no discrepancy
   between ipset and fail2ban (removal from ipset will be managed by fail2ban only, gh-2703)
+* `action.d/badips.py`: script more robust when badips API does not answer (gh-2603)
+* `action.d/badips.py`: new `unbanstop` option to toggle indivual unban when fail2ban stops (gh-2603)
 * `action.d/cloudflare.conf`: fixed `actionunban` (considering new-line chars and optionally real json-parsing
    with `jq`, gh-2140, gh-2656)
 * `filter.d/common.conf`: avoid substitute of default values in related `lt_*` section, `__prefix_line`
--- a/config/action.d/badips.py
+++ b/config/action.d/badips.py
@@ -32,7 +32,7 @@ else: # pragma: 3.x no cover
 	from urllib import urlencode
 
 from fail2ban.server.actions import Actions, ActionBase, BanTicket
-from fail2ban.helpers import splitwords, str2LogLevel
+from fail2ban.helpers import splitwords, str2LogLevel, _as_bool
 
 
 
@@ -69,6 +69,8 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 	updateperiod : int, optional
 		Time in seconds between updating bad IPs blacklist.
 		Default 900 (15 minutes)
+	unbanonstop : bool; optional
+		Unban every IP individually when stopping.
 	loglevel : int/str, optional
 		Log level of the message when an IP is (un)banned.
 		Default `DEBUG`.
@@ -87,12 +89,13 @@ class BadIPsAction(ActionBase): # pragma: no cover - may be unavailable
 
 	TIMEOUT = 10
 	_badips = "https://www.badips.com"
+	_categories = None
 	def _Request(self, url, **argv):
 		return Request(url, headers={'User-Agent': self.agent}, **argv)
 
 	def __init__(self, jail, name, category, score=3, age="24h",
 		banaction=None, bancategory=None, bankey=None, updateperiod=900, 
-		loglevel='DEBUG', agent="Fail2Ban", timeout=TIMEOUT):
+		loglevel='DEBUG', agent="Fail2Ban", unbanonstop=True, timeout=TIMEOUT):
 		super(BadIPsAction, self).__init__(jail, name)
 
 		self.timeout = timeout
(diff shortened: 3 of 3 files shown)


benrubson:
Mmmmh sounds like the test file might need to be updated accordingly.
I'm not sure what's the best way though 😑

sebres:
Well it is indeed part of coverage, this
https://github.com/fail2ban/fail2ban/blob/5bc8d73220831fdbb9abad4e427850c02ea31bc2/fail2ban/tests/action_d/test_badips.py#L156
expecting an error in action at the moment (due to wrong category).
I have no time right now for review, but if the intention is not to throw an error anymore, then the test should be either removed or rewritten with `self.assertEqual` with expected return code.

fail2ban/tests/action_d/test_badips.py
```diff
@@ -153,5 +153,4 @@ def testBanIP(self):
 			# produce an error using wrong category/IP:
 			self.action._category = 'f2b-this-category-dont-available-test-suite-only'
 			aInfo['ip'] = ''
-			self.assertRaises(BadIPsActionTest.pythonModule.HTTPError, self.action.ban, aInfo)
```
sebres: well, this whole removal is wrong (you want surely still call ban, and this assert is a wrapper catching error in callable;), rather you would replace it with something like:
```diff
-self.assertRaises(BadIPsActionTest.pythonModule.HTTPError, self.action.ban, aInfo)
+self.action.ban(aInfo)
```

benrubson: Yes you're right, thank you @sebres for your review and explanations, let's then test this :)

benrubson: Worked nicely, many thanks Sergey 👍

benrubson:
Could we think about merging this please Sergey when you'll have some time to dig into it ?
Thank you very much 👍

benrubson:
I just squashed the various commits and pushed a unique one.

sebres:
One thing by the way: now the action doesn't throw an error (e. g. by get categories) if it is busy (timeout), so our small mock-up wrapper in fail2ban-testcases (which recognizes TO/net failures and skip the tests by network error) could not detect it now, so some tests can fail do to wrong/unexpected data in such cases.

I don't know it is affected but this mock-ups where made because we've basically no interest every time to see Travis CI errors if something got changed, and unrelated badIPs tests failing again with timeout. BadIPs is very unreliable service and we don't have proper mockup's for it.

Also with this "hack" of me (skip on network-error) one sees sporadic errors in CI occasionally, so actually I'm really in favor to get forking off this (and some other) action (or functionality like cymru) to an external repository or separate package.

benrubson:
Thank you for your comment @sebres.

I let `isAvailable` function as you initially implemented it, so should still be OK, BadIPs tests will be skipped if badips.com is not reachable.

I agree that BadIPs is not the most reliable service I've worked with :)

benrubson:
Of course I would be glad if we could consider merging this version like this, and why not, in the future, like you said above, think about another repository / separate package.

Perhaps we could also enable BadIPs CI tests only if the concerned PR modifies BadIPs action ?
That said, perhaps that other modifications made elsewhere could impact BadIPs action.
But then certainly it would impact other actions...

benrubson:
_Travis-CI is OK, coverage decreases by a little as we can see from time to time, so I assume it's OK._

benrubson:
Categories were retrieved 4 times per jail...
My last commit makes then `badips.py` to retrieve categories only one time, speeding-up the startup time 👍

benrubson:
_Mmmh sounds like the Travis error is out of my responsibility, seems related to CI itself._

sebres:
> Mmmh sounds like the Travis error is out of my responsibility, seems related to CI itself.

Yes... restarted the job.

benrubson:
Let's perhaps re-restart the test Sergey ? (as https://github.com/fail2ban/fail2ban/commit/ac8e8db8141a9945c7f435ae042c198e23a6f945 may help)
Thx !

sebres:
in this case it should be rebases (or merged with) - this commit is out of your branch.
(but it is not really necessary as long as we know what happens with 3.8th, so let is as it is).

benrubson:
Mmmh yes you're right... I must be tired !
Let's then let it like this yes 👍

benrubson:
For info, no issue with this script so far on all my servers 👍

benrubson:
_Closing this for now, feel free to re-open if needed._

PR #2287, opened 2018-11-25
setup.py: adding option to ignore tests

angeloc:
Tests files are not always needed especially when installing on low
resource systems like an embedded one.
This patch adds the `--without-tests` option to skip building and
installing of tests files.

PR #2605, opened 2020-01-16
Fixing --withouth-test install option

### LINKED ISSUE #2604: Cannot install testcases with 0.11.1
### Environment:

_- Fail2Ban version (including any possible distribution suffixes): 0.11.1

### The issue:

It appears to be impossible to install the testcases anymore.  There is a --without-testcases option, but it appears to be active by default and there is no --with-testcases option.

#### Steps to reproduce

python3 setup.py install

#### Expected behavior

fail2ban-testcases and tests are installed.

#### Observed behavior
fail2ban-testcases and tests are not install.

Also, the fail2ban-testcases.1 man page does not appear to be handled by the switch.  It is still currently installed, and setting --without-testcases I do not think would change that.

angeloc:
This PR use a different approach to skip installing tests:
* it keeps the setup section intact, so the build phase still work as expected
* it removes the tests files in finalize_options step, before actually installing the files

This way, the test will work as usual with all the other step, but they can be skipped using --without-tests option

closes #2604
(replacement for #2287)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -119,11 +119,9 @@ class install_scripts_f2b(install_scripts):
 class install_command_f2b(install):
 	user_options = install.user_options + [
 		('disable-2to3', None, 'Specify to deactivate 2to3, e.g. if the install runs from fail2ban test-cases.'),
-		('without-tests', None, 'without tests files installation'),
 	]
 	def initialize_options(self):
 		self.disable_2to3 = None
-		self.without_tests = None
 		install.initialize_options(self)
 	def finalize_options(self):
 		global _2to3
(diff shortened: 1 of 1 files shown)


sebres:
Thanks again!
Looks good and build contains tests, whereas `install --dry-run` attempts with and without option `--without-tests` show only tests in the diff of file list:
```diff
  ...
- creating build/lib.linux-x86_64-2.7/fail2ban/tests
- copying fail2ban/tests/filtertestcase.py -> build/lib.linux-x86_64-2.7/fail2ban/tests
- ...
  running build_scripts
  creating build/scripts-2.7
  ...
- copying and adjusting bin/fail2ban-testcases -> build/scripts-2.7
  ...
- changing mode of build/scripts-2.7/fail2ban-testcases from 644 to 755
```
Thus merge,

sebres:
1cefb7fdc6f160c300bfb7fe329b4402703cb2a7 extends now this for build process also (if the tests not expected in package too)

PR #2607, opened 2020-01-17
Kill state after adding IP to PF table

janprzy:
When fail2ban adds an IP to a table of the PF firewall, the state has to be killed afterwards. It this does not happen, the connection can stay open, allowing the attacker to continue attacking.
This might be unnecessary when using fail2ban to protect ssh, since the connection always resets after a failed login attempt. However, I'm using fail2ban to protect an instance of nginx against bots that cause 403 and 404 errors. During my testing, I quickly realised that without this edit, the ban would not immediately take effect.

tl;dr: Sometimes bans didn't take effect immediately because the state wasn't killed.

I got the idea from this post on the FreeBSD forums:
https://forums.freebsd.org/threads/pf-and-fail2ban.51624/

### DIFF
--- a/config/action.d/pf.conf
+++ b/config/action.d/pf.conf
@@ -66,7 +66,7 @@ actioncheck = <pfctl> -sr | grep -q <tablename>-<name>
 #          <time>  unix timestamp of the ban time
 # Values:  CMD
 #
-actionban = <pfctl> -t <tablename>-<name> -T add <ip>
+actionban = <pfctl> -t <tablename>-<name> -T add <ip> && <pfctl> -k <ip>
 
 
 # Option:  actionunban


sebres:
We get similar PRs for some actions every half year :)

The thing is that pf-action can be used also to ban dedicated ports only (not allports case), see https://github.com/fail2ban/fail2ban/blob/d01fe9d22a717a0122143cde41ce5c18a1790f6e/config/action.d/pf.conf#L30
This enhancement can cause a drop of all connections from IP regardless the port, so if you connected to ssh and accidentally due to some false positive (or during a intended test) got banned on http/https, this would reject all your connections inclusive ssh.
But also in case of allports, it may be too "heavy" to drop all established connections (if unexpected or unconscious) - you could be unable to unban yourself (have to wait or to switch an IP).

This does not mean I'm against this, just perhaps it should be extended with some parameter allowing that.
Additionally note that some services would properly reject the established connection after several invalid logins (so you wouldn't need this at least for this jails).
As well as on many systems it is a matter of network configuration (don't know about PF), so it is configurable whether netfilter rule also affects established connections.

But also without this PR it is possible to configure PF (or other actions) to use that trick:
```ini
[jail]
banaction = pf[actionban='<known/actionban> && <pfctl> -k <ip>']
```
There are also another possibilities, like described [in this answer on serverfault](https://serverfault.com/a/998163/488604).
For example this can drop the connections of IP to certain ports only:
```bash
ss -o state established -K dst <ip> 'sport = 80 or sport = 443'
```

janprzy:
> This enhancement can cause a drop of all connections from IP regardless the port, so if you connected to ssh and accidentally due to some false positive (or during a intended test) got banned on http/https, this would reject all your connections inclusive ssh.

You're right, I hadn't thought of this. In that case, I will close this Pull request for now.

Issue #1290, opened 2016-01-01
upcoming firewalld 0.4.0 gains ipset support


aboe76:
The upcoming 0.4.0 version of firewalld will gain ipset support which in turn can benefit fail2ban:
see: http://www.firewalld.org/2015/12/ipset-support/

new commands that can be used for creation:
firewall-cmd --permanent --new-ipset=fail2ban-<name> --type=hash:ip
firewall-cmd --add-rich-rule='rule source ipset=fail2ban-<name> drop'
firewall-cmd --reload

new commands that can be actionban:
firewalll-cmd --ipset=fail2ban-<name> --add-entry=<ip>

Only downside so far is that they have a limitation on timeout option.

please take look and see if this can be added next to the other firewalld actions.

t-woerner:
You can also use the ipset with a timeout:
firewall-cmd --permanent --new-ipset=fail2ban- --type=hash:ip --option=timeout=300
firewall-cmd --permanent --add-rich-rule='rule source ipset=fail2ban- drop'
firewall-cmd --reload

You can then use ipset directly to add entries to the ipset like you did before.

The command "firewalll-cmd --ipset=fail2ban- --add-entry=....." is not usable on ipsets with timeouts, as entries will time out in the ipset itself and I did not want to add an extra timer for the entries to firewalld just to be able to remove the entries in firewalld also to give a correct ipset listing and feedback.

aboe76:
@t-woerner thanks for the clarification.

Issue #1474, opened 2016-07-01
FirewallD Ban Actions


Geraden07:
Apologies for not adhering to the bug report format but I'm not creating this issue for a bug.

This is half a question, half a suggestion. 

All the supplied fail2ban action files for use with firewalld use 'direct' rule configuration to essentially ban IPs in the same way that IP tables would.

However, from their own man page ( https://fedoraproject.org/wiki/FirewallD ) the authors of FirewallD state that ```Direct options should be used only as a last resort when it's not possible to use for example --add-service=service or --add-rich-rule='rule'.```

Considering this, the more appropriate (and significantly simpler) manner in which to ban an IP address in FirewallD is by adding the IP (referred to as the "source") to an appropriate zone, such as "drop" or "block".

The command to do this would be ```firewall-cmd --permanent --zone=drop --add-source=<ip>```. The benefit of this (apart from the simplicity) would be that firewalld would then also be able to display the banned IPs associated with the drop (or block) zone by using the command ```firewall-cmd --info-zone=drop``` or even ```firewall-cmd --list-all-zones```.

No setup or teardown actions would be necessary. Essentially this approach delegates the nitty-gritty details of rule writing to firewalld. Instead of telling firewalld HOW you want to ban an IP, you instead tell firewalld THAT you want to ban an IP and let firewalld deal with the HOW.

If the 'block' zone were to be used, the response to banned IPs would be an icmp-host-prohibited message for IPv4 and icmp6-adm-prohibited for IPv6. If the 'drop' zone were used, the packets would be dropped with no response.

I personally am in favor of using the drop behavior, but it's easy enough that both action's could be and should be supplied as options.

Am I missing something as to why this approach is not preferable? Is there a reason that ```--direct``` rules are necessary and my suggestion is unsuitable? Or is the current behavior simply a result of how iptables rules were ported over to firewalld for compatibility at first release? In which case I strongly suggest this at the very least as an option that should be included with the fail2ban package-provided action files.

sebres:
Thx.
We will continue reorganizing of firewallcmd-actions (#1447, etc.) for our new 0.10 version, so we can see, how this can be optional implemented there.

Geraden07:
So I had some time to play around with this. I believe this works as desired:

```
cat << EOF > /etc/fail2ban/action.d/custom-firewalld.conf
[INCLUDES]
before  =

[Definition]
actionstart =
actionstop =
actioncheck =

[...]

PR #2620, opened 2020-01-30
Using native firewalld ipset implementation

### LINKED ISSUE #2614: Action firewallcmd-ipset using native firewalld ipset implementation
@grooverdan @sebres Created this issue just for curious to know why actions in [firewallcmd-ipset.conf](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/firewallcmd-ipset.conf) fail2ban config do not use native firewalld ipset implementation? As described here for example.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/sec-setting_and_controlling_ip_sets_using_firewalld

Also with warning in the end: 

> Red Hat does not recommend using IP sets that are not managed through firewalld

mspolitaev:
By creating additional action file firewallcmd-ipset-native.conf

Closes #2614, #1290

### DIFF
--- a/config/action.d/firewallcmd-ipset-native.conf
+++ b/config/action.d/firewallcmd-ipset-native.conf
@@ -1,0 +1,77 @@
+# Fail2Ban action file for firewall-cmd using native ipset implementation
+#
+# This requires:
+# ipset (package: ipset)
+# firewall-cmd (package: firewalld)
+#
+# This is for ipset protocol 6 (and hopefully later) (ipset v6.14).
+# Use ipset -V to see the protocol and version.
+#
+# IPset was a feature introduced in the linux kernel 2.6.39 and 3.0.0 kernels.
+#
+# If you are running on an older kernel you make need to patch in external
+# modules.
+
+[INCLUDES]
+
+before = firewallcmd-common.conf
+
+[Definition]
+
+actionstart = firewall-cmd --permanent --new-ipset=<ipmset> --type=hash:ip --option=timeout=<bantime>
+              firewall-cmd --reload
+              firewall-cmd --direct --add-rule <family> filter <chain> 0 <actiontype> -m set --match-set <ipmset> src -j <blocktype>
+
+actionstop = firewall-cmd --direct --remove-rule <family> filter <chain> 0 <actiontype> -m set --match-set <ipmset> src -j <blocktype>
+             firewall-cmd --permanent --delete-ipset=<ipmset>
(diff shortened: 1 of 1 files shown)


sebres:
Thanks!
Merged now, but I'll still try to combine both ipset actions using some parameter like `ipset-type` (set to `native` per default).

sebres:
Hmm... Now I have several questions with this action (after an attempt to combine both actions):
- it uses permanent ipset - persistence issue was touched in #1474 inter alia and I'm unsure it must be targeted here (be part of this PR);
- firewall-cmd seems not to have any sub-command to flush its native created ipset, so it is impossible to implement `actionflush`;
- the action does not support IPv6 (set would miss `--option=family=inet6` by its creation);
- the control of timeout in ipset is not so good (we had several issues already, e. g. #2670, thus now it is under fail2ban control solely)

I'm still trying to "fix" all that in combined action, but the issue with missing flush functionality makes it hardly possible at the moment.

sebres:
Done in c7a86b4616ef1bd43cf96874710b53b42d065684:
- combines actions `firewallcmd-ipset` and `firewallcmd-ipset-native` (parameter `ipsettype=firewalld`), for example:
`banaction = firewallcmd-ipset[ipsettype=firewalld]`
- IPv6-capability for firewalld ipset;
- no internal timeout handling by default;
- no permanent rules yet;

The issue with `actionflush` is still there, therefore I wouldn't recommend to use this in production (let alone I don't think usage of firewalld actions provides any benefit compared with actions using native net-filters like `iptables/ipset` or `nftables`).

mspolitaev:
@sebres glad to help, i am out of context of the problem unfortunately. Hope it's not broke anything, only improve)

sebres:
**Important:** for people landing here to use `firewalld` ipset implementation:
In `action.d/firewallcmd-ipset.conf` old parameter `ipsettype` initially added here for an ipset-backend, renamed since 17daf0ec7887a3c436093b56c8720645ef4ccf62 to `ipsetbackend`, because it is really backend and not a type and because new parameter `ipsettype` will be used for the real set type (see #3760).
So for example one would use this now:
```ini
banaction = firewallcmd-ipset[ipsetbackend=firewalld]
```

*I know it is not quite compatible change, but... (only if someone already used `ipsettype` previously)... 
Anyway, otherwise it'd be really confusing, so a rename was better choice in my opinion.*

PR #2631, opened 2020-02-10
Update Guacamole jail

benrubson:
Hi,

Here's a PR to update Guacamole jail.
It has been done based on their [documentation](https://guacamole.apache.org/doc/gug/configuring-guacamole.html#webapp-logging), and of course has been tested.

Thank you 👍

### DIFF
--- a/fail2ban/tests/files/logs/guacamole
+++ b/fail2ban/tests/files/logs/guacamole
@@ -10,3 +10,8 @@ WARNING: Authentication attempt from 192.0.2.0 for user "null" failed.
 apr 16, 2013 8:32:28 AM org.slf4j.impl.JCLLoggerAdapter warn
 # failJSON: { "time": "2013-04-16T08:32:28", "match": true , "host": "192.0.2.0" }
 WARNING: Authentication attempt from 192.0.2.0 for user "pippo" failed.
+
+# filterOptions: {"logging": "webapp"}
+
+# failJSON: { "time": "2005-08-13T12:57:32", "match": true , "host": "182.23.72.36" }
+12:57:32.907 [http-nio-8080-exec-10] WARN  o.a.g.r.auth.AuthenticationService - Authentication attempt from 182.23.72.36 for user "guacadmin" failed.
--- a/ChangeLog
+++ b/ChangeLog
@@ -44,6 +44,8 @@ ver. 0.10.6-dev (20??/??/??) - development edition
 * ensure we've unique action name per jail (also if parameter `actname` is not set but name deviates from standard name, gh-2686)
 * don't use `%(banaction)s` interpolation because it can be complex value (containing `[...]` and/or quotes), 
   so would bother the action interpolation
+* fixed type conversion in config readers (take place after all interpolations get ready), that allows to 
+  specify typed parameters variable (as substitutions) as well as to supply it in other sections or as init parameters.
 * `action.d/*-ipset*.conf`: several ipset actions fixed (no timeout per default anymore), so no discrepancy
   between ipset and fail2ban (removal from ipset will be managed by fail2ban only, gh-2703)
 * `action.d/cloudflare.conf`: fixed `actionunban` (considering new-line chars and optionally real json-parsing
@@ -61,6 +63,7 @@ ver. 0.10.6-dev (20??/??/??) - development edition
 
 ### New Features and Enhancements
 * new filter and jail for GitLab recognizing failed application logins (gh-2689)
+* `filter.d/guacamole.conf` extended with `logging` parameter to follow webapp-logging if it's configured (gh-2631)
 * introduced new prefix `{UNB}` for `datepattern` to disable word boundaries in regex;
 * datetemplate: improved anchor detection for capturing groups `(^...)`;
 * datepattern: improved handling with wrong recognized timestamps (timezones, no datepattern, etc)
--- a/config/filter.d/guacamole.conf
+++ b/config/filter.d/guacamole.conf
@@ -5,21 +5,47 @@
 
 [Definition]
 
-# Option: failregex
-# Notes.: regex to match the password failures messages in the logfile.
-# Values: TEXT
-#
+logging = catalina
+failregex = <L_<logging>/failregex>
+maxlines = <L_<logging>/maxlines>
+datepattern = <L_<logging>/datepattern>
+
+[L_catalina]
+
 failregex = ^.*\nWARNING: Authentication attempt from <HOST> for user "[^"]*" failed\.$
 
-# Option:  ignoreregex
-# Notes.:  regex to ignore. If this regex matches, the line is ignored.
-# Values:  TEXT
-#
-ignoreregex = 
-
-# "maxlines" is number of log lines to buffer for multi-line regex searches
 maxlines = 2
 
 datepattern = ^%%b %%d, %%ExY %%I:%%M:%%S %%p
               ^WARNING:()**
               {^LN-BEG}
+
+[L_webapp]
+
+failregex = ^ \[\S+\] WARN  \S+ - Authentication attempt from <HOST> for user "<F-USER>[^"]+</F-USER>" failed.
+
+maxlines = 1
+
+datepattern = ^%%H:%%M:%%S.%%f
+
+# DEV Notes:
+#
+# failregex is based on the default pattern given in Guacamole documentation :
+# https://guacamole.apache.org/doc/gug/configuring-guacamole.html#webapp-logging
+#
+# The following logback.xml Guacamole configuration file can then be used accordingly :
+# <configuration>
+#   <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+#     <file>/var/log/guacamole.log</file>
+#     <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+#       <fileNamePattern>/var/log/guacamole.%d.log.gz</fileNamePattern>
+#       <maxHistory>32</maxHistory>
+#     </rollingPolicy>
+#     <encoder>
+#       <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+#     </encoder>
+#   </appender>
+#   <root level="info">
+#     <appender-ref ref="FILE" />
+#   </root>
+# </configuration>
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -440,6 +440,7 @@ backend  = %(syslog_backend)s
 
 port     = http,https
 logpath  = /var/log/tomcat*/catalina.out
+#logpath  = /var/log/guacamole.log
 
 [monit]
 #Ban clients brute-forcing the monit gui login
--- a/fail2ban/client/actionreader.py
+++ b/fail2ban/client/actionreader.py
@@ -38,17 +38,17 @@ class ActionReader(DefinitionInitConfigReader):
 
 	_configOpts = {
 		"actionstart": ["string", None],
-		"actionstart_on_demand": ["string", None],
+		"actionstart_on_demand": ["bool", None],
 		"actionstop": ["string", None],
 		"actionflush": ["string", None],
 		"actionreload": ["string", None],
 		"actioncheck": ["string", None],
 		"actionrepair": ["string", None],
-		"actionrepair_on_unban": ["string", None],
+		"actionrepair_on_unban": ["bool", None],
 		"actionban": ["string", None],
 		"actionreban": ["string", None],
 		"actionunban": ["string", None],
-		"norestored": ["string", None],
+		"norestored": ["bool", None],
 	}
 
 	def __init__(self, file_, jailName, initOpts, **kwargs):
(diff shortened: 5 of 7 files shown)


fail2ban/tests/files/logs/guacamole
```diff
@@ -1,12 +1,2 @@
...
-# failJSON: { "match": false }
-apr 16, 2013 8:32:13 AM org.slf4j.impl.JCLLoggerAdapter warn
-# failJSON: { "time": "2013-04-16T08:32:13", "match": true , "host": "192.0.2.0" }
-WARNING: Authentication attempt from 192.0.2.0 for user "null" failed.
-# failJSON: { "match": false }
-apr 16, 2013 8:32:28 AM org.slf4j.impl.JCLLoggerAdapter warn
-# failJSON: { "time": "2013-04-16T08:32:28", "match": true , "host": "192.0.2.0" }
-WARNING: Authentication attempt from 192.0.2.0 for user "pippo" failed.
+# failJSON: { "time": "12:57:32", "match": true , "host": "182.23.72.36" }
+12:57:32.907 [http-nio-8080-exec-10] WARN  o.a.g.r.auth.AuthenticationService - Authentication attempt from 182.23.72.36 for user "guacadmin" failed.
```
benrubson: The only thing I'm not sure about is here, `time` argument.
I don't know whether or not it will match like this...

benrubson: So, it fails.
Not sure what's the best way to solve this, could be to add the date to the `datepattern` (something like `^%%Y-%%m-%%d %%H:%%M:%%S.%%f`), but then we would go away from the default Guacamole log pattern.
I'll need your advice @sebres. Thx 👍

sebres: 1. Fail2ban always assumes today if it finds a time without date
2. As you can see here, fail2ban test cases are working in "Aug 14 12:00:00 CEST 2005": https://github.com/fail2ban/fail2ban/blob/ac41b8f3be61fedf6cea724d4d1775c07d2045da/fail2ban/tests/utils.py#L359
so you could simply provide full date-time string with `"time": "2005-08-14T12:57:32"`, but if I'm not wrong we had some bug in test-cases that can consider this date as `2005-08-13T12:57:32` (as previous day).

I have to check that.

sebres:
BTW, you changed an existing filter in an incompatible way.
Normally the new versions of filters fail2ban provides are backwards compatible.
You change makes impossible to use fail2ban with older version of guacamole.

As well as I see you've changed the logpath (may be some logging is still different if used in tomcat).

Anyway I really don't know we should do that per default (may be as new filter version or parameter-related action).

fail2ban/tests/files/logs/guacamole
benrubson: Thank you for your detailed answer 👍
I just tested, to be sure / to help, and you're right, it fails as you suspected :
```
AssertionError: guacamole: UTC Time  mismatch 1123930652.0 (2005-08-13T10:57:32) != 1124017052.0 (2005-08-14T10:57:32)  (diff -86400.000 seconds)
```

sebres: > if I'm not wrong we had some bug in test-cases that can consider this date as `2005-08-13`...

OK, not a bug but feature - if parsed time is in the feature `"12:57:32" > now ("12:00:00")` assume yesterday as date, so I guess it will be `2005-08-13T12:57:32` not the 14th.

benrubson:
> BTW, you changed an existing filter in an incompatible way.

Sort of yes (*).

> Normally the new versions of filters fail2ban provides are backwards compatible.

I was not sure about this.

> You change makes impossible to use fail2ban with older version of guacamole.

(*) Older version can still configure their Guacamole logger as proposed in Guacamole documentation, also detailed in the `config/filter.d/guacamole.conf` file of this PR.

The existing filter (before this PR) relies on Tomcat to redirect its console output to `catalina.out`. Which unfortunately is no more true with last Tomcat versions / OS versions, and with Guacamole Docker images. Which implies we now must configure Guacamole logger itself.

> As well as I see you've changed the logpath (may be some logging is still different if used in tomcat).

Yes, old console output (`catalina.out`) vs Guacamole logger itself.
The Guacamole logger path of course is configurable, I tried to choose something relevant...

> Anyway I really don't know we should do that per default (may be as new filter version or parameter-related action).

IMO we could make that per default.
And add a `tomcat-legacy` action if you want to keep the old version ?

sebres:
> Which unfortunately is no more true with last Tomcat versions / OS versions, and with Guacamole Docker images. 

I'm always careful if I see Docker in some issue. What is valid for Docker, must be not necessary valid for usage without Docker container.

> Which implies we now must configure Guacamole logger itself.

So does it really not log anymore without configuration (without container)?

benrubson:
> > Which unfortunately is no more true with last Tomcat versions / OS versions, and with Guacamole Docker images.
> 
> I'm always careful if I see Docker in some issue. What is valid for Docker, must be not necessary valid for usage without Docker container.

Note that their Docker container uses Tomcat 8.5 / Debian 9.9.

> > Which implies we now must configure Guacamole logger itself.
> 
> So does it really not log anymore without configuration (without container)?

I did not try the native (non-Docker) version.
But I went through topics telling that `catalina.out` keeps empty with last Tomcat versions / OS versions, due to for example `systemd` in the path.

I would finally say that, IMO, it's better / much more reliable to configure the application logger itself, than relying on the Tomcat console.

sebres:
> I went through topics telling that catalina.out keeps empty with last Tomcat versions / OS versions, due to for example systemd in the path.

But "systemd in the path" could mean it is logged in journal (strange to expect it in log file in this case).

> it's better / much more reliable to configure the application logger itself, than relying on the Tomcat console.

Normally I would agree, but as long as you're not sure (and I don't think you are), if a PR can affect some working default configurations (there are still some constellations possible where a reconfigure is not expected) - then it breaks backwards compatibility.
I don't think you'd like to see a "lot" of issues like "Guacamole filter does not work after upgrade to 0.10.x"?
Each issue like that cost time and simply detracting our attention from the future development.

> IMO we could make that per default.
> And add a tomcat-legacy action if you want to keep the old version ?

Neither first nor second.
I'd suggest to specify that via parameter `logging`, `format` or even `logtype` (at the moment used to distinguish `file` and `journal` but can be set by filter option in jail), for example something like that:
```ini
[Definition]
logging = catalina
datepattern = <L_<logging>/datepattern>
failregex = <L_<logging>/failregex>

[L_catalina]
...

[L_webapp]
...
```
as well as use conditional filterOptions to cover both variants in `tests/files/logs/guacamole`.
```diff
 ... messages of previous log format ...
+ # filterOptions: [{"logging": "webapp"}]
+ ... messages of new log format ...
```
Need to think about that.

benrubson:
I agree with your proposal @sebres, which then handles "old and new" logging methods.
Do you want me to give this template a try through an additional commit to this PR ?

benrubson:
I then just have pushed a new version based on your template.

However sounds like I have some issues with `maxlines` which is not seen correctly.
May not work like `datepattern` and `maxlines`.

Feel free to advise :)
Many thx !

benrubson:
As a test, enforcing `maxlines = 2` instead of `maxlines = <L_<logging>/maxlines>` makes the tests to succeed, but of course this is not what we want.

sebres:
> As a test, enforcing maxlines = 2 instead of maxlines = <L_<logging>/maxlines> makes the tests to succeed, but of course this is not what we want.

Hmm... I'll take a look what is "wrong" here (possibly the filter factory in test-cases does not expect switch of `maxlines` at the moment).

benrubson:
Thank you @sebres 👍

sebres:
Another question - take a look at 2 messages from my answer https://github.com/fail2ban/fail2ban/issues/2645#issuecomment-592032811.
What do you think about both?
First (with anonymous) is totally different, second miss user name for some reason.

benrubson:
I'm not sure we should catch _anonymous_ messages.
See relevant code [here](https://github.com/apache/guacamole-client/blob/ae4361ab70dae4829b08f34a80b08fac885703ba/guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java#L309-L319).
We would in addition need to log in debug mode.

I'm more concerned by the `[<HOST>(?:,[^\]]*)?\]` writing.
As you can see in this PR, `<HOST>` is not enclosed.
Again, relevant code [here](https://github.com/apache/guacamole-client/blob/ae4361ab70dae4829b08f34a80b08fac885703ba/guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java#L127-L135).
The [Remote IP Valve](https://guacamole.apache.org/doc/gug/proxying-guacamole.html) may have not been configured, leading to proxy + client IPs showing into the log line, which could then lead, depending on the Fail2Ban rule, to a proxy ban, which is certainly not what the Fail2Ban user wants...

And I'm pretty sure lines & rules did not match because of the missing `datepattern = ^%%H:%%M:%%S\.%%f` in the filter definition (which you then added into your last message) 👍

benrubson:
Regarding the Remote IP Valve, there's a PR [here](https://github.com/apache/guacamole-client/pull/469) to enable it by default in the Guacamole container.

sebres:
> And I'm pretty sure lines & rules did not match because of the missing `datepattern = ^%%H:%%M:%%S\.%%f` in the filter definition (which you then added into your last message) 

Nope, I said already why that occurred (anonymous by first and missing "user name" after `for user` by second).
And the `datepattern` does no matter in his case - as long as initial failregex was not anchored (it starts in-between of message, with word boundary) and fail2ban has a default pattern (but without `\.%f`) for time only. I provided it only because I anchored the regex to begin after whole timestamp.
Simply because our filters should be normally universal, so must work in any case, also if someone has different timestamp format and rewrite `datepattern` (or it will be switched to systemd, which does not own datepattern) - otherwise the part matching milliseconds should be matched by failregex if it is anchored by start and it'd bother.

benrubson:
Let's then wait for his feedback after his tests :)
There may however be an issue with the rule you proposed him :
```
failregex = ^\s*\[[^\]]+\] \S+ \S*auth.AuthenticationService - (?:Anonymous )?[Aa]uthentication attempt from \[<HOST>(?:,[^\]]*)?\] (?:for user (?:"[^"]*" )?)?failed\.\s*$
```
Look at the example in this PR :
```
12:57:32.907 [http-nio-8080-exec-10] WARN  o.a.g.r.auth.AuthenticationService - Authentication attempt from 182.23.72.36 for user "guacadmin" failed.
```
Look at the double-space after `WARN`.
Which is due to the log pattern used :
```
<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
```

sebres:
> Look at the double-space after WARN.

I provided it only for 2 messages that he has, Nothing else.
And for your case I assume this simple change can work:
```diff
-\S+ \S*
+\S+\s+\S*
```

benrubson:
> > I then just have pushed a new version based on your template.
> > (...)
> > As a test, enforcing maxlines = 2 instead of maxlines = <L_/maxlines> makes the tests to succeed, but of course this is not what we want.
> 
> Hmm... I'll take a look what is "wrong" here (possibly the filter factory in test-cases does not expect switch of `maxlines` at the moment).

We could also simply create a `guacamole-legacy` filter, and a `guacamole` one.

benrubson:
> > I then just have pushed a new version based on your template.
> > (...)
> > As a test, enforcing maxlines = 2 instead of maxlines = <L_/maxlines> makes the tests to succeed, but of course this is not what we want.
> 
> Hmm... I'll take a look what is "wrong" here (possibly the filter factory in test-cases does not expect switch of `maxlines` at the moment).

Culprit seems to be here :
https://github.com/fail2ban/fail2ban/blob/f2aec43d8aba52d758da57a6473928f83df7db58/fail2ban/server/filter.py#L365-L369
`setMaxLines` only expect an integer and thus does not understand references.

We could simply create a `guacamole-legacy` filter, and a `guacamole` one, we would then avoid `maxlines ` references. If you don't want to make too many "deep" modifications for just one jail...

benrubson:
How could we go further with this @sebres ?
Thank you !

benrubson:
How could we sort this @sebres ?
- solve the `maxlines` references issue ?
- create separated filters, `guacamole-legacy` and `guacamole` ?
- get rid of the old / legacy jail ?

Thank you for your help...

sebres:
I've rebased the branch to current 0.10.

And provided a fix for wrong conversion in 7b05c1ce7a487f2c269815f8a17b7444213ae555.
Although I believe the fix doing that properly and the test-cases should pass now, just I have to think about that a bit (as regards the compatibility in 0.10/0.11).

benrubson:
> And provided a fix for wrong conversion in [7b05c1c](https://github.com/fail2ban/fail2ban/commit/7b05c1ce7a487f2c269815f8a17b7444213ae555).

Really good news, thank you very much !

> I believe the fix doing that properly and the test-cases should pass now

Tests now pass, good news too :+1:

> I have to think about that a bit (as regards the compatibility in 0.10/0.11)

OK :) Certainly that 0.11 changes too many things to literally apply your fix there then.

benrubson:
Many thanks Sergey for your help & support 👍

PR #2635, opened 2020-02-14
Apache, allow patterns to be used from fcgi

benrubson:
Hi,

One may want to use the default Fail2Ban patterns from its own (fcgi/PHP) application, of course for security reasons, and by convenience.
This PR then allows to use Apache `failregex` patterns from fcgi scripts, and for instance from PHP scripts.

_For ref, heres's the `Got error` code (`AH01071`) from Apache source :
https://github.com/apache/httpd/blob/2.4.41/modules/proxy/mod_proxy_fcgi.c#L859
The `PHP message:` prefix comes from PHP, when used in FPM mode :
https://github.com/php/php-src/blob/8202b4ed09fe4160ac9265780816cf380a34b480/sapi/fpm/fpm/fpm_main.c#L609
When used in CGI mode, there's no prefix added :
https://github.com/php/php-src/blob/8202b4ed09fe4160ac9265780816cf380a34b480/sapi/cgi/cgi_main.c#L752_

Thank you 👍


benrubson:
Hi @sebres, could we think about merging this small enhancement please ?
Thank you again 👍

benrubson:
Conflict resolved.
Let's merge this if possible @sebres ?
Many thanks 👍

sebres:
I don't know:
1. The subject and changelog saying about fcgi, but message is pretty clear an PHP error. Can it be some few precise RE (like `(?:Got error '(?:\w+ message: )?)?`) or should it remain PHP?
2. Are you sure that this is global and so belongs to `prefregex` and not to some of `failregex` rather?

benrubson:
Many thanks @sebres for your feedback.

1. The `PHP message:` prefix comes from PHP, when used in FPM mode :
https://github.com/php/php-src/blob/8202b4ed09fe4160ac9265780816cf380a34b480/sapi/fpm/fpm/fpm_main.c#L609
When used in CGI mode, there's no prefix added :
https://github.com/php/php-src/blob/8202b4ed09fe4160ac9265780816cf380a34b480/sapi/cgi/cgi_main.c#L752
I think we could then keep the `PHP` word, to have an explicit / precise reference, but we could also change it for `\w+`, not sure though this would cover other cases. Up to U, just tell me :)

2. Goal is to be able to use any of the `failregex` patterns from applications themselves. `prefregex` then seems to be the proper place for this additional prefix.

Thank you again 👍

benrubson:
Closing this for now, I'll find another way to achieve this goal (I'll certainly use `.local` files). Thx !

PR #2651, opened 2020-03-06
python 3.9 compatibility + CI

### LINKED ISSUE #2646: python 3.9 threading changes
Copied from https://bugzilla.redhat.com/show_bug.cgi?id=1808347

            Bug ID: 1808347
           Summary: fail2ban fails to build with Python 3.9: uses removed
                    threading.Thread.isAlive()
           Product: Fedora
           Version: rawhide
            Status: NEW
         Component: fail2ban
          Assignee: orion@nwra.com
          Reporter: mhroncok@redhat.com
        QA Contact: extras-qa@fedoraproject.org
                CC: anon.amish@gmail.com, axel.thimm@atrpms.net,
                    cstratak@redhat.com, mhroncok@redhat.com,
                    mplch@redhat.com, orion@nwra.com, vonsch@gmail.com
            Blocks: 1785415 (PYTHON39)
  Target Milestone: ---
    Classification: Fedora

fail2ban fails to build with Python 3.9.0a4.
```
ERROR: test_del_file
(fail2ban.tests.filtertestcase.MonitorFailures<FilterPoll>(/tmp/monitorfailures_FilterPoll__nshi9c2fail2ban))
  File "./fail2ban/tests/filtertestcase.py", line 987, in setUp
    self._sleep_4_poll()
  File "./fail2ban/tests/filtertestcase.py", line 1008, in _sleep_4_poll
    Utils.wait_for(self.filter.isAlive, _maxWaitTime(5))
AttributeError: 'FilterPoll' object has no attribute 'isAlive'
```

And more. This looks like:

See https://docs.python.org/3.9/whatsnew/3.9.html#removed

"The isAlive() method of threading.Thread has been removed. It was deprecated
since Python 3.8. Use is_alive() instead."

For the build logs, see:
https://copr-be.cloud.fedoraproject.org/results/@python/python3.9/fedora-rawhide-x86_64/01255081-fail2ban/

For all our attempts to build fail2ban with Python 3.9, see:
https://copr.fedorainfracloud.org/coprs/g/python/python3.9/package/fail2ban/

Testing and mass rebuild of packages is happening in copr. You can follow these
instructions to test locally in mock if your package builds with Python 3.9:
https://copr.fedorainfracloud.org/coprs/g/python/python3.9/

Let us know here if you have any questions.

Python 3.9 will be included in Fedora 33. To make that update smoother, we're
building Fedora packages with early pre-releases of Python 3.9.
A build failure prevents us from testing all dependent packages (transitive
[Build]Requires), so if this package is required a lot, it's important for us
to get it fixed soon.
We'd appreciate help from the people who know this package best, but if you
don't want to work on this now, let us know so we can try to work around it on
our side.

Referenced Bugs:

https://bugzilla.redhat.com/show_bug.cgi?id=1785415
[Bug 1785415] Python 3.9

sebres:
Fixes python 3.9 issue, added python 3.9 to travis CI.
Closes #2646

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
   - python:  3.6
   - python:  3.7
   - python:  3.8
+  - python:  3.9-dev
   - python:  pypy3.5
 before_install:
   - echo "running under $TRAVIS_PYTHON_VERSION"
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,6 +32,7 @@ ver. 0.10.6-dev (20??/??/??) - development edition
   IPv6-capable now.
 
 ### Fixes
+* python 3.9 compatibility (and Travis CI support)
 * restoring a large number (500+ depending on files ulimit) of current bans when using PyPy fixed
 * manual ban is written to database, so can be restored by restart (gh-2647)
 * `filter.d/common.conf`: avoid substitute of default values in related `lt_*` section, `__prefix_line`
--- a/fail2ban/server/jailthread.py
+++ b/fail2ban/server/jailthread.py
@@ -120,3 +120,6 @@ class JailThread(Thread):
 ## python 2.x replace binding of private __bootstrap method:
 if sys.version_info < (3,): # pragma: 3.x no cover
 	JailThread._Thread__bootstrap = JailThread._JailThread__bootstrap
+## python 3.9, restore isAlive method:
+elif not hasattr(JailThread, 'isAlive'): # pragma: 2.x no cover
+	 JailThread.isAlive = JailThread.is_alive
--- a/fail2ban/tests/sockettestcase.py
+++ b/fail2ban/tests/sockettestcase.py
@@ -86,7 +86,7 @@ class Socket(LogCaptureTestCase):
 	def _stopServerThread(self):
 		serverThread = self.serverThread
 		# wait for end of thread :
-		Utils.wait_for(lambda: not serverThread.isAlive() 
+		Utils.wait_for(lambda: not serverThread.is_alive() 
 			or serverThread.join(Utils.DEFAULT_SLEEP_TIME), unittest.F2B.maxWaitTime(10))
 		self.serverThread = None
 
@@ -97,7 +97,7 @@ class Socket(LogCaptureTestCase):
 		self.server.close()
 		# wait for end of thread :
 		self._stopServerThread()
-		self.assertFalse(serverThread.isAlive())
+		self.assertFalse(serverThread.is_alive())
 		# clean :
 		self.server.stop()
 		self.assertFalse(self.server.isActive())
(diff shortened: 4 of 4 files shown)


hobbes1069:
I found my way here from the Fedora Bugzilla ticket. With or without using the pull request as a patch against 0.11.1 I get the following error:
```
ERROR: testBanTimeIncr (fail2ban.tests.observertestcase.BanTimeIncrDB)
  File "./fail2ban/tests/observertestcase.py", line 429, in testBanTimeIncr
    self.assertTrue(ticket.restored)
AttributeError: 'bool' object has no attribute 'restored'

```

sebres:
This PR has nothing to do with that, neither it affects ticket handling nor observer module (functions related)...
Although I cannot reproduce it at all (and CI also shows nothing here), I can imagine some timing issue affecting this test.
But I'll suggest to try latest 0.11th branch firstly.
And which python version is it?

hobbes1069:
It's 3.9 (Fedora Rawhide) but I can also reproduce it in Fedora 32 (Python 3.8)

sebres:
OK, I can exclude timing (sporadic) issue now. It is something other...

This test is unfortunately working with `127.0.0.2`, and per default ignoreself is on, so I assume it is simply (mistakenly) ignored here (this IP could be resolved as your host IP).

Can you confirm seeing `[DummyJail] Ignore 127.0.0.2 by ignoreself rule` immediately before it fails,
if you test it this way:
```
fail2ban-testcases -f -l 10 'testBanTimeIncr.*observer'
```

hobbes1069:
No, I don't see anything like that in the build logs. 

I might be able to test the command manually but it's not easy... I just took over maintenance of the package for Fedora and all package building happens in a chroot. It is possible to shell into the chroot but it is a PITA. I may try applying the referenced commit and see if it "fixes" the problem.

hobbes1069:
I can confirm the referenced commit fixes the problem for me.

sebres:
> It is possible to shell into the chroot but it is a PITA.

You don't need that. For next time, just clone or download fail2ban and invoke this:
```
cd /path/to/fail2ban-dist/
./fail2ban-2to3
PYTHONPATH=. python3 bin/fail2ban-testcases -f -l 10 'testBanTimeIncr.*observer'
```
> I can confirm the referenced commit fixes the problem for me.

Good to know, thx for feedback!

hobbes1069:
Ok, I inherited this package so I'm still figuring everything in the RPM spec file out. It's setup to run 2to3 manually instead of using the supplied script. Could that have been the problem? And should I use the supplied script?

sebres:
I have no idea how a build-script for fedora package does look.
My example was just about how to start a test-suite manually (e. g. single test, with debug level, custom parameters, etc) without to build, entering setup process and without installing python modules.

Sure it'd deviate from complete merged and packaged installation, but it is enough to test some stuff by hand.

And of course 2to3 over code base is only necessary if you would use python3 here.

Issue #2128, opened 2018-05-13
Cloudflare Unban Rule Malfunction


RaidForums:
**I am running the latest cloudflare action file updated 9 months ago but I am having a problem which is the unban rule no longer works. The get CURL request within this unban rule that is supposed to return 1 result for the script to be able to unban the IP instead returns all of page 1 so if you have multiple blocked IPs fail2ban is unable to unban the IP.**

I have about 1300 IPs on my Cloudflare Access Rules right now blocked because the script is no longer working as intended this should be a pretty easy fix.

RaidForums:
Sorry, I originally misclicked without leaving an issue comment.

RaidForums:
After looking into this I realise that no matter what I set the per_page value to on the API It is being reset to 25 when the request is fulfilled by the API. I have made a support ticket with Cloudflare for them to fix this.

RaidForums:
The Cloudflare API has been updated and is working fine but the current unban action is still not working.

RaidForums:
I am currently just making fail2ban run a php script I created to unban the IPs because I didn't see any simple and basic way of fixing the current problem. I guess i'll be waiting until one of the fail2ban developers fix this problem.

wpkc:
My update to the cloudflare action works fine. <https://github.com/wpkc/fail2ban-action-cloudflare-restv4>

I think the use of `cut -d '"' -f6` would be problematic because there is no guarantee that the format of the JSON data returned will always have the IP address in the 6th delimited field.  

My method may be a bit more resource intensive, but I use a single line python script to extract the IP address from the first result in the JSON response data. Then I pipe it to xargs to build and execute the curl statement. I also use the `notes` parameter to make sure Fail2Ban is only deleting firewall rules that it created.  Otherwise, an admin or other program might have set it's own firewall rule only to have Fail2Ban delete it.

```
actionban = curl -s -X POST https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules \
	-H "X-Auth-Email: <cfuser>" -H "X-Auth-Key: <cfkey>" -H "Content-Type: application/json" \
--data '{"mode":"block","configuration":{"target":"ip","value":"<ip>"},"notes":"Banned by Fail2Ban"}'

actionunban = curl -s -X GET -H "X-Auth-Email: <cfuser>" -H "X-Auth-Key: <cfkey>" -H "Content-Type: application/json" \
[...]

PR #2656, opened 2020-03-11
Fix actionunban for cloudflare action

PRIHLOP:
Add command to remove new line character. Needed for working removing rule from cloudflare firewall.

### DIFF
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -60,7 +60,7 @@ actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-
 # API v4
 actionunban = curl -s -o /dev/null -X DELETE -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
               https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | cut -d'"' -f6)
+              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | tr -d '\n' | cut -d'"' -f6)
 
 [Init]
 
--- a/config/action.d/cloudflare.conf
+++ b/config/action.d/cloudflare.conf
@@ -60,7 +60,7 @@ actionban = curl -s -o /dev/null -X POST -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-
 # API v4
 actionunban = curl -s -o /dev/null -X DELETE -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
               https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$(curl -s -X GET -H 'X-Auth-Email: <cfuser>' -H 'X-Auth-Key: <cftoken>' \
-              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | tr -d '\n' | cut -d'"' -f6)
+              'https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=<ip>&page=1&per_page=1' | tr -d '\n' | sed -nE 's/^.*"result"\s*:\s*\[\s*\{\s*"id"\s*:\s*"([^"]+)".*$/\1/p' )
 
 [Init]
 


sebres:
Basically I found the same suggestion on [SF](https://serverfault.com/a/912393/488604) with a comment:
> Looks like Cloudflare has adjusted their API to return formatted JSON (previously the returned JSON was on a single line).

I can merge this PR, but I'd like to know how the output of this line look now:
```bash
cfuser=...; # replace with user
cftoken=...; # replace with token
ip=192.0.2.1; # replace with some currently banned IP
curl -s -X GET -H "X-Auth-Email: $cfuser" -H "X-Auth-Key: $cftoken" \
"https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=$ip&page=1&per_page=1"
```
Can someone, who uses cloudflare action, provide the returned json output, please?

Another thing - it is also good to know what one would get if curl exec is extended with `-H "Content-Type:application/octet-stream"` or `-H "Content-Type:text/plain"`... 
Corresponding [API description](https://api.cloudflare.com/#getting-started-requests) it should return raw data or textual item.
May be we can simplify that a bit.

PRIHLOP:
> Basically I found the same suggestion on [SF](https://serverfault.com/a/912393/488604) with a comment:
> 
> > Looks like Cloudflare has adjusted their API to return formatted JSON (previously the returned JSON was on a single line).
> 
> I can merge this PR, but I'd like to know how the output of this line look now:
> 
> ```shell
> cfuser=...; # replace with user
> cftoken=...; # replace with token
> ip=192.0.2.1; # replace with some currently banned IP
> curl -s -X GET -H "X-Auth-Email: $cfuser" -H "X-Auth-Key: $cftoken" \
> "https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules?mode=block&configuration_target=ip&configuration_value=$ip&page=1&per_page=1"
> ```
> 
> Can someone, who uses cloudflare action, provide the returned json output, please?
> 

I tested and gets this.

```
{
  "result": [],
  "success": true,
  "errors": [],
  "messages": [],
  "result_info": {
    "page": 1,
    "per_page": 1,
    "count": 0,
    "total_count": 0,
    "total_pages": 0
  }
}
```

sebres:
> I tested and gets this.

The output looks weird... Are you sure you invoked it for some IP that was currently banned - result seems to be empty (and count is 0 too)?

Additionally how with my second "wish", can you then (if you get proper result) repeat that also with text/plain content type:
```diff
-curl ... "https://...
+curl ... -H "Content-Type:text/plain" "https://...
```

PRIHLOP:
Sorry @sebres , I don't understand you correctly at first.
Output for banned IP:
```
{
  "result": [
    {
      "id": "221db3858de944aa92cbda30aa2382e6",
      "paused": false,
      "modified_on": "2020-03-16T15:52:47.812883Z",
      "allowed_modes": [
        "whitelist",
        "block",
        "challenge",
        "js_challenge"
      ],
      "mode": "block",
      "notes": "",
      "configuration": {
        "target": "ip",
        "value": "8.8.8.8"
      },
      "scope": {
        "id": "096e33f03ed0e81db9f40c9a4ba7a1cb",
        "email": "$cfuser",
        "type": "user"
      },
      "created_on": "2020-03-16T15:52:47.886115033Z"
    }
  ],
  "success": true,
  "errors": [],
  "messages": [],
  "result_info": {
    "page": 1,
    "per_page": 1,
    "count": 1,
    "total_count": 1,
    "total_pages": 1
  }
}
```
With ```-H "Content-Type:text/plain"``` output is the same.

sebres:
Hmm... 
I guess then usage of `cut` looks too "vulnerable" to me.
I'd suggest something like this:
```bash
... | tr -d '\n' | sed -nE 's/^.*"result"\s*:\s*\[\s*\{\s*"id"\s*:\s*"([^"]+)".*$/\1/p'
```
or even `jq` as suggested in #2128 (and PR #2140)

PRIHLOP:
Ok, no problem. I was made changes as you suggested. I was tested this solution - it works.

baptx:
For information, I used the solution from https://technicalramblings.com/blog/cloudflare-fail2ban-integration-with-automated-set_real_ip_from-in-nginx/ which is based on https://serverfault.com/questions/910940/fail2ban-unban-action-fails-with-cloudflare/912547#912547

sebres:
I rebased it as basis of #2140 to this PR (0.10-based). Will be merged to 0.11 / 1.0 later.
Thx!

baptx:
@sebres by the way, why do you consider the cut command vulnerable? It seems to be the easiest solution to fix the problem.

sebres:
Because I'm a developer... (and know <s>what I do</s> that easiest is often not the same as proper)... Just kidding :)

Because `cut` is just considering the order of fields string (that is JSON). not its real names. It is vulnerable if the order of fields would change tomorrow (by the same structure or named convention), for example some new field going added before the "id".

Both of this JSON's dicts are "equal" - `{"A":1, "B":"2"}` and `{"B":"2", "A":1}` (just another order inside), and using `cut` you'd access the field value (or even name) quasi by its "index", using regexp (or `jq`) by its name:
```
$ echo '{"A":1, "B":"2"}' | cut -d'"' -f6
2
$ echo '{"B":"2", "A":1}' | cut -d'"' -f6
A
$ echo '{"O":0, "A":1, "B":"2"}' | cut -d'"' -f6
B

$ echo '{"A":1, "B":"2"}' | sed -nE 's/^.*"B"\s*:\s*"([^"]+)".*$/\1/p'
2
$ echo '{"B":"2", "A":1}' | sed -nE 's/^.*"B"\s*:\s*"([^"]+)".*$/\1/p'
2
$ echo '{"O":0, "A":1, "B":"2"}' | sed -nE 's/^.*"B"\s*:\s*"([^"]+)".*$/\1/p'
2
```

PR #2702, opened 2020-04-25
New openvpn jail

pburndorfer:
Providing a new jail for openvpn. Tested with openvpn version 2.4.4.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,6 +42,7 @@ ver. 1.1.1-dev-1 (20??/??/??) - development nightly edition
 * `action.d/firewallcmd-rich-*.conf` - fixed incorrect quoting, disabling port variable expansion
   by substitution of rich rule (gh-3815)
 * `filter.d/proxmox.conf` - add support to Proxmox Web GUI (gh-2966)
+* `filter.d/openvpn.conf` - new filter and jail for openvpn recognizing failed TLS handshakes (gh-2702)
 
 ver. 1.1.0 (2024/04/25) - object-found--norad-59479-cospar-2024-069a--altitude-36267km
 -----------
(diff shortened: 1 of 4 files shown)


Aareez01 (approved):

sebres:
Rebased to latest master.
Additionally I fixed the test-cases (added missing tests) and a bit rewrote the filter (better and faster REs).

However I'm unsure they all is really the cases for authentication issues (maybe some of them belongs rather to aggressive mode) and especially that some failure would not counted twice. Here is an example log-excerpt (from web), which illustrates this - it would count basically single failure twice (doe to 2 matches for the first and the last lines):
```
Mar 18 00:36:53 ovpn-server[123]: 192.0.2.123:12345 VERIFY ERROR: depth=2, error=unable to get issuer certificate: <ROOT CA>
Mar 18 00:36:53 ovpn-server[123]: 192.0.2.123:12345 OpenSSL: error:1417C086:SSL routines:tls_process_client_certificate:certificate verify failed
Mar 18 00:36:53 ovpn-server[123]: 192.0.2.123:12345 TLS_ERROR: BIO read tls_read_plaintext error
Mar 18 00:36:53 ovpn-server[123]: 192.0.2.123:12345 TLS Error: TLS object -> incoming plaintext read error
Mar 18 00:36:53 ovpn-server[123]: 192.0.2.123:12345 TLS Error: TLS handshake failed
```

pburndorfer:
I'm not really into this anymore but I'd say any TLS handshake failure or similar are possible attacks/attackers trying to knock on that port etc.? With a valid certificate those shouldn't occur, therefore I wouldn't mind if such an IP address causing that incident is immediately blocked.

sebres:
OK, merged now as it is.

PR #2742, opened 2020-05-30
Update  drupal-auth.conf

aresxc:
Small fix for Drupal 8. D8 uses "Login attempt failed from" while Drupal 7 uses "Login attempt failed for".
The referer part is a must currently, but some requests have an empty one and are not failing.

**[UPD]** Additionally introduces more precise RE, which avoids certain weakness with catch-all's and is injection safe (see new tests).
It would not match in case of user name contains `|` character, but since D8 seems to not encode or quote it properly, we have no chance to make suitable RE (without to accept injectable URIs) and because it would be anyway poor choice of user name, we can accept a small restriction like this.

### DIFF
--- a/config/filter.d/drupal-auth.conf
+++ b/config/filter.d/drupal-auth.conf
@@ -14,7 +14,7 @@ before = common.conf
 
 [Definition]
 
-failregex = ^%(__prefix_line)s(https?:\/\/)([\da-z\.-]+)\.([a-z\.]{2,6})(\/[\w\.-]+)*\|\d{10}\|user\|<HOST>\|.+\|.*\|\d\|.*\|Login attempt failed (?:for|from) .+\.$
+failregex = ^%(__prefix_line)s(?:https?:\/\/)[^|]+\|[^|]+\|[^|]+\|<ADDR>\|(?:[^|]*\|)*Login attempt failed (?:for|from) <F-USER>[^|]+</F-USER>\.$
 
 ignoreregex =
 
--- a/config/filter.d/drupal-auth.conf
+++ b/config/filter.d/drupal-auth.conf
@@ -14,7 +14,7 @@ before = common.conf
 
 [Definition]
 
-failregex = ^%(__prefix_line)s(https?:\/\/)([\da-z\.-]+)\.([a-z\.]{2,6})(\/[\w\.-]+)*\|\d{10}\|user\|<HOST>\|.+\|.+\|\d\|.*\|Login attempt failed for .+\.$
+failregex = ^%(__prefix_line)s(https?:\/\/)([\da-z\.-]+)\.([a-z\.]{2,6})(\/[\w\.-]+)*\|\d{10}\|user\|<HOST>\|.+\|.*\|\d\|.*\|Login attempt failed (?:for|from) .+\.$
 
 ignoreregex =
 
--- a/fail2ban/tests/files/logs/drupal-auth
+++ b/fail2ban/tests/files/logs/drupal-auth
@@ -3,5 +3,15 @@ Apr 26 13:15:25 webserver example.com: https://example.com|1430068525|user|1.2.3
 # failJSON: { "time": "2005-04-26T13:15:25", "match": true , "host": "1.2.3.4" }
 Apr 26 13:15:25 webserver example.com: https://example.com/subdir|1430068525|user|1.2.3.4|https://example.com/subdir/user|https://example.com/subdir/user|0||Login attempt failed for drupaladmin.
 
-# failJSON: { "time": "2005-04-26T13:19:08", "match": false , "host": "1.2.3.4" }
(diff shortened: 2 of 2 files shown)


greggles:
Thanks for your work on this @aresxc and @sebres! I tested the changed regular expression on regexpal and it worked as I expected.

jurgenhaas:
Works for me too. Hope this can be merged soon because Drupal admins are most likely not aware that this filter isn't working anymore.

sebres:
Sure, since this PR is very simple and can be merged, but... This filter has too many catch-all's, so may be we can make it more precise simultaneously.
How about this RE:
```regex
^%(__prefix_line)s(?:https?:\/\/)[^|]+\|[^|]+\|[^|]+\|<ADDR>\|(?:[^|]*\|)*Login attempt failed (?:for|from) <F-USER>\S+</F-USER>\.$
```

jurgenhaas:
@sebres well, that sounds like a different issue to me - but I'm not sure about the policy of this project.

sebres:
The policies don't really matter here.

There are several reasons for do it here and now:
- the (mentioned "vulnerable") catch-all part is changed here (`\|.+\|` -> `\|.*\|`), so indirectly it makes from that not a different issue anymore;
- although we are not in the mode to save a number of issues, but we'd not necessarily appreciate dozen issues or PRs affecting the same filter, moreover the same code line.
- if some line gets touched and some thing (especially a weakness) can be fixed, it must be done
- github allows to change issue title and other stuff and PR can well get a continuation

With other words if you should replace a variable name due some task in code like that:
```diff
- i = j / 0
+ i = i / 0
```
would you really retain a part producing an exception (division by zero), if it is unintended?

jurgenhaas:
Fair enough, so I'll give your proposed new line a try and let you know if it works here.

jurgenhaas:
Looks like it is working if the <F-USER> field is either an IP address or an alpha-numeric username. When the username contains a space though it is not recognized.

sebres:
> When the username contains a space though it is not recognized.

No idea the space is useable in user name in drupal.
You could provide an example how it looks, but if it is plain space (without quotes, escape etc), just change `\S+` to `.+` at end, like here:
```diff
- ^%(__prefix_line)s(?:https?:\/\/)[^|]+\|[^|]+\|[^|]+\|<ADDR>\|(?:[^|]*\|)*Login attempt failed (?:for|from) <F-USER>\S+</F-USER>\.$
+ ^%(__prefix_line)s(?:https?:\/\/)[^|]+\|[^|]+\|[^|]+\|<ADDR>\|(?:[^|]*\|)*Login attempt failed (?:for|from) <F-USER>.+</F-USER>\.$
```
Such a catch-all at end would not really bother, but captured user name can become "incorrect" if this message format changing again later (e. g. get some additional data after user name).

jurgenhaas:
Yes, I had already changed from `\S+` to `.+` and verified that it's working.

The log looks like this:

```
Feb 11 16:49:59 mweb1 drupal_site[24864]: https://www.example.com|1613058599|user|4.9.3.7|https://www.example.com/en/user/login|https://www.example.com/en/user/login|0||Login attempt failed for Jürgen Haas.
```

aresxc:
D8 log example:
`Feb 11 16:55:01 host1 drupal-site: https://example.com|1613062501|user|1.2.3.4|https://example.com/user/login|https://example.com/user/login|0||Login attempt failed from 1.2.3.4.`

sebres:
I have no idea how drupal routing works per default, but If you're already in the mode to check, can you please try some "injection" URL variants, just to ensure they are either rejected by drupal or pipe-char is properly encoded (as `%7c` or somehow else) in log, so something like this:
```
https://www.example.com/en/user/login;test|test
https://www.example.com/en/user/login?test=|&test2=|
https://www.example.com/en/user/login|test|test
```

aresxc:
> I have no idea how drupal routing works per default, but If you're already in the mode to check, can you please try some "injection" URL variants, just to ensure they are either rejected by drupal or pipe-char is properly encoded (as `%7c` or somehow else) in log, so something like this:
> 
> ```
> https://www.example.com/en/user/login;test|test
> https://www.example.com/en/user/login?test=|&test2=|
> https://www.example.com/en/user/login|test|test
> ```

Just the second one:
```
Feb 11 17:13:01 host1 drupal-site: https://example.com|1613063581|user|1.2.3.4|https://example.com/user/login?test=%7C&test2=%7C|https://example.com/user/login?test=|&test2=||0||Login attempt failed from 1.2.3.4.
```

sebres:
Weird.
Well, then `.+` as RE for username (or host in case of "from") is too weak, because can allow such an injection for requests of legitimate users (URI may contain `Login attempt failed ...` followed by some stuff simulating invalid attempt).
But since pipe were not really suitable character for the user name, I think it is good enough to change `.+` with `[^|]+`.

sebres:
Looks good now, tests passed too, so lets merge.
Thx for contribution and testing!

jurgenhaas:
Thx for this great improvement!

PR #2750, opened 2020-06-06
Added filter nginx-bad-request

janprzy:
This filter matches http requests that the server return "400" to. My server is sometimes spammed with garbage requests (often just containing random ASCII characters), so I made this filter to ban the offenders.

It could, if necessary, easily be changed or expanded to match other response codes (like 401 or 403). However, I did not do that here, as I want these codes to be handled differently.


config/filter.d/nginx-bad-request.conf
```diff
@@ -0,0 +1,13 @@
+# Fail2Ban filter to match bad requests to nginx
+#
+
+[Definition]
+
+# The request often doesn't contain a method, only some encoded garbage
+failregex = ^<HOST> \- \S+ \[\] \".+\" 400 .+$
```
sebres, line 7: Part of RE `\".+\" 400 .+$` is a bit vulnerable, e. g. contains a catch-all before 400 and `.+$` which is not really anchor (quasi equivalent to `.`) - either it must be more precise without a not needed anchor like `"[^"]+" 400` (as I guess the quotation mark character is not logged as part of URI, because will be escaped as `\x22`). Or it must be more precise between 400 and anchor `$`.

So I'll suggest something like this:
```diff
-failregex = ^<HOST> \- \S+ \[\] \".+\" 400 .+$
+failregex = ^<HOST> - \S+ \[\] "[^"]*" 400
```

Although I known we've already some old filters with similar "vulnerable" regex, but let us don't make more of that.
We'd not accept unjustified catch-alls anymore.

sebres: Additionally note that an observation of access-log file is ugly (see [wiki :: Best practice](https://github.com/fail2ban/fail2ban/wiki/Best-practice#reduce-parasitic-log-traffic)), so normally one would configure nginx (e. g. by creating of special location or using conditional logging) to log all malformed or not authorized attempts of every kind into different log-file, monitored by fail2ban, where every line (regardless the return code) must be considered as failure. Then such common jail is pretty simply and can look like this:
```ini
[nginx-common]
filter =
logpath = /var/log/nginx-common-errors.log
failregex = ^HOST
...
```

sebres:
Additionally note that an observation of access-log file is ugly (see [wiki :: Best practice](https://github.com/fail2ban/fail2ban/wiki/Best-practice#reduce-parasitic-log-traffic)), so normally one would configure nginx (e. g. by creating of special location or using conditional logging) to log all malformed or not authorized attempts of every kind into different log-file, monitored by fail2ban, where every line (regardless the return code) must be considered as failure. Then such common jail is pretty simply and can look like this:
```ini
[nginx-common]
filter =
logpath = /var/log/nginx-common-errors.log
failregex = ^HOST
...
```

config/filter.d/nginx-bad-request.conf
janprzy: That seems reasonable, I just committed it.

sebres: Artificial escape sequences (`\-` or `\"`) are unneeded here:
```diff
-failregex = ^<HOST> \- \S+ \[\] \"[^\"]+\" 400
+failregex = ^<HOST> - \S+ \[\] "[^"]+" 400
```
Also note that you did not provide a test file for that (which is mandatory).

janprzy: I removed the escape sequences and added a test file

Issue #2802, opened 2020-08-10
Nextcloud Talk - COTURN log - no timestamp


xtrmbuster:
### Environment:

Here is the snippet of log file generated by latest version of [COTURN](https://github.com/coturn/coturn) application.

Running on Ubuntu 20.04LTS and also on Debian 10 x64

`fail2ban-regex --version => 0.10.2`

### The log example :

I'm testing on Debian now but my goal is to manage to match this failing authentication logs on 4-5 lines : 

```
27307: YTUTUyt tyututut
27313: azeaeaeae
27369: handle_udp_packet: New UDP endpoint: local addr 1.2.3.4:3478, remote addr 9.9.9.9:51622
27369: handle_turn_command: STUN method 0x1 ignored
27369: session 001000000000000011: realm <hostname.online> user <>: incoming packet message processed, error 401: Unauthorized
27369: check_stun_auth: Cannot find credentials of user <1595978236:test_user>
27369: session 001000000000000011: realm <hostname.online> user <1595978236:test_user>: incoming packet message processed, error 401: Unauthorized
28402: ZAEAE
```

### The issue / Steps to reproduce
:

So I used this command to test : 

`fail2ban-regex logfile.txt regex.txt -l debug`

Got this error message 

```
W: Found a match for '27369: handle_udp_packet: New UDP endpoint: local addr 1.2.3.4:78, remote addr 9.9.9.9:51622\n27369: handle_turn_command: STUN method 0x1 ignored\n27369: session 001000000000000011: realm <hostname.online> user <>: incoming packet message processed, error 401: Unauthorized\n27369: check_stun_auth: Cannot find credentials of user <1595978236:test_user>\n27369: session 001000000000000011: realm <hostname.online> user <1595978236:test_user>: incoming packet message processed, error 401: Unauthorized'
but no valid date/time found for '27369: session 001000000000000011: realm <hostname.online> user <1595978236:test_user>: incoming packet message processed, error 401: Unauthorized'.
Please try setting a custom date pattern (see man page jail.conf(5)).
If format is complex, please file a detailed issue on https://github.com/fail2ban/fail2ban/issues in order to get support for this format.
```
After several days of research I concluded that it was because there is no timestamp recognized by fail2ban in each line of the log.

The first numbers are the "timestamp" but it's not a real timestamp.

Is there any trick to bypass this condition or a workaround ?

My regex file : 

```
[Init]
maxlines = 4

[Definition]

[...]

PR #2814, opened 2020-08-20
extended datepattern handling (TZ issues, no datepattern, etc)

sebres:
In the last time there were to many time zone issues, so we need to do something here.
This PR trying to solve it by better handling with wrong timestamp as well as some warning signaling user about invalid pattern or zone.

It also contains:

* code review, fix simplest TZ issue - avoid date adjustment by assuming of last year (date without year in the future) by wrong zone (don't adjust by offset up to +24 hours)
* filter: implement mode `inOperation`, which gets activated if filter starts processing of new messages; 
  in this mode a timestamp read from log-line that appeared recently (not an old line), deviating too much from now (up too 24h), will be considered as now (assuming a timezone issue), so could avoid unexpected bypass of failure (previously exceeding `findtime`);
* better interaction with non-matching optional datepattern or invalid timestamps;
* implements special datepattern `{NONE}` - allow to find failures totally without date-time in log messages (filter uses now as timestamp)
closes gh-2802

### DIFF
--- a/fail2ban/server/filter.py
+++ b/fail2ban/server/filter.py
@@ -633,6 +633,9 @@ class Filter(JailThread):
 				fail = element[3]
 				logSys.debug("Processing line with time:%s and ip:%s", 
 						unixTime, ip)
+				# ensure the time is not in the future, e. g. by some estimated (assumed) time:
+				if self.checkFindTime and unixTime > MyTime.time():
+					unixTime = MyTime.time()
 				tick = FailTicket(ip, unixTime, data=fail)
 				if self._inIgnoreIPList(ip, tick):
 					continue
@@ -936,7 +939,7 @@ class FileFilter(Filter):
 					log.setPos(lastpos)
 			self.__logs[path] = log
 			logSys.info("Added logfile: %r (pos = %s, hash = %s)" , path, log.getPos(), log.getHash())
-			if autoSeek:
+			if autoSeek and not tail:
 				self.__autoSeek[path] = autoSeek
 			self._addLogPath(path)			# backend specific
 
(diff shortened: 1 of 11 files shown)


xtrmbuster:
I hope it will be merged soon, thanks for your work !

fail2ban/server/filter.py
```diff
@@ -606,22 +618,59 @@ def processLine(self, line, date=None):
...
+				noDate = True
+				if self.__lastDate and self.__lastDate > MyTime.time() - 60:
+					tupleLine = ("", self.__lastTimeText, line)
+					date = self.__lastDate
+		
+		if self.checkFindTime:
+			# if in operation (modifications have been really found):
+			if self.inOperation:
+				# if weird date - we'd simulate now for timeing issue (too large deviation from now):
+				if (date is None or date < MyTime.time() - 60 or date > MyTime.time() + 60):
```
keestux, line 649: Forgive my late comment. What unit does this 60 represent? Seconds? If yes, then how can 60 seconds have anything to do with timezone?

On Ubuntu 22.04 we now have fail2ban 0.11.2 (with this change), and suddenly these warnings pop up. They pop up if a message is a little over one minute later than expected.

sebres: > What unit does this 60 represent? Seconds? 

Yes.

> If yes, then how can 60 seconds have anything to do with timezone?

It doesn't, it has something with too large deviation (so fail2ban in operation mode starts to use NOW instead of read datepattern in order to handle failure and banning intervals as correct as possible). Such a long delay may be a TZ issue (and more that 90% of users had that), but this warning is rather a signal that something is not alright with the jail/filter/service.

As for confusing warnings - see #2882, and it was changed in #3053.

keestux: Thanks for the explanation

PR #2703, opened 2020-04-25
0.10 / ipset timeout removal

sebres:
Internal handling of timeout in ipset (earlier expiration of ipset entries by timeouts, if it exceeded by time jumps (correction), retarded unban in fail2ban like delays in busy state, etc) is too error-prone and can lead to discrepancy by banned state in fail2ban and ipset (like `already banned` by repeated intrusion attempts after unban in ipset but not in fail2ban's ban-manager).

Candidates to be affected #2670, #2699 (and possibly some others, like #2666).

Several ipset actions fixed (no timeout per default anymore), so no discrepancy between ipset and fail2ban (removal from ipset will be managed by fail2ban only):
- action.d/firewallcmd-ipset.conf
- action.d/iptables-ipset-proto6-allports.conf
- action.d/iptables-ipset-proto6.conf
- action.d/shorewall-ipset-proto6.conf

This should avoid such discrepancy. 
To restore previous handling, parameter `timeout` can be set to `<bantime>` or predefined value `<timeout-bantime>` considering max possible timeout of ipset (2147483 seconds), like here:
```ini
banaction = %(default/banaction)s[timeout='<timeout-bantime>']
```
Also removes a need to specify space `<sp>` by setting of `familyopt` parameter (closes #2579)

PR #2860, opened 2020-10-25
Add action for mikrotik routerOS

a16bitsysop:

- [ X] **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement
- [X ] **CONSIDER adding a unit test** if your PR resolves an issue
- [ X] **LIST ISSUES** this PR resolves
- [ X] **MAKE SURE** this PR doesn't break existing tests
- [ X] **KEEP PR small** so it could be easily reviewed.
- [ X] **AVOID** making unnecessary stylistic changes in unrelated code
- [X ] **ACCOMPANY** each new `failregex` for filter `X` with sample log lines

### DIFF
--- a/config/action.d/mikrotik.conf
+++ b/config/action.d/mikrotik.conf
@@ -4,7 +4,7 @@
 #
 # Author: Duncan Bellamy <dunk@denkimushi.com>
 # based on forum.mikrotik.com post by pakjebakmeel
-#
+##
 # in the instructions:
 # (10.0.0.1 is ip of mikrotik router)
 # (10.0.0.2 is ip of fail2ban machine)
@@ -46,13 +46,15 @@
 
 actionstart =
 
-actionstop =
+actionstop = %(actionflush)s
+
+actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment~\"%(startcomment)s*\"]"
 
 actioncheck =
 
 actionban = %(mtikcommand)s "/ip firewall address-list add list=\"%(mtiklistname)s\" address=<ip> comment=%(mtikcomment)s"
 
-actionunban = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment=%(mtikcomment)s]"
+actionunban = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" address=<ip>]"
 
 mtikcommand = ssh -l %(mtikuser)s -p%(mtikport)s -i %(mtikkeyfile)s %(mtikhost)s
 
@@ -71,9 +73,12 @@ mtikhost =
 # Option: mtiklistname
 # Notes.: name of "address-list" to use on router
 mtiklistname = Auto Fail2Ban
+# Option: startcomment
+# Notes.: used as a prefix to all comments, and used to match for flushing rules
+startcomment = AutoF2B
 # Option: mtikcomment
 # Notes.: comment to use on routerOS (must be unique as used for ip address removal)
-mtikcomment = AutoF2B-<name>-<ip>
+mtikcomment = %(startcomment)s-<name>-<ip>
 
 [Init]
 name="%(__name__)s"
--- a/config/action.d/mikrotik.conf
+++ b/config/action.d/mikrotik.conf
@@ -48,7 +48,7 @@ actionstart =
 
 actionstop = %(actionflush)s
 
-actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment~\"%(startcomment)s*\"]"
+actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment~\"%(startcomment)s-*\"]"
 
 actioncheck =
 
--- a/config/action.d/mikrotik.conf
+++ b/config/action.d/mikrotik.conf
@@ -75,10 +75,10 @@ mtikhost =
 mtiklistname = Auto Fail2Ban
 # Option: startcomment
 # Notes.: used as a prefix to all comments, and used to match for flushing rules
-startcomment = AutoF2B
+startcomment = f2b-<name>
 # Option: mtikcomment
 # Notes.: comment to use on routerOS (must be unique as used for ip address removal)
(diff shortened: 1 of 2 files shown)


Omniflux:
I don't think this line is correct
`actionunban = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment=%(mtikcomment)s]"`

Did you mean
`actionunban = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" address=<ip>]"`

For actionban, you probably want to add a timeout
`actionban = %(mtikcommand)s "/ip firewall address-list add list=\"%(mtiklistname)s\" address=<ip> comment=%(mtikcomment)s timeout=<bantime>"`

In my experience, if you have many bans, because of the setup time for the ssh connection, this causes fail2ban to take a long time when restarting or stopping because each unban requires its own connection. You probably want to add the following lines

`actionstop = %(actionflush)s`
`actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\"]"`

Omniflux:
I just realized you were using the comment to track the address, although it is not obvious why, so the first part of my previous comment may be incorrect.

a16bitsysop:
Yes that is correct, it uses comment to track.  

Then you can have different jails add to the same address list if needed, but I guess doing by address would be faster with alot of bans and different address lists could be used for different jails.  Mikrotik/Router OS is supposed to be fast at finding addresses in address lists.

sebres:
> For actionban, you probably want to add a timeout

I would not do that for several reasons, for instance to hold the time of end of ban under fail2ban control only - if, no matter why, it would be unbanned in mikrotik earlier due to internal expiration, fail2ban would still consider the ticket as banned, in this time it would ignore the next ban (as "already banned") next time a recidive intruder makes a failures again. See #2703 for more info about similar issues.

> In my experience, if you have many bans, because of the setup time for the ssh connection, this causes fail2ban to take a long time when restarting or stopping because each unban requires its own connection.

Newer version of fail2ban has a command `actionflush` to make a "bulk" unban of whole jail. This could be specified, if mikrotik has such possibility (e. g. remove all entries with comment matching `AutoF2B-<name>-*`).

a16bitsysop:
It should be possible to do the action flush with matching comment wild card, I will edit rule and test it out

config/action.d/mikrotik.conf
```diff
@@ -0,0 +1,84 @@
...
+# mhost = 192.168.0.1
+# mlistname = BAD LIST
+
+[Definition]
+
+actionstart =
+
+actionstop = %(actionflush)s
+
+actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment~\"%(startcomment)s*\"]"
```
sebres, line 51: hmm... since `startcomment` and `mtiklistname` don't contain jail name (`<name>`), wouldn't this flush in any lists, regardless the jail? So if some jail gets stopped (restarted), it removes all tickets of other jails.

Shall not it be either:
```diff
-actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment~\"%(startcomment)s*\"]"
+actionflush = %(mtikcommand)s "/ip firewall address-list remove [find list=\"%(mtiklistname)s\" comment~\"%(startcomment)s-<name>-*\"]"
```
or the startcomment get a jail name:
```diff
-startcomment = AutoF2B
+startcomment = f2b-<name>
-mtikcomment = %(startcomment)s-<name>-<ip>
+mtikcomment = %(startcomment)s-<ip>
```

sebres (changes requested):

config/action.d/mikrotik.conf
a16bitsysop: Yes that looks a better way, will change to method 2

sebres (approved):

vitkabele:
Is there any progress on this?

sebres:
> Is there any progress on this?

Progress on what exactly? Can't you use the action?

I can surely merge it, just after all one has to wait for new release anyway.
As for the quality or the functioning I can't say anything since I'm not familiar with that approach, so I was hopping someone else would tell own expertise about.

vitkabele:
Progress on merging to release. It also seems to have merge conflicts now. I'll test the action and report back whether it works for me.

vitkabele:
It works for me, but:

1. I'd change the default parameter for `mtiklistname` to something without a space, e.g. `fail2ban`.

2. The `mtik` prefix from the parameter names could be removed (?). I am not sure what is the scope of action parameter names, however other actions seem to define parameters as if it was private scope. 

3. The `mtiklistname` should imo be just `list`.

I can't however push to the source branch, so it's up to @a16bitsysop if he wants to implement these changes.

a16bitsysop:
sounds okay to me

config/action.d/mikrotik.conf
```diff
@@ -0,0 +1,84 @@
...
+
+actionflush = %(command)s "/ip firewall address-list remove [find list=\"%(list)s\" comment~\"%(startcomment)s-*\"]"
+
+actioncheck =
+
+actionban = %(command)s "/ip firewall address-list add list=\"%(list)s\" address=<ip> comment=%(comment)s"
+
+actionunban = %(command)s "/ip firewall address-list remove [find list=\"%(list)s\" comment=%(comment)s]"
+
+mtikcommand = ssh -l %(user)s -p%(port)s -i %(keyfile)s %(host)s
```
sebres, line 59: The last adjustment made the action broken - the command parameter used in action-tags called `command`, but here it is `mtikcommand`.

config/action.d/mikrotik.conf
```diff
@@ -0,0 +1,84 @@
...
+
+# Option: user
+# Notes.: username to use when connecting to routerOS
+mtikuser =
```
sebres, line 63: same here and other following params (in command they are used without `mtik` prefix)

sebres (changes requested):
please repair

a16bitsysop:
missed those

vitkabele:
Works for me with the changes 👍

vitkabele:
@sebres Hello, it seems that the requested changes are already fixed

sebres (approved):

Issue #2548, opened 2019-10-18
[apache-auth] false alarm with Apache2 mod_evasive module


borzaka:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): `Fail2Ban v0.10.2`
- OS, including release name/version:
`Debian GNU/Linux 10 (buster)`
`Linux hostname 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u1 (2019-09-20) x86_64 GNU/Linux`
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

### The issue:

Fail2Ban `apache-auth` filter is counting `mod_evasive` errors as fails.

#### Steps to reproduce

Install [mod_evasive](https://blog.rapid7.com/2017/04/09/how-to-configure-modevasive-with-apache-on-ubuntu-linux/)

#### Expected behavior

No bans on `mod_evasive` logs.

#### Observed behavior

Banning on `mod_evasive` attacks 8 (DOS).

#### Any additional information

Fix for the `mod_evasive` "client denied by server configuration" error:
https://parkroad.co.za/identifying-and-fixing-modevasive-client-denied-server-configuration-errors

(basically you need to edit your settings to make is less aggressive)

### Configuration, dump and another helpful excerpts

```
# fail2ban-client -d | grep 'apache-auth'
 No filter set for jail monitorix-auth
['add', 'apache-auth', 'auto']
['set', 'apache-auth', 'prefregex', '^\\[\\]\\s\\[(:?error|\\S+:\\S+)\\]( \\[pid \\d+(:\\S+ \\d+)?\\])? \\[client <HOST>(:\\d{1,5})?\\] (?:AH\\d+: )?<F-CONTENT>.+</F-CONTENT>$']
[...]

PR #2875, opened 2020-11-20
apache24 with evasive24

metacreo:
[Fri Nov 20 17:13:45.607707 2020] [evasive24:error] [pid 63747:tid 34386949120] [client .....
Server version: Apache/2.4.46 (FreeBSD)
fail2ban-server -V 0.11.1

### DIFF
--- a/config/filter.d/apache-auth.conf
+++ b/config/filter.d/apache-auth.conf
@@ -13,7 +13,7 @@ before = apache-common.conf
 mode = normal
 
 # ignore messages of mod_evasive module:
-apache-pref-ign-normal = (?!evasive)
+apache-pref-ign-normal = [?!evasive]
 # allow "denied by server configuration" from all modules:
 apache-pref-ign-aggressive =
 # mode related ignore prefix for common _apache_error_client substitution:


sebres:
This is wrong strategy: `(?!evasive)` is a negative lookahead (matches nothing, but should not find `evasive` after previous RE, `[?!evasive]` is a character group match, so should match exactly 1 char from `? ! e v a s i v`.

Set it to empty `apache-pref-ign-normal = ` would do the job you probably expecting. But take a look in #2548 why this was introduced.
Especially https://github.com/fail2ban/fail2ban/issues/2548#issuecomment-543629503 how you can switch behavior with `mode = aggressive`:
```ini
[apache-auth]
# detect mod_evasive (DDoS, brute force) errors as failures:
mode = aggressive
```
This is also covered in our test-suite:

Mode normal:
https://github.com/fail2ban/fail2ban/blob/bb0f732ae69894b22306dd7efa213513e3acd8a2/fail2ban/tests/files/logs/apache-auth#L137-L138

Mode aggressive:
https://github.com/fail2ban/fail2ban/blob/bb0f732ae69894b22306dd7efa213513e3acd8a2/fail2ban/tests/files/logs/apache-auth#L140-L143

As you can see it doesn't match in first and matches in the second test.

So let alone it is not correct implemented, you don't need this change at all - it is really enough to set mode of apache-auth jail to `aggressive`.
Or even to behave the mode (normal), specify parameter `apache-pref-ign-normal` as empty in the filter or directly in jail:
```ini
[apache-auth]
filter = apache-auth[mode=normal, apache-pref-ign-normal='']
```

metacreo:
ok. sorry. I just set mode = aggressive in filter.d/apache-auth and it not woks for me
later I change back filter.d file and set mode in my jail.d/apache.conf and all wok fine
thank you

sebres:
> I just set `mode = aggressive` in filter.d/apache-auth and it not woks for me

Sure, because jail.conf default declaration for filter (in section DEFAULT) contains https://github.com/fail2ban/fail2ban/blob/ffd6b9f6de3b94a255d145017ed6efb3a5c9e79c/config/jail.conf#L120-L126
Which means the option you set in filter will be overwritten with value specified in jail (which is still `normal` per default).

PR #2881, opened 2020-11-25
add nginx-tls-downgrade

stepodev:


sebres:
regex rewritten more precise and conclusive (without catch-all's, etc)...

I am still unsure regarding the jail/filter name:
- the same messages could be produces by port-scanner, wrong protocol, etc;
- filter could be extended later to consider another "attack" vectors

As an idea, maybe simply extend filter `nginx-http-auth` with parameter `mode`, so `aggressive` could target this (and later other) concerns?

stepodev:
> regex rewritten more precise and conclusive (without catch-all's, etc)...
> 
> I am still unsure regarding the jail/filter name:
> 
>     * the same messages could be produced by port-scanner, wrong protocol, etc;
> 
>     * filter could be extended later to consider another "attack" vectors
> 
> 
> As an idea, maybe simply extend filter `nginx-http-auth` with parameter `mode`, so `aggressive` could target this (and later other) concerns?

Thanks for the input. You're right, this seems cleaner. Any other issues?

PR #2898, opened 2020-12-24
Handle postscreen's PREGREET and HANGUP messages

defanor:
Provoking those seems to be a popular activity among spammers.

### DIFF
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -37,7 +37,9 @@ mdre-rbl  = ^RCPT from [^[]*\[<HOST>\]%(_port)s: [45]54 [45]\.7\.1 Service unava
 mdpr-more = %(mdpr-normal)s
 mdre-more = %(mdre-normal)s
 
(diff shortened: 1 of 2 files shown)


sebres:
Isn't this rather something from matter of protection against overload and SPAM defense?
If so I guess it should extend either mode `ddos` or `aggressive` (so belong to related `mdpr-*` then)...

defanor:
Perhaps it is, but not sure if other checks in the "normal" mode are
quite different in that; there is a test case with postscreen already,
handled by a normal mode's filter.

Though I don't have any arguments for its inclusion into the "normal"
mode in particular (I'm using the "aggressive" mode personally, so it
would work either way for me), so if you think it should be in a
different mode, I'll move it there -- just let me know which one would
be more appropriate.

sebres:
let's do it... also don't forget to move the tests to the penultimate section, e. g. after:
https://github.com/fail2ban/fail2ban/blob/696acda8bdc9abd4eca3efb3e306df1ec6b3fdd8/fail2ban/tests/files/logs/postfix#L148

sebres (approved):

sebres (approved):

sebres:
merged, thx.

PR #2913, opened 2021-01-18
Add transport to asterisk RE

brianjmurrell:
Call rejection messages from Asterisk can have the transport prefixed to the IP address.

Signed-off-by: Brian J. Murrell <brian@interlinx.bc.ca>

### DIFF
--- a/fail2ban/tests/files/logs/asterisk
+++ b/fail2ban/tests/files/logs/asterisk
@@ -19,6 +19,8 @@
 [2012-02-13 17:44:26] NOTICE[1638] chan_iax2.c: Host 1.2.3.4 failed MD5 authentication for 'Fail2ban' (e7df7cd2ca07f4f1ab415d457a6e1c13 != 53ac4bc41ee4ec77888ed4aa50677247)
 # failJSON: { "time": "2013-02-05T23:44:42", "match": true , "host": "1.2.3.4" }
 [2013-02-05 23:44:42] NOTICE[436][C-00000fa9] chan_sip.c: Call from '' (1.2.3.4:10836) to extension '0972598285108' rejected because extension not found in context 'default'.
+# failJSON: { "time": "Jan 18 17:39:50", "match": true , "host": "1.2.3.4" }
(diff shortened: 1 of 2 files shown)


brianjmurrell:
CI tests seem to fail with:

```
ValueError: time data 'Jan 18 17:39:50' does not match format '%Y-%m-%dT%H:%M:%S.%f'
540
```

But the date format I put in the example line is that of my Asterisk server.  Am I supposed to contort that into the date format of the fail2ban/tests/files/logs/asterisk file, or is it a bug that it's not recognizing the date format of my Asterisk logging?

The change as it is here is working in my local installation, using the date format from my Asterisk log.

fail2ban/tests/files/logs/asterisk
```diff
@@ -19,6 +19,8 @@
 [2012-02-13 17:44:26] NOTICE[1638] chan_iax2.c: Host 1.2.3.4 failed MD5 authentication for 'Fail2ban' (e7df7cd2ca07f4f1ab415d457a6e1c13 != 53ac4bc41ee4ec77888ed4aa50677247)
 # failJSON: { "time": "2013-02-05T23:44:42", "match": true , "host": "1.2.3.4" }
 [2013-02-05 23:44:42] NOTICE[436][C-00000fa9] chan_sip.c: Call from '' (1.2.3.4:10836) to extension '0972598285108' rejected because extension not found in context 'default'.
+# failJSON: { "time": "Jan 18 17:39:50", "match": true , "host": "1.2.3.4" }
```
sebres, line 22: Test expects time in ISO format here, so it would be `2005-01-18T17:39:50` for `Jan 18 17:39:50`;
2005 is because year is not specified in timestamp and test suite runs in 1124013600 (So Aug 14 12:00:00 CEST 2005).
If it is not significant for you just change the timestamp in test line to `2013-02-05 23:44:43` and in json to `2013-02-05T23:44:42`, so the filter gets the year and would not switch `datepattern` during the test run.

brianjmurrell: OK.  I wasn't sure if the test was supposed to accept all of the date formats that fail2ban accepts.

I will update the date in my PR to the required format.

sebres (approved):

brianjmurrell:
Can this be landed and closed?

sebres:
Sure. Done.

Issue #2920, opened 2021-01-30
xarf-login-attack should not send emails to LACNIC


K2IE:
The [documentation for the abuseix db]([https://docs.abusix.com/abuse-contact-db/contactdb%2Foverview status]) states:

Warning

Please note that LACNIC is the only RIR that does not provide this data, so for LACNIC ranges, the abuse contact abuse@lacnic.net is returned for all queries as we are unable to provide more specific data due to this restriction.

Therefore, no xarf complaints need be sent to LACNIC.  The do in fact reply to each email sent that they can do nothing.

  ```
  LACNIC is an RIR (Regional Internet Registry) for Latin America and
  the Caribbean region. What that means is that LACNIC is responsible for
  the IP address space and ASN allocation/assignment in this region.
  
  As mentioned, the IP address in question was allocated by LACNIC to
  some other organization or ISP in the region. So the abuse activity
  originated in that organization's network, not in LACNIC.
  
  You should query our whois database to get information about the
  source of this abuse activity and the appropriate network contact.
```
 Since there could be others in the future, perhaps we need a way for the call to xarf-login-attack to pass a list of addresses not to bother with should they be returned.

sebres:
FWIW, blacklisting LACNIC (or other regional registries) by its mail address is in principle a workaround (not against PR, just a note).
I guess to exclude registrars at all the proper way would be to consider size of subnet in order to find near responsible or to bypass sections with `OrgTechRef` matching `https://rdap.arin.net/registry/entity/*` (if more than 1 section given).

PR #2924, opened 2021-02-01
Add bcc and dontsend capabilities to xarf-login-attack action

### LINKED ISSUE #2919: Suggestion: bcc parameter for xarf-login-attack
A bcc is useful to both debug as well as monitor activity of the emails sent by xarf-login-attack.  Currently I modify xarf-login-attack.conf to add my email address to the email generated, ie:

            if [ ! -z "$ADDRESSES" ]; then
                oifs=${IFS}; IFS=,; ADDRESSES=$(echo $ADDRESSES)
                IFS=${oifs}
                (printf -- %%b "<header>\n<message>\n<report>\n\n";
                 date '+Note: Local timezone is %%z (%%Z)';
                 printf -- %%b "\n<ipmatches>\n\n<footer>") | <mailcmd> <mailargs> $ADDRESSES,**_myemail@myhost.com_**
            fi

This works fine, but a more elegant solution would be a way to pass this as a parameter from jail.local and to insert it when supplied.  This could look something like:

[sshd]
enabled = true
mode = aggressive
mailargs = -FAbuse -fabuse@myhost.com
**_bcc = my@mail.box_**
action = %(action_)s
         xarf-login-attack-local[mailargs=%(mailargs)s,**_bcc=%(bcc)s_**]

Thanks for considering.

K2IE:
This PR provides the ability to specify an optional bcc for the generated abuse emails.  It also provides the ability to provide an array of abuse addresses for which no email should be generated.  This solves the problem of published abuse addresses which bounce the mail back to the sender or, as in the case of lacnicnet, say that they cannot assist because the are a regional allocation registry.

The dontsend array is specified in jail.local as:

dontsend = ( 'one@address.com' 'two@address.com' 'three@address.com')

The bcc is specified in jail.local as:

bcc = someaddress@myhost.com

Then xarf-login-attack is called as usual, with either of the new ptional oparameters passed as desired:

action = %(action_)s
         xarf-login-attack-local[mailargs=%(mailargs)s,mailcmd=%(mailcmd)s,bcc=%(bcc)s,dontsend=%(dontsend)s]

I also removed a redundant IFS operation on $ADDRESSES.

Resolves #2919 and #2920 .

### DIFF
--- a/config/action.d/xarf-login-attack.conf
+++ b/config/action.d/xarf-login-attack.conf
@@ -49,6 +49,8 @@ actionban = oifs=${IFS};
             IFS=${oifs}
             IP=<ip>
             FROM=<sender>
+            BCC=<bcc>
+            DONTSEND=<dontsend>
             SERVICE=<service>
             FAILURES=<failures>
             REPORTID=<time>@<fq-hostname>
@@ -56,11 +58,16 @@ actionban = oifs=${IFS};
             PORT=<port>
             DATE=`LC_ALL=C date --date=@<time> +"%%a, %%d %%h %%Y %%T %%z"`
             if [ ! -z "$ADDRESSES" ]; then
-                oifs=${IFS}; IFS=,; ADDRESSES=$(echo $ADDRESSES)
-                IFS=${oifs}
+                if [ ! -z "$DONTSEND" ]; then
+                    for ckaddr in "${DONTSEND[@]}"; do
+                        if [[ "$ADDRESSES" == *"$ckaddr"* ]]; then
+                           exit 0   
+                        fi
+                    done
+                 fi
                 (printf -- %%b "<header>\n<message>\n<report>\n\n";
                  date '+Note: Local timezone is %%z (%%Z)';
-                 printf -- %%b "\n<ipmatches>\n\n<footer>") | <mailcmd> <mailargs> $ADDRESSES
+                 printf -- %%b "\n<ipmatches>\n\n<footer>") | <mailcmd> <mailargs> $ADDRESSES $BCC
             fi
 
 actionunban =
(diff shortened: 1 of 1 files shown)


sebres:
Just by the way, please note an amendment to PR description:
```diff
-xarf-login-attack-local[mailargs=%(mailargs)s,mailcmd=%(mailcmd)s,bcc=%(bcc)s,dontsend=%(dontsend)s]
+xarf-login-attack-local[mailargs=%(mailargs)s,mailcmd=%(mailcmd)s,bcc="%(bcc)s",dontsend="%(dontsend)s"]
```
without quotes it would supply wrong (truncated) arguments to the action.

sebres:
> I also removed a redundant IFS operation on $ADDRESSES.

Why it should be redundant (now)? I don't see any code piece that could solve #2372, which could occur again if you would remove this "redundant" operation.

K2IE:
@sebres thanks for your comment.  I have tested extensively without the suggested double-quotes and have no experienced truncation of the arguments, verified by dumping the received arguments to a tempfile for debugging.  Can you please point me in the direction what I'm failing to understand?

K2IE:
> 
> 
> > I also removed a redundant IFS operation on $ADDRESSES.
> 
> Why it should be redundant (now)? I don't see any code piece that could solve #2372, which could occur again if you would remove this "redundant" operation.

I was unaware of #2372.  I will re-add that operation.

sebres:
> Can you please point me in the direction what I'm failing to understand?

```
$ python -c 'from fail2ban.helpers import extractOptions; an, ao = extractOptions("""some-action[a="x, y, z", b=x, y, z]"""); print(ao)'
{'a': 'x, y, z', 'b': 'x'}
```
as you can see `a` contains whole string as enclosed in quotes `x, y, z`, whereas `b` contains only `x`.
As an option value (without quote) only single "word" is allowed (and it cannot contains comma, space or `[]` characters).

As for "not experienced truncation", it is a matter of sequence, so how (or rather when exactly) `%(dontsend)s` gets substituted 
and I'm pretty sure this gets interpolated before it is supplied as parameter to the action in the jail, so it becomes:
```ini
- ..., dontsend=%(dontsend)s
+..., dontsend=( 'one@address.com' 'two@address.com' 'three@address.com')
```

But... trying this:
```
$ python -c 'from fail2ban.helpers import extractOptions; an, ao = extractOptions("""some-action[a=(one@address.com two@address.com three@address.com)]"""); print(ao)'
{'a': '(one@address.com two@address.com three@address.com)'}
```
I see it is indeed correct, so space (alone) is not a separator anymore at least in last version (I changed this extractOptions several times).

Issue #2378, opened 2019-03-22
Apache 404 errors are no longer in Apache 2.4.x error_log


kyz:
Since Apache 2.4, most people's Apache error_log files no longer have any `File does not exist: filename` lines.

[Apache 2.4 release notes](https://www.apache.org/dist/httpd/CHANGES_2.4): "Downgrade error log messages which accompany a 404 request status from loglevel error to info. [PR 35768](https://bz.apache.org/bugzilla/show_bug.cgi?id=35768)"

Apache can be configured with to show them again with `LogLevel INFO` but that brings in a lot of extra messages to error_log that might be unwanted, and it's also not the default config. By default, the only place these errors will appear now is in `access_log` as 404 errors.

This affects the following filters, all of which look at Apache error log for "does not exist" lines:
* apache-botsearch.conf
* apache-nohome.conf
* apache-noscript.conf

What do you recommend to get these filters matching again, by default?

sebres:
> PR 35768

Hmm... this looks like it is merged (and released) long time ago. How all the years this filters work by all the people? *Nobody uses apache nowadays?* :smile: 

> What do you recommend to get these filters matching again, by default?

Shortly - dunno. 
I would definitely not recommend do a monitoring of the access-log (due to possibly too large "parasite" log-traffic in sense of fail2ban).

I don't use apache (why at all if there is a nginx;), so I don't know how exactly it would be possibly in apache, but I'd recommend to make some customization in order to generate the messages in separate log-file (or journal). For example in nginx it would be something like this - https://github.com/fail2ban/fail2ban/issues/1962#issuecomment-343856328

So we are open for suggestions.

0x6772:
Todd Vierling (who I think I remember works on Apache? Maybe via Oracle? This guy: https://stackoverflow.com/users/2644135/todd-vierling) outlined pretty clearly how Apache thinks this ought to work these days in the thread starting here: https://www.jwz.org/blog/2019/03/apache-2-4-1-killed-fail2ban-so-thats-awesome/#comment-197663

sebres:
Well as already said I find the way to observe the default logs as not proper (no matter error or access log).
But changing of a default behavior may be also not good practice too.
So I'm partially agree with both sides in the discussion. 

Is there any simple way to configure the log of apache for selected errors like 404/etc per application or common prefix or whatever is called location in apache? And if yes, which logging-format would get this log-messages (access-log I assume).

kyz:
[...]

PR #2939, opened 2021-02-16
Recognize 404 responses for Apache 2.4.x

jurgenhaas:
Solves #2378

### DIFF
--- a/config/filter.d/apache-botsearch.conf
+++ b/config/filter.d/apache-botsearch.conf
@@ -23,10 +23,7 @@ before = apache-common.conf
 
 [Definition]
 
-prefregex = ^%(_apache_error_client)s (?:AH\d+: )?<F-CONTENT>.+</F-CONTENT>$
-
-failregex = ^(?:File does not exist|script not found or unable to stat): <webroot><block>(, referer: \S+)?\s*$
-            ^script '<webroot><block>' not found or unable to stat(, referer: \S+)?\s*$
+failregex = ^"<ADDR>".+ HTTP/\d\.\d" 404
 
 ignoreregex = 
 


config/filter.d/apache-botsearch.conf
```diff
@@ -23,10 +23,7 @@ before = apache-common.conf
 
 [Definition]
 
-prefregex = ^%(_apache_error_client)s (?:AH\d+: )?<F-CONTENT>.+</F-CONTENT>$
-
-failregex = ^(?:File does not exist|script not found or unable to stat): <webroot><block>(, referer: \S+)?\s*$
-            ^script '<webroot><block>' not found or unable to stat(, referer: \S+)?\s*$
+failregex = ^"<ADDR>".+ HTTP/\d\.\d" 404
```
sebres, line 26: 1. this RE is vulnerable due to catch-all (`.+`), we don't accept filters with unjustified catch-all's anymore (see also https://github.com/fail2ban/fail2ban/pull/1961#pullrequestreview-75940560 for similar PR);
2. this filter does not work so: it searching for certain bots only (declared in include botsearch-common.conf as `block`) for certain `webroot`; both parameters can be also set by user (supplied to filter in jail).
Your variant simply blocks 404 response (regardless the URI, bot-list and root), so I guess it is targeting totally different kind of purposes. I'm not against to make it belonging to this filter (but then not by default, but rather within some mode `aggressive`). 
Anyway to solve #2378 one must change log-format, not the functionality of the filter.

sebres (changes requested):

config/filter.d/apache-botsearch.conf
jurgenhaas: Sorry for my late reply, had been knocked down by some customer projects. Still would like to get this sorted and I'd rather love to get the original filter being adjusted properly for the new Apache log format which was described in #2378

Now, with your comment above I'm not even sure anymore, if we're trying to tackle the correct problem here.

Trying to find out, let me show you what we have in the logs.

**In the error log we have lots of these:**

```
[Wed Mar 17 05:27:05.098401 2021] [error] [pid 9550] AH01071: Got error 'Primary script unknown'
[Wed Mar 17 05:27:05.303839 2021] [error] [pid 9550] AH01071: Got error 'Primary script unknown'
[Wed Mar 17 05:27:05.509085 2021] [error] [pid 9550] AH01071: Got error 'Primary script unknown'
[Wed Mar 17 05:27:05.714101 2021] [error] [pid 9550] AH01071: Got error 'Primary script unknown'
[Wed Mar 17 05:27:05.921099 2021] [error] [pid 9550] AH01071: Got error 'Primary script unknown'
```

Those are triggered by PHP FPM for all requests to PHP files that don't exist. Therefore, each of the above lines has a matching entry in the access log looking like this:

```
"1.9.204.148" - - [17/Mar/2021:05:27:05 +0100] "POST /q.php HTTP/1.1" 404 229 "-" "Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
"1.9.204.148" - - [17/Mar/2021:05:27:05 +0100] "POST /pe.php HTTP/1.1" 404 229 "-" "Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
"1.9.204.148" - - [17/Mar/2021:05:27:05 +0100] "POST /hm.php HTTP/1.1" 404 229 "-" "Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
"1.9.204.148" - - [17/Mar/2021:05:27:05 +0100] "POST /sz.php HTTP/1.1" 404 229 "-" "Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
"1.9.204.148" - - [17/Mar/2021:05:27:05 +0100] "POST /cainiao.php HTTP/1.1" 404 229 "-" "Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
```

My take now is that those requests have nothing to do what is prepared in botsearch-common.conf but they are obviously bot search attempts.

Is apache-botsearch.conf the right approach to block those? If yes, can you give me some hints on how to geth there? If no, what other approach would be correct - if any?

jurgenhaas: Had another look and it seems that apache-badbots is better suited for this task. I tested it with `fail2ban-regex -o row test-apache-access.log apache-badbots` but it doesn't pick up the line above and I can't figure out why.

sebres:
> Is apache-botsearch.conf the right approach to block those?

I already explained why it is not.

> Had another look and it seems that apache-badbots is better suited for this task.

Also this filter is not really suitable to do that, it would at least expect to adjust or extend `badbots` or `badbotscustom` parameters.
So both filters are intended to find failures from **certain** bots only (emphasis on certain) and therefore would not match by default the messages from your examples at all.
Neither I am a friend of such (default) filters, nor I am glad that we have so many confusing apache filters... but they are still there for "backwards compatibility" reasons solely.

>  If no, what other approach would be correct

Since error-log is better suitable than the access-log (see wiki [Best practice](https://github.com/fail2ban/fail2ban/wiki/Best-practice) why it is so) you need a filter matching `AH01071: Got error 'Primary script unknown'` either.
Although RFE #2088 (as well as a small amend 5e980afbb8fc83759e67ab9338b6f176121de708) seems to target exactly this kind of messages, but for some reason you don't have the IP address in this messages at all - is it some custom log-format?

As for the messages from access-log (which also looks a bit customized, because IP address enclosed in quotes), well although one could specify some custom list for the badbots filter (did not check whether the rest of the filter matching basically), but you could create your own failregex like this:
```regex
^\"<HOST>\" [^"]*\"[A-Z]{3,}\s+/(?!site/|cms/)\S+ HTTP/[^"]+" 40(?!1)\d\b
```
where the RE `/(?!site/|cms/)\S+` would match every URI excepting `/site/...` or `/cms/...`, but you can rewrite it with `/\S+\.php` which would match every php-request causing 40x error.
To match for example 404 and 403 only, change `40(?!1)\d` to `40[34]`, however `40(?!1)\d` would bypass 401 only, which must be handled separately, because normally you need to consider authentication "handshake" - phase where user name is not yet supplied by first unauthenticated request, resulting to 401 response (so must ignore `-` as an user name).

As a jail it would look like this:
```ini
[stop-bots]
logpath = /path/to/your/access-log
filter = 
_uri = /(?!site/|cms/)\S+
_code = 40(?!1)\d
failregex = ^\"<HOST>\" [^"]*\"[A-Z]{3,}\s+%(_uri)s HTTP/[^"]+" %(_code)s\b
```
Just it is not really recommended due to the reasons mentioned in the wiki as well as because it can theoretically cause some false positives (if your web-site potentially able to generate some 404th errors), see for example https://stackoverflow.com/a/66444182/7161854 for similar thing.

Thus I'll close this PR as incorrect now.

siggi90:
Would it be possible to somehow recognize multiple 404 log entries within a certain timeframe with Fail2Ban? Its probably possible to cover most cases with a regex, but it would also be nice to ban for example 10-20 404 errors within a minute.

sebres:
> Would it be possible to somehow recognize multiple 404 log entries within a certain timeframe with Fail2Ban? 

Yes, this time window called `findtime` and to specify max attempt count you have to set `maxretrty`. See [wiki :: How fail2ban works](https://github.com/fail2ban/fail2ban/wiki/How-fail2ban-works) for details.
```ini
maxretry = 10
findtime = 1m
```

siggi90:
Alright, thanks. One more question, I've tried both the regex suggested here by jurgenhaas and you and neither seem to get any matches when running fail2ban-regex /var/log/apache2/access.log /etc/fail2ban/filter.d/apache-noscript.conf and no i have it trimmed down to: failregex = <HOST>.*404.* which i think should definitely get a match but I still get zero matches. I have also tried with <ADDR> instead of <HOST> and my log files look like this:

45.146.164.125 - - [01/Jun/2021:10:30:11 +0000] "GET /?XDEBUG_SESSION_START=phpstorm HTTP/1.1" 200 22897 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"
45.146.164.125 - - [01/Jun/2021:10:30:14 +0000] "GET /wp-content/plugins/wp-file-manager/readme.txt HTTP/1.1" 404 5605 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"
45.146.164.125 - - [01/Jun/2021:10:30:16 +0000] "POST /mifs/.;/services/LogService HTTP/1.1" 404 5605 "https://46.22.109.91:443" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"
45.146.164.125 - - [01/Jun/2021:10:30:17 +0000] "GET /console/ HTTP/1.1" 404 5605 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"
45.146.164.125 - - [01/Jun/2021:10:30:17 +0000] "POST /Autodiscover/Autodiscover.xml HTTP/1.1" 404 5605 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"
45.146.164.125 - - [01/Jun/2021:10:30:19 +0000] "GET /_ignition/execute-solution HTTP/1.1" 404 5605 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"

Shouldn't i get matches? Is the HOST/ADDR format not correct in my log file?

siggi90:
I'm sorry the comment stripped out my tags
`failregex = <ADDR>.*404.*`

This is the failregex

sebres:
No idea what did you try exactly, I just removed quotes from afore suggested RE around the `<HOST>` tag and it works as expected:
<details><summary>result for <code>^<HOST> [^"]*\"[A-Z]{3,}\s+/(?!site/|cms/)\S+ HTTP/[^"]+" 40(?!1)\d\b</code></summary>

```
$ fail2ban-regex /tmp/log '^<HOST> [^"]*\"[A-Z]{3,}\s+/(?!site/|cms/)\S+ HTTP/[^"]+" 40(?!1)\d\b'

Running tests
=============

Use   failregex line : ^<HOST> [^"]*\"[A-Z]{3,}\s+/(?!site/|cms/)\S+ HTTP...
Use         log file : /tmp/log
Use         encoding : UTF-8


Results
=======

Failregex: 5 total
|   1) [5] ^<HOST> [^"]*\"[A-Z]{3,}\s+/(?!site/|cms/)\S+ HTTP/[^"]+" 40(?!1)\d\b
`-

Ignoreregex: 0 total

Date template hits:
|- [# of hits] date format
|  [6] Day(?P<_sep>[-/])MON(?P=_sep)ExYear[ :]?24hour:Minute:Second(?:\.Microseconds)?(?: Zone offset)?
`-

Lines: 6 lines, 0 ignored, 5 matched, 1 missed
[processed in 0.02 sec]

|- Missed line(s):
|  45.146.164.125 - - [01/Jun/2021:10:30:11 +0000] "GET /?XDEBUG_SESSION_START=phpstorm HTTP/1.1" 200 22897 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36"
`-
```
</details>

for fail2ban version >= 0.10, replace `<HOST>` with `<ADDR>` (if only IP-addresses expecting).

Regarding `<ADDR>.*404.*`, of course it must work too (as well as `<ADDR>.*404`), just the regex is too vulnerable.

siggi90:
For some reason it works as a separate rule but not when i edited noscript.conf, but everything is working now, thanks.

PR #2984, opened 2021-03-28
Zoneminder filter update

j-marz:
Support new log format, ERR instead of WAR.
Add detection of non-existent user login attempts.


sebres:
Done. Thx!

PR #2989, opened 2021-04-03
readme: Use table for code status

Daniel15:
Before:
image

After:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,11 @@ fail2ban(1) and jail.conf(5)  manpages for further references.
 Code status:
 ------------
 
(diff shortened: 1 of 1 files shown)


sebres:
Thx! Looks indeed better (at least on GH)... 
Just since Travis going periodically broken as well as restricted number of test attempts (e. g. I'm already "banned" there), I tend to adjust this completely to GH-Actions (flows for CI already added, just no coverage bound at the moment).
Will take a look here a bit later.

sebres:
There is only one CI status now (of GHA) - https://github.com/fail2ban/fail2ban#code-status
Thus close as outdated.

PR #2990, opened 2021-04-03
Add Debian roundcube_errors_log path

Daniel15:


config/paths-debian.conf
```diff
@@ -26,3 +26,5 @@ exim_main_log = /var/log/exim4/mainlog
 # was in debian squeezy but not in wheezy
 # /etc/proftpd/proftpd.conf (SystemLog)
 proftpd_log = /var/log/proftpd/proftpd.log
+
+roundcube_errors_log = /var/log/roundcube/errors.log
```
sebres, line 30: Looks like this is default path now (distribution irrelevant)? Doesn't it?

For instance here some hint (at least before https://github.com/roundcube/roundcubemail/issues/7009 it was still called errors) - https://github.com/roundcube/roundcubemail/blob/2321bce4694ee9f155284a7b8c55637c47672e79/INSTALL#L83

If so we'd rather need to change it here -
https://github.com/fail2ban/fail2ban/blob/8b6633842f07ef917e59024aee566af5a14501be/config/paths-common.conf#L93

Daniel15: This is somewhat distro-specific, but maybe it should just be changed in the `paths-common.conf` anyways. By default, the Roundcube logs directory is within the Roundcube app directory. So for example, if you install Roundcube at `/opt/roundcube/`, the logs will be at `/opt/roundcube/logs/`. Debian installs Roundcube to `/var/lib/roundcube/`, and `/var/lib/roundcube/logs/` is a symlink to `/var/log/roundcube`. I'm not sure if all distros do that.

Having said all that, it looks like they started adding the `.log` suffix in Roundcube 1.4 beta in 2017 (https://github.com/roundcube/roundcubemail/pull/313). It looks like Debian Buster still includes the old 1.3 (which doesn't have this option), however backports and Bullseye both include 1.4.

Is there a way to specify two file names? I guess it should use both `/var/log/roundcube/errors.log` and `/var/log/roundcube/errors` to handle both old and new versions of Roundcube.

sebres: yes, it is there:
```ini
roundcube_errors_log = /var/log/roundcube/errors
                       /var/log/roundcube/errors.log
```
unsure it is possible with wildcards in glob pattern (since we need to exclude rotated and zipped logs)

Daniel15:
@sebres Thanks for your feedback. I updated the PR to modify `paths-common.conf` instead 😄

Neustradamus:
Any progress on this PR?

PR #3191, opened 2021-12-24
Display last instead of first 1000 log messages

viaccoz:
With the previous configuration, the expanded default "_grep_logs" command looked as follows:
`logpath="<logpath>"; grep -m 1000 -wF "<ip>" $logpath | tail -n 1000`

As a consequence, the first 1000 matching lines of "<logpath>" were grepped:
`grep -m 1000 -wF "<ip>" $logpath`

And then, the last 1000 lines of the grepped result were kept:
`tail -n 1000`

But since the grepped result only contained the first 1000 matching lines, those are eventually returned. This means that the first 1000 lines are returned, instead of the last 1000 lines.

As a correction, I propose to only keep the last 1000 lines of the result with tail and not filter "-m 1000" with grep:
`logpath="<logpath>"; grep -wF "<ip>" $logpath | tail -n 1000`

### DIFF
--- a/config/action.d/helpers-common.conf
+++ b/config/action.d/helpers-common.conf
@@ -14,4 +14,4 @@ _bypass_if_restored = if [ '<restored>' = '1' ]; then exit 0; fi;
 [Init]
 greplimit = tail -n <grepmax>
 grepmax = 1000
-grepopts = -m <grepmax>
+grepopts = 


sebres:
This is known "restriction" and can be considered as a compromise or a "performance" tweak (especially important if the `logpath` matching multiple log-files). Note that the ban of next intruder will be delayed till this action completed, so by heavy load it can cause a snowball-similar effect (the longer it takes, the more IPs can make much more attempts, the longer it would take in the future calls...).
Moreover since this mailing actions are obsolete (e. g. have several known issues, as well as superseded by `action.d/sendmail-whois-matches` etc), it is not advisable to use them at all.

However if one needs more precise output, one could simply reset `grepopts` in the action parameters in `jail.local`:
```ini
# action = %(banaction)s[port="%(port)s", protocol="%(protocol)s", chain="%(chain)s"]
#          sendmail-whois-lines[sender="%(sender)s", dest="%(destemail)s", logpath="%(logpath)s", chain="%(chain)s", grepopts=""]
action = %(action_mwl)s[grepopts=""]
```
Or even in `action.d/helpers-common.local`:
```ini
[Init]
grepopts = 
```

PR #3210, opened 2022-01-27
Adjusting for updated dovecot log format

karolyi:
This should now match:

`Disconnected: Connection closed: read(size=1003) failed: Connection reset by peer (auth failed, 1 attempts in 0 secs): user=<sales@karolyi.hu>, rip=183.111.188.94, lip=127.0.0.19, session=<Lsz0Oo7WXti3b7xe>`

the issue is the `read(size=1003)` that probably has been added lately and which causes the rule not to discover the log message.


sebres:
Thx for report and PR...

However the proposed solution introduces a small "ambiguity" conflicting with other open parenthesis by `(auth failed, et all)`, because previously non-interesting matching part was completely before open parenthesis, now it could try several branches many times to solve the "conflict" by repeat of `(?: [^ ]+)+` in order to find greedy match...

Let's rewrite that a bit more precise:
```diff
-            ^(?:Aborted login|Disconnected|Remote closed connection|Client has quit the connection)(?::(?: [^ \(]+)+)? \((?:auth failed, \d+ attempts(?: in \d+ secs)?|tried to use (?:disabled|disallowed) \S+ auth|proxy dest auth failed)\):(?: user=<<F-USER>[^>]*</F-USER>>,)?(?: method=\S+,)? rip=<HOST>(?:[^>]*(?:, session=<\S+>)?)\s*$
+            ^(?:Aborted login|Disconnected|Remote closed connection|Client has quit the connection)(?:: (?:[^\(]+|\w+\([^\)]*\))+)? \((?:auth failed, \d+ attempts(?: in \d+ secs)?|tried to use (?:disabled|disallowed) \S+ auth|proxy dest auth failed)\):(?: user=<<F-USER>[^>]*</F-USER>>,)?(?: method=\S+,)? rip=<HOST>(?:[^>]*(?:, session=<\S+>)?)\s*$
```

karolyi:
Um, I tested it with regex101.com, my change extracted the information just right. I don't oppose your change though, since it's more specific than mine. :)

Mine was just a quick (mostly) untested workaround.

PR #3254, opened 2022-04-08
New filter for XRDP

evanlinde:

- [x] **LIST ISSUES** this PR resolves -- none / new feature

### DIFF
--- a/fail2ban/server/datedetector.py
+++ b/fail2ban/server/datedetector.py
@@ -165,8 +165,8 @@ class DateDetectorCache(object):
 		r"%b %d, %ExY %I:%M:%S %p",
 		# ASSP: Apr-27-13 02:33:06
 		r"^%b-%d-%Exy %k:%M:%S",
-		# 20050123T215959, 20050123 215959, 20050123  85959
-		r"%ExY%Exm%Exd(?:T|  ?)%ExH%ExM%ExS(?:[.,]%f)?(?:\s*%z)?",
+		# 20050123T215959, 20050123 215959, 20050123  85959, 20050123-21:59:59
+		r"%ExY%Exm%Exd(?:-|T|  ?)%ExH:?%ExM:?%ExS(?:[.,]%f)?(?:\s*%z)?",
 		# prefixed with optional named time zone (monit):
 		# PDT Apr 16 21:05:29
 		r"(?:%Z )?(?:%a )?%b %d %k:%M:%S(?:\.%f)?(?: %ExY)?",
--- a/fail2ban/tests/files/logs/xrdp
+++ b/fail2ban/tests/files/logs/xrdp
@@ -1,0 +1,37 @@
+#
+# /var/log/xrdp-sesman.log -- should be about the same on any linux distro
+#
+
+# failJSON: { "time": "2022-04-07T12:11:06", "match": true, "host": "10.171.161.151"}
+[20220407-12:11:06] [INFO ] AUTHFAIL: user=badtypist ip=::ffff:10.171.161.151 time=1649351466
+
+# ip injection: 10.171.161.151 should be matched as the host; 192.168.0.1 is an innocent, injected address
+# failJSON: { "time": "2022-04-07T12:11:24", "match": true, "host": "10.171.161.151", "desc": "specifying ip address as username"}
+[20220407-12:11:24] [INFO ] AUTHFAIL: user=192.168.0.1 ip=::ffff:10.171.161.151 time=1649351484
+
+# ip injection: 10.171.161.151 should be matched as the host; 192.168.0.4 is an innocent, injected address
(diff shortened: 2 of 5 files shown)


fail2ban/tests/files/logs/xrdp
```diff
@@ -0,0 +1,37 @@
...
+# failJSON: { "time": "2022-04-07T12:11:06", "match": true, "host": "10.171.161.151"}
+[20220407-12:11:06] [INFO ] AUTHFAIL: user=badtypist ip=::ffff:10.171.161.151 time=1649351466
+
+# ip injection: 10.171.161.151 should be matched as the host; 192.168.0.1 is an innocent, injected address
+# failJSON: { "time": "2022-04-07T12:11:24", "match": true, "host": "10.171.161.151", "desc": "specifying ip address as username"}
+[20220407-12:11:24] [INFO ] AUTHFAIL: user=192.168.0.1 ip=::ffff:10.171.161.151 time=1649351484
+
+# ip injection: 10.171.161.151 should be matched as the host; 192.168.0.4 is an innocent, injected address
+# failJSON: { "time": "2022-04-07T12:22:02", "match": true, "host": "10.171.161.151", "desc": "more devious log injection"}
+[20220407-12:22:02] [INFO ] AUTHFAIL: user=loginjector ip=192.168.0.4 time=123456789\n[20220407-12:16:59] [INFO ] AUTHFAIL: user=endinjection ip=::ffff:10.171.161.151 time=1649352122
```
sebres, line 14: I understand that due to greedy catch-all (`.+`) the regex would match the injected user as "expected".
Just because I see escaped new-line char `\n` I would like to know whether it is just an artificial line or real test attempt.
I mean whether it'd be different possibly (more escapes or vice versa totally unescaped new-line).

fail2ban/server/datedetector.py
```diff
@@ -165,8 +165,8 @@ def _cacheTemplate(self, template):
 		r"%b %d, %ExY %I:%M:%S %p",
 		# ASSP: Apr-27-13 02:33:06
 		r"^%b-%d-%Exy %k:%M:%S",
-		# 20050123T215959, 20050123 215959, 20050123  85959
-		r"%ExY%Exm%Exd(?:T|  ?)%ExH%ExM%ExS(?:[.,]%f)?(?:\s*%z)?",
+		# 20050123T215959, 20050123 215959, 20050123  85959, 20050123-21:59:59
+		r"%ExY%Exm%Exd(?:-|T|  ?)%ExH:?%ExM:?%ExS(?:[.,]%f)?(?:\s*%z)?",
```
sebres, line 169: Not against this enhancement, but...
Would not an extra datepattern in filter help too? I mean something like that instead (or together with that):
```ini
datepattern = ^\[?%%ExY%%Exm%%Exd[-|T]%%ExH:?%%ExM:?%%ExS(?:[.,]%%f)?(?:\s*%%z)?\]?
              ^{DATE}
```
This would also allow to use that filter with older fail2ban versions.
I'll try to adjust the filter a bit.

sebres: Filter extended a bit now.

fail2ban/tests/files/logs/xrdp
evanlinde: That was a literal `\n` entered into the username field and not an actual newline character. (I couldn't figure out how to enter a literal newline character -- at least interactively -- but I'm not sure it's impossible.)

Do you have a suggested alternate to `.+` that would be more appropriate? I'd be happy to test it.

fail2ban/server/datedetector.py
evanlinde: Does the `datepattern` that you've added to the filter extend (rather than replace) what's available from `datedetector.py`? Particularly, I'm asking about with the `^{DATE}` part included. If so, I like this better than messing with `datedetector.py` but I'd want to change the expression to be more particular to xrdp's format.

sebres: `^{DATE}` is the same as default pattern but anchored at begin of line only (default datepatterns would prefer then but use both variants, anchored at begin of line and word boundaries).
As you can see the tests passed and 2nd date format in the tests also found.

fail2ban/tests/files/logs/xrdp
sebres: well I'm unsure it is possible interactively (probably only via command line only or directly with protocol), but ok.

> Do you have a suggested alternate to .+ that would be more appropriate? I'd be happy to test it.

Not really, excepting probably `(?:\S+|.+)` which would also work due to both anchors `^...$`, but could be faster for the case of "proper" user name without the space, because of different branch in PCRE without all the greedy probations.

sebres (approved):

PR #3190, opened 2021-12-24
Simplistic Docker config for deploying the server in a container

MarcPartensky:
Before submitting your PR, please review the following checklist:

- [x] **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement
      against certain release version, choose `0.9`, `0.10` or `0.11` branch,
      for dev-edition use `master` branch
- [ ] **CONSIDER adding a unit test** if your PR resolves an issue
- [ ] **LIST ISSUES** this PR resolves
- [ ] **MAKE SURE** this PR doesn't break existing tests
- [x] **KEEP PR small** so it could be easily reviewed.
- [ ] **AVOID** making unnecessary stylistic changes in unrelated code
- [ ] **ACCOMPANY** each new `failregex` for filter `X` with sample log lines
      within `fail2ban/tests/files/logs/X` file

PR #3255, opened 2022-04-11
Add additional fail-safe rule for action work.

PRIHLOP:
Add additional fail-safe rule for action work.
It is needed for use with Docker.


sebres:
> It is needed for use with Docker.

No. It does not.
1. Docker related stuff can and will be handled separately - see https://github.com/crazy-max/docker-fail2ban (we'll probably move that repo at some point here, as suggested in https://github.com/fail2ban/fail2ban/pull/3190#issuecomment-1001635830)
2. one can simply do that (but correctly) specifying parameter `chain`:
  - either in own derivative action
  - or in related jail needed that (or in default section) overwriting [this parameter](https://github.com/fail2ban/fail2ban/blob/f81f85569d88da74d76f286b431dc54d0ae00748/config/jail.conf#L192):
```ini
chain = FORWARD
```

I see no reason do that by default. Thus closed.

Issue #1930, opened 2017-10-18
Does fail2ban support send email use third-party SMTP server, like Gmail and so on, not use sendmail.


charleslxh:
Does fail2ban support send email use third-party SMTP server, like Gmail and so on, not use sendmail.

I have a gmail account, I want to send email by google email server directly, not setup a local smtp server.

sebres:
Following is possible:
1. Either you can configure sendmail to send mails via external smtp, for example see [here](https://linuxconfig.org/configuring-gmail-as-sendmail-email-relay)
2. We've an pythonic [smpt-action](https://github.com/fail2ban/fail2ban/blob/0.10/config/action.d/smtp.py) (currently used for test purposes only), but possibly you can get it running.
Try something like:
```INI
# jail.local:
[sshd]
action = %(action_)s
         smtp.py[host="host:25", user="my-account", password="my-pwd", sender="sender@example.com", dest="example@example.com", name="%(__name__)s"]
```
3. Or you can create new action, or rewrite any already available mail-action, e. g. like here:
```INI
# action.d/my-mail.local extended "mail-whois-lines.conf":

[INCLUDES]
before = mail-whois-lines.conf

[Init]
# original:
## mailcmd = mail -s
# new:
mailcmd = my-mail-command ... --subject
```
4. If you use fail2ban v. >= 0.10, you can easy redefine each action-parameter of some default action, directly in you `jail.local`:
```INI
[sshd]
action = %(action_)s
         mail-whois-lines[mailcmd="my-mail-command ... --subject", dest="example@example.com", name="%(__name__)s", ...]
```
Note: not all our actions currently have a parameter `mailcmd`, take a look in corresponding action conf-file to check which parameters could be overriden.

shunkica:
I am trying to use method #2 with f2b v0.9.7 but I get this warning:
`fail2ban.transmitter: WARNING Command ['set', 'sshd', 'addaction', 'smtp', '/etc/fail2ban/action.d/smtp.py', '{"dest": "myemail", "host": "myhost", "password": "mypassword", "user": "myuser", "name": "sshd"}'] has failed. Received TypeError("__init__() got multiple values for keyword argument 'name'",)`

sebres:
Well the smpt-action used currently for test-purposes only.
I don't know why the name received twice for you, but you can try it without `, name="%(__name__)s"` inside supplied parameters. Use `fail2ban-client -d | grep smtp` to see which parameters the action will get.

[...]

PR #3268, opened 2022-04-27
Add support for TLS SMTP connections.

Logic-32:
I don't want to install an SMTP server on my host so I'd rather use this file for sending emails.  Since I use Amazon's SES, I need TLS support.  There was a TODO to add support so here is the support :)

### DIFF
--- a/config/action.d/smtp.py
+++ b/config/action.d/smtp.py
@@ -75,7 +75,7 @@ class SMTPAction(ActionBase):
 	"""
 
 	def __init__(
-		self, jail, name, host="localhost", user=None, password=None,
+		self, jail, name, host="localhost", ssl=False, user=None, password=None,
 		sendername="Fail2Ban", sender="fail2ban", dest="root", matches=None):
 		"""Initialise action.
 
@@ -88,6 +88,8 @@ class SMTPAction(ActionBase):
 		host : str, optional
 			SMTP host, of host:port format. Default host "localhost" and
 			port "25"
+		ssl : bool, optional
+			Whether to use TLS for the SMTP connection or not.  Default False.
 		user : str, optional
 			Username used for authentication with SMTP server.
 		password : str, optional
@@ -109,7 +111,7 @@ class SMTPAction(ActionBase):
 		super(SMTPAction, self).__init__(jail, name)
 
 		self.host = host
-		#TODO: self.ssl = ssl
+		self.ssl = ssl
 
 		self.user = user
 		self.password =password
@@ -155,10 +157,19 @@ class SMTPAction(ActionBase):
 		msg['To'] = self.toaddr
 		msg['Date'] = formatdate()
 
-		smtp = smtplib.SMTP()
+		smtp = smtplib.SMTP(self.host)
 		try:
 			self._logSys.debug("Connected to SMTP '%s', response: %i: %s",
 				self.host, *smtp.connect(self.host))
+
+			if self.ssl: # pragma: no cover
+				tls_result = smtp.starttls()[0];
+				if tls_result != 220: # pragma: no cover
+					self._logSys.error(
+						"Failed to starttls() on '%s' for user '%s': %s",
+						self.host, self.user, tls_result)
+					raise Exception("Failed to starttls()")
+
 			if self.user and self.password: # pragma: no cover (ATM no tests covering that)
 				smtp.login(self.user, self.password)
 			failed_recipients = smtp.sendmail(
--- a/config/action.d/smtp.py
+++ b/config/action.d/smtp.py
@@ -157,15 +157,16 @@ class SMTPAction(ActionBase):
 		msg['To'] = self.toaddr
 		msg['Date'] = formatdate()
 
-		smtp = smtplib.SMTP(self.host)
+		smtp = smtplib.SMTP()
 		try:
+			r = smtp.connect(self.host)
 			self._logSys.debug("Connected to SMTP '%s', response: %i: %s",
-				self.host, *smtp.connect(self.host))
+				self.host, *r)
 
 			if self.ssl: # pragma: no cover
-				tls_result = smtp.starttls()[0];
-				if tls_result != 220: # pragma: no cover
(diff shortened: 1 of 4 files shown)


Logic-32:
Just going to mention #1930 as this will help anyone else looking to use smtp.py with ssl.

config/action.d/smtp.py
```diff
@@ -155,10 +157,19 @@ def _sendMessage(self, subject, text):
...
 		try:
 			self._logSys.debug("Connected to SMTP '%s', response: %i: %s",
 				self.host, *smtp.connect(self.host))
+
+			if self.ssl: # pragma: no cover
+				tls_result = smtp.starttls()[0];
+				if tls_result != 220: # pragma: no cover
+					self._logSys.error(
+						"Failed to starttls() on '%s' for user '%s': %s",
+						self.host, self.user, tls_result)
```
sebres, line 170: Since TLS takes place before login, user is probably irrelevant here, isn't it?

Logic-32: Definitely a fair point.  I figure the extra information doesn't hurt.  Though, I suppose if someone tries to setup two SMTP connections to the same host, even with different users, both TLS attempts should fail so the user probably doesn't add anything useful.

Want me to remove it?

sebres: I don't think even the logging is necessarily at all, because raising of error should be enough too.

Logic-32: Done.

Logic-32:
@sebres, I'm not sure why some of the checks on this failed.  Is there anything else you need from me to get this merged?

sebres:
Hmm... something wrong going on here with python 2.x (both 2.7 and pypy2 hanging on some test in GHA CI).

Logic-32:
Indeed.  If I'm reading things right, it was `testBan (fail2ban.tests.action_d.test_smtp.SMTPActionTest)` that had the issue.

I'm new to Python, specifically.  If you can give me some guidance on how to run the tests locally and how to account for different python versions I can try to debug/resolve.

sebres:
You can do it directly from the source code root (unbuilt, uninstalled):
```bash
PYTHONPATH=. python2.7 bin/fail2ban-testcases -vv -l 10 'SMTPActionTest'
```
That will run all SMTPActionTests within the test-suite started with python 2.7 with verbosity 2 and heavydebug (10) log-level.

sebres:
The error is that `self.host` cannot be supplied to `SMTP` constructor, because it can contain port... 
but `connect` takes place few lines below, so I guess 57fb7dec8fd772d9900557dfd1f6b1c0f55767fa would fix that.
Just untested with ssl=True (I'm unsure starttls really requires that SMTP instance should become the host by init).

Logic-32:
Thanks for finding and fixing that!  Do you want me to add a test for ssl=true?

sebres:
I'd never say "no" for more code coverage... So yes, if you can.

Logic-32:
Kk.  My personal life has been a bit crazy lately but I hope to take a stab at this in a couple weeks.

Logic-32:
It turns out that [smtpd is old, deprecated](https://docs.python.org/3/library/smtpd.html), and doesn't support `starttls()`.  The [smtpd-tls lib](https://pypi.org/project/smtpd-tls/), however, does allegedly support `starttls()` and is compatible with python 2.7.  But, [aiosmtpd](https://aiosmtpd.readthedocs.io/en/latest/) is the recommended upgrade path.

What is the recommended way to move forward here?  Not add a TLS test or install/migrate to smtpd-tls (since I'm assuming 2.7 support is required)?

sebres:
> I'm assuming 2.7 support is required

It does not. You can skip the TLS test for 2.7 (and a fallback to smtpd_tls as also OK to me).

Otherwise we can also block this unless we remove support for 2.7.

Logic-32:
I now find myself in a situation that shows off just how little I know of how to work with python properly.
- smtpd-tls supports Python 2.7+ but [you can't install it with pip](https://github.com/tintinweb/python-smtpd-tls/issues/1) in part because I hear pip is no longer available on 2.7?  At least, when I just installed it, it installed for python3 only.  I could download/install from source but where would I document that requirement or any other instructions for getting this working?
- aiosmtpd requires Python 3.6+ and f2b supports 3.2.  Granted, this would just be used in tests only so the only problem here is that using aiosmtpd, you wouldn't be able to test with 3.2 either.

Long story short, I'm happy to continue trying to add the tests.  In part because the test itself seems easy.  But, I could use some guidance how how to best resolve the dependency issue with smtpd-tls at least?

How long is it until you plan to remove support for 2.7? :)

sebres:
> aiosmtpd requires Python 3.6+ and f2b supports 3.2

I have no issue with 3.6+ at all. Python 3.6 is released 2016 and is even end-of-life now.
The support of 2.7 is there to retain compatibility for 2.x (which is hardly incompatible to 3.x in many ways), but only to provide possibility to use fail2ban modules with 2.x based software... 
There is simply no 2.8 update for 2.x, but for 3.x there are (compatible) updates.

> I could use some guidance how how to best resolve the dependency issue

Just put the tests in a conditional block:
```python
try:
  import aiosmtpd
except:
  pass
if aiosmtpd: # sys.version_info >= (3,6):
  ...
```

> How long is it until you plan to remove support for 2.7? :)

Well, I don't know... as long as possible, but it is not so far from out.
What doesn't mean that we need implement some enhancements to support that old version.

sebres:
Rebased to master (resolved conflicts in changelog) + few optimizations and code deduplication:
- combine smtpd and aiosmtpd tests;
- encapsulate smtp facilities to setUpClass/tearDownClass (behaves like a singleton, doesn't start smtp server per test);
- don't generate cert every time (too slow by RSA:2048, use pre-generated short ECC:256 instead);

This also simplifying and drastically speeding-up smtp-action tests:
```diff
  $ bin/fail2ban-testcases --fast -vvv SMTPActionTest
  ...
- Ran 7 tests in 4.058s
+ Ran 10 tests in 0.735s
```

sebres (approved):
looks nice after all, will merge it now

PR #3269, opened 2022-04-27
Adding support for Cloudflare Token API.

### LINKED ISSUE #3080: [RFE]: Use cloudflare API token instead of global key
#### Feature request type
I'd like cloudflare.conf to use a specific API key that only has access to the firewall instead of giving the global API key.

#### Description
There are two ways to access CloudFlare's APIs : The first is the "quick and dirty" way, with a global API key that has access to everything in your account. It is essentially equivalent to your password.
The second way is by using API Tokens, that are made specifically for one application and only has the required permissions. As example, the token for fail2ban would only have the permission to read and write firewall rules.

I have no idea if it is actually feasible to do in fail2ban as you probably need to make multiple requests

Logic-32:
Closes #3080

Copied mostly from cloudflare.conf and adapted for zone-based Token permissions.  I've been running this for awhile now on my server.  It works but, I'm the only one with access to my server so outside of the manual testing I did I can't say how bulletproof it is.

### DIFF
--- a/config/action.d/cloudflare-token.conf
+++ b/config/action.d/cloudflare-token.conf
@@ -1,0 +1,92 @@
+#
+# Author: Logic-32
+#
+# IMPORTANT
+#
+# Please set jail.local's permission to 640 because it contains your CF API token.
+#
+# This action depends on curl.
+#
+# To get your Cloudflare API token: https://developers.cloudflare.com/api/tokens/create/
+#
+# Cloudflare Firewall API: https://developers.cloudflare.com/firewall/api/cf-firewall-rules/endpoints/
+
+[Definition]
+
+# Option:  actionstart
+# Notes.:  command executed on demand at the first ban (or at the start of Fail2Ban if actionstart_on_demand is set to false).
+# Values:  CMD
+#
+actionstart =
+
+# Option:  actionstop
+# Notes.:  command executed at the stop of jail (or at the end of Fail2Ban)
+# Values:  CMD
+#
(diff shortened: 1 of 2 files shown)


config/action.d/cloudflare-token.conf
```diff
@@ -0,0 +1,92 @@
...
+cftarget = ip6
+
+# The firewall mode Cloudflare should use.  Default is "block" (deny access).
+# Consider also "js_challenge" or other "allowed_modes" if you want.
+#
+cfmode = block
+
+# The message to include in the firewall IP banning rule.
+#
+notes = Fail2Ban
```
sebres, line 92: Wouldn't it be better to use jail name in notes? Similar to: https://github.com/fail2ban/fail2ban/blob/f81f85569d88da74d76f286b431dc54d0ae00748/config/action.d/cloudflare.conf#L47
Just for the case the same IP gets banned in two jails at the same time, so the ticket is grouped by `notes` (would be used as an additional identifier during unban).

Logic-32: Done.

sebres: Why did you delete the section `[Init?family=inet6]`? Doesn't cloudflare expect target `ip6` anymore for IPv6?

sebres: Oh right, sorry for noise.

Logic-32:
@sebres, is there anything else you need from me to get this merged?

PR #3299, opened 2022-06-06
Adding pi-hole config

Falc0n2k:
To prevent DNS amplification attacks. This scripted helped me, and I feel like it could benefit the community as well.

### DIFF
--- a/config/filter.d/pihole-dns.conf
+++ b/config/filter.d/pihole-dns.conf
@@ -1,0 +1,26 @@
+# Fail2Ban configuration file
+#
+# original script from www.marek.tokyo
+#
+
+[INCLUDES]
+
+# Read common prefixes. If any customizations available -- read them from
+# common.local
+before = common.conf
+
+
+[Definition]
+
+_daemon = dnsmasq
+
+# log example from /var/log/pihole.log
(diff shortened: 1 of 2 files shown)


config/filter.d/pihole-dns.conf
```diff
@@ -0,0 +1,26 @@
...
+_daemon = dnsmasq
+
+# log example from /var/log/pihole.log
+#Feb 26 04:41:28 dnsmasq[1887]: query[A] 21cl93vlx5n9p.aikoaiko.net from 67.21.36.3
+#(?:DAY )?MON Day 24hour:Minute:Second(?:\.Microseconds)?(?: Year)?
+
+
+
+failregex =  .*query\[A\].*from <HOST>
+                   .*query\[ANY\].*from <HOST>
```
sebres, line 24: Such unprecise regex (containing catch-all's without anchors, especially before the address matching group) are vulnerable and thus PR cannot be accepted as it is.
Furthermore both regex's can be combined to single RE, e. g. either with `\[(?:A|ANY)\]` or still simpler with `\[[^\]]+\]`
I'd suggest something like this instead:
```ini
failregex =  ^%(__prefix_line)squery\[[^\]]+\] \S+ from <ADDR>
```

sebres (changes requested):
I don't think the changes in .gitignore are necessary. 
Also the tests cases (fail2ban/tests/files/logs/pihole-dns) are mandatory.

PR #3315, opened 2022-07-09
Use variables for log paths

nemchik:

The aim with these changes is to avoid using hardcoded paths for `logpath` in jail configurations. Since the master branch now has `paths-*.conf` for various distros, it should be easy for each distro to specify if a specific application's log is expected in a different location than the default value defined in `paths-common.conf`.

An additional benefit (and a goal I am working towards) is the 
```ini
var_log_path = /var/log
logs_path = %(var_log_path)s
```
variables can be easily adjusted for use in docker images/containers.
For example, if a user running fail2ban in docker mounts a volume for the host system `/var/log` directory inside the container as `/host_logs` they could easily change `var_log_path = /var/log` to `var_log_path = /host_logs` using the `paths-overrides.local` file.
A benefit of this approach would allow the user to have other applications inside the docker container that might use `/var/log` and have no need to involve fail2ban, so the user could keep the container application's `/var/log` separate (contained) from their host system's `/var/log`, while having the host system's `/var/log` available to fail2ban.

Another benefit would be that users running fail2ban in docker could aggregate their logs into a specific folder on their host system and mount that folder as a volume to their fail2ban container such as `/aggregated_logs` and set `logs_path = /aggregated_logs`.

It is commonly seen in environments where multiple docker containers are deployed that each container will have a mounted volume (or multiple) that contains persistent user data such as configs and logs, and the data will reside on the host in a location such as `/mnt/user/appdata/<name_of_container>` (this structure is used by unRAID, but a similar structure is often deployed by users of docker containers in other environments when using docker as well) that contains persistent user data such as config and logs. With a structure like this, users can mount subfolders of `/aggregated_logs` (if they choose to use this as their `logs_path`). fail2ban running in a docker container can be easily configured to see all the necessary logs beneath the `logs_path`. For example, `/mnt/user/appdata/nginx/log/` and `/mnt/user/appdata/lighttpd/log/` can be mounted as volumes into a fail2ban docker container as `/aggregated_logs/nginx/` and `/aggregated_logs/lighttpd/` which would allow fail2ban to see the logs in the expected location from `paths-common.conf`.

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -367,7 +367,7 @@ maxretry = 1
 
 filter = openhab
 banaction = %(banaction_allports)s
-logpath = /opt/openhab/logs/request.log
+logpath = %(openhab_request_log)s
 
 
 # To use more aggressive http-auth modes set filter parameter "mode" in jail.local:
@@ -434,19 +434,19 @@ logpath  = %(roundcube_errors_log)s
 [openwebmail]
 
 port     = http,https
-logpath  = /var/log/openwebmail.log
+logpath  = %(logs_path)s/openwebmail.log
 
 
 [horde]
 
 port     = http,https
-logpath  = /var/log/horde/horde.log
+logpath  = %(logs_path)s/horde/horde.log
 
 
 [groupoffice]
 
 port     = http,https
-logpath  = /home/groupoffice/log/info.log
+logpath  = %(groupoffice_log)s
 
 
 [sogo-auth]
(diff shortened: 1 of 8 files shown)


nemchik:
I am not sure how to understand the output of the failing CI check. Any guidance on what I need to change to resolve that issue would be appreciated.

nemchik:
@sebres I've updated to resolve merge conflicts. Can you review?

PR #3329, opened 2022-08-07
Support Python3 built w/o 2to3

tomers:
This is a preliminary work to make the codebase compatible with both Python2 and Python3, without having to run `2to3` to convert the code whenever one wants to run in Python3 environment.
This is not a complete work, and test might break still. I am looking to get the feedback from the tool maintainers first, to see if they are interested at all with such change.

### DIFF
--- a/fail2ban/client/configreader.py
+++ b/fail2ban/client/configreader.py
@@ -26,7 +26,10 @@ __license__ = "GPL"
 
 import glob
 import os
-from ConfigParser import NoOptionError, NoSectionError
+try:
+    from ConfigParser import NoOptionError, NoSectionError
+except ImportError:
+    from configparser import NoOptionError, NoSectionError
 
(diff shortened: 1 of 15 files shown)


sebres:
> I am looking to get the feedback from the tool maintainers first, to see if they are interested at all with such change.

The issue with that approach is that it is hardly possible up to impossible to do it right without `2to3`. For instance the replacement `iteritems()` by `items()` is not correct by 2.x (because it'd duplicate the iterable object implicitly, what is really worse by large lists or collections), etc.

Not to mention that 2.x reached end of life ultimately. So master branch of fail2ban would probably switch completely to 3.x soon. (And we could theoretically use 3to2 facilities to backport them to 2.x pythons).
Anyway I don't see a profit by this PR by any stretch of the imagination.

tomers:
Thanks for this answer! Looking forward for the depracation of the Python2 code base :-)

PR #3336, opened 2022-08-24
Updated firewallcmd-ipset.conf and added discord_notifications.conf

poqdavid:

Added `discord_notifications.conf` so to be able to get notifications of bans, unbans, jails started and stopped
`discord_notifications[webhook=/XXXXXXXXXXXXXXXXXXX/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX, botname=Fail2Ban, discord_roleid=<id>]`

As for the `firewallcmd-ipset.conf` I did a few changes/improvements to the config and tested them

- Added proper values for `familyopt` and `firewalld_familyopt` for ipv4
- Added separate `actionstart` and `actionstop` for `allports` option which I think is a better way to ban all ports
- Improved `ipstype_firewalld` section a bit
- Added `FORWARD` and `OUTPUT` to firewalld rules to make it a bit stricter

Also was thinking to remove the `ipstype_firewalld` section cause it's not needed at least for iptables
and you have to use `--permanent` flag which also needs `firewall-cmd --reload` command or the changes won't take effect and also when you use that it gets rid of the rest of the rules since they aren't marked permanent not to mention the whole adding IP process would be really slow since you need to reload for each IP and that would take a while if you reboot or restart the Fail2Ban server also I feel like it can cause trouble if there is a large number of IPs getting blocked.

As far as I know, even with nftables firewalld has to use the reload command not sure how it works with the nftables but I don't think it's a good idea to reload for each IP, not 100% sure if I am right so please correct me if I am wrong.

### DIFF
--- a/config/action.d/discord_notifications.conf
+++ b/config/action.d/discord_notifications.conf
@@ -1,0 +1,24 @@
+[Definition]
+
+actionstart = curl -sX POST "<webhookurl>" \
+            -H "Content-Type: application/json" \
+            -d '{"username":"<botname>","avatar_url":"https:\/\/avatars.githubusercontent.com\/u\/1087378","content":null,"embeds":[{"title":":bell: Jail Notification :bell:","description":"Jail **[<name>]** has started!","color":65280,"thumbnail":{"url":"https:\/\/github.com\/poqdavid\/f2b_discord_notification\/raw\/master\/check.png"}}]}'
+
+actionstop = curl -sX POST "<webhookurl>" \
+            -H "Content-Type: application/json" \
+            -d '{"username":"<botname>","avatar_url":"https:\/\/avatars.githubusercontent.com\/u\/1087378","content":null,"embeds":[{"title":":bell: Jail Notification :bell:","description":"Jail **[<name>]** has been stopped!","color":16711680,"thumbnail":{"url":"https:\/\/github.com\/poqdavid\/f2b_discord_notification\/raw\/master\/x.png"}}]}'
+
+actionban = curl -sX POST "<webhookurl>" \
+            -H "Content-Type: application/json" \
+            -d '{"username":"<botname>","avatar_url":"https:\/\/avatars.githubusercontent.com\/u\/1087378","content":"<@&<discord_roleid>>","embeds":[{"title":":bell: Ban Notification :bell:","color":16711680,"fields":[{"name":"IP:","value":"```<ip>```","inline":true},{"name":"Reason:","value":"**<failures>** failed attempt(s) on **[<name>]** service.","inline":true},{"name":"Information about the IP","value":"[GreyNoise.io > <ip>](<https:\/\/viz.greynoise.io\/ip\/<ip>>)\n[VirusTotal.com > <ip>](<https:\/\/www.virustotal.com\/gui\/ip-address\/<ip>\/detection>)\n[DB-IP.com > <ip>](<https:\/\/db-ip.com\/<ip>>)"}],"footer":{"text":null},"thumbnail":{"url":"https:\/\/github.com\/poqdavid\/f2b_discord_notification\/raw\/master\/ban.png"}}],"allowed_mentions": { "roles": ["<discord_roleid>"] }}' 
+
+actionunban = curl -sX POST "<webhookurl>" \
+            -H "Content-Type: application/json" \
+            -d '{"username":"<botname>","avatar_url":"https:\/\/avatars.githubusercontent.com\/u\/1087378","content":null,"embeds":[{"title":":bell: Unban Notification :bell:","color":65280,"fields":[{"name":"IP:","value":"```<ip>```","inline":true},{"name":"Reason:","value":"UNBANNED","inline":true},{"name":"Information about the IP","value":"[GreyNoise.io > <ip>](<https:\/\/viz.greynoise.io\/ip\/<ip>>)\n[VirusTotal.com > <ip>](<https:\/\/www.virustotal.com\/gui\/ip-address\/<ip>\/detection>)\n[DB-IP.com > <ip>](<https:\/\/db-ip.com\/<ip>>)"}],"thumbnail":{"url":"https:\/\/github.com\/poqdavid\/f2b_discord_notification\/raw\/master\/check.png"}}]}'
+[Init]
+
+webhookurl = https://discord.com/api/webhooks<webhook>
+
+# DEV NOTES:
+#
+# Author: POQDavid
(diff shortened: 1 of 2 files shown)


sebres:
Just to link with #1474, which has also long discussion about usage of `--permanent` flag, 
I don't follow why fail2ban may need it, especially as you write, if it would still be necessary to invoke `firewall-cmd --reload` additionally.
The entries of fail2ban must be only under the control of fail2ban - e. g. after restart it would restore all actively banned tickets... If some IP gets obsolete (its bantime is outdated), fail2ban doesn't restore such ticket, but if entry remain persistent, who will "unban" it and when?
Let alone it is not really "persistent", since actonstop (and flush) seems to remove the set (and table/rule), so the persistence is unneeded at all?
I just do not understand the necessity of that, but must admit, I'm not really familiar with firewalld (or rather systems using that).

> As far as I know, even with nftables firewalld has to use the reload command not sure how it works with the nftables but I don't think it's a good idea to reload for each IP

It sound weird and if so it's definitely bad idea.

I can only repeat (asked many times before): why, for God's sake, you people using this "interface" (overhang) over the native filters in fail2ban instead of native filters?.. Why simply don't use action of native net-filters directly (iptables-ipset or nftables)?

poqdavid:
> Just to link with #1474, which has also long discussion about usage of `--permanent` flag, I don't follow why fail2ban may need it, especially as you write, if it would still be necessary to invoke `firewall-cmd --reload` additionally. The entries of fail2ban must be only under the control of fail2ban - e. g. after restart it would restore all actively banned tickets... If some IP gets obsolete (its bantime is outdated), fail2ban doesn't restore such ticket, but if entry remain persistent, who will "unban" it and when? Let alone it is not really "persistent", since actonstop (and flush) seems to remove the set (and table/rule), so the persistence is unneeded at all? I just do not understand the necessity of that, but must admit, I'm not really familiar with firewalld (or rather systems using that).

I understand what you are saying and yes there is no need to use the `--permanent` flag because there is no need to use firewalld's ipset interface the native ipset works perfectly fine with the direct rules

> I can only repeat (asked many times before): why, for God's sake, you people using this "interface" (overhang) over the native filters in fail2ban instead of native filters?.. Why simply don't use action of native net-filters directly (iptables-ipset or nftables)?

Actually, that's what I wanted to do, I wanted to just get rid of the `ipstype_firewalld` section and just use `ipstype_ipset` which is the native interface just wasn't sure if you preferred to keep it or not, and as far as nftables for the ipset that also has its own thing which later in the future can be added since as far as I know direct rules work fine even with nftables
so I am going to clean it up and get rid of that section if you are ok with it

sebres:
> I understand what you are saying and yes there is no need to use the --permanent flag because there is no need to use firewalld's ipset interface the native ipset works perfectly fine with the direct rules

Does it mean that `--permanent` flag would be needed for firewalld's ipset? 
Why so?

> ... and just use ipstype_ipset which is the native interface 

I spoke about fully usage of native net-filter (that what firewald calls backend), not ipset only.
Can you explain why you prefer action `firewallcmd-ipset` over the actions `iptables-ipset` or `nftables`?

> just wasn't sure if you preferred to keep it or not

Sure we'll keep it as long as firewalld will support native ipset further.

poqdavid:
> Does it mean that `--permanent` flag would be needed for firewalld's ipset? Why so?
 
Cause when you run the command without `--permanent ` it says `Option can be used only with --permanent.` you literally can't use it without.

> I spoke about fully usage of native net-filter (that what firewald calls backend), not ipset only. Can you explain why you prefer action `firewallcmd-ipset` over the actions `iptables-ipset` or `nftables`?

Oh no I don't prefer `firewallcmd-ipset`, I must have said it wrong or in a confusing way I actually wanted to get rid of the `firewallcmd-ipset` and just use the native ipset for iptables which is just called ipset on ubuntu we can also add `nftables` ipset too in the future 
 
> Sure we'll keep it as long as firewalld will support native ipset further.

and when I said `just wasn't sure if you preferred to keep it or not` it was cause I wanted to remove it and felt like I should check with you before doing so

sebres:
> Cause when you run the command without --permanent it says Option can be used only with --permanent. you literally can't use it without.

Hmm... that's weird. Then I must admit I made probably a mistake (in c7a86b4616ef1bd43cf96874710b53b42d065684) combining `firewallcmd-ipset` with `firewallcmd-ipset-native`, where it was initially added with `--permanent` (and I removed it by melting).
The https://github.com/fail2ban/fail2ban/pull/2620#issuecomment-850898948 describes why.
Strange is that, IIRC, I tested it on some box of me, and it worked without any issue.

@mspolitaev can you also confirm it on your side?

> Oh no I don't prefer firewallcmd-ipset

Well, you do... since you're trying to improve it, I guess you use it.
It's not about native ipset only, I told you all the time about fully native actions [iptables-ipset](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/iptables-ipset.conf) and [nftables](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/nftables.conf) which both working natively without to use firewalld at all.
I mean why one needs to use a overhang interface (what firewalld is) over the native net-filters (called backend in firewalld), if one can use net-filters directly?

> we can also add nftables ipset too in the future

Where? in firewalld action?
Because native nftables action doesn't really need any ipset stuff, it already uses address set internally:
https://github.com/fail2ban/fail2ban/blob/a08b9254689c46e0a32537f1f88430d0cb0446f0/config/action.d/nftables.conf#L58

poqdavid:
> Well, you do... since you're trying to improve it, I guess you use it. It's not about native ipset only, I told you all the time about fully native actions [iptables-ipset](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/iptables-ipset.conf) and [nftables](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/nftables.conf) which both working natively without to use firewalld at all. I mean why one needs to use a overhang interface (what firewalld is) over the native net-filters (called backend in firewalld), if one can use net-filters directly? I don't meant to stop to use firewalld at all, I meant the usage in fail2ban only.

I see what you mean now well I personally had better experience with like firewalld and docker
also its kind of like UFW but with a lot more option and its not complex like iptables so like makes firewall a bit easier to deal with without losing much options

> Where? in firewalld action? Because native nftables action doesn't really need any ipset stuff, it already uses address set internally:
> 
> https://github.com/fail2ban/fail2ban/blob/a08b9254689c46e0a32537f1f88430d0cb0446f0/config/action.d/nftables.conf#L58

Ah I see I had no idea nftables works like that

PR #3349, opened 2022-09-09
IPThreat integration

jjxtra:

- [x] **CONSIDER adding a unit test** if your PR resolves an issue (N/A)
- [x] **LIST ISSUES** this PR resolves (N/A)
      within `fail2ban/tests/files/logs/X` file (N/A)

### DIFF
--- a/config/action.d/ipthreat.conf
+++ b/config/action.d/ipthreat.conf
@@ -1,0 +1,99 @@
+# IPThreat configuration file
+#
+# Added to fail2ban by Jeff Johnson (jjxtra)
+#
+# Action to report IP address to ipthreat.net
+#
+# You must sign up to obtain an API key from ipthreat.net and request bulk report permissions
+# https://ipthreat.net/integrations
+#
+# IPThreat is a 100% free site and service, all data is licensed under a creative commons by attribution license
+# Please do not integrate if you do not agree to the license
+#
+# IMPORTANT:
+#
+# Reporting an IP is a serious action. Make sure that it is legit.
+# Consider using this action only for:
+#   * IP that has been banned more than once
+#   * High max retry to avoid user mis-typing password
+#   * Filters that are unlikely to be human error
+#
+# This action relies on `ipthreat_apikey` being added to the very bottom of this action conf,
+# and the appropriate flags and system set.
+#
+# Example, for ssh bruteforce (in section [sshd] of `jail.local`):
+# ```
+#   action = %(known/action)s
+#            ipthreat[ipthreat_flags="8",ipthreat_system="SSH"]
+# ```
+#
+# `ipthreat_system` is a short name of the system attacked, i.e. SSH, SMTP, MYSQL, PHP, etc.
+#
+# For `ipthreat_flags`, most cases will use 8 (BruteForce), but you could use others.
+# You can use the name or the ordinal.
+# Multiple values are comma separated.
+# ```
+# Name           Ordinal Description
+# Dns            1       Abuse/attack of dns (domain name server)
+# Fraud          2       General fraud, whether orders, misuse of payment info, etc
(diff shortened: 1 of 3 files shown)


config/jail.conf
```diff
@@ -251,6 +251,12 @@ action_cf_mwl = cloudflare[cfuser="%(cfemail)s", cftoken="%(cfapikey)s"]
 #
 action_blocklist_de  = blocklist_de[email="%(sender)s", service="%(__name__)s", apikey="%(blocklist_de_apikey)s", agent="%(fail2ban_agent)s"]
 
+# Report ban via ipthreat.net.
+#
+# See action.d/ipthreat.conf for usage example and details
+#
+action_ipthreat = ipthreat
```
sebres, line 258: This is not really needed... I know we have several entries with similar declaration, but:
- most of them have additional parameters which are either mandatory or pre-defined to some default value supplied from jails;
- and a comment describing where the action can be found is really enough (even that is unneeded in my opinion, many action doesn't noticed in jail.conf at all).

Also the definition is not more complex as without (even much shorter without that substitution):
```diff
 action = %(action_)s
-         %(action_ipthreat)s[ipthreat_flags="8",ipthreat_system="SSH", ipthreat_apikey=...]
+         ipthreat[ipthreat_flags="8",ipthreat_system="SSH", ipthreat_apikey=...]
```

config/action.d/ipthreat.conf
```diff
@@ -0,0 +1,99 @@
...
+actioncheck =
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+#
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionban = curl -sSf "https://api.ipthreat.net/api/report" -X POST -H "Content-Type: application/json" -H "X-API-KEY: <ipthreat_apikey>" -d '{"ip":"<ip>","flags":"<ipthreat_flags>","system":"<ipthreat_system>","notes":"fail2ban"}'
```
sebres, line 83: This will be only substituted correctly if all of the interpolation variables are simple tokens.
If some of them become some special chars, fail2ban would prefer variable escape, so it would be not substituted inside the single quote-string and you can get `"system":"$f2bV_ipthreat_system"` instead of `"system":"..."`.
So possibly something like this would be better:
```diff
- -d '{"ip":"<ip>","flags":"<ipthreat_flags>","system":"<ipthreat_system>","notes":"fail2ban"}'
+ -d "{\"ip\":\"<ip>\",\"flags\":\"<ipthreat_flags>\",\"system\":\"<ipthreat_system>\",\"notes\":\"fail2ban\"}"
```

Also I wonder if ipthreat_system cannot simply contain `<name>` (name of jail) by default. E.g. set this in Init section:
```ini
[Init]
ipthreat_system = <name>
```

config/jail.conf
jjxtra: I've replaced the example in `config/action.d/ipthreat`, does this need to go in jail.conf instead?

config/action.d/ipthreat.conf
jjxtra: I have performed these changes, please let me know if they are sufficient

config/jail.conf
sebres: Although, I'm not against the comments, but don't see the necessity for `action_ipthreat = ipthreat`... I'd indeed prefer to remove everything. A look into `action.d/ipthreat.conf` is enough in my opinion.

> It could still be useful if someone is scrolling through that file or doing a search...

We can put a small notice about every action we have, but... let's avoid duplicating of everything.

config/action.d/ipthreat.conf
sebres: I think it's OK.

sebres:
Merged. Thx!

PR #3362, opened 2022-09-24
Remove hard end of string

invisibleroads:
NGINX now includes `http_x_forwarded_for` in logs after `http_user_agent`, resulting in the badbots filter not working as expected.

```
# nginx.conf
log_format main '$remote_addr - $remote_user [$time_local] "$request" '
                '$status $body_bytes_sent "$http_referer" '
                '"$http_user_agent" "$http_x_forwarded_for"';
```

Removing hard end of string ($) restores this functionality. Test using the following command.

```bash
fail2ban-regex /var/log/nginx/access.log /etc/fail2ban/filter.d/apache-badbots.conf
```


config/filter.d/apache-badbots.conf
```diff
@@ -10,7 +10,7 @@
...
 
-failregex = ^<HOST> -.*"(GET|POST|HEAD).*HTTP.*"(?:%(badbots)s|%(badbotscustom)s)"$
+failregex = ^<HOST> -.*"(GET|POST|HEAD).*HTTP.*"(?:%(badbots)s|%(badbotscustom)s)"
```
sebres, line 13: I don't like such filters at all (and they are better to be considered as an example)...
But let us stop to provide vulnerable regex further - this RE contains several catch-all's (`.*`), but anchored from both sides it can be accepted, since it finds IP at begin of message and agent at end of message.
Removal of this end-anchor breaches this condition and RE becomes vulnerable for false positives and intended manipulation.
With other words one needs to rewrite this RE as precise as possible (without catch-alls).

sebres (changes requested):

Issue #3239, opened 2022-03-17
[RFE]: cloudflare IP lists instead of rules


drzraf:
#### Feature request type
Use Cloudflare mutliple matches ability to ban multiple IP instead as one-rule-per-IP

#### Description
cloudflare module currently create/delete one rule per IP address. This is suboptimal because:
- It clutter the rules instead of restricting all fail2ban processing to its own rule (leaving others ones to other subsystem)
- Because Cloudflare free plan only allows up to 5 rules
- Because rules allow to create unlimited OR-combined match
- Because even one match of one rule can use the `is in` operator to specify multiple comma-separated IP addresses
- Because, even better, we can create "lists" of IP address and use only one match from one rule stating `block if source IP address is part of <list>` (Free plan provide up to 1 list contains up to 10.000 IP addresses per account) 

#### Any additional information
This change would imply that the ban/unban mechanism first fetch the remote value, do the merging (IP address addition/removal) and then call the API the new value.

drzraf:
Mentioning #3239 to @wpkc

sebres:
This would probably imply bulk facilities (#3088) become ready.

wpkc:
- The Fail2Ban CloudFlare action does not use CloudFlare Firewall Rules.  It uses CoudFlare's IP Access Rules, which doesn't seem to have a limit on free accounts.  So the 5 firewall rule limit on free accounts does not apply.
- Fail2Ban temporarily bans IPs based on polling log file data at 1 second intervals.  If you have 10K+ active Fail2Ban IP bans at any one time, you should probably look into other security solutions.
- As a first line of defense, I recommend using `iptables --match-set` along with an auto-generated list of blacklisted IPs, such as this: [https://github.com/stamparm/ipsum](https://github.com/stamparm/ipsum)  This will stop a lot of traffic immediately, before Fail2Ban sees it.

sebres:
> If you have 10K+ active Fail2Ban IP bans at any one time, you should probably look into other security solutions

Have several hosts with multiple jails containing significantly more than 10K active bans each (`iptables-ipset` or `nftables`). And some of nginx jails achieved nearly 100K+ bans (IP-based or `nginx-block-map` actions, limit_req etc). 
What I'm doing wrong?

wpkc:
[...]

PR #3397, opened 2022-10-31
Cloudflare list feature added

Broatcast:
**Add a new cloudflare action with list support**

Since there isn't any solution for this problem, i was adding Cloudflare support back again.
The old solutions didn't work with the new Cloudflare plans.

This action is based on calling an php script to handle all the required logic. In the current state this will be to complex for fail2bans action wrapper.
I know it's not written in python but considering all people who will need this also have php already on the server. 

The installation is easy and supported by the script it self. I will add an wiki page later for this.

**Your api token need this permissions**
  * All accounts - Account Filter Lists:Edit, Account Settings:Read
  * All zones - Zone:Read, Firewall Services:Edit
(for more security add your server ip in the `Client IP Address Filtering`

**Short install instructions:**
Basically you only need to add your Cloudflare token to you `jail.local` into [DEFAULT]

> [DEFAULT]
> cftoken = your-cf-api-token

Then you can run the token-test while looking at your `fail2ban.log`
`php /etc/fail2ban/action.d/cloudflare.php token-test`

This will add the following lines to your logfile
> 2022-10-31 23:02:31,000 fail2ban.actions        [123551]: INFO    [cloudflare] ================
> 2022-10-31 23:02:31,000 fail2ban.actions        [123551]: INFO    [cloudflare] Cloudlfare API token-test
> 2022-10-31 23:02:31,000 fail2ban.actions        [123551]: INFO    [cloudflare] ================
> 2022-10-31 23:02:31,000 fail2ban.actions        [123551]: INFO    [cloudflare] Testing api token
> 2022-10-31 23:02:32,000 fail2ban.actions        [123551]: INFO    [cloudflare] This API Token is valid and active
> 2022-10-31 23:02:32,000 fail2ban.actions        [123551]: INFO    [cloudflare] You can now run setup and permission-test

After this you are ready for the setup process run
`php /etc/fail2ban/action.d/cloudflare.php setup`

This will list all needed information in you `fail2ban.log`
> 2022-10-31 23:08:52,000 fail2ban.actions        [123551]: INFO    [cloudflare] =============
> 2022-10-31 23:08:52,000 fail2ban.actions        [123551]: INFO    [cloudflare] Cloudlfare API setup
> 2022-10-31 23:08:52,000 fail2ban.actions        [123551]: INFO    [cloudflare] =============
> 2022-10-31 23:08:52,000 fail2ban.actions        [123551]: INFO    [cloudflare] Try to list account information
> 2022-10-31 23:08:53,000 fail2ban.actions        [123551]: INFO    [cloudflare] Found account, ID: {your-cf-account-id} Name: example@example.com's Account Type: standard
> 2022-10-31 23:08:53,000 fail2ban.actions        [123551]: INFO    [cloudflare]
> 2022-10-31 23:08:53,000 fail2ban.actions        [123551]: INFO    [cloudflare] Try to list zone id's
> 2022-10-31 23:08:53,000 fail2ban.actions        [123551]: INFO    [cloudflare] Found zone, ID: {your-cf-zone-id} Name: example.com Status: active

The only thing left is to copy your account id and your zone id and write it to you `jail.local` into [DEFAULT]
> [DEFAULT]
> cftoken = {your-cf-api-token}
> cfaccount = {your-cf-account-id}
> cfzone = {your-cf-zone-id}

**Finish**
You can now use this action in any jailconfig you like by overwriting the `banaction` in your `jail.local`.

**Example for `nginx-http-auth`**
> [nginx-http-auth]
> enabled = true
> port    = http,https
> logpath = %(nginx_access_log)s
> banaction = cloudflare-list

*In case of any problems*
Keep an look at your `fail2ban.log`, there is many information logged, api errors get parsed and written to the logfile.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -106,6 +106,7 @@ Raphaël Marichez
 RealRancor
 René Berber
 Robert Edeker
+Robert Kuntz
 Rolf Fokkens
 Roman Gelfand
 Russell Odom
--- a/config/action.d/cloudflare-list.conf
+++ b/config/action.d/cloudflare-list.conf
@@ -1,0 +1,99 @@
+# Fail2Ban configuration file
+#
+# Author: Robert Kuntz
+#
+#
+# INFO
+#
+# This is a new implementation for Cloudflare wich using the list function for Cloudflare
+# to work with the new free plans
+#
+#
+# IMPORTANT
+#
+# Please set jail.local's permission to 640 because it contains your CF API token.
+#
+#
+# API INFORMATIONS
+#
+# API Documentation: https://api.cloudflare.com/
+# API: https://api.cloudflare.com/client/v4/
+# To get your Cloudflare API token: https://developers.cloudflare.com/api/tokens/create/
+
+
+[Definition]
+
+# Option:  actionstart
+# Notes.:  command executed on demand at the first ban (or at the start of Fail2Ban if actionstart_on_demand is set to false).
+# Values:  CMD
+#
+actionstart = php /etc/fail2ban/action.d/cloudflare.php start
+
+# Option:  actionstop
+# Notes.:  command executed at the stop of jail (or at the end of Fail2Ban)
+# Values:  CMD
+#
+actionstop = php /etc/fail2ban/action.d/cloudflare.php stop
+
+# Option:  actioncheck
+# Notes.:  command executed once before each actionban command
+# Values:  CMD
+#
+actioncheck =
+
+# Option:  actionrepair
+# Notes.:  Command executed in order to restore sane environment in error case.
+# Values:  CMD
+actionrepair =
+
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+# Tags:    <ip>  IP address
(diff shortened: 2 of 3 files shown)


Broatcast:
This will also address this Issues:
* #3239

sebres:
> I know it's not written in python but considering all people who will need this also have php already on the server.

I don't think so, but OK.
The question is rather how large is the effort to rewrite it as a pythonic action (combining both `cloudflare-list.conf` and `cloudflare.php` as single `cloudflare-list.py` file written in python)?

sebres:
> In today time if you provide some services on port 80/443 90% would have php on the server

No. 90% of some private hosting - maybe, but I still doubt about the proportion of whole community.
Especially people which may need cloudflare.

> Rather i like to contribute to the project with a working and clean written solution even if its in php, but for this, clean and functional

Sure and thank you!... 
Although I speak php, unfortunately neither I have cloudflare, nor I think it would be good idea to check this php action blindly (without the cloudflare access).

Anyway we can leave it open here up to next regular release, in hope someone'd rewrite it to pythonic action.
People which need it and in mood to experiment could use the files as is from this PR.

Issue #1924, opened 2017-10-17
action.d/pf.conf actionban doesn't kill state  after adding ban.


ghost:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.10.1
- OS, including release name/version: FReebsd 11.1
- [x] Fail2Ban installed via OS/distribution mechanisms

### The issue: 
`actionban = <pfctl> -t <tablename>-<name> -T add <ip>`

actionban should invoke an additional `pfctl -k <ip>` to kill the state entries, since rules are only applied to new connections. Without clearing the state an attacker is able to keep the connection open and to continue the attack

#### Expected behavior
`actionban = <pfctl> -t <tablename>-<name> -T add <ip> && <pfctl> -k <ip>`

sebres:
> Without clearing the state an attacker is able to keep the connection open and to continue the attack

I doubt it, because the ban occurs after failure, recognized from corresponding application. Normally it closes the connection hereafter.

But possibly it would be really a nice feature.

mkeilwaldbrunn:
Hi @sebres,
I really don't want to bother you, but I assume, that you are not familiar with the FreeBSD packet filter.
Please feel free to double-check, for example read the introduction chapter of "A no nonsene guide to PF"

For any rule that has a "keep state" part, which is the default option, the firewall keeps information of packets that already have matched a rule in the "state table". Any new packets are checked  first against the state table and will only be processed by "the rule set evaluation" if there is no match in the "state table". This will lead ato a prior granted "pass .." until the state table TTL expires.

mkeilwaldbrunn:
see also 

https://forums.freebsd.org/threads/51624/

mkeilwaldbrunn:
The "state table tcp.closed timeout" is set to 90s and starts to count down after sshd disconnecting. 
Any subsequent packet whithin 90s from the meanwhile banned host will be allowed by the  entry in the state table.  This will result in ongoing attack and, depending on your loglevel, you'll get a "IP already banned warning" in fail2ban.log.

There's no need to be offensive. 
I'm wondering why you'd developed FreeBSD support in the first place, if you aren't interested in fixing or improving. That's odd, is it not? I simply wanted to support. Fail2ban has this bug for a very long time.

Since you have neither the time nor the desire, please don't bother me with an answer.

Issue closed due to lack of commitment

sebres:
You still not understood me...

> There's no need to be offensive.

Hmmm... where?

[...]

PR #3402, opened 2022-11-07
Fix PF so it will actually kill the session so the ban takes place

VVelox:
Currently PF support is slightly broken. A abusive IP can continue it's action even if it is banned as the state is still good. Fixing this requires calling `pfctl -k <ip>` to kill the states for that IP.

Noticed this when I saw it was constantly failing to ban IPs attempting to use a DNS server for a amp attack. Suricata was flagging it and F2B was shoving the IP into the table, but because it had a existing state the ban would not come into affect.

- [X] **CONSIDER adding a unit test** if your PR resolves an issue

### DIFF
--- a/config/action.d/pf.conf
+++ b/config/action.d/pf.conf
@@ -66,7 +66,7 @@ actioncheck = <pfctl> -sr | grep -q <tablename>-<name>
 #          <time>  unix timestamp of the ban time
 # Values:  CMD
 #
-actionban = <pfctl> -t <tablename>-<name> -T add <ip>
+actionban = <pfctl> -t <tablename>-<name> -T add <ip> | pfctl -k <ip>
 
 
 # Option:  actionunban


sebres:
Closes #1924 
Marked as enhancement now, since I'm firmly convinced it must be optional, at least for multi-port mode (for instance how we made it in #3018 for `ufw`), see https://github.com/fail2ban/fail2ban/issues/1924#issuecomment-953190599 for details.

VVelox:
> Closes #1924 Marked as enhancement now, since I'm firmly convinced it must be optional, at least for multi-port mode (for instance how we made it in #3018 for `ufw`), see [#1924 (comment)](https://github.com/fail2ban/fail2ban/issues/1924#issuecomment-953190599) for details.

Very bad idea to make it optional if you want to actually function.

For UDP it means as long as the attacker keeps going the ban will never take affect. Recently saw this happen on a router failing to ban a 

This also runs into issues with TCP based protocols such as HTTPS2 as well.

Simply put as it stands currently this action is non-workable for any other than TCP based daemon that immediately slam the connection shut upon some sort of issue.

One option is if we want to keep it multiport and not kill stuff, we likely should rename this to pfTCP or the like and drop expectation of it working for anything other than TCP. That and also note that ban will only take place post state connection closure.

Basically the pf action for Fail2ban has never been functional because of the lack of state removal.

VVelox:
If interested, I can take care of the refactoring to break this out into two different actions. One that will work reliably and one that will work for a limited subset of TCP.

sebres:
> Very bad idea to make it optional if you want to actually function.

I already explained my doubts about and don't think I'd repeat all that here.
After all it hardly depends on the type of service, the jail, its concerns etc.

> One option is if we want to keep it multiport and not kill stuff, we likely should rename this to pfTCP or the like and drop expectation of it working for anything other than TCP

So you seem to agree that it must be optional, but for some reason will make from the option a new action type.
This is not necessary.

> If interested, I can take care of the refactoring to break this out into two different actions.

Why? I mean - why a configuration:
```ini
banaction = pf_with_kill
```
is better than something like:
```ini
banaction = pf[kill=true]
```

VVelox:
> > Very bad idea to make it optional if you want to actually function.
> 
> I already explained my doubts about and don't think I'd repeat all that here. After all it hardly depends on the type of service, the jail, its concerns etc.
> 
> > One option is if we want to keep it multiport and not kill stuff, we likely should rename this to pfTCP or the like and drop expectation of it working for anything other than TCP
> 
> So you seem to agree that it must be optional, but for some reason will make from the option a new action type. This is not necessary.
> 
> > If interested, I can take care of the refactoring to break this out into two different actions.
> 
> Why? I mean - why a configuration:
> 
> ```ini
> banaction = pf_with_kill
> ```
> 
> is better than something like:
> 
> ```ini
> banaction = pf[kill=true]
> ```

I actually don't agree with it being optional in the least. My suggestion was on the basis of it makes you happy.

If not using kill by default a very big warning must be given that this very much can not be trusted to prevent attacks in a very large number of circumstances.

Over the weekend I literally watched this current failure mode enable usage in a DDOS as the state was kept alive.

Using this with out killing the state, the user very much should be made aware that this won't function as intended as it is not making proper use of pf.

What I am saying is support for UDP and ICMP should be ripped out of the one is not using pf as intended as it is not clearing the states, because as is this ain't security in the least.

sebres:
> My suggestion was on the basis of it makes you happy.

I'm also not happy with the situation. But unless `pf` doesn't provide proper way to kill the states of certain ports only, we have no other choice. The option is just a compromise.
Killing of the states of any connections from IP is rather a workaround. Don't try to assert it would be a regular solution, because it is not.

> If not using kill by default a very big warning must be given that this very much can not be trusted to prevent attacks in a very large number of circumstances.

Unfortunately the same is valid for vice versa situation - your approach may deviate from approach of many users (which would prefer not to ban someone as to drop someone erroneously in case of false positive) and many use cases where it works without kill (because of either correct handling on application level or with different rules set doesn't allowing established connection to bypass banning tables).
It is just a point of view... In your case it is "Ban is more important than everything else, even if it'd bother legitimate users".
Another point of view - "Never disturb operation of legitimate users" is justified in the same way.
For instance someone has configured some minor jail banning potential intruders on some port for few time (e. g. few seconds after few potentially "bad" attempts) considering that there was no good solution to recognize real evildoers.
Such jail will don't bother legitimate users (which cannot be good distinguished from intruders) if appropriate port-related action is used, but can bother legitimate users if the action abrupt starting to kill all established connections no matter which port. This may be totally unexpected and unwanted for many many fail2ban users.
And hereafter we (not you) will get a lot of issues like "why fail2ban sporadically drops legitimate users of service X even though it is not banned".

And you don't need to convince me using the same already used arguments again and again.

> Using this with out killing the state, the user very much should be made aware that this won't function as intended as it is not making proper use of pf

As already said "proper usage" is the thing everyone defined by oneself. What is proper for you, may be not proper or not always proper for someone else. Also it is the matter of priorities... Just try to understand and accept that for someone the operation may have higher precedence than a supposed "security". Or that accidental out of operation is absolutely unexpected. I don't want to be accountable for that issues, do you?

> If desiring to keep it as one, kill should be the default 

Still again, it depends.

But what is the problem to define this by default (in default section of the jail.local)?
```ini
[DEFAULT]
banaction = pf[kill=true]
```
The thing is - it is then your own decision (and not of the fail2ban devs or maintainers).

VVelox:
You seem the be the one advocating making it be insecure by default here.

I've been willing to accept some people don't value security of their systems and the users of those systems so they don't care if their security software is working properly, so they should be able to set it to insecure by passing a flag to turn off that security bit.

> Or that accidental out of operation is absolutely unexpected. I don't want to be accountable for that issues, do you?

You already are here.

It very much is failing to ban bad actors and is very much non-functional for UDP.

>  It is just a point of view... In your case it is "Ban is more important than everything else, even if it'd bother legitimate users".

Yes. It's called security. If needed the can contact the NOC etc for a fix. It's like when a auth system locks out a account till administratively unlocked based on number of improper failures.

>  And hereafter we (not you) will get a lot of issues like "why fail2ban sporadically drops legitimate users of service X even though it is not banned".

If you did not want to be dealing with, you could of noted that this is very much a known and expected failure mode at the top of the pf.conf file a long time ago given how long you've been aware of this.

Like I said I would be happy to handle the re-factoring that allows it to be ran in a insecure mode. Just that the user should very much be aware they are using pf in a manner that is not secure.

VVelox:
I guess the big question is are you actually open to getting a fix committed if some one takes the time to create one? I've offered to with the requested ability to set it a insecure mode and has just been feeling a lot of push back and lack of interest in this actually being fixed.

It is very frustrating, especially given in that linked to thread, you never though to note the massive possible failure mode for pf at the top of the conf, despite being aware of it. That and the protos outside of TCP are also not meaningfully usable.

That and you are making such a argument about keeping it insecure by default and for making sure it is not clearly documented in terms of the possible failure mode.

I understand where you are coming from with wanting it to be able be configured in such a manner, just I think it should secure by default and properly noted so the person setting it up is aware of the security implications are are opting into it.

Like requiring a option taken to run something as root that should in general not be ran as root.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+#
+# Example..
 #
+# [dovecot]
+# enabled = true
+# filter = dovecot
+# action = pf[actiontype=<allports>,no_kill="true"]
 #
+# The above will mean if some one hits the limit for dovecot, it won't automatically kill their
+# active states, such as maybe SSH.
```
cschuber, line 35: s/some one/someone/

How about putting the comment above the example after the word Example: The example below will not remove the session from the state table therefore not causing the immediate block of the session.

I don't see how the SSH comment adds anything to clarify the example.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+# as the states for that IP were never cleared.
 
 [Definition]
 
 # Option:  actionstart
 # Notes.:  command executed on demand at the first ban (or at the start of Fail2Ban if actionstart_on_demand is set to false).
 # Values:  CMD
 #
-# we don't enable PF automatically; to enable run pfctl -e 
+# we don't enable PF automatically; to enable run pfctl -e
```
cschuber, line 52: Whitespace changes should be a separate commit to avoid cluttering git history by mixing functional and cosmetic changes in the same commit. (This is how I construct my commits.)

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
 #
+# The above will mean if some one hits the limit for dovecot, it won't automatically kill their
+# active states, such as maybe SSH.
+#
+# Where would be doing this be bad? Lets say you have a service using UDP. The attacker can keep it
+# up indefinitely as long they don't let up as UDP is stateless. Similarly with HTTP2, a attacker
+# can keep it going till the web server kills or they finish doing what ever they are doing. This is
+# especially true for a rule that triggers based on attempted data exfiltration.
+#
+# Basially you have the choice of maybe a potetional moment annoyance of having to restart a SSH
```
cschuber, line 42: Another personal pronoun.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+# session or the like or potiential compromise as the pf table change was never allowed to take affect
+# as the states for that IP were never cleared.
```
cschuber, line 44: This paragraph should be rewritten to discuss the risk of long running sessions, such as SSH. However, how could a currently active session cause a ban? Typically if a user has logged in and subsequently caused fail2ban to add the IP to the block list one has more problems than some bad actor attempting to gain access. I would think that such scenario would occur when a valid user has successfully logged in from behind a NAT firewall and another person behind the same NAT firewall has tripped fail2ban. I suppose to be safe the SSH session could be shut down.

Considering another scenario a bad actor could DoS the valid user. For example, when I travel my ISP (cable company) maintains WiFi nodes throughout Canada. I simply connect. However some bad actor could deny me access to my site because this ISP maintains a NAT appliance in each different city.

I'm leaning toward not including this paragraph at all because less experienced users could make the wrong choice because of insufficient information to make an informed decision. Or, this paragraph could be reworded but I think this will confuse the issue even more.

config/action.d/pf.conf
```diff
@@ -84,7 +119,7 @@ actionunban = <pfctl> -t <tablename>-<name> -T delete <ip>
 #
 # Use anchor as jailname to manipulate affected rulesets only.
 # If more parameter expected it can be extended with `pf[pfctl="<known/pfctl> ..."]`
-# 
+#
```
cschuber, line 122: Ditto on whitespace changes.

config/action.d/pf.conf
```diff
@@ -104,6 +139,10 @@ block = block quick
 # Notes.:  internally used by config reader for interpolations.
 # Values:  [ tcp | udp | icmp | ipv6-icmp ] Default: tcp
 #
+# If set to "{ tcp, udp, icmp, ipv6-icmp }", all protocols will be
+# taken care of. Similarly combination of protocols may be handled
+# in a similar manner.
+#
```
cschuber, line 145: Remove the "If." 

What is meant by "taken care of?"

How about: Multiple protocols can be specified by enclosing them in braces. For example "{ tcp, udp, icmp, ipv6-icmp }" will configure the the reader for all protocols.

cschuber (changes requested):

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
 #
 # Author: Nick Hilliard <nick@foobar.org>
 # Modified by: Alexander Koeppe making PF work seamless and with IPv4 and IPv6
+# Further modified by: Zane C. Bowers-Hadley <vvelox@vvelox.net> to make it kill states for a IP so the
+#                      ban takes affect as well as not handling multiple protocols.
+
+# WARNING WARNING WARNING WARNING WARNING
+# WARNING WARNING WARNING WARNING WARNING
+# WARNING WARNING WARNING WARNING WARNING
```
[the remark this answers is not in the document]
VVelox: I'll be honest, I though it looked nice and attention grabbing for a very important bit of detail about how stats are handled.

I am totally open to suggestions on formatting here, but aye! One also works.

cschuber:
> Currently PF support is slightly broken. A abusive IP can continue it's action even if it is banned as the state is still good. Fixing this requires calling `pfctl -k <ip>` to kill the states for that IP.
> 
> Noticed this when I saw it was constantly failing to ban IPs attempting to use a DNS server for a amp attack. Suricata was flagging it and F2B was shoving the IP into the table, but because it had a existing state the ban would not come into affect.
> 
> Before submitting your PR, please review the following checklist:
> 
>     * [x]  **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement
>       against certain release version, choose `0.9`, `0.10` or `0.11` branch,
>       for dev-edition use `master` branch
> 
>     * [x]  **CONSIDER adding a unit test** if your PR resolves an issue
> 
>     * [x]  **LIST ISSUES** this PR resolves
> 
>     * [x]  **MAKE SURE** this PR doesn't break existing tests
> 
>     * [x]  **KEEP PR small** so it could be easily reviewed.
> 
>     * [x]  **AVOID** making unnecessary stylistic changes in unrelated code
> 
>     * [x]  **ACCOMPANY** each new `failregex` for filter `X` with sample log lines
>       within `fail2ban/tests/files/logs/X` file

To say that PF is slightly broken is incorrect and misleading. All stateful firewalls behave the same, inspecting the state table (hash table) before rules are serially matched. (Though iptables allows the user to insert in inspect state table rule in the table, one could put that rule after other rules but that would negate the state table in the first place, therefore we can consider iptables the same as other stateful packet filters.)

BTW, the same problem exists with Checkpoint, CISCO, and Nokia firewalls. This is normal behavior for stateful firewalls. This patch works around this limitation. The comments discuss this as a weakness, suggesting this is buggy behavior. In fact this is a limitation when using stateful firewalls. even stateless firewalls, when with TCP sessions, only the ACK bit is inspected, the ban would take effect once the session has terminated. The ban would take effect at the start of the next session when a SYN packet arrives.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+#
+# This will kill any state for a banned IP. This will ensure the ban actually takes affect.
```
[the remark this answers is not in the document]
VVelox: Interesting. I did not know this bug was more widespread.

And aye, can definitely be re-worded to reflect that.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+# This is need as pf has both the strenght and weakness that changes won't affect current
```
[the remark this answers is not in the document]
VVelox: As in won't take affect in a meaningful manner and/or time frame.

I'm not saying the issue if with PF or the concept, just how it is being used here.

Basically it is saying the way it functions is a double edged sword that posses notable dangers if the abusive states are not killed. There is very good reason for not making it come down immediately, but at the same time time if you don't clean it up post change there is very real risk of continued abuse.

This is what the message is attempting to communicate to the user.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+# states till it is forced to, but unfortunately the only way to make that take affect is
+# to kil them.
+#
+# If some one wishes to fix that, this will requiring writing a tool to grab the session
```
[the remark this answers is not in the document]
VVelox: Hammering all states for that IP instead of just those for a specific port and IP.

Wrote that forgetting about awk off hand. Yeah, definitely fixable and able to hit a specific one.

config/action.d/pf.conf
VVelox: Aye. Was actually meaning to come back to this. Going to remove the braces bit entirely.

The braces should be moved to the init bit, so they are not needed and the user can just give a comma delimited list.

config/action.d/pf.conf
VVelox: This is what the no_kick bit is for.

That said a issue like you mentioned there is easily fixable via just pivoting where you are routing through post leaving the WAP. :P

You are right, that said I disagree about not including it. IMHO you make a great argument for expanding upon it, better listing the failure modes for both to help the user understand when selecting for one.

config/action.d/pf.conf
```diff
@@ -4,16 +4,52 @@
...
+# id for the item in question from 'pfctl -s state -vv'.
+#
+# This is especially important for stateless procols such as UDP as well as TCP protocols
+# that are capable of processing multiple requests in a single session, such as HTTP2.
```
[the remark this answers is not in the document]
VVelox: Aye, was not saying it is layer 4! Brought up HTTP2 as it is a great example of TCP based protocol that the current implementation of the pf filter is lacking when it comes to protecting as the states are hanging around.

But aye, can definitely be rephrased in a way that is more generic.

Basically looking at communicating to the user here that TCP protocols the allow multiple transactions represent a interesting danger here as it allows the attacker to keep the attack going as long as that TCP session is active.

sebres:
> You seem the be the one advocating making it be insecure by default here.

No I'm the one who understand that a workaround (even with many restrictions) is not a solution that make it more secure, rather at end would introduce more issues. This is how abnormal workarounds do.

> > In your case it is "Ban is more important than everything else, even if it'd bother legitimate users".

> Yes. It's called security. 

No it's called "illusive security over everything else" and it is not correct.
First of all you seem not to know what do you talking about... This is how stateful (or even some of stateless) firewalls work.
Many systems exist without real-time monitoring IDS/IPS (say like fail2ban), so later analysis may identify potential intruders and ban them by later occurrence (many times later than a next session) and don't consider such approach as insecure, vice versa it is normal and secure enough for them. 
Many would configure the `maxretry` extremely high, so the ban would take place firstly after 100 or more attempts.
All this are plausible configurations and absolutely justified in certain circumstances.

Absolute security imply that your PC is not connected to the network or even is off. This is also an approach, but not so good sometimes if one needs it for work.

> If you did not want to be dealing with, you could of noted that this is very much a known and expected failure mode at the top of the pf.conf file a long time ago given how long you've been aware of this.

I'm aware of what? That a stateful firewall works as expected and some feature (immediate session termination, what is basically a task of application) cannot be implemented sane using `pf` (or at least I don't know how). Then I am, yes.
And I'm also aware that every action can (and must) be customized for purposes of concrete systems.
In that case it would be enough to specify:
```ini
banaction = pf[actionban='<known/actionban> && pfctl -k <ip>']
```

> Like I said I would be happy to handle the re-factoring that allows it to be ran in a insecure mode. 

No. Again. It is not insecure. Your point of view about the banning behavior and even an approach to kill everything from IP may exist, and it looks probably more secure for you, but has nothing to do with a common security.
Please stop to use populist proverbs or do as you were a security expert, because you are not.

> I guess the big question is are you actually open to getting a fix committed if some one takes the time to create one?

We are. But again, stop to call it "fix", because it is at most a workaround (and controversial in addition).
So a feature, an option, whatever, but not a fix. Savvy?

> It is very frustrating, especially given in that linked to thread, you never though to note the massive possible failure mode for pf at the top of the conf, despite being aware of it. 

It is really enough now. Please stop.

> I understand where you are coming from with wanting it to be able be configured in such a manner, just I think it should secure by default and properly noted so the person setting it up is aware of the security implications are are opting into it.

No you don't. And your approach to consider a dirty workaround as a solution may exist, but what sounds reasonable for small VPS is not necessarily justified for industrial purposes.
Security is **always** a balance between "Everything is closed" (read a sentence above about "Absolute security") and "Spice must flow". I hope the emphasis is clear.

And I'm ready to make proper fix "by default", but not a workaround like this.

VVelox:
> <snip>

I am just going to cut to the chase and point out your first reply was basically very strongly caming across as telling me to screw off it is a known issue because are already aware of it being an issue and every one knows everything you know, so there for why should there ever have to be a comment at the top of the file noting the potential failure mode as was as a potential solution, even if it is not one you would choose.

And I've already conceded it may not be for every one, yet you keep accusing me saying it should be the same for every one. All I've said is the user should have a choice and be aware.

All I wanted to do was fix a failure mode and you told me it was never an issue because it is is common.

Since the word go you've been belittling and insulting the entire time. Basically you've approached some one finding a failure mode and just attacked them from the offset for wanting to fix it while the entire time denying it is a exploitable failure mode on the basis of it being "known" as well as a common one because you know of it.

I'm sorry if I mis-read your initial reply, resulting in the conversation spiraling out and us both getting increasingly needlessly pissy with each other.

I just felt the above would be good to point out with how this all has gone and the like in a attempt to get it more back on track and de-escalate.

As to not making it the default, I will concede if a major version bump is not being done that would be a bad idea to make it the default. This way users don't end up surprised.

On a amusing note, you mentioned IPS, this has been a bit of my work with fail2ban and how I initially found this issue. My previous PR was actually one with a set of filter rules for some common Suricata fast log items based on class etc. That said closed it as while useful, that is being moved to Lilith for the purpose of automated filter generation and will allow far more meaningful control.

VVelox:
@cschuber Dude! Thanks for the input and all done.

The kick mode is also failsafe when using multiport, incase some one passes it a range or non-numeric port. Will add the IP and exit non-zero with a warning message, so it gets logged.

In regards to your comment about the new possible failure mode being able to used in a targeted DOS, I think this honestly results more from network design, IMHO. Basically concentrating services this behavior in which this would be problematic with ones it is not. Would be curious to hear your thoughts on that.

Reworked the doc bit at the part to be more subdued, even though I think this is a notably under appreciated failure mode, especially if wanting to handle UDP as well. Leaves most of the details about what can go wrong with out kicking out.

@sebres If you are happy with how it looks, I can begin updating the tests.

config/action.d/pf.conf
```diff
@@ -64,9 +79,24 @@ actioncheck = <pfctl> -sr | grep -q <tablename>-<name>
...
+#          <port>  ports being banned, if using multiport action
 # Values:  CMD
 #
-actionban = <pfctl> -t <tablename>-<name> -T add <ip>
+actionban = <pfctl> -t <tablename>-<name> -T add <ip>; if [ "<kick>" = "true" ]; then
+           RANGE_TEST=`echo "<port>" | sed "s/://"`
+           NUM_TEST=`echo "<port>" | sed "s/[A-Za-z]//"`
+           if [ "<actiontype>" = "any" ]; then
+               pfctl -k <ip>
+           elif [ "$RANGE_TEST" != "<port>" ]; then
```
sebres, line 90: I don't think it is good idea to validate action parameters here (every time for every ban)...
1. Banning actions are not supposed to produce an error (unfortunately we don't have currently a facility to validate parameters).
2. There is `actionstart` which will be executed once

sebres (changes requested):
It looks good to me (excepting mentioned point about parameter validation)...
But I don't have pf (and no pf-kernel capable boxes at the moment to hand to check it)...
May be @cschuber or someone of @fail2ban/contributors can check/review that.
Thx.

szepeviktor:
Do we have current connection killing for other actions?
I use [ss](https://manpages.debian.org/bullseye/iproute2/ss.8.en.html) to kill connections.

sebres:
> Do we have current connection killing for other actions?

Yes, for example ufw https://github.com/fail2ban/fail2ban/blob/bbfff1828061514e48395a5dbc5c1f9f81625e82/config/action.d/ufw.conf#L35-L43
(see #3018)

Just here it is pf-kernel stuff.

opoplawski:
I have a setup where internal fail2ban hosts apply the bans to a pfsense firewall.  I recently added `/sbin/pfctl -k` to that as well, so it's definitely a useful thing to do.

config/action.d/pf.conf
VVelox: Ohh! Yeah. That would definitely be a problem.

So better to fail during action start then.

Fewer context switches as well.

karolyi:
not sure if it helps anybody, but I'm preparing something for myself exactly on this front. only I do it by filtering the PF states and killing the sessions by id.

something along the lines of:
```
pfctl -s states -vv|PORT=443 grep -A 4 -E ":$PORT) <- .*:ESTABLISHED"
```

then to take the `id: XXX` lines and kill them with `pfctl -k id -k XXX` one by one.

that's an external script that gets executed as a separate action, and will log to syslog w/ `logger`.

Issue #881, opened 2014-12-07
fail2ban network


szepeviktor:
The same botnet is attacking different servers administered by me.
Is there a way to syncronize banned IPs?
Thus banning the other server's attackers.

grinapo:
I don't want to disturb the waters but I usually prefer _secure_, _reliable_ and _flexible_ transport between servers. There are a few message passing solutions (esp. if you're already possess some clustering running) but the simplest out of the top of my head is serf (https://www.serfdom.io/) where you basically setup a mesh of serf nodes (either autodetected or manual, multicast or unicast) then use the client a a few simple "scripts" to broadcast bans and unbans. Doesn't care if servers are offline, can queue up messages or not, etc. And pretty simple to set up.
[Note to self: someday I shoulda follow my own advice and set it up. :-P]

buanzo:
serf is quite awesome indeed.

but if you need to keep things simple, and also if you do not necessarily
have an established relationship between nodes (for instance,
self-organized attack prevention/detection human group), then
secure/reliable/flexible is https.

leeclemens:
I'd like to add that it is not that two nodes don't trust each others' identities (https - Authentication), but that they need to determine their Authorization (to 'post' data).

leeclemens:
The nsupdate action may meet the goals, secure (uses a tsig for auth), reliable, flexible, https://github.com/fail2ban/fail2ban/pull/925

klepsydra:
My solution to this was simpler than a privately distributed p2p ban list.

My script simply applies the global attacker lists (sources are configurable, using a list of URLs) which my servers using fail2ban also contribute to as well via my blocklist.de account:

https://gist.github.com/klepsydra/ecf975984b32b1c8291a

So in that sense the whole world is my P2P fail2ban network. 

If other servers around the world get attacked, and the global blacklists find out about their IP addresses, then my servers will also block that addresses, for at least a day, until they get cleared from the global blacklists.

(BTW, the folks who took notice at Serverfault.com couldn't seem to comprehend this situation.
http://serverfault.com/questions/647891/make-fail2ban-use-public-blacklists )

szepeviktor:
@klepsydra What had you got -5 on serverfault?

klepsydra:
Apparently the question landed in a forum I thought was for network/system admins, but not really. Did you read their comments etc. to find out why?

szepeviktor:
After reading those I setup somethin like an RBL but text based.
[...]

PR #3407, opened 2022-11-16
Create crowdsec.conf

ne20002:
Hi @sebres 

This is a draft for the crowdsec agent. Any feedback welcome.

### DIFF
--- a/config/action.d/crowdsec.conf
+++ b/config/action.d/crowdsec.conf
@@ -1,0 +1,115 @@
+# Crowdsec configuration file
+#
+# Action to report IP address to Crowdsec local API
+#
+# NOTE: These reports may include sensitive Info.
+#
+# IMPORTANT: This action relies on a credentials being added to the above action conf,
+# and the appropriate categories set.
+#
+# Example, for ssh bruteforce (in section [sshd] of `jail.local`):
+#   action = %(known/action)s
+#            crowdsec[lapi_url="http://local-api:8080", machine_id="ReverseProxy", password="******"]
+#
+
+[INCLUDES]
+
+#before = mail-whois-common.conf
+
+[Definition]
+ 
+# bypass action for restored tickets
+norestored = 1
+ 
+# Option:  actionstart
+# Notes.:  command executed on demand at the first ban (or at the start of Fail2Ban if actionstart_on_demand is set to false).
+# Values:  CMD
+#
+actionstart =
+ 
+# Option:  actionstop
+# Notes.:  command executed at the stop of jail (or at the end of Fail2Ban)
+# Values:  CMD
+#
+actionstop =
+ 
+# Option:  actioncheck
+# Notes.:  command executed once before each actionban command
+# Values:  CMD
+#
+actioncheck =
+ 
+# Option:  actionban
+# Notes.:  command executed when banning an IP. Take care that the
+#          command is executed with Fail2Ban user rights.
+#
+# Tags:    See jail.conf(5) man page
+# Values:  CMD
+#
+actionban = WHOIS_RESULT=$(whois <ip>)
+            WHOIS_COUNTRY=$(echo "$WHOIS_RESULT" | grep -a -i -m 1 Country | sed -n -e 's/^.*[Cc]ountry:\s*//p')
+            WHOIS_ASN=$(echo "$WHOIS_RESULT" | grep -a -i Origin | tail -1 | sed -n -e 's/^.*[Oo]rigin.*:\s*AS//p')
+            WHOIS_ASNAME=$(echo "$WHOIS_RESULT" | grep -a netname | tail -1 | sed -n -e 's/^.*netname:\s*//p')
+            TIME=$(LC_ALL=C date -d "@<time>" +"%%Y-%%m-%%d, %%H:%%M:%%S")
+            JSON_STRING=$(/usr/bin/jq -s -R -n \
+                  --arg ip "<ip>" \
+                  --arg scenario "fail2ban/<name>" \
+                  --arg msg "Ip <ip> performed illegal connection attempt." \
+                  --arg bantime "<bantime>s" \
+                  --arg ti "$TIME" \
+                  --arg cn "$WHOIS_COUNTRY" \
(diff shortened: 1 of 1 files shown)


ne20002:
> Can you clarify please the field of application for that (necessity to report IP address to Crowdsec)... Are you using two IPS systems together? Why?

I'm indeed using two IPS. But not on the same machine.

Crowdsec consists of agents detecting ips wihich then give this information enriched with other information to the Crowdsec local API. The local APi does not need to be running on the same machine. The local API uploads detected ips to Crowdsec central API from which it also retrieves a list of checked ips detected by other users. The blocking is done be bouncers getting the list of ips to block from the local API.

In my case I have a local API running, a bouncer on my edge router and agents on two Nginx server. Due to missing availablility of a recent agent implementation for Debian stable I still have fail2ban running on my reverse proxy.
Also, I feel the agents a bit heavy (with the enriching). So I figured this out to send the information about ips detected by fail2ban to my local API, which then updates the blocklist on the router (and therefor blocks the ips for all open systems in the network) and also uploads the info to Crowdsec central API for sharing.
I see this a bit like sending the information about blocked ips to abuseipdb (which is also a fail2ban action).

At the moment I have about 18k ips in the bouncer blocklist retrieved form central API from which 80 - 300 really get blocked by the bouncer and this only on port 80, 443). I also have around 30 ips not blocked by Crowdsec but detected by fail2ban because they are blocked on the reverse proxy. Those are definitly crawlers, spider or whatever you call em. So I felt it to be good idea to report them to Crowdsec central API.

_Using fail2ban as an agent for Crowdsec does possibly not follow the itentional idea of fail2ban nor Crowdsec. But in my case it closes a gap so I consider it a valid use case._

I'm sure the action can be optimized.

ne20002:
> Regarding the action, I don't think it is justified to use whois in that manner:

Isn't this what e.g. the mail action does? Calling whois and put the info into the mail body for each blocked ip. Look to me very similar.

ne20002:
Actually yes. I believe for the use case I have (add ips detected by fail2ban to Crowdsec local api for blocking) the whois is not needed, it will work with ip only.
It's just not showing the country information in Crowdsec metrics or dashboard.
Maybe this is a way to go for this agent?

And yes: Crowdsec does it's own blocking. This is what I want to make use of. Crowdsec is designed to take information from multiple sources to create blocklists for bouncers which may run on firewall devices or e.g. an Nginx. The sharing is the part fail2ban does not allow. 
So with Crowdsec I can run mutiple agents on different services/servers I have and share the bad ip detection and blocking decission by blocking on my edge router for all services all together. And also share the information with the Crowdsec community.

I now like to use fail2ban as a Crowdsec agent to feed the information about the detected ip to the Crowdsec system.

sebres:
> Maybe this is a way to go for this agent?

Dunno. Alternatively one can do the whois lookup as an option.

> The sharing is the part fail2ban does not allow.

At the moment, yes. There is an RFE #881 for that.

ne20002:
I'll suggest to split this into two agents: one simple for usage without whois plus a second one with whois (similar to mail agents).
I feel a need for a way to use fail2ban with Crowdsec. To outnumber the bad I (as a user) appriciate if I have the option to use both systems working together to give best results.

sebres:
> I'll suggest to split this into two agents: one simple for usage without whois plus a second one with whois (similar to mail agents).

No. I don't like it for many reasons, especially because of code duplication and necessity to support/maintain 2 actions hereafter...
Sure we have many mail-actions but it is historical and I don't think it is justified to continue this handling further.

Let alone the way with option is pretty simple here:
```ini
[Definition]

...

# Option `lookup` (default `none`) - supply `whois` to perform a lookup for country and ASN using whois command
lookup = none

actionban = <lookup-<lookup>/lookupinfo>
            TIME=$(LC_ALL=C date -d "@<time>" +"%%Y-%%m-%%d, %%H:%%M:%%S")
            JSON_STRING=...
            ...


[lookup-none]
lookupinfo = WHOIS_COUNTRY=''
            WHOIS_ASN=''
            WHOIS_ASNAME=''

[lookup-whois]
lookupinfo = WHOIS_RESULT=$(whois <ip>)
            WHOIS_COUNTRY=$(echo "$WHOIS_RESULT" | grep -a -i -m 1 Country | sed -n -e 's/^.*[Cc]ountry:\s*//p')
            WHOIS_ASN=$(echo "$WHOIS_RESULT" | grep -a -i Origin | tail -1 | sed -n -e 's/^.*[Oo]rigin.*:\s*AS//p')
            WHOIS_ASNAME=$(echo "$WHOIS_RESULT" | grep -a netname | tail -1 | sed -n -e 's/^.*netname:\s*//p')
```
Moreover it makes possible to extend the action with new ways implementing lookup using other facilities (e.g. GeoIP databases or SaaS's, etc).

And configuration for that would be also very simple (since one needs to specify other parameters anyway):
```ini
action = %(known/action)s
         crowdsec[lookup=whois, other_params=...]
```
(or even set `lookup` in `action.d/crowdsec.local`)

*Although as already said if I were you, I'd still prefer pythonic action...*

ne20002:
> Although as already said if I were you, I'd still prefer pythonic action...
Didn't really get this. I'm no experienced in phyton.

ne20002:
I changed the action so that call of whois is optional (and off by default). 

In terms of Crowdsec the Crowdsec agents install geoip db with em to do the resolution. Crowdsec considers this information as relevant for finding decission on ban or not to ban. The info is also forwarded to the Crowdsec local API for statistics and display. 
As for fail2ban the decission to ban is already done when calling the action, the geo information is not needed. Also, Crowdsec seems to work an enhancements on local API side to take away such tasks as geoip lookup if not needed on the agent in future.

I'll test if I find a way to at least put the crowdsec calling part into background. I do agree with your statements made in #3422 about fail2ban to be as lightweight as possible. But for DOS or DDOS situations with lots of ips to ban I beleive this is something that is not fail2ban to handle. Usually fail2ban is acting on log entries which usually is after a request is processed. For heavy requests situations rate limiting on a reverse proxy seems much more suitable.

PR #3410, opened 2022-11-19
Fixed the pattern to work with Dante v1.4.2 and tested on the real Dante instance under Ubuntu 22.04.1 LTS

maximmasiutin:
Fixed the pattern to work with Dante v1.4.2 and tested on the real Dante instance under Ubuntu 22.04.1 LTS.

Please see the real examples of the log lines (danted is configured to log to syslog (journalctl)), the hostnames, usernames and IPs are redacted:
```
Nov 19 21:03:46 srv danted[3360179]: info: block(4): tcp/accept ]: 198.51.100.55.58211 198.51.100.83.52483
Nov 19 21:12:57 srv danted[3374579]: info: block(1): tcp/accept ]: 212.28.70.169.8490 198.51.100.83.52483: error after reading 31 bytes in 2 seconds: pam_authenticate() for user "socks5_user" failed: Authentication failure
Nov 19 21:12:57 srv danted[3374579]: info: block(1): tcp/accept ]: 212.28.70.169.8494 198.51.100.83.52483: error after reading 3 bytes in 1 second: eof from local client
```
The string 2 and 3 did match (blocking by invalid user or other failure after connecting), but the string 1 (blocking by IP) not. Therefore, this pull request addresses the issue.

### DIFF
--- a/config/filter.d/dante.conf
+++ b/config/filter.d/dante.conf
@@ -1,7 +1,8 @@
 # Fail2Ban filter for dante
 #
-# Make sure you have "log: error" set in your "client pass" directive
-#
+# Make sure you have "log: error" set in your dante config "client pass" directive to ban by invalid credentials,
+# and "log: connect error" in your "client block" directive to ban by invalid IPs
+
 
 [INCLUDES]
 before = common.conf
@@ -9,7 +10,7 @@ before = common.conf
 [Definition]
 _daemon = danted
 
-failregex = ^%(__prefix_line)sinfo: block\(1\): tcp/accept \]: <HOST>\.\d+ [\d.]+: error after reading \d+ bytes? in \d+ seconds?: (?:could not access |system password authentication failed for )user "<F-USER>[^"]+</F-USER>"
+failregex = ^%(__prefix_line)sinfo: block\(\d+\): tcp/accept \]: <HOST>\.\d+ [\d.]+(?:: error after reading \d+ bytes? in \d+ seconds?: (?:could not access |system password authentication failed for )user "<F-USER>[^"]+</F-USER>")?
 
 [Init]
 journalmatch = _SYSTEMD_UNIT=danted.service


sebres:
> The string 2 and 3 did match (blocking by invalid user or other failure after connecting)

No it does not. 
There is no match for `pam_authenticate()` for the 2nd and no `could not access` or `system password authentication failed` for 3rd (which were mandatory previously). 
But I'm unsure the 3rd failure must be matched at all, rather I guess it should not (it looks to me like the same attempt - the time and a session ID 3374579 are equal for 2 and 3).

To match 2nd line, the filter needs at least:
```diff
- failregex = ^%(__prefix_line)sinfo: block\(\d+\): tcp/accept \]: <HOST>\.\d+ [\d.]+: error after reading \d+ bytes? in \d+ seconds?: (?:could not access |system password authentication failed for )user "<F-USER>[^"]+</F-USER>"
+ failregex = ^%(__prefix_line)sinfo: block\(\d+\): tcp/accept \]: <HOST>\.\d+ [\d.]+: error after reading \d+ bytes? in \d+ seconds?: (?:could not access |(?:system password authentication failed|pam_authenticate\(\)) for )user "<F-USER>[^"]+</F-USER>"
```
(although a bit weird logging "wording", since the word "failed" jumps left and right around "for user",
compare "system password authentication **failed** for user X" vs. "pam_authenticate for user X **failed**")

> but the string 1 (blocking by IP) not.

Are there really no other errors for the session (3360179) with first attempt? It is a bit strange that only the IPs are logged (and there is no message after IP at all).

> Therefore, this pull request addresses the issue.

Well, the PR makes the part of message after IPs optional, so now would match everything in the log in following format:
```
Nov 19 21:12:57 srv danted[123456]: info: block(1): tcp/accept ]: 192.0.2.111.1234 192.0.2.5.5555: no matter which message is here
```
(emphasis is at "no matter which message is here")

So I don't think it is good idea to do that.

maximmasiutin:
@sebres - thank you for your attention! I may send you real Dante logs by email if you wish. Feel free to decide what to do with the pull request, I may always edit the config to accommodate my needs. Anyway, I hope that the full log might have been useful for you.

sebres:
Partially done in 3190febb2723bca7cd38b6086751b26bdb7e0e68 - 2nd line from your excerpt matched now.
I don't think 3rd line should be matched at all (e. g. it's an after effect of same failure).
As for the first line - it doesn't look like authentication failure, therefore if it should be matched, this must be added for mode `ddos` or `aggressive` only. 
Thus close this.

maximmasiutin:
@sebres Thank you very much!

PR #3421, opened 2022-11-28
Fixed docs: Added missing escape backslash in documentation tip

mccoderpy:
Added missing \\ in documentation tip
image

### DIFF
--- a/FILTERS
+++ b/FILTERS
@@ -270,7 +270,7 @@ Take note of  -l heavydebug  / -l debug  and -v as they might be very useful.
      call graphs and can browse different versions.
 
 .. TIP::
-     Some applications log spaces at the end. If you are not sure add \s*$ as
+     Some applications log spaces at the end. If you are not sure add \\s*$ as
      the end part of the regex.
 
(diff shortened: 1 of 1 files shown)


sebres:
Where did you get that screenshot? 
I'm unsure FILTERS file is supposed to be markdown formatted. 
But if yes (or we'd allow that):
- either it must get more fixes (there are several other tokens with a single backslash)
- I'd like to use enclose the code in \`, so to have `\s` instead of \\s either
- it must probably get `<!-- vim: syntax=Markdown -->` (like our ChangeLog) or renamed to FILTERS.md

mccoderpy:
> Where did you get that screenshot? I'm unsure FILTERS file is supposed to be markdown formatted. But if yes (or we'd allow that):
> 
> * either it must get more fixes (there are several other tokens with a single backslash)
> * I'd like to use enclose the code in `, so to have `\s` instead of \s either
> * it must probably get `<!-- vim: syntax=Markdown -->` (like our ChangeLog) or renamed to FILTERS.md

I took the screenshot directly from https://fail2ban.readthedocs.io/en/latest/filters.html

PR #3469, opened 2023-02-28
New filter: routeros-auth.conf

### LINKED ISSUE #3458: [FR]: MikroTik RouterOS login failure
Add filter for MikroTik RouterOS logs, detecting login failures.

The MikroTik RouterOS enables sending logs to a remote syslog server, where the analysis can be performed by fail2ban.

Entries like:
```
Feb 15 11:29:03 gw.local system,error,critical login failure for user admin from 192.168.33.2 via web
Feb 15 11:57:42 gw.local system,error,critical login failure for user  from 2001:470:1:c84::24 via ssh
```

(Note the empty username in the second row.)

vitkabele:
Add filter to detect failed login attempts in the log produced by MikroTik RouterOS

- Add test for new filter
- Mention the new filter in jail.conf

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -958,6 +958,9 @@ port    = http,https
 logpath = %(syslog_authpriv)s
 backend = %(syslog_backend)s
 
+[routeros-auth]
+port = ssh,http,https
+logpath = /var/log/MikroTik/router.log
 
 [zoneminder]
 # Zoneminder HTTP/HTTPS web interface auth
--- a/fail2ban/tests/files/logs/routeros-auth
+++ b/fail2ban/tests/files/logs/routeros-auth
@@ -1,0 +1,15 @@
+# RouterOS v7.5
+# failJSON: { "time": "2005-02-15T11:25:46", "match": true , "host": "192.168.88.6", "user": "admin" }
+Feb 15 11:25:46 gw.local system,error,critical login failure for user admin from 192.168.88.6 via web
+
+# RouterOS v7.5
+# failJSON: { "match": false }
+Feb 15 11:26:15 gw.local system,info log rule changed by admin
+
+# RouterOS v7.5
+# failJSON: { "time": "2005-02-15T11:57:42", "match": true , "host": "2001:470:1:c84::24", "user": "" }
+Feb 15 11:57:42 1234.hostname.cz system,error,critical login failure for user  from 2001:470:1:c84::24 via ssh
+
+# RouterOS v7.5
+# failJSON: { "time": "2005-03-02T09:09:46", "match": true , "host": "1.2.3.4", "user": "user with space" }
+Mar  2 09:09:46 gw.local system,error,critical login failure for user user with space from 1.2.3.4 via ssh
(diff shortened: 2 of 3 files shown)


config/jail.conf
```diff
@@ -958,6 +958,9 @@ port    = http,https
 logpath = %(syslog_authpriv)s
 backend = %(syslog_backend)s
 
+[routeros-auth]
+port = ssh
```
sebres, line 962: Hmm... the banning port is ssh, but the message in your excerpt says "login failure from ... **via web**".
Doesn't it imply a http/https also? I mean a multi-port banning action will only ban the access from IP to port 22, so possible access to 80/443 are still allowed after ban (also if locally login takes place by ssh internally, but it is allowed due to local IP address, and it is good so).

fail2ban/tests/files/logs/routeros-auth
```diff
@@ -0,0 +1,9 @@
+
+# failJSON: { "match": true , "host": "192.168.88.6" }
+Feb 15 11:25:46 gw.local system,error,critical login failure for user admin from 192.168.88.6 via web
```
sebres, line 3: See the notice above.

config/jail.conf
vitkabele: Ah you are right. I missed that. I will change it to `port = ssh,http,https`.

fail2ban/tests/files/logs/routeros-auth
vitkabele: Anything special to do here besides fixing the port directive in `jail.conf`?

sebres: At first look noting...
Although the construction `(system,?|error,?|critical,?)+` looks a bit weird to me.
Are there really constellation where the log doesn't contain `system,error,critical` exactly, as in your tests? The order may change? Some keyword is optional?
Anyway if wanted such a strange dynamic here, I'd rewrite it like that:
```diff
- ... (system,?|error,?|critical,?)+ ...
+ ... (?:,?(?:system|error|critical)){1,3} ...
```
The question is if it really dynamical (free ordered and optional in parts) and not so important, why just not use `\S+` instead. So it would simply ignore up to first space and search `login failure` hereafter.
Otherwise (e. g. the order is fix, but for instance only `system` and `critical` are optional) I'd simply write something like that instead:
```diff
- ... (system,?|error,?|critical,?)+ ...
+ ... (?:system,)?error(?:,critical)? ...
```
This would avoid cyclic branching, thus whole RE becomes faster, especially because of non-greedy match using `.*?`

sebres: Absolutely no - `.+` is a catch-all, which is:
- really bad before a capturing groups matching foreign input (`.*?` is already bad enough, which is justified because RE is anchored from both sides)
- completely opens a left-anchor - `^.+ login ... $` is almost the same as `login ... $`
- due to unanchored nature makes usage of another catch-all (non-greedy `.*?`) unjustified.

Don't do that, please.

vitkabele: Thank you for the feedback. I went through the documentation once again and collected more log samples from all over the internet.
The `system,error,critical` sequence appears identical in all the logs I found, so my original attempt was probably just an over-engineering. I also extended the example log messages for the test purposes.

I pushed a new version to my branch.

vitkabele:
Hello, could you merge this PR? I'd like to mark it as done.

PR #3472, opened 2023-03-05
add filter to catch env lookups in nginx logs

bensig:

### DIFF
--- a/config/filter.d/nginx-env.conf
+++ b/config/filter.d/nginx-env.conf
@@ -1,0 +1,13 @@
+# fail2ban filter configuration for nginx 
+# searches for any requests for .env in nginx logs
+
+[Definition]
+failregex = ^<HOST>(?::\d+)?\s+.*.env .*
+
+ignoreregex = 
+
+datepattern = {^LN-BEG}
+
+# DEV NOTES:
+# .env files can contain sensitive data... anyone looking for those is prob bad
+# Author: Ben Sigman
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -395,6 +395,10 @@ logpath  = %(nginx_error_log)s
 port    = http,https
 logpath = %(nginx_access_log)s
 
+[nginx-env]
+port    = http,https
+logpath = %(nginx_access_log)s
+
 # Ban attackers that try to use PHP's URL-fopen() functionality
 # through GET/POST variables. - Experimental, with more than a year
 # of usage in production environments.


bensig:
This filter checks nginx logs for anyone searching for `.env` files. Simple, but powerful.

It has been massively successful in blocking scans. Banned 1000's of addresses with this filter - would be good to include in `fail2ban` master so others can use. 

In many cases, webhackers are able to find sensitive info in these `.env` files that are left exposed on servers. Discovered an increase in probes for `.env` files and have been running this on production for about a year now.

config/filter.d/nginx-env.conf
```diff
@@ -0,0 +1,13 @@
+# fail2ban filter configuration for nginx 
+# searches for any requests for .env in nginx logs
+
+[Definition]
+failregex = ^<HOST>(?::\d+)?\s+.*.env .*
```
sebres, line 5: Hmm...
this RE looks a bit vulnerable to me (too many catch-alls), let alone dot-char in `.env` matches any character, so I guess it's not really what you want.

sebres (commented):
Sorry, but I don't think we really need filter for the env-lookup only.
Maybe some generalized filter to search unexpected requests?... 
Something like this:
```ini
[Definition]
_file = [^"]*
_uri = /(?:[^/]*/)*%(_file)s
_bad_req = [^"]*(?=" %(_bad_code)s)
_bad_code = 4(?:00|44)
_req = (?:%(_bad_req)s|[A-Z]+ %(_uri)s [^"]*)
_code = (?!401)[45]\d\d
failregex = ^<ADDR> - \S+ [^"]*"%(_req)s" %(_code)s\b
```
Pros:
- it is a good and safe generalized solution matching also bad (400) and rejected (444) requests.
- one can simply customize it to match specific uris or files, for instance:
```ini
filter = nginx-unexpected-req[_file='\.\w*env']
```
would match what you're suggesting here, but safe against false positives (different .env only and only for expected http-codes, e. g. like not found etc).

Just since monitoring of access-log is not recommended (see [wiki :: Best practice](https://github.com/fail2ban/fail2ban/wiki/Best-practice), especially "Reduce parasitic log-traffic"), I'm unsure such general jail is good idea.
Better would be special nginx-location or conditional access-log (see the [wiki](https://github.com/fail2ban/fail2ban/wiki/Best-practice#reduce-parasitic-log-traffic)), logging only failures, so the jail would monitor almost bad traffic only.

PR #3515, opened 2023-05-22
fail2ban.service: use RuntimeDirectory instead of hardcoding /run

smod:
Also, remove PIDFile option, it's not needed for a Type=simple process run in foreground.

IMHO, fail2ban should not even create the PID file when run with "-f".


sebres:
> fail2ban should not even create the PID file when run with "-f".

Well, fail2ban can be also started from command line (`?sudo? fail2ban-client ...`), also accidentally. So IMHO to avoid such a mistaken "start", usage of pid file is still justified.

PR #3521, opened 2023-06-07
Zoneminder jail.conf and zoneminder.conf update

clipo1979:
Updated Zoneminder rules and regex due to changes in the upstream Zoneminder package starting with version 1.35

### DIFF
--- a/fail2ban/tests/files/logs/zoneminder
+++ b/fail2ban/tests/files/logs/zoneminder
@@ -1,3 +1,15 @@
+ZM >1.35 Succesful login attemp
+
+2023/06/25 05:36:32.232620 web_php[2716946].INF [127.0.0.1] [Login successful for user "test-user1"] at /usr/share/zoneminder/www/api/app/Controller/AppController.php line 86
+
+ZM >1.35 Failed login
+
+2023/06/25 10:32:02.778987 web_php[2958965].ERR [127.0.0.1] [Could not retrieve user test-user2 details] at /usr/share/zoneminder/www/includes/auth.php line 322
+2023/06/25 10:32:57.008059 web_php[2717014].ERR [127.0.0.1] [Could not retrieve user test-uset2 details] at /usr/share/zoneminder/www/includes/auth.php line 322
+
+
+ZM <1.35 Logs
+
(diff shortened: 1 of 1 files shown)


config/jail.conf
```diff
@@ -963,11 +963,22 @@ logpath = %(syslog_authpriv)s
...
+#port    = http,https
+#logpath = %(apache_error_log)s
+
 [zoneminder]
-# Zoneminder HTTP/HTTPS web interface auth
-# Logs auth failures to apache2 error log
+# Zoneminder >1.35 HTTP/HTTPS web interface auth
+# Logs auth failures to Zoneminder web_php.log and apache error log
+# See notes in filter.d/zoneminder.conf
+enabled = false
```
sebres, line 978: `enabled = false` is default for all jails and unneeded here.

config/jail.conf
```diff
@@ -963,11 +963,22 @@ logpath = %(syslog_authpriv)s
...
 port    = http,https
-logpath = %(apache_error_log)s
+logpath = /var/log/zm/web_php.log
```
sebres, line 980: This is basically the single option that changed, right?

config/filter.d/zoneminder.conf
```diff
@@ -1,27 +1,42 @@
...
-#
 # Option:  failregex
-# Notes.:  regex to match the login failure and non-existent user error messages in the logfile.
 
-prefregex = ^%(_apache_error_client)s (?:ERR|WAR) <F-CONTENT>\[(?:Login denied|Could not retrieve).*</F-CONTENT>$
+#Uncomment the below regex for use with ZM < 1.34
+#failregex = ^%(_apache_error_client)s WAR \[Login denied for user "[^"]*"\]
+
+#Uncomment the below regex for use with ZM > 1.34
+failregex = ^\s*web_php\[\d+\]\.ERR \[<HOST>\] \[(?:Could not retrieve user|Login denied for user) \S+
```
sebres, line 27: Since the regex is different, one'd need to adjust the test-cases in [fail2ban/tests/files/logs/zoneminder](https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/zoneminder).

sebres (changes requested):

PR #3523, opened 2023-06-09
action.d/*.conf: moved persisted-storage files to /var/lib/fail2ban directory

drTr0jan:
There's a case when `/var/run` dir is mounted as a tmpfs (ex, diskless environment). So `/var/run/fail2ban` dir can be absent after system boot and some action-script wouldn't read and write their persisted-storages.

The proposed patch changes persisted-storage of `action.d/bsd-ipfw.conf` and `action.d/dummy.conf`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,6 +20,8 @@ ver. 1.0.3-dev-1 (20??/??/??) - development nightly edition
   if available for platform and uses DNS to find local IPv6 as a fallback only
 * improve `ignoreself` by considering all local addresses from network interfaces additionally to IPs from hostnames (gh-3132)
 * `action.d/mikrotik.conf` - new action for mikrotik routerOS, adds and removes entries from address lists on the router (gh-2860)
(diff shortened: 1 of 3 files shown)


sebres:
Well, this is the sense of usage `/var/run` for both actions... Especially because the unban all or flush anyway removes everything and should not leave something after restart.
Let alone the persistence of such entries for `action.d/dummy.conf` (which is test dummy action) is very questionable.
And `actionstop` of `action.d/bsd-ipfw` removes the startstatefile either way.
Anyway at the moment it doesn't no matter where the files are placed at all.

For the case someone needs it anyway for some reason, it is simply configurable in jail.local, for instance:
```ini
banaction = bsd-ipfw[startstatefile=/var/lib/fail2ban/ipfw-started-table_<table>]
```

So I'll close this PR as unwanted.
May be something like that would be nice:
```ini
- startstatefile = /var/run/fail2ban/ipfw-started-table_<table>
+ startstatepath = /var/run/fail2ban
+ startstatefile = <startstatepath>/ipfw-started-table_<table>
```
so a configuration could be shorter:
```ini
banaction = bsd-ipfw[startstatepath=/var/lib/fail2ban]
```

PR #3527, opened 2023-06-13
Fixed startup on debian bookworm (using systemd as sshd_backend)

vafgoettlich:
On Debian Bookworm (and newer) rsyslog is not installed as default anymore:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1018788

Because of that systemd should be preferred if possible. I checked sshd and postfix, both of them are logging to journal (Maybe more, but I am not sure).


alvgalrus:
This is still relevant for Debian 12 / Bookworm.

PR #3532, opened 2023-06-22
Restore bans from the database with full ticket information

lade-odoo:
In the case where we use the `ignoreCommand` with information from the ticket, we want these informations to also be included when restoring the bans from the database (the ticket may no longer be required to be banned).


sebres:
Thx for PR.
Indeed, it's reasonable. Tests passed too, thus will merge it now.

PR #3533, opened 2023-06-22
Allow unbanip of tuple IDs

lade-odoo:
As specified [here], IDs in Fail2ban can be tuples. However, when a tuple ID is banned, there is no way to remove it via the `unbanip` command line. If tried, the tuple will be considered as a list with each element being an ID. Instead, we should try to parse the string to see if it represents an object.

[here]: https://github.com/fail2ban/fail2ban/blob/226a59445a8046b7f86c3bc072be1d78555ccdb0/fail2ban/server/failregex.py#L313


sebres:
Confirmed.
Just `eval` is evil, thus I'd rewrite it somehow, either with `ast.literal_eval` or with RE parsing...
Since `literal_eval` is safe against "injections" and other kind of malicious evaluations, I'd probably prefer that way:
```diff
+ from ast import literal_eval
...
		# IPs can be represented in string format (e.g.: tuples)
-		is_ip_parsed = True
-		if isinstance(ip, str):
+		if isinstance(ip, str) and str[:1] in ("(","["):
			try:
-				ip = eval(ip)
+				ip = literal_eval(ip) 
-			except Exception:
+			except SyntaxError:
-				is_ip_parsed = False
+				pass
...
-			if not is_ip_parsed and not isinstance(ip, IPAddr):
+			if not isinstance(ip, (IPAddr, tuple, list)):
```
However I don't think `actions.py` is the proper module to fix this. Rather it must be fixed somewhere in https://github.com/fail2ban/fail2ban/blob/48c91dfb6bd8478968c226d0c6ca965bdce2768d/fail2ban/server/server.py#L531

sebres (changes requested):
As already wrote... actions doesn't look like appropriate place to fix that.

lade-odoo:
Changed the modification with a small refactoring in order to avoid the recursive method. This would cause issues as it would recurse on the `TUPLE_ID` which must be considered as an `ID` and not a simple list/tuple.

sebres:
Done.
However there is probably still one case to "solve" - `banip`, that must also accept tuple IDs now for consistent reasons.

lade-odoo:
Solved the issue of `banip` and other command that also needed to be adapted.

fail2ban/server/filter.py
```diff
@@ -468,7 +468,7 @@ def performSvc(self, force=False):
 
 	def addAttempt(self, ip, *matches):
 		"""Generate a failed attempt for ip"""
-		if not isinstance(ip, IPAddr):
+		if not isinstance(ip, (IPAddr, list, tuple)):
```
sebres, line 471: Does it really bother somewhere? If so then we'd need to fix that in `IPAddr`, because normally `IPAddr` can safe handle raw values (not IPs, inclusive tuple or lists) without any problem. But has additional facility to check whether it's IP, etc. 
So we'd always prefer `IPAddr` as identifier/IP in internal filters/actions functionality if possible.
```python
>>> from fail2ban.server.ipdns import IPAddr
>>> id = ('A','B')
>>> ip = IPAddr(id); ip
('A', 'B')
>>> ip == id
True
>>> ip.family == IPAddr.CIDR_RAW
True
>>> ip.raw
('A', 'B')
```

fail2ban/server/server.py
```diff
@@ -523,27 +522,19 @@ def getAction(self, name, value):
 	def setBanTime(self, name, value):
 		self.__jails[name].actions.setBanTime(value)
 	
-	def addAttemptIP(self, name, *args):
-		return self.__jails[name].filter.addAttempt(*args)
+	def addAttemptIP(self, name, ip, failures):
```
sebres, line 525: I'm not sure we should change the API here (`*failures` can be supplied similarly from outside).

sebres (commented):
Thx!
I can't approve the changes right now... see my comments.

fail2ban/server/server.py
sebres: at least this would be better:
```diff
- def addAttemptIP(self, name, ip, failures):
+ def addAttemptIP(self, name, ip, *args):
```

sebres:
FWIW, restored as new branch with small review, - [gh-3533--tuple-fid](https://github.com/fail2ban/fail2ban/compare/c54d505dea7c18608481bf481cda88cbf908b61c...sebres:gh-3533--tuple-fid).

PR #3545, opened 2023-07-28
New jail for postfix: screendnsbl 

pbiering:
this extension adds a new postfix jail which can be optionally dedicated enabled.

It is triggered by a log line like

```
Jul  1 03:55:34 xxx postfix/postscreen[188902]: DNSBL rank 6 for [192.0.2.30]:52340
```

It catches a lot of unwanted connections and log volume decreased a lot.

### DIFF
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -52,6 +52,10 @@ mdre-aggressive = %(mdre-auth2)s
 mdpr-errors = too many errors after \S+
 mdre-errors = ^from [^[]*\[<HOST>\]%(_port)s$
 
+# Extra mode "screendnsbl", triggered on postfix/postscreen[<PID>]: DNSBL rank <NUM> for [<HOST>]:<PORT>
+mdpr-screendnsbl = DNSBL rank \d+
+mdre-screendnsbl = for \[<HOST>\]%(_port)s
+
 
 failregex = <mdre-<mode>>
 
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -599,6 +599,16 @@ backend  = %(postfix_backend)s
 maxretry = 1
 
 
+[postfix-screendnbl]
+
+filter   = postfix[mode=screendnsbl]
+port     = smtp,465,submission
+logpath  = %(postfix_log)s
+backend  = %(postfix_backend)s
+maxretry = 1
+bantime  = 4h
+
+
 [sendmail-auth]
 
 port    = submission,465,smtp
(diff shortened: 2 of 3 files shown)


pbiering:
any update on this PR

config/filter.d/postfix.conf
```diff
@@ -52,6 +52,10 @@ mdre-aggressive = %(mdre-auth2)s
 mdpr-errors = too many errors after \S+
 mdre-errors = ^from [^[]*\[<HOST>\]%(_port)s$
 
+# Extra mode "screendnsbl", triggered on postfix/postscreen[<PID>]: DNSBL rank <NUM> for [<HOST>]:<PORT>
+mdpr-screendnsbl = DNSBL rank \d+
+mdre-screendnsbl = for \[<HOST>\]%(_port)s
```
sebres, line 57: The resulting failregex would be unanchored, what is no go for the PRs.
Please change:
```diff
mdre-screendnsbl = for \[<HOST>\]%(_port)s
mdre-screendnsbl = ^for \[<HOST>\]%(_port)s
```

config/filter.d/postfix.conf
```diff
@@ -52,6 +52,10 @@ mdre-aggressive = %(mdre-auth2)s
 mdpr-errors = too many errors after \S+
 mdre-errors = ^from [^[]*\[<HOST>\]%(_port)s$
 
+# Extra mode "screendnsbl", triggered on postfix/postscreen[<PID>]: DNSBL rank <NUM> for [<HOST>]:<PORT>
```
sebres, line 55: I guess it is wrong mode name for DNS-based Blackhole List, correct would be `dnsbl`, but much better why not combine it with `rbl`?

config/jail.conf
```diff
@@ -599,6 +599,16 @@ backend  = %(postfix_backend)s
 maxretry = 1
 
 
+[postfix-screendnbl]
```
sebres, line 602: I'm still unsure it'd really expect a new jail.
I know there is one for RBL, but it is more for backwards compatibility reasons, however obsolete since mode rbl is included in mode normal too.

sebres (changes requested):

config/filter.d/postfix.conf
pbiering: intention was to implement an optional feature, not extending/overwriting existing definitions leading to by others unexpected behavior

config/jail.conf
pbiering: intention was to implement an optional feature, not extending/overwriting existing definitions leading to by others unexpected behavior

config/filter.d/postfix.conf
pbiering: fixed

Issue #927, opened 2015-01-26
Support to ban subnets


K1LLUM1N471:
Hello,

in which case does fail2ban support to ban subnets.
I used the instruction from http://cup.wpcoder.de/fail2ban-ip-blacklist/ and entered the subnet instead of the single IP, but fail2ban seems not to handle them at all.

Kind regards,
K1LLUM1N471

leeclemens:
I'm not sure where on that page it refers to subnets being blacklisted (ignoring /32 as a subnet). Generally only one IP fails to authenticate and subsequently gets blacklisted by fail2ban. Do you have a specific use case you are attempting to achieve? How would the subnet be determined from a single IP failing to log in?

K1LLUM1N471:
There is no specific use case, just a server in a sensitive area of
research and development, which has to be secured. The main reason is,
to block iterated unauthorised access from foreign countries e.g. China,
USA).
The determination of the associating subnet is currently done by hand
(Google + Subnet calculator).

Kind regards,
K1LLUM1N471

Am 28.01.2015 um 01:27 schrieb Lee Clemens:

sebres:
@K1LLUM1N471 I have a branch doing that, at the moment written relative #716 (as part of increment feature - ban-time-incr within factor plugin for observer). It can be easy configured in `[jail]` resp. `[default]` or as include in it, here is an example:
```python
...
# If geo feature enabled (dictionaries "geo.country" or "geo.region" are specified):
#   - for each failure this factor appears like a simple "divider" for "maxretry" (inside of "findtime" interval);
#   - for time of each ban it is a simple multiplier coefficient besides "bantime.factor" ("jail.Factor"), 
#     to change default behavior use "ban.Factor" in expression "bantime.formula".

# The larger value of factor leads to faster ban (few failures) and longest time of ban by increase.

# "geo.country" - dictionary to define factor by country of IP address (country:factor pair)
#               (using of this method needs a Country, Region, or City database)
#geo.country =
geo.country = default:10
              DE:1 RU:2
...
```
Example above means - the factor for each country is 10, however for DE it is 1 and RU it is 2.

This feature is configurable to use `geoip` databases or cymru-like dns resolvers...

I will commit it this or next week and let know if as far as.

leeclemens:
[...]

PR #3291, opened 2022-05-30
nftables.conf - add support for cidr notation

ttyS4:
Currently when trying to add an address like: 141.98.11.0/24 it fails with:

```
fail2ban.utils          [720]: ERROR   7fe8c36f6630 -- exec: nft add element inet f2b-table addr-set-custom \{ 141.98.11.0/24 \}
fail2ban.utils          [720]: ERROR   7fe8c36f6630 -- stderr: "Error: You must add 'flags interval' to your set declaration if you want to add prefix elements"
```

After adding 'flags interval' one can ban ranges now as expected.

Before submitting your PR, please review the following checklist:

- [ ] **CHOOSE CORRECT BRANCH**: if filing a bugfix/enhancement
      against certain release version, choose `0.9`, `0.10` or `0.11` branch,
      for dev-edition use `master` branch
- [ ] **CONSIDER adding a unit test** if your PR resolves an issue
- [ ] **LIST ISSUES** this PR resolves
- [ ] **MAKE SURE** this PR doesn't break existing tests
- [ ] **KEEP PR small** so it could be easily reviewed.
- [ ] **AVOID** making unnecessary stylistic changes in unrelated code
- [ ] **ACCOMPANY** each new `failregex` for filter `X` with sample log lines
      within `fail2ban/tests/files/logs/X` file

PR #3573, opened 2023-09-15
feat: support interval in nftables.conf

alexgarel:
Adding the "flags interval;" in addr-set definition enables the manual ban of ip addresses interval using a subnet mask (eg: 10.0.0.0/24).

Relates to: 
- #927


sebres:
Duplicate of #3291

alexgarel:
Oups ! Sorry @sebres.

PR #3580, opened 2023-09-24
GHA: update GitHub action

stweil:
- remove two jobs for Python 3.6 (no longer supported)
- update versions for used actions
- remove whitespace at line ending


stweil:
There is no need to waste processing time on Python 3.6 which is no longer supported.

sebres:
There is no reason to remove it, is there one?
Think about boxes they never get an update for 3.6...
The change would be justified if support of 3.6 should bother somewhere, does it... 

Also you removed pypy which we still official support, why?

stweil:
Python 3.6 reached upstream end-of-life on December 23rd, 2021, and no longer receives security or bug fixes. Apps still using it contain potential security vulnerabilities. Boxes which still depend on Python 3.6 typically will contain more old software with severe security issues. But they still can use all tagged versions of fail2ban (although I doubt that will help much).

For Git master there is simply no reason why Python 3.6 should be supported. Removing it saves resources and is a necessary step to allow an update of the CI runner to a recent Ubuntu version which no longer provides Python 3.6.

The current PyPI configuration "pypy3" tests Python 3.6, too, so there are currently two CI jobs with Python 3.6, and I suggest to drop both for Git master (not for the branches!). The action setup-python@v4 no longer supports "pypy3". If a replacement is desired for Git master, I suggest to add a more specific "pypi3.10" (or any other release).

sebres:
> recent Ubuntu version which no longer provides Python 3.6.

And the reason to switch to recent Ubuntu version is? (I know at some point it'd be necessary but...)

> I suggest to add a more specific "pypi3.10" (or any other release).

Agree. Can you do that in this PR?

.github/workflows/main.yml
```diff
@@ -22,15 +22,15 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, '3.10', '3.11', pypy3]
+        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
```
sebres, line 25: some pypy version would be nice

sebres (changes requested):

stweil:
> Agree. Can you do that in this PR?

There are several [cached releases in the Ubuntu runner](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#pypy). I suggest to select one or two of those. Which ones would you prefer?

[More pypy releases](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#pypy) are supported, but they might require more processing time because they are not cached..

stweil:
I now updated the PR with a new job for pypy3.10.

sebres (approved):

PR #3589, opened 2023-10-02
Fix ipset fails when fail2ban restarts

tomers:
Fixes fail2ban fails when restarting, due to the ipset being persistent, and the `actionstart` command tries to recreate existing set, and fails since this set already exists.


sebres:
> Fixes fail2ban fails when restarting, due to the ipset being persistent

How so? 
Where is it persistent? 
It looks either like a misunderstanding or misconfiguration to me.
Because corresponding this:
https://github.com/fail2ban/fail2ban/blob/8eac5f597aaf0c993f4ba006beeeafa28b18bafe/config/action.d/iptables-ipset-proto4.conf#L46

the set gets removed (`actionstop` will be invoked by restart too).

As for misconfiguration, for instance it may be mistakenly supplied common `<name>` for two different jails.
Or some error by stop that avoid the set removal.

Neustradamus:
@tomers: Have you seen @sebres comment?

PR #3590, opened 2023-10-03
filter.d/postfix.conf: add Sender address rejected: Malformed DNS server reply

sharewax:
If we have some mailformed MX record like:
warning: malformed domain name in resource data of MX record for exmaple.com
```
;; ANSWER SECTION:
example.com.	120	IN	MX	10 .
```
and someone put somewhere your host as MX so you will receive a lot of connections for it.


sebres (commented):
Can you provide a message example (every RE expects at least one test in f2b)... or better add a sample with JSON to [fail2ban/tests/files/logs/postfix](https://github.com/sharewax/fail2ban/edit/patch-1/fail2ban/tests/files/logs/postfix?pr=%2Ffail2ban%2Ffail2ban%2Fpull%2F3590).

Neustradamus:
@sharewax: Any news on your PR?

sharewax:
It seems that I've added needed info, but can't find how to squash all commits in one :( Should I do it on my side or it's possible to do it from yours? Thank you in advance!

PR #3597, opened 2023-10-06
Fix Dropbear filter when logging to STDOUT

MichaIng:
Since Debian Bookworm, the distribution ships Dropbear with a native systemd service instead of the default upstream init.d service, and accordingly uses the `-F` and `-E` flags, to run it in foreground and have it logging to STDOUT instead of syslog.

As usual, timestamps and also the PID are now included by the log message emitted by Dropbear, in addition to the systemd journal log prefix.

The Dropbear filter hence does not match anymore. This commit adds the PID and timestamp as optional pattern between prefix and fail log text, to support Dropbear on Debian Bookworm and newer (and likely new versions of other distros) without breaking the old pattern when running Dropbear without `-E` flag.

Additionally, for performance reasons, this commit adds a `journalmatch` entry, matching Debian's and Fedora's `dropbear.service` with `dropbear` executable/identifier, the most likely match for a Dropbear systemd service.

- [x] I can add a line to https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/dropbear, but is there a way to get this `failJSON`?

### DIFF
--- a/config/filter.d/dropbear.conf
+++ b/config/filter.d/dropbear.conf
@@ -23,13 +23,15 @@ before = common.conf
 
 _daemon = dropbear
 
-prefregex = ^%(__prefix_line)s<F-CONTENT>(?:[Ll]ogin|[Bb]ad|[Ee]xit).+</F-CONTENT>$
+prefregex = ^%(__prefix_line)s(\[\d+\] [A-Z][a-z]+ \d\d \d\d:\d\d:\d\d )?<F-CONTENT>(?:[Ll]ogin|[Bb]ad|[Ee]xit).+</F-CONTENT>$
 
 failregex = ^[Ll]ogin attempt for nonexistent user ('.*' )?from <HOST>:\d+$
             ^[Bb]ad (PAM )?password attempt for .+ from <HOST>(:\d+)?$
             ^[Ee]xit before auth \(user '.+', \d+ fails\): Max auth tries reached - user '.+' from <HOST>:\d+\s*$
 
 ignoreregex = 
+
+journalmatch = _SYSTEMD_UNIT=dropbear.service + _COMM=dropbear
 
 # DEV Notes:
 #
--- a/config/filter.d/dropbear.conf
+++ b/config/filter.d/dropbear.conf
@@ -23,7 +23,7 @@ before = common.conf
 
 _daemon = dropbear
 
-prefregex = ^%(__prefix_line)s(\[\d+\] [A-Z][a-z]+ \d\d \d\d:\d\d:\d\d )?<F-CONTENT>(?:[Ll]ogin|[Bb]ad|[Ee]xit).+</F-CONTENT>$
+prefregex = ^%(__prefix_line)s(?:\[\d+\] \w{2,3} [\d:\s]+)?<F-CONTENT>(?:[Ll]ogin|[Bb]ad|[Ee]xit).+</F-CONTENT>$
 
 failregex = ^[Ll]ogin attempt for nonexistent user ('.*' )?from <HOST>:\d+$
             ^[Bb]ad (PAM )?password attempt for .+ from <HOST>(:\d+)?$
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,8 @@ ver. 1.1.1-dev-1 (20??/??/??) - development nightly edition
 * `filter.d/freeswitch.conf` - bypass some new info in prefix before [WARNING] (changed default `_pref_line`),
   FreeSWITCH log line prefix has changed in newer versions (gh-3143)
 * `filter.d/postfix.conf` - consider CONNECT and other rejected commands as a valid `_pref` (gh-3800)
+* `filter.d/dropbear.conf`:
+  - recognizes extra pid/timestamp if logged into stdout/journal, added `journalmatch` (gh-3597)
 * `filter.d/recidive.conf` - restore possibility to set jail name in the filter, _jailname is positive now (gh-3769)
 * `filter.d/roundcube-auth.conf` - improved RE better matching log format of roundcube version 1.4+ (gh-3816)
 * `filter.d/sshd.conf`:
--- a/fail2ban/tests/files/logs/dropbear
+++ b/fail2ban/tests/files/logs/dropbear
@@ -13,3 +13,6 @@ Jul 27 01:04:12 fail2ban-test dropbear[1335]: Bad password attempt for 'root' fr
 Jul 27 01:04:22 fail2ban-test dropbear[1335]: Exit before auth (user 'root', 10 fails): Max auth tries reached - user 'root' from 1.2.3.4:60588
 # failJSON: { "time": "2005-07-27T01:18:59", "match": true , "host": "1.2.3.4" }
 Jul 27 01:18:59 fail2ban-test dropbear[1477]: Login attempt for nonexistent user from 1.2.3.4:60794
+
+# failJSON: { "time": "2005-07-10T23:53:52", "match": true , "host": "1.2.3.4" }
+Jul 10 23:53:52 fail2ban-test dropbear[825]: [825] Jul 10 23:53:52 Bad password attempt for 'root' from 1.2.3.4:52289
--- a/fail2ban/tests/files/logs/dropbear
+++ b/fail2ban/tests/files/logs/dropbear
@@ -14,5 +14,5 @@ Jul 27 01:04:22 fail2ban-test dropbear[1335]: Exit before auth (user 'root', 10 
 # failJSON: { "time": "2005-07-27T01:18:59", "match": true , "host": "1.2.3.4" }
 Jul 27 01:18:59 fail2ban-test dropbear[1477]: Login attempt for nonexistent user from 1.2.3.4:60794
 
-# failJSON: { "time": "2005-07-10T23:53:52", "match": true , "host": "1.2.3.4" }
+# failJSON: { "time": "2005-07-10T23:53:52", "match": true , "host": "1.2.3.4", "desc": "extra pid/timestamp may be logged into journal, gh-3597" }
 Jul 10 23:53:52 fail2ban-test dropbear[825]: [825] Jul 10 23:53:52 Bad password attempt for 'root' from 1.2.3.4:52289


sebres:
> I can add a line to fail2ban/tests/files/logs/dropbear, but is there a way to get this failJSON?

The captured values from following lines must simply match the parameter in JSON (time, host etc), so just copy-paste it from another lines and adjust the values inside to the values of your line...
By time it is important to consider that if the line doesn't contain the year or TZ, the test suite is working in
https://github.com/fail2ban/fail2ban/blob/8170e9fe75fd2c2c4c51a1d9972b683401cddccb/fail2ban/tests/utils.py#L52
what is basically `So Aug 14 12:00:00 CEST 2005`, so it can be 2004 (if date without year is before that time) and/or in CET if the TZ is different at parsed timestamp.

Or just add it to this PR (or to file direct without failJSON), I'd extended it by myself.

MichaIng:
Not sure whether I understood in which case it can be 2004. But following the other entries, I think it should work 😅.

While producing those, I recognised two other issues:
```
Jul 10 23:57:29 VM-Bookworm dropbear[825]: [825] Jul 10 23:57:29 Bad password attempt for 'root' from 192.168.1.3:52289
Jul 10 23:57:29 VM-Bookworm dropbear[825]: [825] Jul 10 23:57:29 Exit before auth from <192.168.1.3:52289>: (user 'root', 10 fails): Max auth tries reached - user 'root'
Jul 10 23:58:41 VM-Bookworm dropbear[826]: [826] Jul 10 23:58:41 Login attempt for nonexistent user
Jul 10 23:59:24 VM-Bookworm dropbear[826]: [826] Jul 10 23:59:24 Exit before auth from <192.168.1.3:52325>: Max auth tries reached - user 'is invalid'
```
- 🈯 The 1st is the most important one, which does now match.
- 🈴 The 2nd is after 10 failed login attempts, and does not match anymore, since `from <HOST>` is now located at the end anymore.
- 🈴 Login attempts for nonexistent users do not contain the host anymore.
- 🈺 Max retries for invalid users did never match, but could be consequently added.

Max retry failures are IMO not important, since usually, the failed attempts before this happens match and lead to a ban already, before max retries are even hit. Invalid user logins however IMO are important, and I am not sure why those do not contain the host anymore. I'll check back with Dropbear. This is however unrelated to the systemd/STDOUT /`-E` matching, i.e. I'll open a new PR, once I verified changed output in recent (upstream) Dropbear versions, and whether the missing host is intended or a bug.

MichaIng:
I opened an issue about the broken matches: #3791
The 2nd case can be fixed easily, the 4th could be additionally added, the 3rd was a mistake Dropbear end, as far as I can see.

config/filter.d/dropbear.conf
```diff
@@ -23,7 +23,7 @@ before = common.conf
 
 _daemon = dropbear
 
-prefregex = ^%(__prefix_line)s<F-CONTENT>(?:[Ll]ogin|[Bb]ad|[Ee]xit).+</F-CONTENT>$
+prefregex = ^%(__prefix_line)s(\[\d+\] [A-Z][a-z]+ \d\d \d\d:\d\d:\d\d )?<F-CONTENT>(?:[Ll]ogin|[Bb]ad|[Ee]xit).+</F-CONTENT>$
```
sebres, line 26: The issue with that additional timestamp - it looks like a default date-time format, so that for example it may depend on current locale.
Basically it can be any of 
https://github.com/fail2ban/fail2ban/blob/8170e9fe75fd2c2c4c51a1d9972b683401cddccb/fail2ban/server/datedetector.py#L126-L157
Also `[A-Z][a-z]+ \d\d` would not match `Mär 10` (umlaut) or `Jul  1` (space and single digit)...
There is no good RE to cover it (no special chars or separator after the timestamp), and catch-all like `.*` is too ugly.
I'd suggest something like `(?:\[\d+\](?:\s+\w{2,3}(?:(?:\s+|[:\-\./])\w{1,3}){0,2})?(?:(?:\s+|[:\-\./])\d+){0,4}\d )?` but it also looks a bit like a workaround to me. Although since all date formats mostly ends with a digit (and followed by space), I think it may be acceptable and `\d ` can be considered as an "anchor" for the timestamp, before the content of message begins.

sebres (commented):
> Another thing: For systemd services like used on Debian, shall I add the journalmatch right here?

Sure.

config/filter.d/dropbear.conf
MichaIng: Dropbear is no localised. However, of course it could use a system library to generate the timestamp, which is. I just applied German locale on a test system, and added `LC_ALL=de_DE.UTF-8` to the system's and unit's environment, and added `ExecStartPre=/bin/dash -c 'date; locale'` to assure it is all applied:
```
Mär 01 13:38:19 VM-Bookworm systemd[1]: Starting dropbear.service - Lightweight SSH server...
Mär 01 13:38:19 VM-Bookworm dash[848]: Fr 1. Mär 13:38:19 CET 2024
Mär 01 13:38:19 VM-Bookworm dash[849]: LANG=de_DE.UTF-8
Mär 01 13:38:19 VM-Bookworm dash[849]: LANGUAGE=
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_CTYPE="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_NUMERIC="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_TIME="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_COLLATE="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_MONETARY="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_MESSAGES="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_PAPER="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_NAME="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_ADDRESS="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_TELEPHONE="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_MEASUREMENT="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_IDENTIFICATION="de_DE.UTF-8"
Mär 01 13:38:19 VM-Bookworm dash[849]: LC_ALL=de_DE.UTF-8
Mär 01 13:38:19 VM-Bookworm systemd[1]: Started dropbear.service - Lightweight SSH server.
Mär 01 13:38:19 VM-Bookworm dropbear[850]: [850] Mar 01 13:38:19 Failed loading /etc/dropbear/dropbear_dss_host_key
Mär 01 13:38:19 VM-Bookworm dropbear[850]: [850] Mar 01 13:38:19 Not backgrounding
```
As can be seen, despite German locale being applied, `journalctl` and `date` (within the unit) returning the problematic timestamp, Dropbear sticks to the expected one.

But otherwise, would `{^LN-BEG}` or `%(datepattern)s` (from `common.conf`) cover all cases, to be failsafe, in case builds on other distros, or if the build time locale(config have an effect?

sebres: > As can be seen, despite German locale being applied, journalctl and date (within the unit) returning the problematic timestamp

In reality it doesn't really (see below). It's just a text representation... in journal entry it is stored in numeric (binary) form.

> But otherwise, wouldn't `{DATE}` cover all cases, to be failsafe...

No, it would not.
1. `datepattern` only used by file-related backends, not for systemd backend, where the time will be obtained directly from journal entry (in form of datetime object without to parse it at all).
2. therefore the filter will not search the date at all (neither the first, what is simply representation in text-form, nor the second, logged by the service logger in text form) and therefore it'd not cut it out from message before `prefregex` or `failregex` are applied. No date expected, no extraction happens.

MichaIng: > In reality it doesn't really (see below). It's just a text representation... in journal entry it is stored in numeric (binary) form.

What I mean is that in login shell (`/etc/default/locale`), as well as the systemd unit's context, the German locale is applied, but Dropbear still logs in international date format.

> No, it would not.

I see. So `{DATE}` is not a generic date pattern for free usage, but one to match the log timestamp so that Fail2Ban can apply its `findtime` rules and such? _Since there is already the `datedetector` with a regex template, it would be actually nice to make it available for filters, independent of whether/how Fail2Ban extracts the timestamp for its own purpose. When daemons do not provide a particular journal logging mode, but STDOUT for logging to foreground console/fd in the first place, doubled timestamps in journal are quite common. But this is a different topic 🙂._

Regarding the question whether Dropbear logs its timestamps in different format depending on host OS, build environment/config or such, before guessing around or applying a possibly unnecessary complex regex, maybe @mkj can answer this faster than I am able to check and understand the code?

mkj: Dropbear is logging with `strftime` `"%b %d %H:%M:%S`.

https://github.com/mkj/dropbear/blob/2674736bce847bc51a30a07c191562047659ac1f/src/svr-session.c#L338-L341

sebres: Better would be surely not to log `[825] Jul 10 23:57:29` at all (for instance by something like `-F -E --log-msg-only` or `--journal`), since it simply doesn't make sense.

MichaIng: Okay great, so we know the two digit day and time format for sure. The only thing is that `%b` should use the current locale. Not sure why it does not in my tests, or it is set/overridden elsewhere. To be failsafe, we could use `\w+` or `\w{2,3}` instead `[A-Z][a-z]+`?

A journal logging mode for Dropbear would be great, in combination with an upstream systemd service, for completion. However, we would need to support the additional timestamp in any case, so it does not matter for this PR.

sebres: > To be failsafe, we could use `\w+` or `\w{2,3}` instead `[A-Z][a-z]+`?

`\w{2,3}` is fully enough. To simplify the whole thing - `(?:\[\d+\] \w{2,3} [\d:\s]+)?`. Note `(?:...)` for non-capturing group.

> However, we would need to support the additional timestamp in any case, so it does not matter for this PR.

No, we would not... the logger can simply avoid to write this part at all in that cases.
The journal entry need only contain the message, everything else is made by the wrapper stderr to journal automatically.
Here is for instance the code how it's done in fail2ban (for better understanding):
https://github.com/fail2ban/fail2ban/blob/8170e9fe75fd2c2c4c51a1d9972b683401cddccb/fail2ban/server/server.py#L739-L743

For this particular case it'd be enough something like that in https://github.com/mkj/dropbear/blob/2674736bce847bc51a30a07c191562047659ac1f/src/svr-session.c#L346:
```diff
+		if (!opts.logmsgonly) {
-		fprintf(stderr, "[%d] %s %s\n", getpid(), datestr, printbuf);
+			fprintf(stderr, "[%d] %s %s\n", getpid(), datestr, printbuf);
+		} else {
+			fprintf(stderr, "%s\n", printbuf);
+		}
```

MichaIng: > To simplify the whole thing - `(?:\[\d+\] \w{2,3} [\d:\s]+)?`. Note `(?:...)` for non-capturing group.

I saw the syntax by times, but looked it up just when you posted the complex regex above. Makes sense if there is no `\1`-like reference done, of if we do not know parent or child pattern groups 👍. Let's go with that one.

> No, we would not... the logger can simply avoid to write this part at all in that cases.

What I mean is that, even when Dropbear implemented this logging mode now, it would take years until it trickled down into stable releases of major distros. And even then, we cannot be sure which Dropbear version is really used, and whether this logging mode is really used, even if available. I do not know the policy about how compatible upstream Fail2Ban filters shall be, but I personally would just keep supporting all possible cases, even if this logging option gets implemented.

PR #3641, opened 2023-11-28
Added firewallcmd-rich-rules_new.conf which supports multiport and allports types

iprok:
The direct interface used in regular firewallcmd actions has been deprecated. So I updated firewallcmd-rich-rules.conf to support type={multiport,allports}

### DIFF
--- a/config/action.d/firewallcmd-rich-rules_new.conf
+++ b/config/action.d/firewallcmd-rich-rules_new.conf
@@ -1,0 +1,58 @@
+# Fail2Ban configuration file
+#
+# Author: Donald Yandt 
+# Changed by: Ivan Prokudin
+# 
+# Because of the rich rule commands requires firewalld-0.3.1+
+# This action uses firewalld rich-rules which gives you a cleaner iptables since it stores rules according to zones and not
+# by chain. So for an example all deny rules will be listed under <zone>_deny. 
+#
+# If you use the --permanent rule you get a xml file in /etc/firewalld/zones/<zone>.xml that can be shared and parsed easliy
+#
+# Example commands to view rules:
+# firewall-cmd [--zone=<zone>] --list-rich-rules
+# firewall-cmd [--zone=<zone>] --list-all
+# firewall-cmd [--zone=zone] --query-rich-rule='rule'
+
+[INCLUDES]
+
+before = firewallcmd-common.conf
+
+[Definition]
+
+# Option:  type
+# Notes.:  type of the action.
+# Values:  [ multiport | allports ]  Default: multiport
+#
+type = multiport
+
+actionstart = 
+
+actionstop = 
+
+actioncheck = 
+
+#you can also use zones and/or service names. 
+#
+# zone example: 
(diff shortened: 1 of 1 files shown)


config/action.d/firewallcmd-rich-rules_new.conf
```diff
@@ -0,0 +1,58 @@
+# Fail2Ban configuration file
```
sebres, line 1: If it gets deprecated, why not update the original action instead?

sebres (changes requested):

config/action.d/firewallcmd-rich-rules_new.conf
iprok: Hello! I've done it. Please see.

sebres: Hmm...
Now almost everything got changed in firewallcmd-actions.
Unexpected to solve few "deprecations".

iprok: That was my first thought. That's why I have made separate file at first. So what's the right way to go to make it merged?

iprok: @sebres Hello! Please tell me what's the right way you want me to do? I would like to make it merged.

sebres:
Well, as I was asking "why not update the original action instead", I meant why not update `firewallcmd-rich*` action(s) only instead of new action `firewallcmd-rich-rules_new`. If I correct understand the first commit the difference between your `firewallcmd-rich-rules_new` and `firewallcmd-rich-rules` was only the `type` (`allports`)?
Now you modified all `firewallcmd-*` actions, what is possibly not bad. Just the question is - **what is different now and which cons and pros the switch to rich rules does have?**
I know it is deprecated and that [release description](https://firewalld.org/2021/06/the-upcoming-1-0-0) shows:
*Many things can be added with minimal effort via rich rules.*
but also:
*Don’t fret. Deprecation is not removal. The direct interface will not be removed for a long time. It just means its use is discouraged because in most cases policy objects are a better choice.*

I'm not really familiar with firewalld (never understand the necessity to use in fail2ban an intermediate layer instead of native net-filter directly), so I don't follow which consequences a switch to rich rules may have for people used `firewallcmd-*` actions with direct interface previously. Do you?

> Please tell me what's the right way you want me to do? I would like to make it merged.

Sorry, right now I have no time to review...
But first look at that:
1. backwards compat - you deleted many actions, may be a rewrite with include like [action.d/iptables-multiport.conf](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/iptables-multiport.conf) and [action.d/iptables-allports.conf](https://github.com/fail2ban/fail2ban/blob/master/config/action.d/iptables-allports.conf) would be better.
2. ...TBD... (as said no time)
3. CHANGELOG (can be made at end if everything else solved)

However **firstly I'd like to clarify the questions from begin of this comment**.

Maybe the replacement of existing `firewallcmd-rich-rules` with `firewallcmd-rich-rules_new` would be fully enough (if they basically compatible)? And mark other actions as deprecated (and a changelog entry, so people would be informed about).

unknowndevQwQ:
@iprok What can I do for this PR?

PR #2723, opened 2020-05-14
Add SoftEtherVPN jail

benrubson:
Hi,

This PR adds a new jail, for [SoftEtherVPN](https://github.com/SoftEtherVPN/SoftEtherVPN).

Thank you 👍

PR #3669, opened 2024-01-27
Enable multi line support. Doesn't work with versions < 0.9

kovalroma:
This is my first pull request. I'm sorry if I did it wrong. In short, the current configuration is not working because SoftEtherVPN devs changed the log format. I'm not sure from which version onwards.

The current log looks like the following:

```
2024-01-23 20:36:33.809 The connection "CID-21" (IP address: 8.8.8.8, Host name: google-dns.com, Port number: 55947, Client name: "SoftEther VPN Client", Version: 4.42, Build: 9798) is attempting to connect to the Virtual Hub. The auth type provided is "Password authentication" and the user name is "admin".
2024-01-23 20:36:33.809 Connection "CID-21": User authentication failed. The user name that has been provided was "admin".
```

### DIFF
--- a/config/filter.d/softethervpn.conf
+++ b/config/filter.d/softethervpn.conf
@@ -1,9 +1,16 @@
 # Fail2Ban filter for SoftEtherVPN
 # Detecting unauthorized access to SoftEtherVPN
 # typically logged in /usr/local/vpnserver/security_log/*/sec.log, or in syslog, depending on configuration
+# Config for versions < 0.9 made by quixrick (https://reddit.com/u/quixrick) and jonisc
+# Further reference: http://www.vpnusers.com/viewtopic.php?f=7&t=6375&sid=76707e8a5a16b0c9486a39ba34763901
+
+#Enable multi line support. Doesn't work with versions < 0.9
+[Init]
+maxlines = 2
 
 [INCLUDES]
 before = common.conf
 
 [Definition]
-failregex = ^%(__prefix_line)s(?:(?:\([\d\-]+ [\d:.]+\) )?<SECURITY_LOG>: )?Connection "[^"]+": User authentication failed. The user name that has been provided was "<F-USER>(?:[^"]+|.+)</F-USER>", from <ADDR>\.$
+datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S.%%f
+failregex = IP address: <HOST>.*\n.*User authentication failed.*


PizzaProgram:
Thank you very much for creating this pull req. :-)
This is a MUST to secure the server at least a bit.

Personally I was shocked when I've realized that SoftEtherVPN has no counter to at least **show** how many failed attempts are happening in real time. _(Not to mention to block those IPs)_

**fail2ban** will be the perfect tool to do this instead.

config/filter.d/softethervpn.conf
```diff
@@ -1,9 +1,16 @@
...
+[Init]
+maxlines = 2
 
 [INCLUDES]
 before = common.conf
 
 [Definition]
-failregex = ^%(__prefix_line)s(?:(?:\([\d\-]+ [\d:.]+\) )?<SECURITY_LOG>: )?Connection "[^"]+": User authentication failed. The user name that has been provided was "<F-USER>(?:[^"]+|.+)</F-USER>", from <ADDR>\.$
+datepattern = ^%%Y-%%m-%%d %%H:%%M:%%S.%%f
+failregex = IP address: <HOST>.*\n.*User authentication failed.*
```
sebres, line 16: This RE is very weak (no anchors, catch-alls, etc), thus vulnerable.
Also note there is a tag `<SKIPLINES>`, used for skipping multiple lines (especially for `maxlines > 2`).

sebres:
1. Our test suite needs a coverage for every regex (see [fail2ban/tests/files/logs/softethervpn](https://github.com/fail2ban/fail2ban/blob/master/fail2ban/tests/files/logs/softethervpn)), otherwise it'd fail.
2. Old multiline filters (with `maxlines` buffer window) have many disadvantages, so normally we don't accept this kind of filters if it is possible with `<F-MLFID>` (see example below).
3. I don't think the "current" log-format is actual (probably some older version of softethervpn?), simply because it looks like it were older than excerpt used for current f2b filter.

New multiline filter (without `maxlines` buffer window) would look like this:
```ini
[Definition]

failregex = ^\s*The <F-NOFAIL>connection</F-NOFAIL> "<F-MLFID>CID-\d+</F-MLFID>" \(IP address: <ADDR>,
            ^\s*Connection "<F-MLFID>CID-\d+</F-MLFID>": User authentication failed. The user name that has been provided was "<F-USER>[^"]+</F-USER>"
```

sebres:
For the record: the pull request [SoftEtherVPN#1122](https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1122) provides the enhancement with `from <ip>` by the message (seems to be released with 5.02.5180).
I don't see something is changed since that by the logging of failed auth - https://github.com/SoftEtherVPN/SoftEtherVPN/blob/dc2b11918abe0566de69b7ae7bf5450cfbaa3b6d/src/Cedar/Protocol.c#L2098

kovalroma:
Well, it's strange. Based on GitHub, the latest version of SoftEtherVPN is 5.02.5181. However, if we check on [https://www.softether.org/](https://www.softether-download.com/en.aspx?product=softether), the latest version is 4.43. Probably, these two versions have different log formats.

sebres:
> Maybe we need to split filter and log for Stable and Dev branches of SoftEtherVPN ?

Why dev branches? As I already wrote this version is [released May 2021](https://github.com/SoftEtherVPN/SoftEtherVPN/releases/tag/5.02.0), so it is stable.
The maintainers of distro repositories may surely follow own policies so the update can take longer, but it is not our issue.

The problem with different versions (many distros and soft versions) is hardly solvable within stock fail2ban - we have neither the resources nor the time to support all the filters to follow every software change across all distributions.
Additionally note that if we'll write a filter supporting other (old) log-format, in the meantime before it got released here and later in related distro, the software may get an update in the distribution and the filter gets obsolete anyway. The effort is simply not justified.
And the more filters fail2ban would get, the more complex the issue will be.
With other words it is very expensive for an open source project like fail2ban, especially if it doesn't get any financial assistance.
Some adjustments can be surely made by the maintainers of fail2ban for certain distributions, but I guess even that may be too heavy.

Therefore the decision at the moment is to support the log format of latest released version. Or optionally parametrized filter that allow to control format (if they are not too different).

But one could simply use the filter I provided above or even write own filter/regex.

kovalroma:
> Why dev branches? As I already wrote this version is [released May 2021](https://github.com/SoftEtherVPN/SoftEtherVPN/releases/tag/5.02.0), so it is stable.

@sebres Please take a look at the description of the repository on which you are referring to. 

> About
Cross-platform multi-protocol VPN software. Pull requests are welcome. The stable version is available at https://github.com/SoftEtherVPN/SoftEtherVPN_Stable.

This [repo](https://github.com/SoftEtherVPN/SoftEtherVPN/) is not Stable, but Dev. The stable repo is [here](https://github.com/SoftEtherVPN/SoftEtherVPN_Stable).

I want to clarify that the current version of the filter and logs only covers the Dev repository, not the Stable one. Is this the intended behavior?

sebres:
> This [repo](https://github.com/SoftEtherVPN/SoftEtherVPN/) is not Stable, but Dev. The stable repo is [here](https://github.com/SoftEtherVPN/SoftEtherVPN_Stable).

Hmm, never saw that before... 2 repos for the same thing...
Under normal circumstances a release is release.

Well, I hope the issue is very clear and comprehensible - it is hardly possible to control any filter for every service across all platforms/distros. Also what is stable for example for debian, may be already obsolete for other platforms.

> Why does fail2ban support the Dev version but not the stable one?

Because the author of filter (see #2723) is also the author of enhancement of the the enhancement (pull request [SoftEtherVPN#1122](https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1122))?
Fail2ban (and/or its filters) is community driven thing, so if someone provides a new filter, we accept it, because should trust the contributors. Especially if I'm not familiar with that service (I never used SoftEtherVPN and do not intend to use it later).
From other point of view - fail2ban is just a tool and its configuration (inclusive filters/REs) is more or less the matter of user.

> Is this the intended behavior?

Seems like you missed another point I provided above - if new version of f2b gets released (in the distro you use), SoftEtherVPN gets probably updated, so is the effort really justified?

One can try to contact the maintainers of fail2ban for your distro, possibly they would provide a patch for that.

But I would not say "No", if someone would provide a PR for "old" stable version. 
This PR is closed because:
1. the RE is vulnerable
2. the filter uses old multiline mode (with `maxlines > 1`)
3. we already have newer version... mistake or not, it is released in 0.11.2/0.10.6, so can be already used by someone, so the changes on filter must be backwards compatible.
4. many filters of the same thing is not welcome here, so in best case the version should be parametrized somehow.

@benrubson some thoughts?

kovalroma:
I've got your point. I will open an Issue to discuss this problem in a wide audience.

kovalroma:
Let's continue our discussion [here](https://github.com/fail2ban/fail2ban/issues/3670)

benrubson:
> @benrubson some thoughts?

Best would have been for @dnobori to merge [this](https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/pull/8) in the so called _Stable_ version...
But he's almost unresponsive...

PR #3690, opened 2024-03-07
Add to postfix accepted logs

karolyi:
Hey,

I use a `-o syslog_name=postfix/submission-587` setting in my postfix daemon to be able to distinguish between daemons on various ports. This results in `Mar  7 16:12:50 myhostname postfix/submission-587/smtpd[51501]: connect from unknown[2.57.122.55]:59164` loglines, that fail2regex doesn't recognize.

This patch accounts for the extra hyphens in the daemon name.

### DIFF
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -10,7 +10,7 @@ before = common.conf
 
 [Definition]
 
-_daemon = postfix(-\w+)?/\w+(?:/smtp[ds])?
+_daemon = postfix(-\w+)?/[^/\[:]+(?:/smtp[ds])?
 _port = (?::\d+)?
 _pref = [A-Z]{4}
 
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -10,7 +10,7 @@ before = common.conf
 
 [Definition]
 
-_daemon = postfix(-\w+)?/[^/\[:]+(?:/smtp[ds])?
+_daemon = postfix(-\w+)?/[^/\[:\s]+(?:/smtp[ds])?
 _port = (?::\d+)?
 _pref = [A-Z]{4}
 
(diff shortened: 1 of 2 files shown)


sebres:
> I use a -o syslog_name=postfix/submission-587 setting in my postfix daemon

1. And `submission587` instead of `submission-587` would be not OK?
2. If someone (you) would use `submission.587` by syslog_name setting, this would again not match the pattern
3. If it is anyway a local customization (isn't it?), why just not add it as parameter to the filter for jail:
```ini
[postfix]
filter = %(known/filter)s[_daemon='postfix(-\w+)?/[-\w]+(?:/smtp[ds])?']
```

With other words why in this way, and not for instance something like that:
```diff
- _daemon = postfix(-\w+)?/\w+(?:/smtp[ds])?
+ _daemon = postfix(-\w+)?/[^/\[:]+(?:/smtp[ds])?
```

karolyi:
I used the hyphen for readability purposes, but you're right in that others might use different characters for separating the words.

I already have `filter = postfix[mode=aggressive]` for my postfix jail, does the further expansion/parameter specification `[_daemon='postfix(-\w+)?/[-\w]+(?:/smtp[ds])?']` go along well with that?

```diff
- _daemon = postfix(-\w+)?/\w+(?:/smtp[ds])?
+ _daemon = postfix(-\w+)?/[^/\[:]+(?:/smtp[ds])?
```

is fine with me as well, as long as it works. Needless to say, reading/understanding the regexes is quite convoluted, by their nature.

Which route would you like to take?

sebres:
Definitions:
```ini
filter = postfix[mode=aggressive, _daemon='...']
# or:
filter = postfix[mode=aggressive][_daemon='...']
# or:
mode = aggressive
filter = %(known/filter)s[_daemon='...']
```
are the same after substitution.

(Valiant with `%(known/filter)s` may be more preferred, because it gets updated by itself if tomorrow a [default filter-value](https://github.com/fail2ban/fail2ban/blob/3047572701cbba0dca4008b00440e9bf03bab24d/config/jail.conf#L167) gets enhanced in stock jail.conf)

> Needless to say, reading/understanding the regexes is quite convoluted, by their nature.

`\w` - alphanumeric (same as `[[:alnum:]]`
`[^/\[:]` - means any char but not `/`,  `[` and `:` (all 3 chars has in daemon prefix special meaning).

> Which route would you like to take?

I'm OK with both variants, just in case of `[-\w]+` I'd prefer to escape minus-char in set, so `[\-\w]+`, just to emphasize it is a char and not a range. 
If you'd add this escape backslash and also a short note to changelog, I'd merge the PR.

karolyi:
Definitions: duly noted, thanks. :)

I'll modify the PR with your provided regex suggestion.

I can read regexes, it's just not straightforward when you don't work with them all day long :) Most of the time I just read up some long-time-not-read documentation again for it.

PR #3691, opened 2024-03-08
A new filter to ban bad path reflective of hacking attempts

jackdeguest:

### DIFF
--- a/config/filter.d/apache-badpath.conf
+++ b/config/filter.d/apache-badpath.conf
@@ -1,0 +1,36 @@
+# Fail2Ban apache-badpath filter
+
+[INCLUDES]
+
+before = apache-common.conf
+
+[Definition]
+ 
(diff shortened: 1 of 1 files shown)


sebres:
Such filters are too individual, thus it'd not be the part of stock fail2ban.

Furthermore, this filter is a typical example of "bad" practice:
- the REs are too vulnerable (containing many catch-alls)
- too many REs (where they all can be combined to single RE)
- the paths will be checked regardless of the response code (404, 403 or whatever), so would consider it as failure even for 200 (found URL).
- it mixes 2 approaches (bad agents and bad paths), what is not a problem for some own custom jail, but is not good for a stock filter.

And last but not least, our test suite expects a test-file for the coverage of every RE.

As for combining of REs and better non-vulnerable RE, see for instance https://github.com/fail2ban/fail2ban/pull/1961#pullrequestreview-75940560

PR #3692, opened 2024-03-10
Change journalmatch postfix

pingou2712:
On Debian 12, Postfix operates through several sub-services rather than a single service. The general _SYSTEMD_UNIT=postfix.service in Fail2Ban only tracks basic service start-stop activities, missing out on detailed logs. To capture the full scope of Postfix's activities, targeting a more specific systemd unit like postfix@-.service in Fail2Ban is essential for effective monitoring and security.

### DIFF
--- a/config/filter.d/postfix.conf
+++ b/config/filter.d/postfix.conf
@@ -76,6 +76,6 @@ ignoreregex =
 
 [Init]
 
-journalmatch = _SYSTEMD_UNIT=postfix@-.service
+journalmatch = _SYSTEMD_UNIT=postfix.service _SYSTEMD_UNIT=postfix@-.service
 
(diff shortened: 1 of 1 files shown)


sebres:
Is it a backwards compatible change. I mean would it also consider messages logged with `postfix` unit?

pingou2712:
Indeed, _SYSTEMD_UNIT=postfix@-.service considers messages only logged with Postfix sub-units and not with the main Postfix unit. In my new commit, I updated journalmatch to _SYSTEMD_UNIT=postfix.service _SYSTEMD_UNIT=postfix@-.service, ensuring that Fail2Ban monitors both the primary Postfix service and its sub-units, thereby enhancing log coverage while maintaining backward compatibility.

Issue #3473, opened 2023-03-05
[FR]: Suggestion for a recidive FOUND filter (Double Tap)


SDuesterhaupt:
By default, the Jail threshold is 5 events in 10 minutes.  A further reduction of the trigger rate (matches per time) makes no sense. The usability suffers.

Since quite some time I regularly notice double tap events in my logs. These double-tap events occur every few minutes and are caused by several IPs that are probably part of botnets. Thus, these events are below the radar. Many pennies make a dollar. At least the server signals: I’m watching you.

The recidive-filter searches the fail2ban.log for BAN events.

Below is a filter suggestion for the interested community that looks for FOUND events instead:

/etc/fail2ban/filter.d/recidive-found.local

```
# Fail2Ban filter for repeating FOUND events
#
# This filter monitors the fail2ban log file, and enables you to add long
# time bans for ip addresses that get found by fail2ban multiple times.
#
# Reasons to use this: block very persistent attackers for a longer time,
# stop receiving email notifications about the same attacker over and
# over again.
#
# This jail is only useful if you set the 'findtime' and 'bantime' parameters
# in jail.conf to a higher value than the other jails. Also, this jail has its
# drawbacks, namely in that it works only with iptables, or if you use a
# different blocking mechanism for this jail versus others (e.g. hostsdeny
# for most jails, and shorewall for this one).

[INCLUDES]

# Read common prefixes. If any customizations available -- read them from
# common.local
before = common.conf

[Definition]

_daemon = (?:fail2ban(?:-server|\.actions)\s*)

# The name of the jail that this filter is used for. In jail.conf, name the jail using
# this filter 'recidive', or supply another name with `filter = recidive[_jailname="jail"]`
_jailname = recidive-found

failregex = ^%(__prefix_line)s(?:\s*fail2ban\.filter\s*%(__pid_re)s?:\s+)?INFO\s+\[(?!%(_jailname)s\])(?:.*)\]\s+Found\s+<HOST>.*$

datepattern = ^{DATE}

ignoreregex =

journalmatch = _SYSTEMD_UNIT=fail2ban.service PRIORITY=5

# Author: Tom Hendrikx, modifications by Amir Caspi
# further modificated by: Stephan Düsterhaupt, Ivo Noack
```

The appropriate jail config:

/etc/fail2ban/jail.d/recidive-found.local

```
[recidive-found]
enabled = true
filter = recidive-found

bantime.increment = false
bantime.rndtime =
bantime.maxtime =
[...]

PR #3693, opened 2024-03-10
Change Regex Recidive and journalmatch For Systemd Match

pingou2712:
**Proposed Changes due to Two Issues with Fail2Ban/Logging/Recidive under SYSTEMD:**

1. Issue with Original Regex in SYSTEMD Logs:

The original regex used by Fail2Ban does not match the log format when logging under SYSTEMD. The SYSTEMD log entries look like this:

Mar 10 10:50:07 namemachine fail2ban[15699]: [postfix-sasl] Ban 99.99.99.99

However, the original regex expects entries to contain .actions and NOTICE, which are not present in these logs. Therefore, I propose modifying the regex to make .actions and NOTICE optional, ensuring compatibility with SYSTEMD log formats.

2. Default Ban Priority Issue:

On my newly installed system (debian12), Fail2Ban is logging bans with priority level 4, not the expected level 5. Instead of specifying a particular priority level in journalmatch, I suggest using a more general approach:

journalmatch = _SYSTEMD_UNIT=fail2ban.service

This change will avoid specifying the priority level, reducing the potential need for future adjustments if the default priority level changes again. Additionally, omitting the priority level does not significantly increase the number of lines to analyze and offers more flexibility for future log format changes.

### DIFF
--- a/config/filter.d/recidive.conf
+++ b/config/filter.d/recidive.conf
@@ -19,7 +19,7 @@
 # common.local
 before = common.conf
 
-[Definition]
+[DEFAULT]
 
 _daemon = (?:fail2ban(?:-server|\.actions)\s*)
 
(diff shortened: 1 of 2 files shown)


sebres:
Hmm... Strange.
Which distribution is it?

If I look at 2 possibilities to log to journal, I don't see where the log-format should be changed this way.
However it can be probably `SYSTEMD-JOURNAL` (instead of `SYSLOG` or `SYSOUT`), so the formatting is not applied because of
https://github.com/fail2ban/fail2ban/blob/383adec83cc33d9dcb12e94594440e35f75ba3df/fail2ban/server/server.py#L749-L750

Since this filter is obsolete (due to already implemented `bantime.increment` and new backend proposal for recidive jail), I have no problem with that, however it can be rewritten better considering `logtype`.

pingou2712:
Nice rewrite considering the log type.

Indeed, I use logtarget = SYSTEMD-JOURNAL instead of SYSLOG or SYSOUT. Is this a bad practice? I was under the impression that SYSTEMD-JOURNAL was the preferred option when using systemd for everything?

I wasn't aware that this filter had become obsolete. Could you elaborate on the new backend proposal for the recidive jail that you mentioned? I seem to be out of the loop on this information.

Thank you very much for your assistance.

sebres:
> I've chosen to use logtarget = SYSTEMD-JOURNAL instead of SYSLOG or SYSOUT. Is this a bad practice?

No. I just never tried it, so did not known its format is restricted this way.

> Could you elaborate on the new backend proposal for the recidive jail that you mentioned?

New backend for recidive jail shall not monitor any file/journal at all, instead it would be direct notified from observer module (new event-driven thing implemented for `bantime.increment` feature) about every failure.
This is in PoC at the moment, details in https://github.com/fail2ban/fail2ban/issues/3473#issuecomment-1456143123

PR #3701, opened 2024-03-21
filter.d/exim.conf: rewrite host line regex for all varied exim's log_selector states

### LINKED ISSUE #3263: [FR]: Exim: lines from encrypted connections (TLS, log entries X= and CV=) not matched
### Environment:

- Fail2Ban version  : 0.10.2-2.1
- OS, including release name/version : Debian 10.12 (buster)

#### Service, project or product which log or journal should be monitored

- Name of filter or jail in Fail2Ban (if already exists) : exim, exim-spam
- Service, project or product name, including release name/version : exim4 4.92-8+deb10u6

Exim log lines are _normally_ matched correctly, but not when the remote has successfully issued a STARTTLS command. Log lines then have `X=(TLS cipher suite)` and `CV=(certificate validation status)` entries that are not currently matched.

### Relevant lines from monitored log files:

```
2022-04-22 06:40:33 H=(pmhtcr) [<HOST>] F=<<EMAIL>> rejected RCPT <<EMAIL>>: relay not permitted
2022-04-22 00:23:03 H=(3E64W6mP) [<HOST>] X=TLS1.2:ECDHE_ECDSA_AES_256_GCM_SHA384:256 CV=no F=<<EMAIL>> rejected RCPT <<EMAIL>>: relay not permitted
```

(I redacted the IP and email addresses in those lines, using `<HOST>` and `<EMAIL>` placeholders, respectively)

The second line doesn't match because `exim-common.conf` only looks for `I=(local interface used)`, `U=(local user or RFC 1413 identity)`, `P=(protocol used)` and `F=(sender address)` entries:

```
host_info_suf = (?::\d+)?(?: I=\[\S+\](:\d+)?)?(?: U=\S+)?(?: P=e?smtp)?(?: F=(?:<>|[^@]+@\S+))?\s
```

The `X=` and `CV=` entries fit between `P=` and `F=`; the following definition for `host_info_suf` fixes the issue for me (added to my filters.d/exim-common.local file):

```
host_info_suf = (?::\d+)?(?: I=\[\S+\](:\d+)?)?(?: U=\S+)?(?: P=e?smtps?)?(?: X=\S+)?(?: CV=(?:yes|dane|no))?(?: F=(?:<>|[^@]+@\S+))?\s
```

I used https://www.exim.org/exim-html-current/doc/html/spec_html/ch-log_files.html as the basis for the field definitions.

### Further considerations

* See the relevant [source code in exim4's `src/receive.c`](https://github.com/Exim/exim/blob/97c83a31f1269ac154408a571b9207c6f3552fc9/src/src/receive.c#L4044-L4061)
* I disregarded the `DN=` and `SNI=` entries, as they are optional, while `tls_cipher` and `tls_certificate_verified` are on by default.
* On a separate note: the `P=(protocol used)` regex should probably be extended to also cover `esmpts` (extended  smpt over SSL). I tacked on a `s?` to my `(?: P=e?smtps?)?` non-capturing group, above.

bes-internal:
Depending on Exim's log_selector settings, log lines may contain additional information about the connection. And also the line itself with the address of the remote host can vary greatly. 
But fortunately, all states can be found in the Exim code itself and taken into account. 
Makes it easier to add new regexps. 

Closes #3263

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -28,9 +28,11 @@ ver. 1.0.3-dev-1 (20??/??/??) - development nightly edition
   (value read from `/proc/sys/net/ipv6/conf/all/disable_ipv6`) if available, otherwise seeks over local IPv6 from network interfaces
   if available for platform and uses DNS to find local IPv6 as a fallback only
 * improve `ignoreself` by considering all local addresses from network interfaces additionally to IPs from hostnames (gh-3132)
+* `filter.d/exim.conf`:
+  - rewrite host line regex for all varied exim's log_selector states (gh-3263)
+  - fixed "dropped: too many ..." regex, also matching unrecognized commands now (gh-3502)
 * `action.d/mikrotik.conf` - new action for mikrotik routerOS, adds and removes entries from address lists on the router (gh-2860)
 * `action.d/smtp.py` - added optional support for TLS connections via the `ssl` arg.
-* `filter.d/exim.conf` - fixed "dropped: too many ..." regex, also matching unrecognized commands now (gh-3502)
 * `filter.d/nginx-forbidden.conf` - new filter to ban forbidden locations, e. g. using `deny` directive (gh-2226)
 * `filter.d/sshd.conf`:
   - avoid double counting for "maximum authentication attempts exceeded" (gh-3502)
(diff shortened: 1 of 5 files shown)


sebres:
As always exemplary PR, thanks!
I simplified and "optimized" it a bit, thus...
Let's merge it now.

PR #3727, opened 2024-04-27
Fix cloudflare-token.conf on Nixos

wangxiaoerYah:
fix on nixos bug:
uses unban error:
 `id for ip cannot be found using target ip`

The reason why the error occurs on nixos is that the special operating environment of nixos causes the commands before and after it to be executed separately. Subsequent commands do not inherit the previous command. Merging it solves the problem without affecting other distributions.

### DIFF
--- a/config/action.d/cloudflare-token.conf
+++ b/config/action.d/cloudflare-token.conf
@@ -55,7 +55,7 @@ actionunban = id=$(curl -s -X GET "<_cf_api_url>" \
               <_cf_api_prms> \
                   | awk -F"[,:}]" '{for(i=1;i<=NF;i++){if($i~/'id'\042/){print $(i+1)}}}' \
                   | tr -d ' "' \
-                  | head -n 1)
+                  | head -n 1); \
               if [ -z "$id" ]; then echo "<name>: id for <ip> cannot be found using target <cftarget>"; exit 0; fi; \
               curl -s -X DELETE "<_cf_api_url>/$id" \
                   <_cf_api_prms> \


sebres:
> the special operating environment of nixos causes the commands before and after it to be executed separately. Subsequent commands do not inherit the previous command.

What?!..

There are really shells which consider new-line as a separate shell invocation? Like (compound-list) subshell execution in posix shells?..

So that:
```bash
id=123; echo $id
```
and
```bash
id=123;
echo $id
```
are not the same?
Really?

wangxiaoerYah:
On my machine it was indeed necessary to add a punctuation mark, not sure about others.

vincelegithub:
I'm running fail2ban on docker (within SWAG reverse-proxy and under Openmediavault) and I experienced the same thing. With this fix, everything is OK, the IP is removed from Cloudflare.

wangxiaoerYah:
I will disable this pr if I think it is unnecessary.

wangxiaoerYah:
> > the special operating environment of nixos causes the commands before and after it to be executed separately. Subsequent commands do not inherit the previous command.
> 
> What?!..
> 
> There are really shells which consider new-line as a separate shell invocation? Like (compound-list) subshell execution in posix shells?..
> 
> So that:
> ```bash
> id=123; echo $id
> ```
> and
> ```bash
> id=123;
> echo $id
> ```
> are not the same?
> Really?

There's nothing wrong with it in principle, but I did go through it. And the modifications I made didn't destroy the original functionality. Indeed it works fine after the modifications.
If it doesn't make sense, I'll close it.

vincelegithub:
If this merge request is closed, there is still something to investigate. Unbanning IPs was not working for me before this change. It works after this change. This is on a pretty standard installation (Docker with Swag+Debian based distro).

Something is broken, and not just on Nixos.

PR #3775, opened 2024-06-25
Enhancing regex for ngxin_bad_requests

elegraf007:
Sorry, my bad, please reject this PR as I did come from master and wanted to commit to 0.11  .....

I will restart the PR

PR #3776, opened 2024-06-25
NGINX bad request enhancement

elegraf007:
Enhance the NGINX bad request filter for real world attack situations

Enhancing the searching regex for 3 common situations:

1. Clients or bots trying to access resources not being hosted to explore critical files. -> Error 404 is returned

2. Clients or bots trying to exploit services by pinging URIs with methods not allowed (i.e. POST instead of GET) -> Error 405 is returned

3. Clients or bots explore dynamic hosted sites by randomly sending parameters to exploit the API (for all APIs from old CGI to REST service proxied) -> Error 50x is returned
The latter is the weekest criteria bc. the return code is at best a expected exception and then defined by the API designer or at worst depending on the system behind. CGI for example let nginx throw 519 on harsh errors (which is not covered by the regex at all)

If desired I can provide similar to the Apache filter also

### DIFF
--- a/config/filter.d/nginx-bad-request.conf
+++ b/config/filter.d/nginx-bad-request.conf
@@ -5,7 +5,7 @@
 
 # The request often doesn't contain a method, only some encoded garbage
 # This will also match requests that are entirely empty
-failregex = ^<HOST> - \S+ \[\] "[^"]*" 400
+failregex = ^<HOST> - \S+ \[\] "[^"]*" [4-5]0[0-9]
 
 datepattern = {^LN-BEG}%%ExY(?P<_sep>[-/.])%%m(?P=_sep)%%d[T ]%%H:%%M:%%S(?:[.,]%%f)?(?:\s*%%z)?
               ^[^\[]*\[({DATE})
--- a/fail2ban/tests/files/logs/nginx-bad-request
+++ b/fail2ban/tests/files/logs/nginx-bad-request
@@ -21,3 +21,6 @@
 
 # failJSON: { "time": "2014-12-12T22:59:02", "match": true , "host": "2.5.2.5" }
 2.5.2.5 - tomcat [12/Dec/2014:22:59:02 +0100] "GET /cgi-bin/tools/tools.pl HTTP/1.1" 400 162 "-" "-" "-"
+
+# failJSON: { "time": "2014-12-12T22:59:02", "match": true , "host": "192.168.144.1" }
+192.168.144.1 - - [12/Dec/2014:22:59:02 +0100] "GET /adsf HTTP/1.1" 404 118 "-" "-" "-"


elegraf007:
Replaces https://github.com/fail2ban/fail2ban/pull/3775

Sorry again!

config/filter.d/nginx-bad-request.conf
```diff
@@ -5,7 +5,7 @@
 
 # The request often doesn't contain a method, only some encoded garbage
 # This will also match requests that are entirely empty
-failregex = ^<HOST> - \S+ \[\] "[^"]*" 400
+failregex = ^<HOST> - \S+ \[\] "[^"]*" [4-5]0[0-9]
```
sebres, line 8: The issue with that: strictly seen only 400 (and possibly 444) is bad request response code.
Someone would probably need to consider 403 (Forbidden) or 40[56] (Not allowed, Not acceptable) also as such "bad request", but definitely not 404, not to mention 401.

The issue with 404 is: it may be too heavy: few "broken" or outdated urls to some resources may cause a ban for fully legitimate users.
The issue with 401 is still worse, because it can be part of authentication handshake and such RE must additionally consider the user name (e. g. whether it is not empty).
Anyway we could accept something like this:
```diff
+ code = 400
- failregex = ^<HOST> - \S+ \[\] "[^"]*" 400
+ failregex = ^<HOST> - \S+ \[\] "[^"]*" <code>
```
so it'd be possible to set the code via filter in jail, e. g.:
```ini
filter = %(known/filter)s[code="(?!401)[45]\d{2}"]
```
but definitely not such code range by default... Sorry, no chances.

fail2ban/tests/files/logs/nginx-bad-request
```diff
@@ -20,4 +20,7 @@
 7.8.9.10 - root [20/Jan/2015:01:17:07 +0100] "CONNECT 123.123.123.123 HTTP/1.1" 400 162 "-" "-" "-"
 
 # failJSON: { "time": "2014-12-12T22:59:02", "match": true , "host": "2.5.2.5" }
-2.5.2.5 - tomcat [12/Dec/2014:22:59:02 +0100] "GET /cgi-bin/tools/tools.pl HTTP/1.1" 400 162 "-" "-" "-"
\ No newline at end of file
+2.5.2.5 - tomcat [12/Dec/2014:22:59:02 +0100] "GET /cgi-bin/tools/tools.pl HTTP/1.1" 400 162 "-" "-" "-"
+
+# failJSON: { "time": "2014-12-12T22:59:02", "match": true , "host": "192.168.144.1" }
+192.168.144.1 - - [12/Dec/2014:22:59:02 +0100] "GET /adsf HTTP/1.1" 404 118 "-" "-" "-"
```
sebres, line 26: As already said, 404 is not bad request, so no chance by default for that.
However if you'd implement above suggested enhancement (with code) this test may remain, but would expect an option, for instance:
```diff
+ filterOptions: [{"code": "(?!401)[45]\d{2}"}]
  # failJSON: ...
  ... 404 ...
```

sebres (changes requested):

config/filter.d/nginx-bad-request.conf
dacook: I think this general rule can help identify  abuse (as described) and ban accordingly. But it's very general so as you say it's not appropriate to be used by default.

What is the definition of a "bad" request? It seems to be an opinion. The only technical definition I'm aware of is the specification which defines "errors":

> - [4xx (Client Error)](https://httpwg.org/specs/rfc9110.html#status.4xx): The request contains bad syntax or cannot be fulfilled
> - [5xx (Server Error)](https://httpwg.org/specs/rfc9110.html#status.5xx): The server failed to fulfill an apparently valid request

But as you suggest it's important to consider how these are used in the real world. 
So... can we refine this rule to make it useful for general use?

* It sounds like there's a good reason to treat 401s with care.
* 404s: the only example of legitimate use I can think of is a web crawler looking for a large number of pages that have been removed. Banning/slowing down the server would be bad if it meant your stale content was still showing up in search results. I'm unsure if that would happen but it seems worth avoiding just in case.

So if all other error codes are uncontroversial, this PR has potential?

And should it adjust this existing rule, or would it be better to create a new separate filter for this more general rule (as I suggested in https://github.com/fail2ban/fail2ban/pull/3980). What do you think @sebres ?

Issue #3428, opened 2022-12-11
[RFE]: Hide sensetive info when reporting to third party


hsntgm:
#### Feature request type

Hide sensetive data when reporting to third party ip-db servers e.g. abusedipdb, badips.com via extarnal script or via action.d/abuseipdb.conf

#### Description

When reporting to third party, sensetive data like IP, hostname, mail address, MAC address exposes.

example **ufw-port-scan** matches sent to third party
`2022-12-11T22:07:58.230240+03:00 mail kernel: [52836.383408] [UFW BLOCK] IN=eth0 OUT= MAC=must_hide SRC=x.x.x.x DST=**must_hide**`

example **dovecot** matches sent to third party
`Dec 05 23:50:17 auth-worker(211446): Info: conn unix:auth-worker (pid=211445,uid=97): auth-worker<2>: sql(**must_hide**,2.57.122.10): Password mismatch (given password: **must_hide**)`

#### Considered alternatives
Alternatively user able to rewrite **matches** in jails. Something like  **tp_comment** for just clean and secure reporting to third party server.

```
[ufw-port-scan]
enabled = true
port = all
filter = ufw-port-scan
logpath = /var/log/ufw.log
tp_comment = Port probing on unauthorized port $PORT $PROTOCOL $maxretry in $findtime sec <-- Suggestion
maxretry = 5
findtime = 300
action = %(action_mwl)s
         %(action_abuseipdb)s[abuseipdb_apikey="xxx", abuseipdb_category="14"]
```
```
[dovecot]
enabled  = true
mode     = aggressive
port     = pop3,pop3s,imap,imaps,submission,465,sieve
logpath  = /var/log/dovecot.log
tp_comment = Email auth brute force attack $maxretry in $findtime sec  <-- Suggestion (or more generic one for all kind of matches)
backend  = polling
findtime = 10800
maxretry = 3
bantime  = 7200
action   = %(action_mwl)s
           %(action_abuseipdb)s[abuseipdb_apikey="xxx", abuseipdb_category="18"]
```

#### Any additional information
abuseipdb.conf

> Option:  actionban
> Notes.:  command executed when banning an IP. Take care that the
>           command is executed with Fail2Ban user rights.
> 
>           ** IMPORTANT! **
> 
>           By default, this posts directly to AbuseIPDB's API, unfortunately
>           this results in a lot of backslashes/escapes appearing in the
>           reports. This also may include info like your hostname.
>           If you have your own web server with PHP available, you can
[...]

PR #3797, opened 2024-07-16
Fix: Remove sensitive information from logs sent to AbuseIPDB

momohammadi:
**Updated the actionban ensure it correctly handles the removal of sensitive information such as server IP
addresses before reporting to AbuseIPDB's API.**

This change improves security and aligns the configuration with best practices
for handling sensitive information in logs.

This enhancement allows users to customize which information is removed using options like dst_ips_prefix,
hostname, and other_removed, preventing potential exposure of sensitive data.


config/action.d/abuseipdb.conf
```diff
@@ -68,24 +68,30 @@ actionstop =
...
+#          the report, you can set options like dst_ips_prefix, hostname, and other_removed
+#          in the jail.local file or by passing arguments in your jail action line.
 #
 # Tags:    See jail.conf(5) man page
 # Values:  CMD
 #
-actionban = lgm=$(printf '%%.1000s\n...' "<matches>"); curl -sSf "https://api.abuseipdb.com/api/v2/report" -H "Accept: application/json" -H "Key: <abuseipdb_apikey>" --data-urlencode "comment=$lgm" --data-urlencode "ip=<ip>" --data "categories=<abuseipdb_category>"
+
+actionban = lgm=$(printf '%%.1000s\n...' "<matches>" | \
+            sed -r 's/<source_ips_field>([0-9]{1,3}\.){1}[0-9]{1,3}/<source_ips_field>[REMOVED]/g' | \
```
sebres, line 85: `source_ips_field` is undefined in init section, should it be `dst_ips_prefix` or some related RE instead?

config/action.d/abuseipdb.conf
```diff
@@ -102,3 +108,22 @@ actionunban =
 # Register for abuseipdb [https://www.abuseipdb.com], get api key and set below.
 # You will need to set the category in the action call.
 abuseipdb_apikey =
+
+# Option: dst_ips_prefix
+# Notes:  Specifies the prefix used before destination IPs in your log for removing the first two octets of DST IP from match logs.
+# Example: If your iptables log entry is DST=xxx.xxx.xxx.xxx, set dst_ips_prefix to DST=
+# dst_ips_prefix = DST=
+# Values:  STRING  Default: None
+dst_ips_prefix =
```
sebres, line 117: It is completely unused in action at the moment.

config/action.d/abuseipdb.conf
sebres: What would happen if any of the new parameters is undefined (by default they are empty)?

config/action.d/abuseipdb.conf
```diff
@@ -102,3 +108,22 @@ actionunban =
...
+# Notes:  Specifies text to be removed from matches log entries.
+# Example: To remove hostname and app from iptables log entries, set other_removed to HOSTNAME APPNAME:
+# other_removed = myhost kernel:
+# Values:  STRING  Default: None
+other_removed =
+
+# Option: hostname
+# Notes:  Use this option to remove hostname from matches log.
+# Values:  STRING  Default: None
+hostname =
```
sebres, line 129: There are already dynamic action tags for the hostname - see https://github.com/fail2ban/fail2ban/blob/7fd097d73fb995daa39a834879b553a53d1f64ca/fail2ban/server/action.py#L63-L64

So one could use `<fq-hostname>` and/or `<sh-hostname>` instead (and they would not need init definition and should not be supplied as parameters).

sebres (changes requested):

sebres:
For the record, a RFE #3428

By the way, I'm unsure it is possible to correct remove "sensitive" info with the approach like that.
1. it may be too error prone - e. g. sensitive data may get formatted or changed in unexpected way in the log, so you'd be unable to remove it with simple sed-REs, with other words some info may remain untouched;
2. the information is extremely dependent on the service (or jail)
3. what is not sensitive for you may be sensitive for others and vice versa;
5. etc

I guess it is hardly possible to create a generalized solution, so it is too individual in my opinion...

The only "correct" way at the moment is something like described in https://github.com/fail2ban/fail2ban/issues/3428#issuecomment-1443566485 (replacing matches by a fixed comment).

PR #3816, opened 2024-08-09
Fix roundcube login failregex for roundcube 1.4+

Skamasle:
roundcube change log format from 1.4 version as you can see here https://github.com/roundcube/roundcubemail/blob/e92d8e31a342653fb65d12cf5fb706cac4031841/program/lib/Roundcube/rcube_imap.php#L194

### DIFF
--- a/config/filter.d/roundcube-auth.conf
+++ b/config/filter.d/roundcube-auth.conf
@@ -15,7 +15,7 @@ before = common.conf
 
 prefregex = ^\s*(\[\])?(%(__hostname)s\s*(?:roundcube(?:\[(\d*)\])?:)?\s*(<[\w]+>)? IMAP Error)?: <F-CONTENT>.+</F-CONTENT>$
 
-failregex = ^(?:FAILED login|Login failed) for <F-USER>.*</F-USER> from <HOST>(?:(?:\([^\)]*\))?\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
+failregex = ^(?:FAILED login|Login failed) for <F-USER>.*</F-USER></F-USER>(?:against \S+)? from <HOST>(?:(?:\([^\)]*\))?\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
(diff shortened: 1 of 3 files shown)


sebres:
Huh, related to https://github.com/roundcube/roundcubemail/commit/130f0cde3e13fecbc70c707d9bce78ea9eed71fc it was 2017, already.

config/filter.d/roundcube-auth.conf
```diff
@@ -15,7 +15,7 @@ before = common.conf
 
 prefregex = ^\s*(\[\])?(%(__hostname)s\s*(?:roundcube(?:\[(\d*)\])?:)?\s*(<[\w]+>)? IMAP Error)?: <F-CONTENT>.+</F-CONTENT>$
 
-failregex = ^(?:FAILED login|Login failed) for <F-USER>.*</F-USER> from <HOST>(?:(?:\([^\)]*\))?\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
+failregex = ^(?:FAILED login|Login failed) for <F-USER>.*</F-USER></F-USER>(?:against \S+)? from <HOST>(?:(?:\([^\)]*\))?\. (?:(?! from ).)*(?: user=(?P=user))? in \S+\.php on line \d+ \(\S+ \S+\))?$
```
sebres, line 18: This format will be already matched by current RE (probably due to `(?:(?:\([^\)]*\))?\. (?:(?! from ).)*`)...
But OK, I'll merge it after small review because of new tests and for the consistence. Thx.

By the way, there is accidentally additional `</F-USER>`, what would bother the RE (since unmatched tag causing also unmatched closing parenthesis)...

sebres: Fixed now in 8e0a2366f067ee33d0840b57fcd2810b3281b378, tests passed too, so I'll merge it. Thx again.

sebres:
Since it has only 1 `failregex` now, one could rewrite it together with `prefregex` to single RE (and remove the latter), but it'd not be backwards compatible anymore (if someone extended or customized the `failregex` of filter or jail locally), so it is retained for compatibility reasons...

PR #3856, opened 2024-10-13
style: cut main config random trailing space

unknowndevQwQ:

### DIFF
--- a/config/fail2ban.conf
+++ b/config/fail2ban.conf
@@ -63,7 +63,7 @@ pidfile = /var/run/fail2ban/fail2ban.pid
 
 # Options: dbfile
 # Notes.: Set the file for the fail2ban persistent data to be stored.
-#         A value of ":memory:" means database is only stored in memory 
+#         A value of ":memory:" means database is only stored in memory
 #         and data is lost when fail2ban is stopped.
 #         A value of "None" disables the database.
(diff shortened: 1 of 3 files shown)


sebres:
Sorry, but no.
This repository would never accept pure cosmetic changes like that (without any needs). 
This means - one can do that if there are also non-cosmetic functional amends.
Also if it looks harmless, it is not and it is simply the question of principle to reject such PRs.
Basically it is called collaboration hygiene, and unwanted because simply introducing conflicts across branches etc.

unknowndevQwQ:
> Sorry, but no. This repository would never accept pure cosmetic changes like that (without any needs). This means - one can do that if there are also non-cosmetic functional amends. Also if it looks harmless, it is not and it is simply the question of principle to reject such PRs. Basically it is called collaboration hygiene, and unwanted because simply introducing conflicts across branches etc.

How much functional change do I need to make to justify cosmetic changes? Adding new actions? Or only when it affects specific configuration file additions and deletions?

unknowndevQwQ:
> > Or only when it affects specific configuration file additions and deletions?
> 
> Right. Ideally only the lines that are really changed, but since the lines are mostly only comments, it is enough to improve the file itself to justify such cosmetic amends.

That is to say, if there is no specific check, the opportunity for modification will be missed or even more useless spaces will be added?

Issue #3687, opened 2024-03-02
[RFE]: ASN/Country based ban using IPinfo's free IP to Country ASN MMDB database


abdullahdevrel:
#### Feature request type

I work for IPinfo but I have fail2ban installed in all my servers.

The existing implementation of ASN/Country bans seems a bit confusing to me, and the per [IP lookup mechanism](https://github.com/fail2ban/fail2ban/blob/3047572701cbba0dca4008b00440e9bf03bab24d/ChangeLog?plain=1#L1418) could be slow for large scale attacks. I would like to suggest incorporating the [IPinfo IP to Country ASN database](https://ipinfo.io/developers/ip-to-country-asn-database) into Fail2Ban as a native download. The IPinfo IP to Country ASN database provides a static/locally downloaded MMDB database that can be extremely fast to query and provide country and ASN-based metrics in the same response, which is quite effective in deterring SSH access attacks.

#### Description

Consider my ban list. I am using [IPinfo's CLI](https://github.com/ipinfo/cli) to run summarize IP metadata of IP addresses and bulk IP geolocation/asn enrichment. I am using [JQ](https://jqlang.github.io/jq/) to wrangle my JSON responses.

**ASNs:** The following query summarizes my banned IP log and prints out the top 5 ASNs of the banned IP addresses.

```bash
cat /var/log/fail2ban.log | grep Ban | ipinfo grepip -o | ipinfo summarize -j | jq -r .asns
```

```json
{
  "AS132203 Tencent Building, Kejizhongyi Avenue": 24,
  "AS14061 DigitalOcean, LLC": 13,
  "AS4134 CHINANET-BACKBONE": 6,
  "AS45090 Shenzhen Tencent Computer Systems Company Limited": 17,
  "AS4766 Korea Telecom": 4
}
```

**Countries:** The following query summarizes my banned IP log and prints out the top 5 countries of the banned IP addresses.

```bash
cat /var/log/fail2ban.log | grep Ban | ipinfo grepip -o | ipinfo summarize -j | jq -r .countries
```

```json
{
  "CN": 35,
  "JP": 7,
  "KR": 8,
  "SG": 15,
  "US": 10
}
```

[...]

PR #3868, opened 2024-10-29
Add GeoIP support for banning IPs

### LINKED ISSUE #3790: [RFE]: add ip-api.com info to fail2ban.log for jail members

05bmckay:
Fixes #3790

This should be a good fix for issue #3790, ip-api is a pretty nice API but I haven't had a ton of time to play with it so I might be missing something. I would greatly appreciate any review on this if someone wanted to help.

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -166,6 +166,11 @@ mode = normal
 #
 filter = %(__name__)s[mode=%(mode)s]
 
+# "ignoregeo" specifies the countries to ignore based on geolocation.
+#ignoregeo = 
+
+# "geoipdb" specifies the path to the GeoIP database.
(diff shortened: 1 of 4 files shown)


pano9000:
not an expert on how Python http requests work, but I think this lacks some error handling:
E.g. if you query an invalid IP there is no "country" field in the JSON response

https://ip-api.com/#abcdef

Same with: what if the request fails, because the server is done, etc.

abdullahdevrel:
If you are interested, you can upvote this feature request: https://github.com/fail2ban/fail2ban/issues/3687.

From my experience of running fail2ban, geoip alone does not cut it. ASN is also an important metadata.

PR #3903, opened 2024-12-20
docs: explicitly list supported status flavors

pano9000:
Hi,

I've noticed that the status command mentions "flavors", but is not really telling you what flavors exists.
I've changed that in the man page and the protocol.py file based on the list found in the `status` method of the `actions.py`
```
supported_flavors = ["short", "basic", "stats", "cymru"]
```

I've sorted them alphabetically in the docs though for easier reading and appended the current default value ("basic") with 'default'.


melroy89 (approved):

melroy89:
Thank you!

Issue #1060, opened 2015-05-31
Multiple chains (INPUT+FORWARD; via intermediate chain) / PREROUTING?


blueyed:
I would like to use multiple chains (INPUT and FORWARD) by default, but
especially for the "recidive" jail.

I am trying the following now, where I use an intermediate chain
(`my-f2b-main-chain`), which is added as a first rule into INPUT and FORWARD
then:

    iptables -N my-f2b-main-chain
    iptables -A my-f2b-main-chain -j RETURN
    iptables -I INPUT -j my-f2b-main-chain
    iptables -I FORWARD -j my-f2b-main-chain

This chain then gets used for fail2ban, when both chains should get handled.

Does this make sense?
Can something like this get provived/shipped with fail2ban?

I've also thought about using the PREROUTING chain, but that might not make
sense, and then there seems to be no way currently to pass the '-t nat'
argument.
Although "chain = PREROUTING -t nat" might work?

leeclemens:
This discussion may be better suited for the mailing list, since it's not really an issue with fail2ban (i.e. bug). Providing the action configuration you are using and the jail config, at a minimum, seem necessary to help further.

blueyed:
It seems to be a common scenario, where you have the host route most traffic to containers (through iptable's FORWARD, using nat), but should block it at the front (INPUT).
fail2ban runs on the host and reads log output from the containers.

My jail config is pretty much the default for Debian, with several jails enabled via `jails.local`.

leeclemens:
Like I said, more specific information is required to address your concern here (your information will not work with the default iptables action, and "pretty much the default" doesn't help anyone know how you have it configured). I'd suggest checking this out:
http://www.fail2ban.org/wiki/index.php/HOWTO_Seek_Help

Since this doesn't seem to be a bug per se, I'm going to close this issue.

blueyed:
The relevant config is to change the default (or specific) chain to `my-f2b-main-chain` (instead of INPUT).
This was not meant to report a bug, but to discuss / suggest a new feature/action.

PR #3909, opened 2024-12-27
Enhancement on iptables: allow bans to be effective on multiple chains at the same time

avcbvamorec:
This patch allows the ban to be applied on the INPUT and the FORWARD chains at the time. May be useful at least on routing devices and on docker hosting machines.

### DIFF
--- a/config/action.d/iptables.conf
+++ b/config/action.d/iptables.conf
@@ -65,19 +65,19 @@ rule-jump = -j <_ipt_rule_target>
 # Several capabilities used internally:
 
 _ipt_for_proto-iter = for proto in $(echo '<protocol>' | sed 's/,/ /g'); do
-_ipt_for_chain-iter = for chain in $(echo '<chains>' | sed 's/,/ /g'); do
+_ipt_for_chain-iter = for iteredchain in $(echo '<chain>' | sed 's/,/ /g'); do
 _ipt_for_chain-done = done;
 _ipt_for_proto-done = done
 
 _ipt_add_rules = <_ipt_for_proto-iter>
               <_ipt_for_chain-iter>
-              { %(_ipt_check_rule)s >/dev/null 2>&1; } || { <iptables> -I $chain %(_ipt_chain_rule)s; }
+              { %(_ipt_check_rule)s >/dev/null 2>&1; } || { <iptables> -I $iteredchain %(_ipt_chain_rule)s; }
               <_ipt_for_chain-done>
               <_ipt_for_proto-done>
 
 _ipt_del_rules = <_ipt_for_proto-iter>
               <_ipt_for_chain-iter>
-              <iptables> -D $chain %(_ipt_chain_rule)s
+              <iptables> -D $iteredchain %(_ipt_chain_rule)s
               <_ipt_for_chain-done>
               <_ipt_for_proto-done>
 
@@ -88,7 +88,7 @@ _ipt_check_rules = <_ipt_for_proto-iter>
               <_ipt_for_proto-done>
 
 _ipt_chain_rule = <pre-rule><ipt_<type>/_chain_rule>
-_ipt_check_rule = <iptables> -C $chain %(_ipt_chain_rule)s
+_ipt_check_rule = <iptables> -C $iteredchain %(_ipt_chain_rule)s
 _ipt_rule_target = f2b-<name>
 
 [ipt_oneport]
@@ -106,12 +106,12 @@ _chain_rule = -p $proto <rule-jump>
 
 [Init]
 
-# Option:  chains
+# Option:  chain
 # Notes    specifies the iptables chains to which the Fail2Ban rules should be
 #          added. May be a sigle chain (eg. INPUT) or a comma separated list
 #          (eg. INPUT, FORWARD)
 # Values:  STRING  Default: INPUT
-chains = INPUT
+chain = INPUT
 
 # Default name of the chain
 #
--- a/config/action.d/iptables.conf
+++ b/config/action.d/iptables.conf
@@ -65,22 +65,30 @@ rule-jump = -j <_ipt_rule_target>
 # Several capabilities used internally:
 
 _ipt_for_proto-iter = for proto in $(echo '<protocol>' | sed 's/,/ /g'); do
+_ipt_for_chain-iter = for chain in $(echo '<chains>' | sed 's/,/ /g'); do
+_ipt_for_chain-done = done;
 _ipt_for_proto-done = done
 
 _ipt_add_rules = <_ipt_for_proto-iter>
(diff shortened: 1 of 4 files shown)


sebres:
> May be useful at least on routing devices and on docker hosting machines

Searching for prerouting chain? See https://linux-training.be/networking/ch14.html#idp69806048
In-between both iptables and nftables have this chain.

Related issue #1060

avcbvamorec:
AFAIK prerouting chain was designed to make conntracks (raw table) and modify packets (nat and mangle tables), this is why it is not present by default in the filter table. Anyway, the patch is simple and add the generic feature of applying ban rules to a list of chains.

I did not see the issue related, sorry for not referencing it.

config/action.d/iptables.conf
```diff
@@ -98,11 +106,12 @@ _chain_rule = -p $proto <rule-jump>
...
-# Option:  chain
-# Notes    specifies the iptables chain to which the Fail2Ban rules should be
-#          added
+# Option:  chains
+# Notes    specifies the iptables chains to which the Fail2Ban rules should be
+#          added. May be a sigle chain (eg. INPUT) or a comma separated list
+#          (eg. INPUT, FORWARD)
 # Values:  STRING  Default: INPUT
-chain = INPUT
+chains = INPUT
```
sebres, line 114: > the patch is simple and add the generic feature of applying ban rules to a list of chains

Well, it is... But this rename makes the action (and its derivates) backwards incompatible - so better would be to rename it back (as any others parameters like `port`, `protocol` etc that are singular).

sebres (changes requested):

config/action.d/iptables.conf
avcbvamorec: Your are absolutely right.

sebres (commented):
The problem with that: on the whole it remains backwards-incompatible, well conditionally if it is used as an inheritance:
and when for actions like this it may work: https://github.com/fail2ban/fail2ban/blob/9dde3d019e26c5ab604de234361c62305a41087a/config/action.d/iptables-xt_recent-echo.conf#L10
(at least as long as it'd be used with single chain),
for some other derivatives (especially users custom actions) it could fail.

Also there are quite likely several others issues, for instance `_ipt_check_rules` is checking existence of 2 rules in 2 different chains now. What would happen if only one is missing, and in which order?..

The strategy may be:
- either a new action, but then good working inherited actions (like probably `iptables-ipset`) would not benefit from this RFE;
- or the action shall be extended with optional parameter (similar to `type`), so it remains completely compatible used as single chain configuration.

sebres:
> Is there any official workaround to get multichains behavior like this? 

At the moment, no - there is no such default action here, however:
- you can create your own action doing that (probably the best choice currently);
- to workaround this, you could use 2 `iptables-ipset` actions, like this:
```ini
banaction = iptables-ipset[type=multiport]
banaction_allports = iptables-ipset[type=allports]
action_ = %(banaction)s[name=%(__name__)s-inp, port="%(port)s", protocol="%(protocol)s", chain="INPUT"]
          %(banaction)s[name=%(__name__)s-fwd, port="%(port)s", protocol="%(protocol)s", chain="FORWARD"]
```
  (ideally if they would share the same set, but it'd expect more tweaks)
- or single action where you overwrites some internal handlers, which would put 2 chains, that would reference 1 set within, e. g. something like this (untested):
```ini
action = iptables-ipset[type=multiport, port="%(port)s", protocol="%(protocol)s", multi_chains="INPUT, FORWARD", chain="$chain", _ipt_for_proto-iter="for chain in $(echo '<multi_chains>' | sed 's/,/ /g'); do <known/_ipt_for_proto-iter>", _ipt_for_proto-done="<known/_ipt_for_proto-done>; done"]
```

> will chain=PREROUTING work?

Conditionally... It depends on configuration, platform/distro, etc. In worse case it'd miss something or affect more chains as expected. Normally raw- or mangle-prerouting as well as connection (state) tracking happens before input/forward chains, but not necessarily the NAT-prerouting (if it is mangling direct to input chain, it'd bypass NAT prerouting) etc.
In any case there is no generalized common answer.

> looking for solution in my use case as part of services are on docker, and part on bare metal.

Can not you simply set `chain = INPUT` for bare metal jails and `chain = FORWARD` (or rather `chain = DOCKER-USER`) for docker jails, as described in [wiki :: Fail2Ban and Docker](https://github.com/fail2ban/fail2ban/wiki/Fail2Ban-and-Docker)?
Or do you have some jail monitoring logs/journal of both types? Because the question of multiple chains is interesting only in this particular case.

AllirWasTaken:
Logs from multiple services, either from docker or BM, end up in same log files in current config. I will experiment with setting up my own actions for that matter. Thank you for quick response.

sebres (approved):

sebres:
I rebased it to latest master and adjusted it a bit to make the feature available by almost all derivations. 
Also fixed the test coverage for all affected actions.
Tests pass also now, so I'd merge it later (if no objections follow and we get tested them a bit).

@AllirWasTaken you can test it with thin action (just make a backup and replace iptables.conf from this PR in `/etc/fail2ban/action.d`) and then you'd be able to do something like this:
```ini
# jail for bare metal and docker:
[some-jail]
...
banaction = iptables-ipset
chain = INPUT,DOCKER-USER
```
(and welcome with feedback)

sebres:
> This patch does not work at stable release (1.1.0)

"Does not work" doesn't exist in IT world. What an error do you get? Please provide something confirming your claims (error message by actionstart or actionban, dump of `fail2ban-client -d | grep $jail`, etc).
Maybe an attempt to patch was not successful?

However you didn't need to "patch" the action - it'd be enough to replace whole file [iptables.conf](https://raw.githubusercontent.com/fail2ban/fail2ban/52d239483d193a949b97e72b3ea001aae267528e/config/action.d/iptables.conf), which is completely independent (doesn't have any include) and must be compatible to v.1.1 or v.1.0.

The only difference I see in your action is the split of both for-loops into 2 separate lines, where in the merged action they belong to the same line:
```diff
-- for chain in $(echo '<chain>' | sed 's/,/ /g'); do for proto in $(echo '<protocol>' | sed 's/,/ /g'); do
++ for chain in $(echo '<chain>' | sed 's/,/ /g'); do
++ for proto in $(echo '<protocol>' | sed 's/,/ /g'); do
-- done; done
++ done
++ done
```
How it shall help to "fix" something is completely unclear to me.
With other word, something like this shall not work in your shell, doesn't it?
```bash
for chain in $(echo 'INPUT,FORWARD' | sed 's/,/ /g'); do for proto in $(echo 'tcp,udp' | sed 's/,/ /g'); do
  echo "test $chain ($proto)"
done; done
```

PR #3948, opened 2025-02-22
Add AbuseIPDB reporting actionban script for Fail2Ban

### LINKED ISSUE #3428: [RFE]: Hide sensetive info when reporting to third party

hsntgm:
---

💡 **NOTE:**

This PR has drifted from its original scope with time and the
integration has grown well beyond the scope of a single actionban
script — what started as a **fix for sensitive info leaking**  into
third party reports turned into a concurrency-safe solution for high-volume
**actionban** triggers. That also makes it a hard-to-review PR to
ask maintainers to sign off on. Given that scope, I've moved active
development to a standalone repository:

**[Fail2Ban AbuseIPDB Reporter](https://github.com/psaux-it/fail2ban-abuseipdb)**

The new README covers installation, the full architecture, database
schema, and troubleshooting in more detail than fits comfortably in
an action.d script. It may also be a useful reference for anyone
thinking about concurrency-safe patterns for future **actionban** bash scripts
in core.

I'll leave this PR open for just reference — thanks to @sebres and
@orlitzky for the review early on, and @CyberSteve99 for the
production feedback.

---

### Description:
Introduces a new **actionban script** for Fail2Ban to report
banned IP addresses to **AbuseIPDB**, with support for **custom comments**.
It prevents **duplicate reporting** by maintaining a local list of already
reported IPs and ensures that the IP is not reported again if it has
been previously reported and is still listed in AbuseIPDB.

#### Why we need?
- Functionality to report IPs to AbuseIPDB with a customizable
  comment for each ban to hide sensetive data.
- To check both the local list and AbuseIPDB to avoid
  redundant reporting and to ensure the IP is not reported multiple
  times.

This enhances the integration with **AbuseIPDB** while ensuring
that Fail2Ban reporting remains efficient and effective, avoiding
API rate limits by reducing duplicate entries.

### Fixes:
- Fixes issue [#3428](https://github.com/fail2ban/fail2ban/issues/3428) — [RFE]: Hide sensetive info when reporting to third party

### Checklist:
- [x] Chose `master` as the base branch for a new feature.
- [x] Added unit tests (if applicable).
- [x] List issues this PR resolves: Fixes #3428
- [x] Verified that the PR doesn't break existing tests.
- [x] Kept the PR small and focused.
- [x] No unrelated stylistic changes.
- [x] Added example log lines for new `failregex` (if applicable).


files/fail2ban-abuseipdb.sh
```diff
@@ -0,0 +1,255 @@
...
+fi
+
+# Check runtime dependencies
+dependencies=("curl" "jq" "flock")
+for dep in "${dependencies[@]}"; do
+    if ! command -v "${dep}" &>/dev/null; then
+        log_message "FATAL: -${dep} is not installed. Please install -${dep} to proceed."
+        exit 1
+    fi
+done
```
sebres, line 141: Firstly, I don't think it is good idea to check the dependencies by every ban (it'd be probably OK on actionstart, but definitelly not on actionban/actionunban).

files/fail2ban-abuseipdb.sh
```diff
@@ -0,0 +1,255 @@
...
+#   If you restart your server often this script prevents duplicate reporting to AbuseIPDB by maintaining a local list of already reported IPs.
+#   Before reporting, it checks both the local list and AbuseIPDB to ensure the IP hasn't
+#   been reported previously, thereby avoiding redundant entries and potential API rate limiting.
+#
+# Integration with Fail2Ban:
+#   - Place this script in the 'action.d' directory of Fail2Ban.
+#   - Configure Fail2Ban to use this script as the `actionban`.
+#   - First edit 'abuseipdb.conf' in '/etc/fail2ban/action.d/abuseipdb.conf' and add following rule,
+#   - actionban = /etc/fail2ban/action.d/fail2ban-abuseipdb.sh \
+#         "<abuseipdb_apikey>" "<matches>" "<ip>" "<abuseipdb_category>" "<bantime>"
```
sebres, line 15: Hmm... questionable thing to integrate this script with the action.
Especially because it is not advised to adjust stock configuration files (because of possible conflicts by update).
One can surely create `action.d/abuseipdb.local` instead and put it there, but...

files/fail2ban-abuseipdb.sh
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env bash
+#
+# Description:
+#   This script serves as a Fail2Ban `actionban` to report offending IP addresses to AbuseIPDB with custom COMMENTS.
+#   Mainly whenever Fail2Ban restarts, it calls the actionban function for each IP stored in the database file.
+#   If you restart your server often this script prevents duplicate reporting to AbuseIPDB by maintaining a local list of already reported IPs.
```
sebres, line 6: The same thing can be done with:
```ini
norestored = true
```
See: https://github.com/fail2ban/fail2ban/blob/882e6d5e0011fe12bac8a356caea0394bc19b26c/ChangeLog?plain=1#L1023-L1028

files/fail2ban-abuseipdb.sh
hsntgm: I agree. While `curl` and `flock` (provided by `util-linux`) are usually available even in slim images, `jq` often isn’t and  potentially cause issues, maybe we can keep it.

files/fail2ban-abuseipdb.sh
hsntgm: Thanks to point `norestored = true`, let me check.

files/fail2ban-abuseipdb.sh
hsntgm: You're absolutely right, placing the configuration in `action.d/abuseipdb.local` is a better choice to keep the default untouched. 

I also wanted to clarify that the purpose of this PR is not to modify the core functionality of `abuseipdb.conf` directly. The script is intended as a isolated helper script, and it’s entirely up to you whether to keep it completely isolated in repo or link it to the `abuseipdb.conf` as the default configuration.

I’ve been using this script in a production environment for over a year with positive results, which is why I thought it might be helpful to share it with others through this PR.  https://www.abuseipdb.com/user/50466

files/fail2ban-abuseipdb.sh
sebres: I wasn't speaking about some particular dependency, the checks are basically OK...
What I wanted to tell you - it is worse to do that on every ban (in `actionban`), where it can be done once in `actionstart` (executed only at start of jail).
Since it is intended to be synchronous action (why actually not `nohup ... &`?) it'd delay banning action, and then every extra check is an additional reason for extra latency...
Just put it to `actionstart` and it'd be done only once.
However in my opinion the description about dependencies is fully enough.

files/fail2ban-abuseipdb.sh
hsntgm: Ok, now the user can choose to rely on Fail2Ban's `norestored` setting using `BYPASS_FAIL2BAN` in `abuseipdb.local`. I have set it to `false` by default, ensuring that `actionban` will always trigger, allowing management at the script level.

files/fail2ban-abuseipdb.sh
hsntgm: Sorry I don't have deep knowledge how things trigger in Fail2Ban but I think I got the idea and re-organized .sh. Please let me know If I missed something.

hsntgm:
@sebres I've made several commits to the PR. After working with 'actionstart' and 'actionban' through single bash scripting, I've realized it's not a easy journey, but I enjoyed the challenge of handling high concurrency and optimizing latency. I've frozen the code for now and will test it in production. 

@orlitzky when you have a moment, could you take a quick look at the script? I’d love to get your thoughts on how it handles high concurrency. And of course, if there’s a bash script involved, a Gentoo dev has to be there! 😄🐧🔧

files/abuseipdb.local
```diff
@@ -0,0 +1,74 @@
...
+# Notes.:  Uncomment and leave as-is
+# actionstart = nohup /etc/fail2ban/action.d/fail2ban-abuseipdb.sh \
+#     "--actionstart" "<SQLITE_DB>" "<LOG_FILE>" &
+
+
+# Option:  actionban
+######################
+# Notes.:  Uncomment and leave as-is
+# actionban = /etc/fail2ban/action.d/fail2ban-abuseipdb.sh \
+#     "<abuseipdb_apikey>" "<matches>" "<ip>" "<abuseipdb_category>" "<bantime>" "<restored>" "<BYPASS_FAIL2BAN>" "<SQLITE_DB>" "<LOG_FILE>"
```
orlitzky, line 67: This will leak the API key in the `ps` output which could be an issue on a web server with SSH users. The usual workaround is to pass the secret in the environment but I don't know off the top of my head how to do that here.

files/fail2ban-abuseipdb.sh
```diff
@@ -0,0 +1,472 @@
...
+    SQLITE_DB="${8:-/var/lib/fail2ban/abuseipdb/fail2ban_abuseipdb}"
+    LOG_FILE="${9:-/var/log/abuseipdb/abuseipdb.log}"
+fi
+
+log_message() {
+    local message="$1"
+    echo "$(date +"%Y-%m-%d %H:%M:%S") - ${message}" >> "${LOG_FILE}"
+}
+
+LOCK_INIT="/tmp/abuseipdb_actionstart_init.lock"
```
orlitzky, line 85: Should use `mktemp()` here to avoid malicious users hijacking these paths.

orlitzky:
How much concurrency do you expect? :)

I would guess that the one-insert-at-a-time model is what's going to slow it down, if anything. And TBH bash is not a great choice if you're really worried about performance. Python is already a given with fail2ban, so that would be my recommendation is bash is slowing you down. There are some other advantages to this:

* Python has its own SQLite API that is a little bit smarter about connection handling than simply launching `sqlite3` over and over again
* Python has its own https/jquery libraries that avoid new dependencies on curl and jq
* Python has library functions for things like date mangling and temporary file handling that avoid launching new processes
* All that aside, python is just plain faster than bash

But it really depends on how serious you want to get about performance.

files/abuseipdb.local
sebres: Well, it is questionable (at least if used in fail2ban) - it is running as root, so basically possible with ps & co for root or sudoers only.
But they are anyway able to find the key in fail2ban configuration files, so it doesn't matter.
Regarding the env-vars, fail2ban can supply arguments as variables to shell commands of action, but at the moment only if the argument contains some special character. At the moment, there is no possibility to force it for plain arguments (what, I think, the api-key is).
Maybe we shall indeed implement some option, to enforce then vars usage.
Alternatively, one could set the var in env of the systemd-unit and use that var here, but... it is like using a sledge-hammer to crack a nut, in my opinion for abovementioned reason.

orlitzky: > Well, it is questionable (at least if used in fail2ban) - it is running as root, so basically possible with ps & co for root or sudoers only.

The full process list is "usually" still available to all users. (I use quotes because I don't know what every distro is doing.) On linux, you can mount `/proc` with `hidepid=1` to prevent other users from seeing root processes, but it's not the kernel default. My regular user can see the argument to all root processes, e.g.

```
$ ps aux | grep root
...
root     43667  0.2  0.0 106168 55616 pts/4    S+   08:22   0:00 emacs --no-window-system hello-world.txt
```

Ref: https://www.redhat.com/en/blog/hidepid-linux-hide-pid

hsntgm:
@orlitzky I appreciate your review. Actually, concurrency varies based on the attack surface, as the PR attempts to improve the third-party reporting workflow with shell scripting while maintaining its own local database, which triggers with each ban action. On a heavily attacked IP blocks, this can occur hundreds of times and more. On my IP block, I typically see around 20-30 ban actions per minute, which is relatively low, so I haven't had the chance to test PR in a very high-concurrency scenario. While the PR handles current ban load, based on your considerations, bash scripting or general skeleton may not be the best approach here.

@sebres, I'm interested in this aspect of Fail2Ban (third-party reporting) because I believe that reporting to third parties is key in the fight against 🖤🎩 - So, I would appreciate the opportunity to contribute to better handling and deeper integration of third-party reporting in your roadmap.

hsntgm:
@CyberSteve99 

You are welcome, vary on use cases there are many opportunities to improve reporting workflow. I'm glad the startpoint helped you. I shaped the abuseipdb reporting architecture to be simple as much as possible (even possibly current approach still not simple enough). The PR origin was actually solving issue -hiding sensetive info when reporting third party then with time I improved the workflow a bit and added new isolated IP DB logic. Here there are more simple [PR](https://github.com/fail2ban/fail2ban/pull/4089) handle origin issue differently. So this PR now looks like a bit overcomplicated but still works like a charm for edge cases for spesific use cases like yours.

**The seperate DB's Exact Role in This Architecture**

The isolated DB is useful in one specific scenario: when the same IP gets banned again.

Here is the full decision flow:
```
Fail2Ban triggers actionban (IP banned)
            │
            ▼
   Is IP in local SQLite DB?
   ┌─────────────────────────────┐
   │                             │
  YES                            NO
   │                             │
   ▼                             ▼
Check AbuseIPDB API         Insert IP into DB
(/v2/check)                      │
   │                             ▼
   ├── Still listed?        Report to AbuseIPDB
   │   → SKIP (done)        (/v2/report)
   │                             │
   └── No longer listed?    If report fails
       → RE-REPORT           → Delete from DB
```

**The DB Solves Two Problems**

**1. API Quota Protection (primary purpose)**

The /v2/check and /v2/report endpoints have daily limits. Without the DB, every repeat ban of the same IP would blindly hit /v2/report again, wasting quota and spamming AbuseIPDB with duplicate reports.
The DB acts as a fast local gate:

> "Did we ever report this IP?" → yes → then ask AbuseIPDB if it's still needed
> "Never seen this IP?" → report immediately, no API check needed

**2. Fail2Ban Restart Isolation (secondary purpose)**

With **BYPASS_FAIL2BAN=1**, after a Fail2Ban restart the script ignores Fail2Ban's own session restoration (RESTORED flag) and uses its own DB to independently decide what was already reported. **This gives you full independence from Fail2Ban's internal state.**

**The isolated 3party IP DB Is NOT**

- A live ban list
- A record of currently active Fail2Ban bans
- An expiry tracker
- A replacement for fail2ban-client status

I use this approach till **Feb 22, 2025**, here some results. In my env the approach works stable so far.

```
mail ~ # wc -l /var/log/abuseipdb/abuseipdb.log
305619 /var/log/abuseipdb/abuseipdb.log
```

```
2026-06-28 03:40:02 - INFO: IP 103.174.**.** has already been reported and remains on AbuseIPDB.
2026-06-28 03:49:40 - INFO: IP 130.12.***.** has already been reported and remains on AbuseIPDB.
2026-06-28 03:51:23 - INFO: IP 148.251.***.** has already been reported and remains on AbuseIPDB.
2026-06-28 03:51:43 - INFO: IP 178.151.**.*** has already been reported and remains on AbuseIPDB.
2026-06-28 03:51:44 - INFO: IP 178.151.**.*** has already been reported and remains on AbuseIPDB.
2026-06-28 03:54:14 - SUCCESS: Reported IP 20.106.***.** to AbuseIPDB.
2026-06-28 03:59:14 - INFO: IP 51.68.**.*** has already been reported and remains on AbuseIPDB.
2026-06-28 04:02:47 - INFO: IP 216.73.**.** has already been reported and remains on AbuseIPDB.
2026-06-28 04:14:42 - INFO: IP 45.148.**.*** has already been reported and remains on AbuseIPDB.
2026-06-28 04:14:56 - INFO: IP 45.148.**.*** has already been reported and remains on AbuseIPDB.
2026-06-28 04:15:47 - SUCCESS: Reported IP 52.154.**.** to AbuseIPDB.
2026-06-28 04:23:41 - INFO: IP 114.119.**.** has already been reported and remains on AbuseIPDB.
```

```
mail ~ # sqlite3 /var/lib/fail2ban/abuseipdb/fail2ban_abuseipdb "SELECT COUNT(*) FROM banned_ips;"
130036
```

Best~Hasan

files/fail2ban-abuseipdb.sh
hsntgm: They are now located securely in the service-isolated runtime directory.

hsntgm:
After reviewing a year's worth of logs, I identified a few lingering pain points and have force-pushed three targeted fixes to address them.

1. **Bounded async process‑based worker pool** – A semaphore using numbered lock files limits the number of concurrent actionban background processes.

3. **Per‑IP cooldown** – A local database check enforces AbuseIPDB’s 15‑minute same‑IP reporting window before making any API call.

5. **Per‑IP mutual exclusion** – A file lock per IP prevents parallel handling of the same IP, eliminating duplicate API calls and SQLite contention.

PR #3965, opened 2025-03-12
Add an nft ipset backend for firewalld-ipset

spauka:
This PR adds a nftables-ipset backend to the `firewalld-ipset` action. Previously, this backend supported either using `firewall-cmd` directly or the iptables based `ipset` command. Running `firewall-cmd` is painfully slow for adding/removing large numbers of IP addresses, hence using `ipset` or `nft` is preferable for this purpose.

In addition, we allow setting the action via the normal firewalld zones, since the nftables-ipset backend is compatible with the firewalld ipsets.

TODO:
- [ ] Initial version only supports allports, should add support for multiport via a ip,port ipset.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -55,6 +55,8 @@ ver. 1.1.1-dev-1 (20??/??/??) - development nightly edition
   by substitution of rich rule (gh-3815)
 * `filter.d/proxmox.conf` - add support to Proxmox Web GUI (gh-2966)
 * `filter.d/openvpn.conf` - new filter and jail for openvpn recognizing failed TLS handshakes (gh-2702)
+* `action.d/firewallcmd-ipset.conf` - add an nft ipset backend, which replaces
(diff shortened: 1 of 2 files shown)


sebres:
If the nftables is anyway the backend of firewalld, the question is then - why don't you use nftables action directly and prefer the "proxy" like firewalld instead? Which sense does the usage of additional layer here?

spauka:
The main motivation would be that it is nice to keep all firewall rules visible and administer-able in one place. That said, I don't think that this is a good solution. It doesn't actually interact well with firewalld with banned IPs disappearing if firewalld touches the ipset, and so this adds burden rather than reduces it.

It may be that calling the firewalld DBUS socket directly for speed is a better solution that I will explore, since the overhead seems to be in spinning up a new python process for each new IP to be added or removed. For now, I will close this PR.

PR #3980, opened 2025-04-02
Add filter for nginx HTTP errors (4xx and 5xx)

dacook:
An excessive number of HTTP errors is likely abuse of some kind. 
This may include application login errors, if the app responds with an appropriate HTTP status code.

We use this filter to manage abusive traffic on a production server, so I wanted to contribute it for general use. 
This filter seems rather rudimentary compared to other built-in filters, so I'm open to any feedback about it. 
I see that fail2ban's focus is on "authentication errors", so maybe this is too general. But it seems useful enough as an optional filter.

If agreed, I am happy to go ahead and add tests.

ChangeLog: use PR title.

### DIFF
--- a/config/filter.d/nginx-http-errors.conf
+++ b/config/filter.d/nginx-http-errors.conf
@@ -1,0 +1,7 @@
+# Nginx client and server errors (HTTP 4xx or 5xx)
+#
+# Author: David Cook
+# Inspired by https://serverfault.com/a/849859/959055
+
+[Definition]
+failregex = <HOST> -.*"(GET|POST|HEAD).*HTTP.*" [4-5][0-9][0-9]


config/filter.d/nginx-http-errors.conf
```diff
@@ -0,0 +1,7 @@
+# Nginx client and server errors (HTTP 4xx or 5xx)
+#
+# Author: David Cook
+# Inspired by https://serverfault.com/a/849859/959055
+
+[Definition]
+failregex = <HOST> -.*"(GET|POST|HEAD).*HTTP.*" [4-5][0-9][0-9]
```
sebres, line 7: Sorry, but no...
This regex is vulnerable (unanchored, has several catch-alls, etc)
The response code range 400-599 is too heavy by default, e. g. at least 401 (and probably 404) shall be handled differently.

Something like this possibly:
```ini
[Definition]
requri = /\S*
rescode = (?!40[14])[45]\d+
failregex = ^<ADDR> [^"]*"[A-Z]+\s+%(requri)s\s+[^"]*" %(rescode)s
```

config/filter.d/nginx-http-errors.conf
```diff
@@ -0,0 +1,7 @@
+# Nginx client and server errors (HTTP 4xx or 5xx)
+#
+# Author: David Cook
+# Inspired by https://serverfault.com/a/849859/959055
```
sebres, line 4: Don't write whereby it's inspired, write what this filter is doing and which issues one can expect by activating of such jails (false positives and banning of legitimate users for instance).

sebres (changes requested):
1. Fix the vulnerable RE
2. Every filter expected new jail entry in jail.conf, and a test-file in test-factory (in fail2ban/tests/files/logs/), see other pull requests for examples.
3. Entry in changelog please

dacook:
Thanks for the feedback, I will comment on the other PR to discuss the best direction forward

PR #3982, opened 2025-04-03
Added jail for Paperless ngx

LearningSpot:
Jail for  [paperless-ngx](https://github.com/paperless-ngx/paperless-ngx) unsuccessful authentication attempts

### DIFF
--- a/config/filter.d/paperless-ngx.conf
+++ b/config/filter.d/paperless-ngx.conf
@@ -1,0 +1,9 @@
+# Fail2Ban filter for unsuccessful paperless-ngx authentication attempts
+# Logged in /var/log/paperless.log
+
+# Author: LearningSpot
+
+[Definition]
+
+failregex = ^\s*(?:\[\]\s*)?\[INFO\] \[paperless\.auth\] Login failed for user `<F-USER>\S+</F-USER>` from (private )?IP `<ADDR>`\.$
+ignoreregex =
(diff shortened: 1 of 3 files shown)


sebres:
DMS?
I'm unsure we really need to have a preinstalled jail for every DMS across the world.
The filters are principally OK (it's just additional file unless referenced), but the `jail.conf` will be always used, and every jail costs time (even if not enabled)...
I must think about some encapsulation of jails/filters in new version, either by some additional groups or probably much better:
a remodeling moving jail info (default port, logpath, etc) to the filter and remove all "non-generic" (or even completely all) jails from jail.conf, so if one would add new section for the jail in `jail.local`, everything else would work out-of-the-box.

sebres:
Not yet, maybe later... IIRC, the jail is mandatory right now.

LearningSpot:
> Not yet, maybe later... IIRC, the jail is mandatory right now.

Just a thought, how about commenting out the jails no need to be there always. That shouldn't affect the time consumption. Will it work?

sebres:
Well, doesn't matter currently, to be covered correctly by the test-suite it shall be with a jail, so if jail gets commented it'd produce at least this error:
https://github.com/fail2ban/fail2ban/blob/c76e90fbb17115b90ca5b641c2183c0e3e644942/fail2ban/tests/clientreadertestcase.py#L889

LearningSpot:
> Well, doesn't matter currently, to be covered correctly by the test-suite it shall be with a jail, so if jail gets commented it'd produce at least this error:
> https://github.com/fail2ban/fail2ban/blob/c76e90fbb17115b90ca5b641c2183c0e3e644942/fail2ban/tests/clientreadertestcase.py#L889

Well I also thought it won't be that easy or you would have implemented that already. I think it's because the verification on line 873

PR #3993, opened 2025-04-16
Add filter to block unintended directory browsing

### LINKED ISSUE #1176: Apache directory index requests
I think it is left out of the Apache noscript filter. It's a HTTP/403.

```
[Mon Aug 31 22:24:19.775235 2015] [autoindex:error] [pid 25091] [client 4.6.4.1:53525] 
AH01276: Cannot serve directory /home/*/wp-content/plugins/tinymce-advanced/: No matching DirectoryIndex (index.html,index.cgi,index.pl,index.php,index.xhtml,index.htm) found, and server-generated directory index forbidden by Options directive
```

https://wiki.apache.org/httpd/ListOfErrors
May I send the PR?

gonwan:
Fixes #1176.

### DIFF
--- a/config/filter.d/apache-noindex.conf
+++ b/config/filter.d/apache-noindex.conf
@@ -1,0 +1,19 @@
+# Fail2Ban filter to block unintended requests for directory browsing
+#
+# Directory browsing is ontrolled by Indexes Option in apache2. The option
+# can by set per directory. This filter helps to block unintended requests.
+# for direcotry browsing.
+
+[INCLUDES]
+
+# overwrite with apache-common.local if _apache_error_client is incorrect.
+before = apache-common.conf
+
+[Definition]
+
+failregex = ^%(_apache_error_client)s ((AH01276: )?.*Cannot serve directory .* directory index forbidden by Options directive$|Directory index forbidden by Options directive: .*$)
+
+ignoreregex = 
+
+# Author: Binhao Qian <gonwan@gmail.com>
+
--- a/fail2ban/tests/files/logs/apache-noindex
+++ b/fail2ban/tests/files/logs/apache-noindex
@@ -1,0 +1,8 @@
+# Apache 2.2
+# failJSON: { "time": "2025-04-16T16:18:16", "match": true , "host": "13.82.231.149" }
+[Sat Apr 16 16:18:16 2025] [error] [client 13.82.231.149] Directory index forbidden by Options directive: /some/path/to/public/wp-includes/
+
+# Apache 2.4
+# failJSON: { "time": "2025-04-16T16:18:17", "match": true , "host": "13.82.231.149" }
+[Wed Apr 16 16:18:17.927866 2025] [autoindex:error] [pid 27978] [client 13.82.231.149:11585] AH01276: Cannot serve directory /some/path/to/public/wp-includes/: No matching DirectoryIndex (index.html,index.php) found, and server-generated directory index forbidden by Options directive
+


config/filter.d/apache-noindex.conf
```diff
@@ -0,0 +1,19 @@
...
+# for direcotry browsing.
+
+[INCLUDES]
+
+# overwrite with apache-common.local if _apache_error_client is incorrect.
+before = apache-common.conf
+
+[Definition]
+
+failregex = ^%(_apache_error_client)s ((AH01276: )?.*Cannot serve directory .* directory index forbidden by Options directive$|Directory index forbidden by Options directive: .*$)
```
sebres, line 14: Sorry, but no, this regex is too vulnerable has many catch-alls and therefore the decision-making parts are not anchored... (let alone it'd be slow on parasitic messages).
Something like this maybe?
```diff
- failregex = ^%(_apache_error_client)s ((AH01276: )?.*Cannot serve directory .* directory index forbidden by Options directive$|Directory index forbidden by Options directive: .*$)
+ failregex = ^%(_apache_error_client)s (?:(?:AH01276: )?Cannot serve directory \S+: No matching DirectoryIndex \([^\)]*\) found, and server-generated directory|Directory) index forbidden by Options directive\b
```

sebres (changes requested):
Please fix the RE, but...

I'm not sure whether it shall be part of stock fail2ban filters at all: neither it is an authentication issue, nor it serves the needs preventing some attack vector. The "forbidden" message may confuse here - but it is more or less just a non-existing URI ending with `/`.
I don't think it is good idea to ban it.

sebres (approved):

sebres (changes requested):
There are few typos... Please fix them, so the spellchecker becomes quiet.

sebres:
> On my wordpress host, hundreds of directory listing requests are logged in one hours. An attacker just wants to find some unprotected resources for an entrance. A vulnerable wordpress plugin can be on of them. 

This approach would be called "[security through obscurity](https://en.wikipedia.org/wiki/Security_through_obscurity)" and not recommended... With other words - nobody shall conceive a "protection" based on hiding of something. Especially because for instance in this particular case some botnet with hundreds of IPs could find such hidden vulnerable plugin even with fail2ban (and even if it'd use much fewer IPs, it has a potential to succeed, just would take longer).
However this can be considered as a helper to stop flood in logs or to avoid unwanted traffic and extra load by evildoers. Just... exactly like other questionable filters (as such detecting 404 etc), it may pretty simple cause false positives (ban legitimate users) or unintentionally allow scenarios described in this comment - https://github.com/fail2ban/fail2ban/pull/1070#issuecomment-113232258

Also it contravenes a bit against [wiki :: Best practice](https://github.com/fail2ban/fail2ban/wiki/Best-practice) - basically instead of several apache jails each "protecting" against single "attack" vector, ideally, it shall be exactly 1 common jail finding every vector of particular host with few REs.

But people build even honeypots and other similar things (that quasi provoke or attract the attacks and even generate more load on the host), so who am I to forbit it...

> Well, it all depends on you.

I was spoken only about it as a stock filter/jail... As an example filter (and also if all risks of its usage are known) it is OK for me (so maybe a short sentence in comments of filter would be good, or maybe an article in the wiki describing the risks and a comment referring to it).
Also I plan small restructuring of fail2ban jail/filters (the filters can be then added without the jails), so let pause it for a while.

gonwan:
Thanks for your kind explanation!

sebres (approved):

PR #4020, opened 2025-06-17
Update sendmail-reject.conf

billfor:
Fixes:

- change the F-MLFID to only capture the sendmail sequence number rather than the full prefix line because when using BSD verbose the multiline messages come in with different pri and will break the group (e.g. <mail.info> and <mail.notice>)
- add regex for lost input channel messages and update it for the current sendmail (after mail vs after rcpt)

[sample.txt](https://github.com/user-attachments/files/20782189/sample.txt)

### DIFF
--- a/config/filter.d/sendmail-reject.conf
+++ b/config/filter.d/sendmail-reject.conf
@@ -20,10 +20,10 @@ before = common.conf
 [Definition]
 
 _daemon = (?:(sm-(mta|acceptingconnections)|sendmail))
-__prefix_line = %(known/__prefix_line)s<F-MLFID>(?:\w{14,20}: )?</F-MLFID>
-prefregex = ^%(__prefix_line)s<F-CONTENT>.+</F-CONTENT>$
-
+__prefix_line = %(known/__prefix_line)s(?:\w{14,20}: )?
 addr = (?:(?:IPv6:)?<IP6>|<IP4>)
+
+prefregex = ^\s*(?:<mail\.[^\>]+> )?<F-MLFID>%(__prefix_line)s</F-MLFID><F-CONTENT>.+</F-CONTENT>$
 
 cmnfailre = ^ruleset=check_rcpt, arg1=(?P<email><\S+@\S+>), relay=(\S+ )?\[%(addr)s\](?: \(may be forged\))?, reject=(?:550 5\.7\.1(?: (?P=email)\.\.\.)?(?: Relaying denied\.)? (?:IP name possibly forged \[(\d+\.){3}\d+\]|Proper authentication required\.|IP name lookup failed \[(\d+\.){3}\d+\]|Fix reverse DNS for \S+)|[45]5[13] [45]\.1\.8(?: (?P=email)\.\.\.)? Domain of sender address \S+ does not (?:exist|resolve)|550 5\.[71]\.1 (?P=email)\.\.\. (Rejected: .*|User unknown))$
             ^ruleset=check_relay(?:, arg\d+=\S*)*, relay=(\S+ )?\[%(addr)s\](?: \(may be forged\))?, reject=421 4\.3\.2 (Connection rate limit exceeded\.|Too many open connections\.)$
--- a/config/filter.d/sendmail-reject.conf
+++ b/config/filter.d/sendmail-reject.conf
@@ -22,6 +22,11 @@ before = common.conf
 _daemon = (?:(sm-(mta|acceptingconnections)|sendmail))
 __prefix_line = %(known/__prefix_line)s(?:\w{14,20}: )?
 addr = (?:(?:IPv6:)?<IP6>|<IP4>)
+# mta_dname -- matches name of MTA daemon (typically specified in DAEMON_OPTIONS),
+# normally something without spaces like MTA-v4 or Deamon0, etc. If it'd contain spaces, one can
+# rewrite it in jail using  `filter = %(known/filter)s[mta_dname="[^,]+"]` or in .local overwrite
+# of the filter. (we would not use catch-alls here to satisfy obscure artificial case).
+mta_dname = \S+
 
 prefregex = ^\s*(?:<mail\.[^\>]+> )?<F-MLFID>%(__prefix_line)s</F-MLFID><F-CONTENT>.+</F-CONTENT>$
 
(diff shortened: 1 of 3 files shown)


sebres:
You have to provide the related matching test message (with JSON) for every RE in [tests/files/logs/sendmail-reject](https://github.com/billfor/fail2ban/edit/sendmail/fail2ban/tests/files/logs/sendmail-reject#L4?pr=%2Ffail2ban%2Ffail2ban%2Fpull%2F4020).

config/filter.d/sendmail-reject.conf
```diff
@@ -20,23 +20,24 @@ before = common.conf
...
             ^ruleset=check_relay(?:, arg\d+=\S*)*, relay=(\S+ )?\[%(addr)s\](?: \(may be forged\))?, reject=421 4\.3\.2 (Connection rate limit exceeded\.|Too many open connections\.)$
             ^rejecting commands from (\S* )?\[%(addr)s\] due to pre-greeting traffic after \d+ seconds$
             ^(?:\S+ )?\[%(addr)s\]: (?:(?i)expn|vrfy) \S+ \[rejected\]$
-            ^<[^@]+@[^>]+>\.\.\. No such user here$
-            ^<F-NOFAIL>from=<[^@]+@[^>]+></F-NOFAIL>, size=\d+, class=\d+, nrcpts=\d+, bodytype=\w+, proto=E?SMTP, daemon=\S+, relay=\S+ \[%(addr)s\]$
+            ^<[^@]+@[^>]+>\.\.\. (?:No such user here|User unknown)$
+            ^<F-NOFAIL>from=<[^@]+@[^>]+></F-NOFAIL>, size=\d+, class=\d+, nrcpts=\d+,(?: bodytype=\w+,)? proto=E?SMTP, daemon=MTA(?:-v[46])?, relay=(?:\S+ )?\[%(addr)s]$
```
sebres, line 32: I'd like to retain escaping slash before closing bracket (`\]`), just more readable as paired to `\[`.

sebres (approved):
Looks good to me.
Any chance to get ChangeLog entry for that?

billfor:
All done and I reverted my change to look for daemon=MTA(?:-v[46) because I must have had an older version of sendmail-reject that looked for daemon=MTA, but apparently you have a recent commit https://github.com/fail2ban/fail2ban/commit/b710d5b6c7b0249a257246b15ce6408630a5fe9c#diff-32449c7802383924eb815aab4a5e0ca4cef0e56b1dc0f9442c3f14c65efa3821R33 which changed it to daemon=\S+ That is actually better because even thought the default is MTA-v[46] now, the user might change it in the sendmail.mc with DAEMON_OPTIONS, so your change is more compatible.

fail2ban/tests/files/logs/sendmail-reject
```diff
@@ -115,3 +122,12 @@ Mar 29 22:51:43 server sendmail[3529565]: xA32R2PQ3529565: [192.0.2.2] did not i
 Mar 29 22:51:45 server sm-mta[50437]: 06QDQnNf050437: example.com [192.0.2.3] did not issue MAIL/EXPN/VRFY/ETRN during connection to IPv4
 # failJSON: { "time": "2005-03-29T22:51:46", "match": true , "host": "2001:DB8::1", "desc": "IPv6" }
 Mar 29 22:51:46 server sm-mta[50438]: 06QDQnNf050438: example.com [IPv6:2001:DB8::1] did not issue MAIL/EXPN/VRFY/ETRN during connection to IPv6
+
+# filterOptions: {"mode": "aggressive"}
+
+# failJSON: { "time": "2005-06-17T13:03:43", "match": true, "host": "127.0.0.1" }
+Jun 17 13:03:43 <mail.notice> robin sm-mta[26864]: 55HH324M026864: lost input channel from localhost [127.0.0.1] to MTA-v4 after rcpt
```
sebres, line 129: Hmm... I don't think it is correct, if loopback means here the local address, does it?
Or rather you simply tested the case from localhost?

config/filter.d/sendmail-reject.conf
```diff
@@ -20,23 +20,24 @@ before = common.conf
 [Definition]
 
 _daemon = (?:(sm-(mta|acceptingconnections)|sendmail))
-__prefix_line = %(known/__prefix_line)s(?:\w{14,20}: )?
+__prefix_line = %(known/__prefix_line)s<F-MLFID>(?:\w{14,20}: )?</F-MLFID>
```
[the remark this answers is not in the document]
sebres: > Are you saying that it can't be relocated between the ML-FID because of side effects when log lines like NOQUEUE and STARTTLS are written, containing no ID

Yes. But the correct answer is rather dunno... I'm not familiar with sendmail, especially for every platform. 
But if some doesn't produce that IDs in logs, it'd be not backwards compatible anymore, yes.

> If that is true then I guess there is no way to support the BSD format for mail logs where you need to match lines without IP

No, there are several options, but I guess we could rewrite the RE to cover both states, just with rewrite of prefix-line. I'll try to extend it now.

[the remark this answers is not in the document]
billfor: Here is a sample log (mildly modified to include lines without the ID and consecutive lines without ids). We want the filter to correctly block the ip address of the bad user (192.168.1.44). It appears to work correctly for me (bans the IP). I didn't add extra lines though.

[mail.txt](https://github.com/user-attachments/files/22099567/mail.txt)

```
Hi,

The IP 192.168.1.44 has just been banned by Fail2Ban after
2 attempts against sendmail-reject.

```

fail2ban/tests/files/logs/sendmail-reject
sebres: So it was an artificially constructed line either... Then good.
You could use some IP from test subnet 192.0.2.0/24.

config/filter.d/sendmail-reject.conf
```diff
@@ -20,23 +20,25 @@ before = common.conf
...
-            ^<F-NOFAIL>from=<[^@]+@[^>]+></F-NOFAIL>, size=\d+, class=\d+, nrcpts=\d+, bodytype=\w+, proto=E?SMTP, daemon=\S+, relay=\S+ \[%(addr)s\]$
+            ^<[^@]+@[^>]+>\.\.\. (?:No such user here|User unknown)$
+            ^<F-NOFAIL>from=<[^@]+@[^>]+></F-NOFAIL>, size=\d+, class=\d+, nrcpts=\d+,(?: bodytype=\w+,)? proto=E?SMTP, daemon=\S+, relay=(?:\S+ )?\[%(addr)s\]$
 
 mdre-normal =
 
 mdre-extra = ^(?:\S+ )?\[%(addr)s\](?: \(may be forged\))? did not issue \S+ during connection
 
 mdre-aggressive = %(mdre-extra)s
+                  ^lost input channel from (?:\S+ )?\[%(addr)s\] to .*? after rcpt$
```
sebres, line 40: Is a catch-all `.*?` (especially non-greedy) really needed here?..

Cannot it be either:
```diff
-                  ^lost input channel from (?:\S+ )?\[%(addr)s\] to .*? after rcpt$
+                  ^lost input channel from (?:\S+ )?\[%(addr)s\] to \S+ after rcpt$
```

Or if a part `after rcpt` is not important (means no matter what it is), could not we use right unanchored RE too:
```diff
-                  ^lost input channel from (?:\S+ )?\[%(addr)s\] to .*? after rcpt$
+                  ^lost input channel from (?:\S+ )?\[%(addr)s\]
```
Sure, it can lost the channel every time, e. g. even if the connection is unstable, but it is `aggressive` mode, so by large `maxretry` shall not generate false positives en masse. 
But I'm not really familiar with sendmail, just don't like catch-alls where they are not expected.

billfor: Yeah sorry I put the other one back to the way you had it with \S+, but I forgot to update that one. If the daemon name is empty or has lots of spaces, then it matched. Is the issue with lazy .*? related to performance? I guess just put it back to \S+

billfor: also yeah if you want to eliminate it entirely you can truncate everything after "to". If you do that, you will catch "after mail" in addition to "after rcpt"  . I used to have (mail|rcpt), but got rid of the mail, because I think that will only show up if you are sending outbound, so its probably user error and not something intentional.

sebres:
> I guess that no longer relies on the defaults in common.conf

Why? It still does. Almost all parts of common prefix are optional, and adding of `(?:<mail\.[^\>]+> )?` also optional changes nothing for any other formats, excepting that it'd match `<mail.infp>`, `<mail.notice>` etc before `<F-MLFID>` begins.

> Certainly the user could redefine the service with SYSLOG_FACILITY

And then it'd not be default anymore and may expect customization of __prefix_line. I don't care about non standard configuration, because it is simply almost impossible to consider everything, especially across all distributions and versions.
This is pretty simple - customized logging needs customized filters.

There is no way to fix it "properly" without to provide this as a parameter (so would expect user configuration) or alternatively to rewrite `(?:\w{14,20}: )?` as mandatory... Dunno if it is possible and would not break some distros or MTA versions, which doesn't log this (session?) ID by default. Thus it is currently minimal invasive way.

And last but not least, as [wiki :: Best practice](https://github.com/fail2ban/fail2ban/wiki/Best-practice) says:
- *use more precise regex parts as predefined in filter or common config-files substitutions, e. g. redefine `__prefix_line` with your own shorter and more efficient RE exactly matching related part of your log-messages, because interpolation of default RE for the `__prefix_line` may be too long (due to all targeting systems or distributions and backwards-compatibility reasons) and therefore not optimal for application on your system.*

PR #4025, opened 2025-06-25
Filters created by conquestace for services

conquestace:

### DIFF
--- a/config/filter.d/gitea-auth.conf
+++ b/config/filter.d/gitea-auth.conf
@@ -1,3 +1,3 @@
 [Definition]
-failregex = .*Failed authentication attempt.*remote_addr=.*$
+failregex = ^.*Failed authentication attempt.*from <HOST>[:0-9]*:.*$
 ignoreregex =
--- a/config/filter.d/gitea-auth.conf
+++ b/config/filter.d/gitea-auth.conf
@@ -1,0 +1,3 @@
+[Definition]
+failregex = .*Failed authentication attempt.*remote_addr=.*$
+ignoreregex =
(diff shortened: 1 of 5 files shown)


config/filter.d/gitea-auth.conf
```diff
@@ -0,0 +1,3 @@
+[Definition]
+failregex = .*Failed authentication attempt.*remote_addr=.*$
```
sebres, line 2: The RE doesn't contain `<HOST>` or `<ADDR>`
Beside that such REs cannot be accepted - they are vulnerable because unanchored, has an interim catch-all, etc.

config/filter.d/homeassistant-auth.conf
```diff
@@ -0,0 +1,3 @@
+[Definition]
+failregex = Login attempt or request with invalid authentication from .*
```
sebres, line 2: Same as above.

sebres (changes requested):
Almost all REs must be fixed.
The PR expects test cases (at least 1 line with JSON for every match/every RE).

sebres (changes requested):
1. `^.*` and `.*$` are not anchors. Thus the RE is slow (especially on parasitic messages), unprecise and the vulni cannot be precluded.
2. another catch-alls (e. g. `.*`) are not justified and shall be rewritten more precise.
3. test cases are still missing.

I could also adjust the PR, but I need a matched log-line (complete line with date etc) each RE.

Sorry but it remains draft unless fixed.

conquestace:
I understand what I need to do. I am going to make a PR for a single filter at a time with the proper testing.

PR #4037, opened 2025-07-20
filter.d/asterisk: fix regex to match "No matching endpoint found" with retry info

kusaka-0107:

## Issue
This PR fixes the regex pattern that fails to match newer Asterisk log entries for "No matching endpoint found" events.

## Current behavior
The regex doesn't match when Asterisk logs use the newer format with detailed retry information:
```
Request 'REGISTER' from '"8464" <[sip:8464@34.97.118.227](https://hoge.foo/hoge/d98e5e63-9371-4af8-9884-2d5334de7940)>' failed for '117.55.202.19:7505' (callid: 4250754195) - No matching endpoint found after 7 tries in 1.037 ms
```

## Fix
Update the regex pattern to handle both:
- Legacy format: `No matching endpoint found from '192.168.1.100' (message ignored)`
- New format: `Request 'REGISTER' from ... failed for 'IP:PORT' ... - No matching endpoint found after X tries in Y ms`


kusaka-0107:
> I reverted the RE back, but relieved the end-anchor - `\b[^']*` does the job, and would ignore any text at and without single quote inside (also `after X tries in Y ms`), as well as prevent false match on theoretic inject on foreign data (since there is a non greedy catch-all inside and completely open end would make RE too weak). Test (inclusive new one) pass, thus I will merge it now. Thx!

@sebres 
Thank you for the improvement and merge!

Your solution using `\b[^']*` is much cleaner and more robust than my approach. I learned a lot about balancing flexibility and security in regex patterns. The explanation about preventing injections while handling format variations was really helpful.

Thanks for taking the time to review and improve my PR. This has been a great learning experience!

PR #4047, opened 2025-08-04
Update nginx-limit-req filter

billfor:
Updating to include delayed requests and connection limiting.


config/filter.d/nginx-limit-req.conf
billfor, line 1: failregex = ^%(__prefix_line)s(?:limiting|delaying) (?:request|connection)s?(?:, excess: [\d\.]+,?)? by zone "(?:%(ngx_limit_req_zones)s|%(**ngx_limit_con_zones**)s)", client: <ADDR>,

If you don't want ngx_limit_con_zones then take the variable out (you left it in but not used). Normally I define these as two separate things.

billfor (commented):
Looks like you kept ngx_limit_con_zones = [^"]+ but it is not referenced any more. Normally you would have a separate zone for the connection limits as opposed to the request limits, which is why I had two lines. 

(?:%(ngx_limit_req_zones)s)|

config/filter.d/nginx-limit-req.conf
sebres: done (ngx_limit_con_zones removed)

PR #4048, opened 2025-08-04
update nginx limit-req filter again

billfor:
Remove the extra variable and rename the consolidated variable.

### DIFF
--- a/config/filter.d/nginx-limit-req.conf
+++ b/config/filter.d/nginx-limit-req.conf
@@ -35,8 +35,7 @@ before = nginx-error-common.conf
 #
 #   ngx_limit_req_zones = lr_zone|lr_zone2
 #
-ngx_limit_req_zones = [^"]+
-ngx_limit_con_zones = [^"]+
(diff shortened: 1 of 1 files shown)


sebres:
Sorry, but better no renames here, because the parameter can be supplied from outside:
```ini
[jail1]
filter = nginx-limit-req[ngx_limit_req_zones=zone1]

[jail2]
filter = nginx-limit-req[ngx_limit_req_zones=zone2]
```
so the change would be backwards incompatible.

I'll adjust it with deletion of `ngx_limit_con_zones` only.

billfor:
i suppose the name of the filter being limit-req also is a bit confusing since I extended it to limit connections per ip as well

sebres:
I know, just backwards compat is must in minor releases and a break of it can be justified only with some grave bug fix.
It can be renamed later in the next major release, or could be done with a revert of `nginx-limit-req` back to original and introduction of common filter like `nginx-limit` combining both limits.

PR #4050, opened 2025-08-10
Use systemd as default backend for Debian in .deb packages

luisg123v:
Even though it was added to the `master` branch when merging #3527 (in 0b255a8), it is missing from the `debian` branch.

### DIFF
--- a/config/paths-debian.conf
+++ b/config/paths-debian.conf
@@ -13,6 +13,7 @@ banaction = nftables
 banaction_allports = nftables[type=allports]
 
 sshd_backend = systemd
+postfix_backend = systemd
 
 syslog_mail = /var/log/mail.log
 


luisg123v:
@sebres thanks for the info.

> Branch `debian` gets normally updated (and master merged) implicitly before release.

Is that expected to occur soon? If no, would you consider merging this one as a workaround?

> Also note that this branch can only affect our .deb package released here.

I'm aware of that, I'm installing the .deb package from here.

> The official debian release and repository is in https://salsa.debian.org/python-team/packages/fail2ban/

Thanks for pointing to the exact repo. I've submitted this change there:

https://salsa.debian.org/python-team/packages/fail2ban/-/merge_requests/18

Unfortunately, it probably won't be accepted because they are  handling these configurations from several other files and patches.

Issue #4074, opened 2025-09-20
[FR]: froxlor


ikaRus1:
Hi,

Debian 12 ships with Fail2Ban v1.0.2 which has a froxlor-auth.conf filter that does not detect failures.

This is a sample fail line from syslog, I tried to build a new regex but didn’t get very far…
```
Sep 20 14:44:52 hostname froxlor[181107]: froxlor.WARNING: Unknown user tried to login. {"source":"login","action":"50","user":"141.226.1.2"} []
```
Help appreciated.

TIA & regards,
Dudi

para-do-x:
Hello ikaRus1, 

you are right it is not working. Looks like the froxlor logging was change. 
I did worked today on the regex and got it work.

First what I found is to change in  /etc/fail2ban/jail.conf the logpath syslog_user:

```
[froxlor-auth]

port    = http,https
logpath  = %(syslog_user)s
backend  = %(syslog_backend)s
```

Second was the new regex. I learned things.
 
I modified the froxlor-auth.conf file and test it in my froxlor installation. Here is the file, so you can test it. (remove the .txt from file) 
[froxlor-auth.conf.txt](https://github.com/user-attachments/files/22454915/froxlor-auth.conf.txt)

btw, I'm installed version 1.1.0 from the deb file from this repository, to my Debian 12, and it works. 
[fail2ban_1.1.0](https://github.com/fail2ban/fail2ban/releases/download/1.1.0/fail2ban_1.1.0-1.upstream1_all.deb)

Hope it will help you. 

Br,

Para-do-x

ikaRus1:
Thank you!

I’ll test soon.

Regards,
Dudi

ikaRus1:
Hi,

I had to change the hostname in regex to match any hostname, but this works.

Thanks & regards,
Dudi

para-do-x:
Hello Dudi, 

great that it work. 
What did you changed now?
The hostname was already out, but may not to not worked with any. 
I could think problem is with the first 'froxlor'?

Br,

Para-do-x

para-do-x:
Hi, 

OK, this froxlor is not the hostname, it's like in your example 'froxlor[181107]'. 
But Understand, my hostname was only one word, so '+' did work. 
Right would be then the '*' to cover complex hostname's also.

Thank you for the feedback. 

Br,

Para-do-x

Issue #2926, opened 2021-02-02
Froxlor filter not matching version 0.10.23.1-3


Martin-vH:
### Environment:

- Fail2Ban version (including any possible distribution suffixes): 0.10.2 (0.10.2-2.1)
- OS, including release name/version: Debian Buster 4.19.0-13-amd64 #1 SMP Debian 4.19.160-2 (2020-11-28) x86_64 GNU/Linux
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

### The issue:

Hello folks, I don't know, whether the report is at the right place here, but I found no other quick information, how fail2ban filters for 3rd-party tools come into the Debian package.

The issue is quite simple; the filter provided for froxlor-auth is not matching current logging output.

See my adjustments below. Thanks in advance.

BR Martin

### Configuration, dump and another helpful excerpts

Filter provided by @jomu78 with my testing adjustments:

```
# Fail2Ban configuration file to block repeated failed login attempts to Froxlor installation(s)
#
# Froxlor needs to log to Syslog User (e.g. /var/log/user.log) with one of the following messages
# <syslog prefix> froxlor.WARNING: Unknown user '<USER>' tried to login. {"source":"login","action":"50","user":"<HOST>"} []
# <syslog prefix> froxlor.WARNING: User '<USER>' tried to login with wrong password. {"source":"login","action":"50","user":"<HOST>"} []
#
# Author: Joern Muehlencord
#

[INCLUDES]

# Read common prefixes. If any customizations available -- read them from
# common.local
before = common.conf


[Definition]

#_daemon = Froxlor
_daemon = froxlor

# Option:  failregex
# Notes.:  regex to match the password failures messages in the logfile. The
#          host must be matched by a group named "host". The tag "<HOST>" can
#          be used for standard IP/hostname matching and is only an alias for
#          (?:::f{4,6}:)?(?P<host>[\w\-.^_]+)
# Values:  TEXT
#

#prefregex = ^%(__prefix_line)s\[Login Action <HOST>\] <F-CONTENT>.+</F-CONTENT>$
prefregex = ^%(__prefix_line)s.+ froxlor\.WARNING\: <F-CONTENT>.+</F-CONTENT>$

#failregex = ^Unknown user \S* tried to login.$
#            ^User \S* tried to login with wrong password.$
[...]

PR #4075, opened 2025-09-22
Froxlor auth update

para-do-x:

- [x] **LIST ISSUES** this PR resolves or describe the approach in detail
      #4074 
      Update:
      #2926
  failregex[1]:    
`^%(__prefix_line)s\S+ froxlor\[\S+\]: froxlor.WARNING: Unknown user tried to login. {"source":"login","action":"50","user":"<HOST>"\S* \[\]
`
log[1]:
`2025-09-21T17:46:18.311379+02:00 hostname froxlor[1055219]: froxlor.WARNING: User tried to login with wrong password. {"source":"login","action":"50","user":"1.2.3.4"} []`

  failregex[2]:
`^%(__prefix_line)s\S+ froxlor\[\S+\]: froxlor.WARNING: User tried to login with wrong password. \{"source":"login","action":"50","user":"<HOST>"\S* \[\]`
log[2]:
`2025-09-21T16:30:13.118232+02:00 hostname froxlor[1054438]: froxlor.WARNING: Unknown user tried to login. {"source":"login","action":"50","user":"1.2.3.4"} []`
    a sample user.log is attached, with other froxlor entries. 
[user.log](https://github.com/user-attachments/files/22477566/user.log)

Update2:
Changed test file (failJSON)

Update3: 
Changed <ADDR> back to <HOST> which match like before and the test file.
As here will always be a ip, I would more go for <ADDR>. 
But not sure if <HOST> would be more standard. So I go for this now.  

- [x] **PROVIDE ChangeLog** entry describing the pull request
   updated the regex to the new logging situation for froxlor and changed logpath in jail.conf. 

   Some update to this PR, for a better Quality:
  - issue #2926 is also related to this PR. It would saved a bit time, if I would found it before, but I learn in this way more. 
   - Also the regex's will better match and will have less false positive.
   - I also see the thoughts in this old issue. But better to have a working filter which could be used, as to have one, which not work and  you could think it works. 
   - Froxlor is a server management, so if someone try to attack it, it would be better to block. Also if this could hit some users, which forgotten the password. At the end the server admin should think which risk is higher.   

And at the end, sorry for quick PR info this morning.

Br,

Para-do-x


sebres:
The change `<ADDR>` to `<HOST>` would change nothing here.
1. the value `1.2.3.4` is correct IP address (`<HOST>` in opposite to `<ADDR>` would additionally match a FQDN hostname)
   So it was pretty correct previously.
2. the [error in test](https://github.com/fail2ban/fail2ban/actions/runs/17969130238/job/51125949723#step:8:581) says:
  `UTC Time  mismatch 1758465013 (2025-09-21T14:30:13) != 1116685813.0 (2005-05-21T14:30:13)`
which going to the wrong timestamp in failJSON:
```diff
- # failJSON: { "time": "2005-05-21T16:30:13", "match": true , "host": "1.2.3.4" }
+ # failJSON: { "time": "2025-09-21T16:30:13", "match": true , "host": "1.2.3.4" }
2025-09-21T16:30:13.118232+02:00 hostname froxlor[1054438]: froxlor.WARNING: Unknown user tried to login. ...
```

para-do-x:
Hello sebres,

ChangeLog updated now. 

Files looking smarter now. ;-)
Tested on my froxlor server and it works. 

A generally question:
Out from [froxlor's Security Policy](https://github.com/Froxlor/Froxlor?tab=security-ov-file)

Supported versions

    ️✅ 2.2.x (main git-branch)
    ️✅ 2.1.x (v2.1 git-branch)
    ❌ 2.0.x (2.0.x-tags)
    ❌ 0.10.x (0.10.x-tags)
    ❌ other git-branches

So already in the version 4 years ago, the log format was changed and this is not in support. So is this may it is unnecessary ballast?

Thank you.

Br,

Para-do-x

sebres:
> So is this may it is unnecessary ballast?

You mean the `type=1` thing? As long as it remains low cost (as it is right now) - it doesn't bother here at all.

Test pass, thus I'll merge it now.

PR #4091, opened 2025-10-23
Add subnet filters

GAS85:
Add IP Ranges block filter based on https://wiki.sitnikov.eu/doku.php?id=howto:fail2ban&s[]=fail2ban#fail2ban_ip_ranges
User can add different IP ranges to the `/var/log/ip-blacklist.log` in format `10.10.10.10/8 [$(date +"%Y-%m-%d %H:%M:%S")]`. They will be banned by e.g. iptables all ports.

Solves: #927

- [X] **LIST ISSUES** this PR resolves or describe the approach in detail
- [X] **PROVIDE ChangeLog** entry describing the pull request


sebres:
1. it doesn't really solve #927.
2. the approach is questionable, who would write in that file and how it is better than:
  - `fail2ban-client set "$jail" attempt 192.0.2.1/24 ...` 
  - `fail2ban-client set "$jail" banip 192.0.2.1/24`

config/filter.d/filter-ip-blacklist.conf
```diff
@@ -0,0 +1,43 @@
+[Definition]
+ 
+# Option:  failregex
+# Notes :  Detection of blocked ip addresses.
+# Values:  TEXT
+#
+ 
+failregex = ^(?P<ip4>\d{1,3}(?:\.\d{1,3}){3}/\d{1,2})\s+
```
GAS85:
```suggestion
failregex = ^(?P<ip4>\d{1,3}(?:\.\d{1,3}){3}/\d{1,2})\s+\[(?P<timestamp>[^\]]+)\].*$
```

GAS85 (commented):
Add timestamp to the regex

config/filter.d/filter-ip-blacklist.conf
sebres: 1. For which reason this may be needed? 
    1.a. Normally it shall be matched by `datepattern`
    1.b. The part of messages matched by `datepattern` is cut out from message before `failregex` gets applied (see the note in [the manpage](https://manpages.debian.org/stable/fail2ban/jail.conf.5.en.html#failregex~2))
    1.c. The timestamp shall be better at 0 position in log if possible.
2. `.*$` is not an anchor and shall be removed or replaced with some non catch-all (like `\S*$`).

PR #4107, opened 2025-11-13
Action: Implement cloudflare IP list 

rvojcik:
Cloudflare IP List action which implement action which add and remove in global IP Lists.
You could use IP Lists in cloudflare Access rules as you wish.

This implemntation manage IP addresses in IP List. 

You could also create different rules for different domains based on your needs.

### DIFF
--- a/config/action.d/cloudflare-list.conf
+++ b/config/action.d/cloudflare-list.conf
@@ -54,15 +54,11 @@ actionban = curl -s -X PUT "<_cf_api_url>" \
 #          <time>  unix timestamp of the ban time
 # Values:  CMD
 #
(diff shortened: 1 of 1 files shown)


Skamasle:
Hi, why remove original jq ?

`{ jq -r '.result[0].id' 2>/dev/null || `

I now isnt installed by default, but in my opinion if available, it should be used as the first option.

rvojcik:
@Skamasle thank you for the comment. I have no explanation, it somehow drop off from new configuration file, probably it's just my mistake. 
I will put it back and test it

rvojcik:
I also flip order of some curl arguments to make it more like original cloudflare.conf for better read.
I tested it, it should work as expected with or without `jq` command.

PR #4133, opened 2026-01-07
Update jail.conf.5 documentation for action tags

B1gG:
Missing tags added to the man page:
family - IP address family string
ip-rev - Reverse DNS PTR record
ip-host - Hostname of the IP
fid - Failure/ticket ID
bantime - Effective ban duration
bancount - Number of times IP has been banned
restored - Flag for restored tickets (0 or 1)
F-* - Free-form filter match tags
raw-ticket - Raw ticket representation
jail.banned - Currently banned count
jail.banned_total - Total ban count
jail.found - Current failure count
jail.found_total - Total failure count

### DIFF
--- a/man/jail.conf.5
+++ b/man/jail.conf.5
@@ -452,28 +452,67 @@ two commands to be executed.
 The following tags are substituted in the actionban, actionunban and actioncheck (when called before actionban/actionunban) commands.
 .TP
 .B ip
-IPv4 IP address to be banned. e.g. 192.168.0.2
+IPv4 or IPv6 IP address to be banned. e.g. 192.168.0.2 or 2001:db8::1
+.TP
+.B family
+IP address family as a string. e.g. inet4 or inet6
+.TP
+.B ip-rev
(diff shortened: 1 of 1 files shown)


sebres (approved):

sebres:
Looks good, thanks!
I'll merge it later if no objections follow.
I'd also like to adjust it a bit, e. g. a bit more about `ip` and `fid` - for instance they are the same if ID of ticket is an IP (no `<F-ID>` tag used), but may differ if filter uses `<F-ID>...</F-ID>` to capture failure identifier and it is not IP (captured by `<ADDR>`). Also they can contain not IP only, but also subnet, depending on used capture-tag in regex - see #2560.
And that tags `bantime` and `bancount` get updated asynchronously (for known as bad evildoers, if `bantime.increment` is enabled), so may contain initial values in `actionban` and guaranteed correct only within `actionprolong` (and `actionunban`).

zmvs9spnq6-maker (approved):

PR #4149, opened 2026-02-26
fix: replace 20 bare except clauses with except Exception

haosenwang1018:
Replace bare `except:` clauses with `except Exception:` for PEP 8 compliance. Bare except catches SystemExit, KeyboardInterrupt, and GeneratorExit which is rarely intended.

### DIFF
--- a/fail2ban/tests/actiontestcase.py
+++ b/fail2ban/tests/actiontestcase.py
@@ -241,14 +241,14 @@ class CommandActionTest(LogCaptureTestCase):
 		setattr(self.__action, 'ac', "<a><b>")
 		setattr(self.__action, 'ab', "<ac>")
 		setattr(self.__action, 'x?family=inet6', "")
-		# produce self-referencing properties except:
+		# produce self-referencing properties except Exception:
 		self.assertRaisesRegex(ValueError, r"properties contain self referencing definitions",
 			lambda: self.__action.replaceTag("<a><b>", 
 				self.__action._properties, conditional="family=inet4")
 		)
 		# remote self-referencing in props:
 		delattr(self.__action, 'ac')
-		# produce self-referencing query except:
+		# produce self-referencing query except Exception:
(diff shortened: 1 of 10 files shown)


fail2ban/tests/actiontestcase.py
```diff
@@ -241,14 +241,14 @@ def testReplaceTagSelfRecursion(self):
 		setattr(self.__action, 'ac', "<a><b>")
 		setattr(self.__action, 'ab', "<ac>")
 		setattr(self.__action, 'x?family=inet6', "")
-		# produce self-referencing properties except:
+		# produce self-referencing properties except Exception:
```
sebres, line 244: Since this (and one bellow) is a comment, I guess you used find&replace, didn't you?

If yes, an approach with automatic replacement may be questionable, because in constructs like below it doesn't matter whether the exception is an error or keyboard interrupt or exit:
```python
...
while True:
  try:
    ...
  except:
    do_something_by_unexpected_end(); # because we'd leave the loop.
    break
...
```

fail2ban/tests/actiontestcase.py
```diff
@@ -241,14 +241,14 @@ def testReplaceTagSelfRecursion(self):
...
 		self.assertRaisesRegex(ValueError, r"properties contain self referencing definitions",
 			lambda: self.__action.replaceTag("<a><b>", 
 				self.__action._properties, conditional="family=inet4")
 		)
 		# remote self-referencing in props:
 		delattr(self.__action, 'ac')
-		# produce self-referencing query except:
+		# produce self-referencing query except Exception:
```
sebres, line 251: same as above.

sebres (changes requested):

haosenwang1018:
CI failures require manual investigation. Closing for now.

sebres:
> CI failures require manual investigation.

I did not see failures in tests of CI. The only "blocker" was my request to change from a review (comment lines).
I'd also like to get the answer to the question I asked above.

PR #4178, opened 2026-04-28
Fix SMTP line length to comply with RFC 5321/5322

mohamedsolaiman:
## Summary

Fixes #4170

The SMTP action (`config/action.d/smtp.py`) previously sent email bodies without line wrapping, which could result in lines exceeding the RFC 5321 maximum of 998 characters. This is particularly problematic when ban notifications include long log lines, whois output, or match data.

## Changes

- Added a `_wrap_text()` static method to the `SMTPAction` class that wraps long lines at 78 characters (per RFC 5322 recommendation for readability)
- Integrated `textwrap.TextWrapper` with settings to preserve existing paragraph structure, blank lines, and formatting
- Applied line wrapping in `_sendMessage()` before constructing the `EmailMessage`

## Technical Details

- Uses Python’s built-in `textwrap` module (no new dependencies)
- `replace_whitespace=False` preserves intentional whitespace in log output
- `drop_whitespace=False` maintains paragraph breaks
- `expand_tabs=False` keeps tab formatting intact
- Short lines and blank lines are passed through unchanged
- Only lines exceeding 78 characters are wrapped

## Testing

The existing test suite in `fail2ban/tests/action_d/test_smtp.py` continues to pass, as the default email messages are short enough to not require wrapping. Long lines from whois output, match logs, or other dynamic content will now be properly wrapped.

## RFC References

- [RFC 5321 Section 4.5.3.1.6](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.6): Lines MUST NOT exceed 998 characters
- [RFC 5322 Section 2.1.1](https://datatracker.ietf.org/doc/html/rfc5322#section-2.1.1): Lines should be limited to 78 characters for readability

### DIFF
--- a/config/action.d/smtp.py
+++ b/config/action.d/smtp.py
@@ -19,6 +19,7 @@
 
 import socket
 import smtplib
+import textwrap
 import email.policy
 from email.message import EmailMessage
 from email.utils import formatdate, formataddr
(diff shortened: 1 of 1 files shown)


sebres (changes requested):
Looks good, however... 

- too many unneeded changes, e. g. whitespace in padding are changed (tabs replaced with spaces?)
- doesn't EmailMessage use a policy where `max_line_length` is set to 78 characters by default anyway? If it doesn't (by default), maybe some policy can be used instead of textwrap package?

PR #4174, opened 2026-04-17
docs: Improve INCLUDES section documentation in jail.conf.5

vincent067:
## Summary

This PR addresses issue #4076 by improving the documentation for the `[INCLUDES]` section in the jail.conf man page.

## Changes

- Added explanation that included files are resolved relative to the configuration directory
- Documented support for multiple files (space-separated) and globbing patterns
- Provided practical examples for common use cases:
  - Including a single file
  - Including multiple files with both `before` and `after`
  - Using glob patterns to include multiple matching files
- Added a note about the current behavior when included files don't exist

## Rationale

The current documentation only briefly mentions the `before` and `after` directives without explaining:
1. How file paths are resolved
2. That multiple files can be included
3. That globbing is supported
4. What happens when files are missing

This has led to confusion for users trying to modularize their fail2ban configurations.

## Testing

- Verified the man page syntax is correct
- Examples follow the same format as other sections in the man page

Closes #4076

Please let me know if you'd like any adjustments! 😊

Issue #4076, opened 2025-09-28
[BR]: INCLUDES sections are poorly documented and failed inclusions are not reported


kyz:
### Environment:

- Fail2Ban version  : 1.1.0-8
- OS, including release name/version :  Debian 13 "Trixie"
- [x] Fail2Ban installed via OS/distribution mechanisms
- [x] You have not applied any additional foreign patches to the codebase
- [x] Some customizations were done to the configuration (provide details below is so)

### The issue:

I've noticed some people having great trouble getting the INCLUDES section to work.

Firstly, it is not clear, nor documented in the `jail.conf` manpage, how to define multiple files for inclusion. No example is given.

Secondly, fail2ban doesn't fail or even warn you, if any included file does not exist.

#### Steps to reproduce

Enter something like this in config:
```
[INCLUDES]
before = file1.conf, file2.conf
```
or even just this:
```
[INCLUDES]
before = file1.conf
```
when file1.conf is in a different directory

#### Expected behavior
fail2ban should warn, e.g.
```
# fail2ban-client --test
2025-09-28 01:23:45,678 fail2ban.configparserinc[420]: WARNING File 'file1.conf, file2.conf' included by '/etc/fail2ban/jail.conf' doesn't exist
```

#### Observed behavior
fail2ban does not warn on this config error, nor does it document how to use it in the manpage.

#### Any additional information
My suggested change for the manpage:
```diff
--- jail.conf.5.orig	2025-09-28 01:14:36.679351925 +0100
+++ jail.conf.5	2025-09-28 02:10:52.709529223 +0100
@@ -83,10 +83,27 @@
 
 .TP
 .B before
-indicates that the specified file is to be parsed before the current file.
+indicates that the specified file(s) are to be parsed before the current file.
+Filenames should be given one per line, and can either be absolute paths, or
+relative to the file doing the including.
 .TP
 .B after
-indicates that the specified file is to be parsed after the current file.
+indicates that the specified file(s) are to be parsed after the current file
+and included in its definition. One use-case for this is to designate another
+file as the official location to make edits that extend or override a filter.
+Files listed here do not need to exist.
+.RE
+
+Example of usage:
+.RS
+.nf
+[INCLUDES]
+before = ../common/common.conf
+         extras.conf
+         /usr/local/etc/fail2ban/common-local.conf
+
+after = user-overrides.conf
+.fi
 .RE
 
[...]

PR #4179, opened 2026-04-28
Improve INCLUDES section documentation in jail.conf

mohamedsolaiman:
## Summary

Fixes #4174

The `[INCLUDES]` section in `jail.conf` previously had minimal documentation, making it unclear to users how distribution-specific paths are loaded and how to customize them for their setup.

## Changes

Added comprehensive documentation to the `[INCLUDES]` section explaining:

1. **`before` vs `after` directives** - How each directive works and their precedence rules ("before" files are loaded first and can be overridden; "after" files override the current file)

2. **Purpose of paths-\<distro\>.conf files** - What they provide (distribution-specific log file locations and backend preferences)

3. **List of all available distribution paths files** - Debian, Fedora, openSUSE, Arch, FreeBSD, and macOS with descriptions of which distributions they cover

4. **Customization guidance** - How users can create `paths-overrides.local` or override individual variables in `jail.local`

5. **Practical examples** - Switching to Fedora paths, overriding a specific log path

## Motivation

The lack of clear documentation has been a source of confusion for users who:
- Don't know which paths file to select for their distribution
- Don't understand the difference between `before` and `after` includes
- Don't know how to properly customize log file paths

This is a documentation-only change with no functional impact.

### DIFF
--- a/config/jail.conf
+++ b/config/jail.conf
@@ -32,6 +32,41 @@
 
 [INCLUDES]
 
+# The INCLUDES section controls which configuration files are loaded before
+# or after this file. This is primarily used to load platform-specific log
+# file paths that vary between Linux distributions.
+#
+# Directive "before": The referenced file is loaded BEFORE the current file.
+#   Values in the current file will override those in the "before" file.
+#   This is useful for providing defaults that can be overridden.
+#
+# Directive "after": The referenced file is loaded AFTER the current file.
+#   Values in the "after" file will override those in the current file.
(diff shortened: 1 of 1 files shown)


sebres:
> Fixes #4174

How it fixes #4174, when it is also a PR improving the documentation in man-page. Did you probably mean #4076?

Well, I don't think everything belongs to the comment in config (the man-page is exactly for that purposes). 
I'm not against some shirt summarization, but config is the wrong place for the whole stuff in my oppinion.

PR #4180, opened 2026-04-28
Add Cockpit web management interface filter

mohamedsolaiman:
## Summary

Adds a new fail2ban filter for **Cockpit**, the popular web-based server management interface that is included by default in many Linux distributions (Fedora, RHEL, CentOS, Rocky, AlmaLinux, Ubuntu).

Cockpit is one of the most widely deployed web-based server management tools, yet fail2ban does not include a filter for it. This leaves servers running Cockpit vulnerable to brute-force authentication attacks without fail2ban protection.

### New Filter: `config/filter.d/cockpit.conf`

Matches authentication failures from multiple PAM modules:
- `pam_unix` (default on most systems)
- `pam_sss` (SSSD/FreeIPA environments)
- `pam_ldap` (LDAP authentication)

Features:
- Uses `prefregex` for efficient prefix matching
- Extracts remote host (`rhost`) and username from PAM log lines
- Includes `journalmatch` for systemd journal backend support
- Only matches the `cockpit` PAM service (won't accidentally match sshd or other services)

### Test Samples: `fail2ban/tests/files/logs/cockpit`

Includes positive and negative test cases:
- pam_unix auth failure with/without username
- pam_sss (SSSD) auth failure
- pam_ldap auth failure
- IPv6 rhost support
- Negative tests: authentication success, wrong PAM service

## Example Jail Configuration

```ini
[cockpit]
enabled = true
port = http,https,9090
filter = cockpit
logpath = %(syslog_authpriv)s
backend = %(syslog_backend)s
maxretry = 3
bantime = 1h
```

## Typical Log Lines Matched

```
Mar  8 09:37:44 server cockpit-session[12345]: pam_unix(cockpit:auth): authentication failure; logname= uid=0 euid=0 tty= ruser= rhost=192.0.2.1  user=admin
Mar 10 14:15:30 server cockpit-session[34567]: pam_sss(cockpit:auth): authentication failure; logname= uid=0 euid=0 tty= ruser= rhost=192.0.2.3  user=testuser
```

## Testing

Regex patterns have been validated against typical Cockpit log output. The filter follows the same patterns as existing filters like `proxmox.conf` for PAM-based authentication matching.

### DIFF
--- a/config/filter.d/cockpit.conf
+++ b/config/filter.d/cockpit.conf
@@ -1,0 +1,63 @@
+# Fail2Ban filter for Cockpit web-based server management
+#
+# Cockpit is a web-based graphical interface for servers.
+# It is included by default in many Linux distributions including
+# Fedora, RHEL, CentOS, Rocky, AlmaLinux, and Ubuntu.
+#
+# Authentication failures are typically logged via PAM to the system
+# auth log (e.g., /var/log/auth.log or /var/log/secure) or to
+# the systemd journal.
+#
+# To enable Cockpit logging, ensure that Cockpit is configured to
+# use PAM authentication (default).
+#
+# Jail example:
+#    [cockpit]
+#    enabled = true
+#    port = http,https,9090
+#    filter = cockpit
+#    logpath = %(syslog_authpriv)s
+#    backend = %(syslog_backend)s
+#    maxretry = 3
+#    bantime = 1h
+
+[INCLUDES]
+
+before = common.conf
+
+[Definition]
+
+_daemon = cockpit-(?:session|ws)
+
+prefregex = ^%(__prefix_line)s<F-CONTENT>.+</F-CONTENT>$
+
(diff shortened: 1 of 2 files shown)


sebres:
I'm unsure we need such filter at all, since the vector is covered by `pam-generic` filter (and jail).
Just try with:
```bash
$ fail2ban-regex -vv /tmp/log 'pam-generic[__pam_auth="pam_\w+"]'
```
If the intention is only to match PAM authentication failures from `cockpit` only (e. g. in extra jail), one can simply use it within new jail:
```ini
[cockpit]
filter = pam-generic[__pam_re="pam_\w+\(cockpit:auth\):"]
port = http,https,9090
logpath = %(syslog_authpriv)s
backend = %(syslog_backend)s
enabled = true
```
This shall do the same job. So basically it'd be enough to add this to `jail.conf` (without `enabled = true`).

Issue #2810, opened 2020-08-13
Need firewalld zone selection criteria?


Paladin:
This could be a feature request or a bug (quite possibly a bug in my own understanding, for that matter).

#2217 # Environment:

_Fill out and check (`[x]`) the boxes which apply. If your Fail2Ban version is outdated, 
and you can't verify that the issue persists in the recent release, better seek support 
from the distribution you obtained Fail2Ban from_

- Fail2Ban version (including any possible distribution suffixes): 0.11.0 / 9.el7.2 (current version in CentOS repo)
- OS, including release name/version: CentOS 7.8.2003
- [ x] Fail2Ban installed via OS/distribution mechanisms
- [ x] You have not applied any additional foreign patches to the codebase
- [ ] Some customizations were done to the configuration (provide details below is so)

### The issue:

fail2ban always issues ban in default zone, which then fails to ban attacks coming from non-default zones.

#### Steps to reproduce

Server has two net interfaces, each assigned to a different zone
Attacker tries to break in over ssh via the interface assigned to the non-default zone
fail2ban issues ban on the default zone

#### Expected behavior

Attacker gets banned

#### Observed behavior

Attacker does not get banned

#### Any additional information

Not sure if this is readily solvable. For extreme precision, fail2ban would need to know which interface the attack came from (or which zone it was coming from) in order to issue the correct ban. It could be solved acceptably if I could provide it with a list of zones for the ban to be imposed in, but I don't readily see how that is done in the current product, and firewalld will not allow multiple default zones. Adding a config param to the jail (e.g., "firewalld-zones") for the action to use would probably be enough. Alternatively, adding the IP as source to the drop/reject zone might also be good enough.

As it sits, fail2ban seems to be only able to protect one firewalld zone. Rather like going into battle with a helmet and no body armor.

#### Any customizations done to /etc/fail2ban/ configuration
```
```

#### Relevant parts of /var/log/fail2ban.log file:
_preferably obtained while running fail2ban with `loglevel = 4`_

```
```

#### Relevant lines from monitored log files in question:

```
```

Paladin:
I forgot to add, the workaround I'm currently using is to hardcode

    zones = zone1,zone2

in firewallcmd-common.local and then create firewallcmd-rich-rules.local from the conf changing these actions:

[...]

PR #4184, opened 2026-05-18
As per example noted in the firewallcmd-rich-rules.conf file, properly specify the zone

wolfc:

- [X] **CONSIDER adding a unit test** if your PR resolves an issue
- [X] **LIST ISSUES** this PR resolves or describe the approach in detail
~~**ACCOMPANY** each new `failregex` for filter `X` with sample log lines
      (and `# failJSON`) within `fail2ban/tests/files/logs/X` file~~
- [X] **PROVIDE ChangeLog** entry describing the pull request

In the current `firewallcmd-rich-rules.conf` action the zone setting is disregarded. This results in the rich rule always applying to the default zone as opposed to the one configured.

By amending the command with a zone argument the proper zone is used for the rich rule.

### DIFF
--- a/config/action.d/firewallcmd-rich-rules.conf
+++ b/config/action.d/firewallcmd-rich-rules.conf
@@ -37,8 +37,8 @@ actioncheck =
 
 fwcmd_rich_rule = rule family=\"<family>\" source address=\"<ip>\" port port=\"$p\" protocol=\"<protocol>\" %(rich-suffix)s
 
-actionban = ports="<port>"; for p in $(echo $ports | tr ", " " "); do firewall-cmd --add-rich-rule="%(fwcmd_rich_rule)s"; done
+actionban = ports="<port>"; for p in $(echo $ports | tr ", " " "); do firewall-cmd --zone="<zone>" --add-rich-rule="%(fwcmd_rich_rule)s"; done
 	   
-actionunban = ports="<port>"; for p in $(echo $ports | tr ", " " "); do firewall-cmd --remove-rich-rule="%(fwcmd_rich_rule)s"; done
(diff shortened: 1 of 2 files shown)


sebres (commented):
Is zone `public` really default one? I mean, is this change completely backwards compatible?
In the [docu](https://firewalld.org/documentation/zone/default-zone.html) default zone is described as something different, if I understand it correct.

wolfc:
The default zone is configured at https://github.com/fail2ban/fail2ban/blob/1.1.0/config/action.d/firewallcmd-common.conf#L37 (aka `public`)

sebres:
I did not ask what is default in your action. I asked what is default by filrewalld without your enhancement (without zone supplied)... Please, read attentively the question, before you're trying to answer.

wolfc:
Ah my bad, we're talking the default zone of firewall-cmd if nothing is specified. That would be system specific and can be overridden using `--set-default-zone`. Meanwhile fail2ban specifies an explicit default zone being `public`. So effectively the backwards compatibility or rather the current functionality of fail2ban is unspecified at best as it is conflicting.

The zone configuration option in fail2ban was added through https://github.com/fail2ban/fail2ban/commit/7e54cee8d61be84bfb8f3eeff70f9f86b90eef9e but apparently has never worked correctly. See also #2810 
Note that removing the option does make fail2ban use the firewall-cmd default zone correctly but is equally non-backwards compatible in terms of provided functionality. It would constitute a zero-admin change though.

In the end relying on a bug in default behavior is bad. This PR fixes this discrepancy.

sebres:
Understand.
OK, thus merged. Thx!

PR #4208, opened 2026-07-20
fix: fail config test on invalid directives without '='

### LINKED ISSUE #4207: [BR]: Configuration test does not report invalid directive in jail.local
### Environment:

* Fail2Ban version: v1.0.2
* OS, including release name/version: Ubuntu 24.04 LTS
* [x] Fail2Ban installed via OS/distribution mechanisms
* [x] You have not applied any additional foreign patches to the codebase
* [x] Some customizations were done to the configuration

### The issue:

`fail2ban-client -t` reports the configuration as successful even when an invalid directive is present in `jail.local`.

#### Steps to reproduce

Create a configuration like:

```ini
[DEFAULT]
192.168.1.10 10.0.0.2
bantime = 3600
```

where `ignoreip =` was accidentally omitted.

Run:

```bash
sudo fail2ban-client -t
```

#### Expected behavior

The configuration test should report an error or at least a warning that the line is not a valid configuration directive.

#### Observed behavior

The output is:

```text
OK: configuration test is successful
```

The invalid line appears to be ignored silently, which can make configuration mistakes difficult to notice.

I'm not sure if this is expected behavior. If so, please feel free to close this issue. I thought it might be worth reporting because a warning from the configuration test could help users catch simple mistakes such as a missing `ignoreip =`.

#### Any customizations done to /etc/fail2ban/ configuration

```ini
[DEFAULT]
192.168.1.10 10.0.0.2
bantime = 3600
```

#### Relevant parts of /var/log/fail2ban.log file:

```
N/A
```

#### Relevant lines from monitored log files:

```
N/A
```

H-TTTTT:
## Problem
A bare line in `jail.local` (for example an IP list where `ignoreip =` was accidentally omitted) is not a valid config directive. Users expect `fail2ban-client -t` to report this clearly and fail the configuration test.

## Fix
Catch `configparser.ParsingError` while loading shared config files in `SafeConfigParserWithIncludes`. Log a clear **invalid directive** warning (filename, line, content) and treat the file as unread so the existing "Could not read config files" path fails the test—consistent with how unreadable/decoding-error files are handled.

## Test plan
- [x] `python3 bin/fail2ban-testcases fail2ban.tests.fail2banclienttestcase.Fail2banServerTest.testServerTestFailStart`
- [x] `python3 bin/fail2ban-testcases fail2ban.tests.clientreadertestcase`
- Manual: jail.local with `[DEFAULT]` / `192.168.1.10 10.0.0.2` / `bantime = 3600` → `fail2ban-client -t` exits non-zero and logs `Invalid directive`

Fixes #4207

### DIFF
--- a/fail2ban/client/configparserinc.py
+++ b/fail2ban/client/configparserinc.py
@@ -31,7 +31,7 @@ from ..helpers import getLogger
 
 # SafeConfigParser deprecated from Python 3.2 (renamed to ConfigParser)
 from configparser import ConfigParser as SafeConfigParser, BasicInterpolation, \
-	InterpolationMissingOptionError, NoOptionError, NoSectionError
+	InterpolationMissingOptionError, NoOptionError, NoSectionError, ParsingError
 
 # And interpolation of __name__ was simply removed, thus we need to
 # decorate default interpolator to handle it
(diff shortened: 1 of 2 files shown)


fail2ban/client/configparserinc.py
```diff
@@ -215,6 +215,26 @@ def _map_section_options(self, section, option, rest, defaults):
...
+
+	def _readSharedFile(self, cfg, filename):
+		"""Read a single config file; treat parse errors like unreadable files."""
+		try:
+			return cfg.read(filename, get_includes=False)
+		except ParsingError as e:
+			# Same idea as UnicodeDecodeError handling: don't apply a partial
+			# file and surface a clear error so `fail2ban-client -t` fails.
+			self._logParsingError(e)
+			return []
```
sebres, line 236: How adding an exception handling shall fix the issue (where previously the exception was throwing to the caller)?

sebres:
The issue #4207 is closed as incorrect, since the test with `fail2ban-client -t` already fails and show the line.

PR #4212, opened 2026-07-30
New XARF v4 reporting action, deprecate legacy xarf-login-attack

### LINKED ISSUE #3820: [FR]: Update the xarf action
(Is this really a filter request? I mean... it's more like an AR)

### Environment:

- Fail2Ban version  : Fail2Ban v1.0.2
- OS, including release name/version : CentOS Stream 9

#### Action information
- Name : xarf-login-attack
- Repository or URL (if known) : https://github.com/fail2ban/fail2ban/blob/master/config/action.d/xarf-login-attack.conf

#### Log or journal information
Not relevant

Our xarf filter https://github.com/fail2ban/fail2ban/blob/master/config/action.d/xarf-login-attack.conf is targeted for XARF 0.2, which is a very, very old version not supported by anyone. We should definitely seek to update to one of the versions at https://github.com/abusix/xarf.

A sample of the new format is https://github.com/abusix/xarf/blob/master/samples/positive/3/loginattack_sample_optional_api_info.json. It is definitely more involved, but since we *know* python is definitely present (hey, fail2ban is written in py!), we can be sure that there's a JSON library on the computer.

tknecht:
## Summary

fail2ban's only XARF capability today is `action.d/xarf-login-attack.conf`: a shell action emitting the obsolete **x-arf 0.2** format (hardcoded `Version: 0.2`, a schema URL that no longer reflects the current spec) via unescaped shell/`printf` string concatenation.

This PR adds a new `xarf` action that reports login-attack bans as **XARF v4** (https://xarf.org), and marks the legacy action deprecated (kept fully functional, no breaking change).

Closes #3820

- `fail2ban/server/xarfreport.py` — a reusable, dependency-free generator (`build_login_attack()`) that builds a XARF v4 `login_attack` document from ban data using only the stdlib. If the official `xarf` PyPI library is installed and exposes the expected v4 generator API, it's used instead for full schema validation; otherwise (or on any validation error) it falls back to the stdlib builder, so a report is always produced. `setup.py` gains no new hard dependency.
- `config/action.d/xarf.py` — a Python `ActionBase` action (modeled on `smtp.py`). Resolves the banned IP's abuse contact via the existing Abusix Contact DB TXT lookup, builds the v4 report, and sends it using the official XARF email transport (RFC 5965 `multipart/report; report-type=feedback-report`, built with Python's `email` package rather than shell concatenation). Best-effort `source_port` extraction from the matched log line where available (e.g. sshd). DNS-resolved contact strings are sanitized so they can't be smuggled in as argv flags to the mail command.
- `config/jail.conf` — new `action_xarf_v4` shortcut alongside the existing `action_xarf`; identity (`xarf_reporter_*`/`xarf_sender_*`) is set by the operator in `jail.local`.
- `config/action.d/xarf-login-attack.conf` — marked deprecated (header comment + runtime warning on `actionstart`), still fully functional for existing configs.
- `ChangeLog` — entries for the new action and the deprecation.

## Test plan

- [x] New unit tests for the generator (`fail2ban/tests/xarfreporttestcase.py`) covering both the stdlib and optional-library code paths, fallback-on-validation-error, and field mapping.
- [x] New action-level tests (`fail2ban/tests/servertestcase.py::XarfV4ActionTest`) covering identity validation, abuse-contact resolution/sanitization, XARF v4 email structure, and end-to-end `ban()`.
- [x] `bin/fail2ban-testcases` — full suite passes; the only failures are pre-existing and unrelated to this change (verified they also occur on unmodified `master`: live-DNS-dependent tests, a `setuptools`-missing test, and a shell-quoting test — all environment-dependent, not touched by this PR).
- [x] `pyflakes` clean on the new/changed Python files.


fail2ban/tests/servertestcase.py
```diff
@@ -2411,3 +2411,161 @@ def testComplexMailActionMultiLog(self):
 					ticket = _actions.Actions.ActionInfo(ticket, dmyjail)
 					action.ban(ticket)
 					self.assertLogged(*tests[test], all=True)
```
sebres, line 2413: There is folder [fail2ban/tests/action_d/](https://github.com/fail2ban/fail2ban/tree/master/fail2ban/tests/action_d) which is supposed to test actions.
Simply add new test class as `fail2ban/tests/action_d/test_*.py` like:

https://github.com/fail2ban/fail2ban/blob/1ebc369b05d7e871fa90e545dcd5308623c6a087/fail2ban/tests/action_d/test_smtp.py#L32

fail2ban/tests/xarfreporttestcase.py
```diff
@@ -0,0 +1,142 @@
...
+
+import base64
+import hashlib
+import unittest
+
+from ..server import xarfreport
+from .utils import LogCaptureTestCase
+
+
+class XarfEvidenceTest(unittest.TestCase):
```
sebres: Same here - just move it to `fail2ban/tests/action_d/test_*.py`.

fail2ban/tests/utils.py
```diff
@@ -382,6 +382,7 @@ def gatherTests(regexps=None, opts=None):
 	from . import samplestestcase
 	from . import fail2banclienttestcase
 	from . import fail2banregextestcase
+	from . import xarfreporttestcase
```
sebres: This and changes below would be unnecessary too if moved to `fail2ban/tests/action_d/test_*.py` (they are loaded automatically).

sebres (changes requested):
Thx for PR!

fail2ban/tests/servertestcase.py
tknecht: Moved — all XARF tests now live in `fail2ban/tests/action_d/test_xarf.py` (d40ae1ef), following the `test_smtp.py` convention. The manual registration in `utils.py` is gone too.

fail2ban/tests/xarfreporttestcase.py
tknecht: Moved — all XARF tests now live in `fail2ban/tests/action_d/test_xarf.py` (d40ae1ef), following the `test_smtp.py` convention. The manual registration in `utils.py` is gone too.

fail2ban/tests/utils.py
tknecht: Moved — all XARF tests now live in `fail2ban/tests/action_d/test_xarf.py` (d40ae1ef), following the `test_smtp.py` convention. The manual registration in `utils.py` is gone too.

config/action.d/xarf.py
```diff
@@ -0,0 +1,264 @@
...
+		"""Best-effort: extract the attacker source port from matched log text.
+
+		fail2ban generally knows only the attacked (destination) port; some
+		services (e.g. sshd) log the source port in the failure line. When
+		present, the most recent (triggering) match is used. Returns an int
+		in 1..65535, or None when unavailable.
+		"""
+		if not self._source_port_cre or not text:
+			return None
+		for val in reversed(self._source_port_cre.findall(text)):
```
sidhommoez: The reversed() here picks the wrong port on the stock sshd filter. sshd.conf:107 has
`cfooterre = ^<F-NOFAIL>Connection from</F-NOFAIL> <HOST>` and filter.py merges nofail-matches
into matches, so with LogLevel VERBOSE the ticket contains:

    Connection from 115.249.163.77 port 51353 on 127.0.0.1 port 22

findall gives ['51353', '22'] and we take 22 — the sshd listening port, reported as the attacker's.
Ran the real filter over tests/files/logs/sshd:206 to confirm.

Anchoring to the IP would fix it, something like `re.escape(str(aInfo['ip'])) + r'\s+port (\d{1,5})'`.
Given XARF calls this field critical for CGNAT attribution, I'd rather omit it than guess wrong.

Side note: `matches` defaults to ipmatches, which is merged across all jails, so the port can even
come from a different service's log line.

config/action.d/xarf.py
```diff
@@ -0,0 +1,264 @@
...
+		ts = self._iso(aInfo.get('time') or time.time())
+		evidence = aInfo.get(self.matches) or aInfo.get('matches') or ''
+		failures = aInfo.get('failures')
+		data = {
+			"source_identifier": str(aInfo['ip']),
+			"timestamp": ts,
+			"first_seen": ts,
+			"protocol": self.protocol,
+			"service": self.service,
+			"attempt_count": int(failures) if failures is not None else None,
```
sidhommoez, line 233: This fires for manual bans too. addBannedIP builds a bare BanTicket with
`{'matches': [], 'failures': 0}`, it's not restored so norestored doesn't skip it, and 0 is not
None so we end up with attempt_count: 0 and no evidence array. What
`fail2ban-client set sshd banip 203.0.113.9` actually sends:

    {"source_identifier": "203.0.113.9", "type": "login_attack",
     "attempt_count": 0, "evidence_source": "log", ...}

Zero evidence, and we're explicitly claiming zero attempts. Same path blocklist imports use.
Probably want to bail out early in ban() if there's no evidence.

config/action.d/xarf.py
```diff
@@ -0,0 +1,264 @@
...
+		}
+		if dport is not None:
+			data["destination_port"] = dport
+		sport = self._extract_source_port(evidence)
+		if sport is not None:
+			data["source_port"] = sport
+		return data
+
+	def _sendmail(self, recipients, msg):
+		cmd = shlex.split(self.mailcmd) + ['--'] + list(recipients)
```
sidhommoez: No -f here, so the envelope sender ends up as root even though the comment on 63-64 says
envelope_from is the envelope. It's only ever the From: header (line 144). Means SPF/DMARC won't
align and bounces go to root@localhost.

The old action passed `-f <sender>` (xarf-login-attack.conf:123), same as sendmail-common. Worth
adding it back, maybe with a mailargs option.

config/jail.conf
```diff
@@ -220,13 +229,28 @@ action_mw = %(action_)s
...
+# identity (xarf_*_org/contact/domain) set in jail.local. Install the official
+# XARF v4 `xarf` python package (pip install xarf) for full schema validation;
+# a built-in generator is used otherwise.
+action_xarf_v4 = %(action_)s
+             xarf.py[service="%(__name__)s", port="%(port)s", protocol="%(protocol)s",
+                     reporter_org="%(xarf_reporter_org)s", reporter_contact="%(xarf_reporter_contact)s",
+                     reporter_domain="%(xarf_reporter_domain)s",
+                     sender_org="%(xarf_sender_org)s", sender_contact="%(xarf_sender_contact)s",
+                     sender_domain="%(xarf_sender_domain)s"]
+
```
sidhommoez, line 254: Missing `sender="%(sender)s"` — action_mw, action_mwl and the old action_xarf all pass it. Without
it we always fall back to fail2ban@getfqdn() and quietly ignore whatever the admin configured.

Also xarf.py:73 uses socket.getfqdn() directly; we've got ipdns.getfqdn() / DNSUtils.getHostname()
for exactly the cases where that returns localhost or a PTR.

fail2ban/server/xarfreport.py
```diff
@@ -0,0 +1,137 @@
...
+	for key in ("evidence_source", "service", "destination_ip",
+			"destination_port", "source_port", "attempt_count"):
+		val = data.get(key)
+		if val is not None:
+			kwargs[key] = val
+	evtext = data.get("evidence_text")
+	if evtext:
+		kwargs["evidence"] = [_xarf.create_evidence(
+			"text/plain", evtext, description="fail2ban log matches")]
+	result = _xarf.create_report(
```
sidhommoez, line 129: create_report ignores the timestamp we pass on line 100 and stamps its own. Same input,
two paths:

    stdlib -> "2025-01-11T12:17:20Z"   (ban time)
    lib    -> "2026-07-30T08:20:54..." (now)

first_seen survives, so the lib document is inconsistent with itself. Probably needs setting on the
returned model. Might be worth a bug on the lib as well.

config/action.d/xarf.py
```diff
@@ -0,0 +1,264 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: t :
```
sidhommoez: modeline should end `t -*-`, not `t

fail2ban/server/xarfreport.py
```diff
@@ -0,0 +1,137 @@
...
+from ..helpers import getLogger
+
+logSys = getLogger(__name__)
+
+try:
+	import xarf as _xarf
+	# Only use the library when it exposes the expected XARF v4 generator API
+	# (guards against an unrelated/older package squatting the `xarf` name).
+	_HAVE_XARF = hasattr(_xarf, "create_report") and hasattr(_xarf, "create_evidence")
+except ImportError:  # pragma: no cover - depends on optional install
```
sidhommoez: only catches ImportError, but a pydantic mismatch would blow upload_python_module and kill the whole action

fail2ban/server/xarfreport.py
```diff
@@ -0,0 +1,137 @@
...
+	Uses the official ``xarf`` library when available; on any error falls
+	back to the stdlib builder so a report is still produced.
+	"""
+	if _HAVE_XARF:
+		try:
+			return _build_login_attack_lib(data)
+		except Exception as e:  # noqa: broad - never block reporting
+			logSys.warning(
+				"xarf library report build failed (%s: %s); "
+				"falling back to stdlib builder", type(e).__name__, e)
```
sidhommoez, line 171: This catch swallows the validation error from line 120 and then sends the report anyway via the stdlib builder, which doesn't validate. So we fall back exactly when the lib has told us the document is broken. Saw this live on a normal sshd ban:

    xarf library report build failed (ValueError: xarf validation failed:
      'source_port' is a required property); falling back to stdlib builder

...and the invalid JSON went out. Could we split the two cases — fall back on import/runtime errors, but drop the report on a real validation failure?

config/action.d/xarf.py
```diff
@@ -0,0 +1,264 @@
...
+				p = int(val)
+			except (TypeError, ValueError):
+				continue
+			if 1 <= p <= 65535:
+				return p
+		return None
+
+	def _aInfoToData(self, aInfo):
+		try:
+			dport = int(self.port)
```
sidhommoez: int(self.port) throws for every stock port value, and we swallow it with no log line, so
destination_port just silently disappears. jail.conf passes port="%(port)s" and the real values are
`ssh`, `http,https`, `0:65535`.

Also the default port="0" on line 59 does parse, and gives destination_port: 0 — which violates
minimum: 1. getservbyname for names, first entry for comma lists, skip ranges, and default to None?

sidhommoez (changes requested):
Transport looks right — I checked it against xarf.org/docs/email-transport and the 3-part multipart/report, Feedback-Type: xarf and the xarf.json attachment all match. Sanitising of the DNS contacts holds too, I tried CRLF and shell meta characters and nothing gets through. 🥇 

Four things I'd want fixed first though, all around source_port:

The login_attack schema makes source_port required whenever source_identifier is an IP (the conditional at the bottom of connection-login-attack.json). It's always an IP for us, but we can only scrape the port out of log text — and out of the 98 shipped filter samples, only sshd has it.
So most jails emit an invalid report. Worse, when the lib catches that, build_login_attack falls back to the stdlib builder and mails the same invalid doc anyway.

And where we do find a port, it's often the wrong one — with LogLevel VERBOSE the sshd filter pulls "Connection from <ip> port 51353 on 127.0.0.1 port 22" into the matches, and "last match wins" picks 22.

Last one: `fail2ban-client banip` sends a complaint with attempt_count 0 and no evidence at all.

Since the port genuinely isn't there for most filters, might be worth raising it on the spec side too rather than only working around it here.

Comments inline. Also the ChangeLog entries are missing a gh- ref.

tknecht:
Thanks for the thorough review — all four `source_port` issues plus the rest are addressed in affe8c28:

- **Wrong port on verbose sshd lines**: `source_port` extraction is now anchored to the banned IP itself, so `Connection from <ip> port 51353 on 127.0.0.1 port 22` no longer yields the local port 22. Added a regression test reproducing your exact scenario.
- **Manual bans (`banip`) with no evidence**: `ban()` now skips entirely (before even resolving an abuse contact) when there's no log evidence, instead of sending `attempt_count: 0`.
- **Missing `-f` envelope sender**: new `mailargs` option, defaulting to `-f <envelope_from>` (overridable).
- **`sender=` dropped in jail.conf**: `action_xarf_v4` now passes it through; also switched the default envelope-sender resolution from raw `socket.getfqdn()` to `DNSUtils.getHostname()`.
- **Lib ignores our `timestamp`**: now enforced on the returned dict so it can't drift from `first_seen`.
- **Bare `except ImportError`**: broadened to `except Exception`, with a test simulating a broken install via a fake `xarf` module to confirm it degrades instead of crashing.
- **Falling back to stdlib after a real validation rejection**: this was the one I'd flag as most important — added `XarfValidationError`, raised only when the library actually validated and rejected the data (as opposed to being unavailable/erroring). `build_login_attack()` now returns `None` in that case and nothing is sent, rather than shipping the same invalid document unvalidated via the stdlib path.
- **`int(self.port)` throwing on real jail.conf values**: replaced with a real parser handling service names, comma lists, and treating ranges as unknown. Also fixes `destination_port: 0` from the `port="0"` default, which you didn't call out directly but is the same root cause.
- Modeline and `ChangeLog` `gh-` refs fixed too.

On raising `source_port` on the spec side: already done — [xarf/xarf-spec#19](https://github.com/xarf/xarf-spec/issues/19). It landed as of yesterday (`source_port` is now `x-recommended`, not required, for `login_attack` and the other aggregate connection types), though only on `main` — the latest tagged release (`v4.2.0`) predates it, so the `xarf` PyPI library will still enforce the old requirement until it picks up a release built against the new schema.

10 new regression tests, one per bug, in `fail2ban/tests/action_d/test_xarf.py`.

sidhommoez (approved):
Thank you 
LGTM

tknecht:
@sebres — circling back on this one. The test-location feedback from your review (moving things into `fail2ban/tests/action_d/test_xarf.py`) was addressed back in d40ae1ef, and the field-mapping/delivery bugs sidhommoez found in a second pass were fixed in affe8c28 (details above). sidhommoez re-reviewed and approved. Whenever you get a chance, would appreciate another look — happy to address anything else that comes up.

sebres:
Unfortunately I have no time to review, too busy...

But, 2 things I noted right now:
- the entries in changelog should be moved to new sections (they are placed in 1.1.1 which is already released now)
- why do you need fail2ban/server/xarfreport.py‎ (in this folder)? Cannot the facilities be placed in action or test (depending on where it is needed)?

tknecht:
Thanks for the pointers, both addressed:

- ChangeLog entries moved into the new `1.1.2-dev-1` section (they were stuck under the now-released `1.1.1`, since this branch had gone stale — also rebased the whole branch onto current `master` while I was at it).
- Folded `fail2ban/server/xarfreport.py` directly into `config/action.d/xarf.py` — you're right that it had exactly one consumer, and `smtp.py` (the only other Python action) is fully self-contained with no bespoke helper module in `fail2ban/server/`. Matches that convention now; happy to pull it back out into a shared module if a second XARF-based transport ever needs it.

No rush — whenever you get a chance.

PR #4216, opened 2026-08-13
Add Cowrie honeypot JSON log filter

jarihu:
Adds a new filter for the Cowrie (https://github.com/cowrie/cowrie) SSH/Telnet honeypot that parses its JSON-format logs natively via a Python filter, rather than forcing fragile regex matching over JSON.

**Changes**
- config/filter.d/cowrie_json.conf - filter definition with a syntax-valid failregex (matching "src_ip": "<HOST>") and disabled datepattern, so the raw JSON line is handed intact to the Python parser.
- config/filter.d/cowrie_json.py - custom FileFilter subclass that json.loads each line, matches eventid == "cowrie.session.connect", extracts src_ip, and registers the failure via failManager.addFailure.

**Why**
Cowrie's JSON output contains dynamic fields and escapes that make pure-regex filters brittle. Parsing JSON directly gives reliable IP extraction and lets the filter target specific eventids cleanly.


config/filter.d/cowrie_json.conf
```diff
@@ -0,0 +1,10 @@
+[Definition]
+
+# 1. Reset prefregex so Fail2ban passes the entire line to Python
+prefregex = ^<F-CONTENT>.*</F-CONTENT>$
```
sebres: such prefregex is unneeded - it doesn't filter something and captures whole line

config/filter.d/cowrie_json.conf
```diff
@@ -0,0 +1,10 @@
...
+
+# 2. Syntax-valid regex containing <HOST> so Fail2ban's compiler doesn't throw an error
+failregex = ^.*"src_ip"\s*:\s*"<HOST>".*$
```
sebres: such weak REs are considered as vulnerable and would be not accepted:
- please anchor RE, `^.*` and `.*$` are not anchors at all
- please rewrite it without catch-alls
- if it contains IP only, use `<ADDR>` instead of `<HOST>`

config/filter.d/cowrie_json.py
```diff
@@ -0,0 +1,32 @@
...
+                    ticket = self.getFailTicket(ip, line=line)
+                    self.failManager.addFailure(ticket)
+                    logSys.info("[cowrie_json] Custom JSON match found IP: %s", ip)
+
+        except Exception as e:
+            # Silently ignore non-JSON or malformed log lines
+            pass
+
+        # Return super call so Fail2ban completes line processing
+        return super().processLine(line, flags)
```
sebres: This is unneeded for filter and unexpected here.

It looks like a new backend (to parse json from log), but it is wrong directory (backends are normally in fail2ban/server)... As well as the real backend may need more stuff as only simple class with overridden processLine method.
Guess some strange AI (which hallucinating) behind this PR.

sebres (changes requested):
You seem to follow 2 different approaches here:
- to write a new filter for some log
- to write a new generic backend to parse json-logs

The former is possible without the latter, as well as vice versa.

Please fix below mentioned things...

Also as our PR template says every new filter needs a test file, see for example any in [fail2ban/tests/files/logs](https://github.com/fail2ban/fail2ban/tree/master/fail2ban/tests/files/logs)
So create fail2ban/tests/files/logs/cowrie_json and provide few tests for matched (and possibly not matched) lines.

config/filter.d/cowrie_json.py
jarihu: Thanks for the input. Feedback was noted and new commits address all highlighted issues.

config/jail.conf
```diff
@@ -995,3 +995,8 @@ logpath = /var/log/vaultwarden.log
 [xrdp]
 port    = 3389
 logpath = /var/log/xrdp-sesman.log
+
+[cowrie]
+filter  = cowrie_json
```
sebres: By default, the filter name is always equal jail name in fail2ban. Would be good, if you'd rename cowrie_json to cowrie (and remove this line here).

sebres (changes requested):

.gitignore
```diff
@@ -11,3 +11,4 @@ __pycache__
 .vagrant/
 .idea/
 .venv/
+graphify-out
```
sebres: I'll remove this (must anyway rebase this PR in order to move log-entry) - you can add this to your local git-ignore (e. g. `.git/info/exclude`)...

sebres (approved):
rebased (changelog entry moved) + few improvements added

PR #4218, opened 2026-08-14
filter.d/postfix: catch "hostname ... does not resolve to address" in aggressive mode

### LINKED ISSUE #4078: [RFE]: Postfix should catch "warning: hostname ... does not resolve to address"
#### Feature request type
It is not uncommon to have postfix configured to reject senders that fake their hostname. It would be nice to match these for the jail.

#### Description
Example journal line
```txt
Sep 30 00:01:49 myhost.example.com postfix/smtpd[1069473]: warning: hostname ip-static.nexlogic.ph does not resolve to address 103.206.80.131: Name or service not known
```

My system is Ubuntu24. This currently has fail2ban `1.0.2`, which has a known problem with the `postfix` jail listening to the wrong systemd unit. But even the latest `filter.d/postfix.conf` doesn't catch anything.

Although I'm quite familiar with regex I find it extremely difficult to digest the postfix filter (or any other filter I must say). So, I'm hoping that one of the experts can come up with a regex for the above line.

VXNCXNX:
Postfix logs this when a client's reverse DNS does not resolve back to its address:

```
warning: hostname bad.example.com does not resolve to address 192.0.2.44: Name or service not known
```

The `warning:` prefix is already accepted via `mdpr-auth`, so the prefregex matches; only the failregex was missing.

This is the regex you gave in gh-4078, which @keestux confirmed works. It is added to `mdre-aggressive` only — a legitimate sender with broken rDNS trips this message, so it would be a false-positive source in the default mode.

Closes gh-4078

## Verification

```
stock,   mode=aggressive : 0 matched, 1 missed
patched, mode=aggressive : 1 matched, 0 missed
patched, default mode    : 0 matched, 1 missed   <- unchanged, as intended
```

Adjacent warnings do not newly match:

| line | matches |
|---|---|
| `hostname X does not resolve to address 192.0.2.7` | yes |
| `hostname X verification failed` | no |
| `192.0.2.9: address not listed for hostname X` | no |
| `connect from unknown[192.0.2.8]` | no |

`<ADDR>` rather than `<HOST>` since the tail is always an IP here.

`python3 bin/fail2ban-testcases` → `Ran 540 tests, OK (skipped=14)`.

The fixture is asserted, not skipped: changing its expected host fails the suite, and so does reverting the filter change on its own.

One question — you wrote this as a jail-local config rather than a filter change, so if you meant it to stay user-side, say the word and I will close this.

### DIFF
--- a/fail2ban/tests/files/logs/postfix
+++ b/fail2ban/tests/files/logs/postfix
@@ -154,6 +154,9 @@ Jan 14 16:18:16 xxx postfix/smtpd[14933]: warning: host[192.0.2.5]: SASL CRAM-MD
 # failJSON: { "time": "2004-11-04T09:11:01", "match": true , "host": "192.0.2.152", "user": "admin", "desc": "reason unavailable" }
 Nov  4 09:11:01 mail postfix/smtpd[1234]: warning: unknown[192.0.2.152]: SASL LOGIN authentication failed: (reason unavailable), sasl_username=admin
 
+# failJSON: { "match": true , "host": "192.0.2.44", "desc": "hostname does not resolve to address, aggressive only (gh-4078)" }
+Aug 14 10:00:01 mail postfix/smtpd[1234]: warning: hostname bad.example.com does not resolve to address 192.0.2.44: Name or service not known
+
 # ---------------------------------------
 # Test-cases of postfix DDOS mode:
 # ---------------------------------------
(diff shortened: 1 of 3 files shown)


fail2ban/tests/files/logs/postfix
sebres, line 149: we have already aggressive section here, so...

sebres, line 156: ... just move your new test here.

fail2ban/tests/files/logs/postfix
```diff
@@ -198,3 +198,7 @@ Sep 19 12:10:58 hostname postfix/smtpd[14059]: warning: Message delivery request
 # filterOptions: [{}, {"mode": "ddos"}, {"mode": "aggressive"}]
 # failJSON: { "match": false, "desc": "don't affect lawful data (sporadical connection aborts within DATA-phase, see gh-1813 for discussion)" }
 Feb 18 09:50:05 xxx postfix/smtpd[42]: lost connection after DATA from good-host.example.com[192.0.2.10]
+
+# filterOptions: [{"mode": "aggressive"}]
+# failJSON: { "match": true , "host": "192.0.2.44", "desc": "hostname does not resolve to address, aggressive only (gh-4078)" }
+Aug 14 10:00:01 mail postfix/smtpd[1234]: warning: hostname bad.example.com does not resolve to address 192.0.2.44: Name or service not known
```
sebres: \^ move it to line 157 (without new filterOptions aggressive section)

sebres (changes requested):
Thx for the PR.

VXNCXNX:
Moved into the existing aggressive section, thanks. The sample now sits after the "reason unavailable" case at line 157, and the extra filterOptions block at the end is gone.

Full suite still green (540 tests, 20 skipped). I checked the case is actually exercised rather than just present: changing the expected host to a wrong value makes it fail, restoring it passes.

sebres (approved):

PR #4219, opened 2026-08-14
fix: replace bare except with except Exception

TrueFurina:
### Summary

Six bare `except:` clauses catch everything, including
`KeyboardInterrupt`/`SystemExit`:

- `fail2ban/client/configparserinc.py` (2) — fall back when a defaults-map
  assignment fails
- `fail2ban/compat/asyncore.py` (4) — the vendored asyncore compatibility
  layer's `handle_error()` paths (already guarded by
  `except _reraised_exceptions: raise` before the bare except)

Replace with `except Exception:` — same intended behavior (fallback /
`handle_error()`), but `KeyboardInterrupt`/`SystemExit` and programming
errors propagate instead of being routed through error handlers.

### Verification

- `python -m py_compile` on both files → OK
- `grep -c "except:"` on both files → 0 remaining bare handlers

### Files changed

- `fail2ban/client/configparserinc.py` (2 lines)
- `fail2ban/compat/asyncore.py` (4 lines)

### DIFF
--- a/fail2ban/client/configparserinc.py
+++ b/fail2ban/client/configparserinc.py
@@ -202,11 +202,11 @@ after = 1.conf
 				rplcmnt = 1
 				try: # set it in map-vars (consider different python versions):
 					defaults[sopt] = v
-				except:
+				except Exception:
 					# try to set in first default map (corresponding vars):
 					try:
 						defaults._maps[0][sopt] = v
-					except: # pragma: no cover
+					except Exception: # pragma: no cover
 						# no way to update vars chain map - overwrite defaults:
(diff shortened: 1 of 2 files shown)


sebres (commented):
I guess you used something like find&replace, didn't you?

If yes, an approach with automatic replacement may be questionable, because in constructs like below it doesn't matter whether the exception is an error or keyboard interrupt or exit:
```python
...
while True:
  try:
    ...
  except:
    do_something_by_unexpected_end(); # because we'd leave the loop.
    break
...
```

PR #4224, opened 2026-08-16
paths-arch: define default banactions

felixonmars:
jail.conf no longer defines active banaction defaults since 1.1.1, and paths-debian.conf now supplies Debian’s nftables defaults. Arch packaging switches jail.conf to paths-arch.conf, which currently leaves %(banaction)s undefined and breaks stock jail parsing/tests. This restores the historical iptables defaults for the Arch path file.

### DIFF
--- a/config/paths-arch.conf
+++ b/config/paths-arch.conf
@@ -8,6 +8,8 @@ after  = paths-overrides.local
 
 
 [DEFAULT]
+banaction = iptables-multiport
+banaction_allports = iptables-allports
 
 apache_error_log = /var/log/httpd/*error_log
 


sebres:
> This restores the historical iptables defaults for the Arch path file.

1. IIRC, Arch like Debian uses nftables as the underlying kernel packet filtering framework by default... It would surely handle userspace compatibility through an nftables-based iptables compatibility layer. But why shall fail2ban use such mockup layer instead of native net-filter by default?
2. Although the idea was to leave this up to the maintainers of related distro, but if we starting to configure it here... Could we also provide the default actions for all other main distros?

sebres:
Done within deffcbcd1712a3a0cbd79c1da324096ebf36d306 together with other `paths-*.conf`. Thx!

PR #4220, opened 2026-08-15
Add support for systemd socket activation

h7x4:
Before submitting your PR, please review the following checklist:

- [X] **CONSIDER adding a unit test** if your PR resolves an issue
- [ ] **LIST ISSUES** this PR resolves or describe the approach in detail
      (There were no pre-existing issues about this that I could find)
- [X] **MAKE SURE** this PR doesn't break existing tests
- [X] **KEEP PR small** so it could be easily reviewed
- [X] **AVOID** making unnecessary stylistic changes in unrelated code
- [ ] **ACCOMPANY** each new `failregex` for filter `X` with sample log lines
      (and `# failJSON`) within `fail2ban/tests/files/logs/X` file
      (No new failregexes)
- [X] **PROVIDE ChangeLog** entry describing the pull request

---

This PR adds support for letting systemd manage fail2ban's socket's creation and lifecycle by providing the socket using fd-passing. This allows any services dependent on communicating with fail2ban via this socket to be started in parallel with fail2ban during boot, as well as making restarts more graceful by pausing incoming traffic on the socket while fail2ban restarts.

PR #4225, opened 2026-08-20
Fix systemd socket activation being broken

Deric-W:

- [x] **PROVIDE ChangeLog** entry describing the pull request

---

While trying to use fail2ban with the provided `fail2ban.socket` file I noticed that the resulting socket activation is broken in two ways:
1. the socket passed to fail2ban is shut down (by fail2ban) when fail2ban is stopped, resulting in systemd logging errors since it can no longer listen on it
2. the socket is created in the runtime directory of `fail2ban.service`, resulting in its removal (by systemd) when the service is stopped

This PR solves them as follows:
1. remember whether the socket was passed via systemd and skip shutdown in this case
2. instruct systemd to preserve the runtime directory (fail2ban still performs some cleanup, for example removing the pid file)


sebres:
Thx! Looks plausible to me...
Request to the author of #4220 @h7x4 - any objections?

h7x4:
Yeah, seems reasonable. Thank you for the fix :)

PR #4226, opened 2026-08-20
Add default banactions

arjendekorte:
According to the changelog for 1.1.1, default `banactions` need to be specified in paths-*.conf (maintainer level) now.

### DIFF
--- a/config/paths-opensuse.conf
+++ b/config/paths-opensuse.conf
@@ -9,6 +9,9 @@ after  = paths-overrides.local
 
 [DEFAULT]
 
+banaction = nftables
+banaction_allports = nftables[type=allports]
+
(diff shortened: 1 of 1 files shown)


arjendekorte:
Should use iptables instead of nftables (will update later)

sebres:
Huh? IIRC, modern openSUSE uses `nftables` as native underlying kernel packet-filtering framework, however managed typically through `firewalld` (which in turn uses `nftables` directly as its modern backend framework).
`iptables` is legacy and used as mock-up compatibility layer only.

Wouldn't it be better to extend #4224 to include these (and other) distributions instead?

arjendekorte:
You're right, see #4229. This also synchronizes back the paths in openSUSE Factory as of today.

PR #4239, opened 2026-09-16
fix SMTP ban durations and add optional notification batching

crazy-max:
SMTP ban notifications currently overwrite the ticket ban duration with the jail default. Preserve the duration supplied by `ActionInfo` so increased and permanent durations are reported correctly when already available.

Add optional batching to reduce notification bursts and allow time for asynchronous ban escalation. Tickets become eligible after a configurable age or prolongation, then send together when the count or time threshold is reached. Immediate delivery remains the default, and pending notifications are flushed on shutdown. A fixed age alone does not guarantee escalation has finished.


sebres:
The problem here is that the ticket's bantime change is completely asynchron to time point of ban... Here is how it could see:
1. jail triggered a ban for the ticket
2. the ban would be send to actions immediately to execute `actionban` as soon as possible.
3. the observer (running in different thread) recognize the IP of ticket as evil (recidive) and decided to increment the `bantime`
4. in this case the actions would additionally execute `actionprolong`, where `bantime` is guaranteed to be prolonged.

Thereby on step 2 there is no guarantee at all whether `bantime` has the initial value of the jail or increased by observer thread value. 
There was an idea to retard evaluation of `actionban` (depending on some parameter) to wait for check in step 3, but it is still PoC branch right now. 
So at the moment for things like mailer actions the only possibility is to send firstly 1 mail (banned) and hereafter another (prolonged)... Or to retard (and probably cumulate several tickets to single mail) inside the action. I think the letter is the best option.

crazy-max:
@sebres Hum I see, this doesn't resolve the timing race :thinking:. It only prevents SMTP from overwriting an already-set ticket duration with the jail default. Would you prefer delayed notifications to be included in this PR as well?

sebres:
> this doesn't resolve the timing race

Sure, I just wanted to explain that it doesn't make much sense to send `bantime` here.

> Would you prefer delayed notifications to be included in this PR as well?

I never say no to such suggestions. :)
Ideally, it'd be something like delay and accumulation for tickets with some burst:
1. send only tickets that are older than parameter "age" (e. g. 2 secs, to guarantee increment/prolongation take place), or which are already prolonged
2. either min count (e. g. 20 "old" tickets) or min latency (e. g. 10 seconds since last mail) is reached.

crazy-max:
Pushed extra commit following your suggestion. It batches notifications by ticket age or prolongation, then sends when the count or time threshold is reached. It's opt-in atm, so the default behavior stays unchanged. Does this look closer to what you had in mind?

sebres:
Ugh, nice!
It looks really close, if I understand everything correct (however I imagined more of some internals of fail2ban instead of the action, but OK)... Need to review that all later.

At first glance, the few points I would like to mention:
- it uses 1 additional thread per action (per jail), what is not an issue, just by many jails... maybe one could combine them to single thread;
- time.monotonic() is good, but at the moment it'd be affected by some timing issues in test-suite (because for instance it may be abrupt slow during the test)... we using MyTime for that, and in test MyTime.setTime to move the timing related part of test further in time;
- do I understand correctly - it doesn't  send the IP list by bulk mail, just jail-name and the count? I thought it'd be similar to single ban mail.

Someone other than me needs to test it extensively as well (I cannot do that in real production - too many bans/IPs, so I could only play some simulation in best case).

